PR #4, opened 2020-08-19
Check version

jensfr:
Make sure the kata-operator-payload image can be installed on the version of OCP in use.
This is done by comparing the clusterversion against the tag of the SourceImage field in the kataconfig custom resource. If they don't match the reconcile will fail with an error until the custom resource is changed. 

Signed-off-by: Jens Freimann <jfreiman@redhat.com>

### DIFF
--- a/deploy/crds/kataconfiguration.openshift.io_v1alpha1_kataconfig_cr.yaml
+++ b/deploy/crds/kataconfiguration.openshift.io_v1alpha1_kataconfig_cr.yaml
@@ -2,6 +2,9 @@ apiVersion: kataconfiguration.openshift.io/v1alpha1
 kind: KataConfig
 metadata:
   name: example-kataconfig 
+spec:
+  config:
+     sourceImage: quay.io/isolatedcontainers/kata-operator-payload:4.5.4
 #spec:
 #  kataConfigPoolSelector:
 #    matchLabels:
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -34,8 +34,7 @@ spec:
       containers:
         - name: kata-operator
           # Replace this with the built image name
-          #image: quay.io/isolatedcontainers/kata-operator:v1.0
-          image: quay.io/harpatil/kata-operator:v.0.1 
+          image: quay.io/isolatedcontainers/kata-operator:v2.0
           command:
           - kata-operator
           imagePullPolicy: Always
--- a/pkg/controller/kataconfig/openshift_reconciler.go
+++ b/pkg/controller/kataconfig/openshift_reconciler.go
@@ -155,7 +155,7 @@ func (r *ReconcileKataConfigOpenShift) processDaemonsetForCR(operation DaemonOpe
 					Containers: []corev1.Container{
 						{
 							Name:            "kata-install-pod",
-							Image:           "quay.io/isolatedcontainers/kata-operator-daemon:v1.0",
+							Image:           "quay.io/isolatedcontainers/kata-operator-daemon:v2.0",
 							ImagePullPolicy: "Always",
 							SecurityContext: &corev1.SecurityContext{
 								Privileged: &runPrivileged,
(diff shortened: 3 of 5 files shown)


pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -694,3 +715,95 @@ func (r *ReconcileKataConfigOpenShift) monitorKataConfigInstallation() (reconcil
...
+func checkCompatibility(clusterversion string, sourceImageVersion string) (bool, error) {
+	/* map OCP version to the minimum version of the sourceImage
+	 * that can be used on it. For example if a user tries
+	 * to install a sourceImage tagged with 4.5.3 on an OCP cluster
+	 * of version 4.5.4 the installation would fail because the RPM
+	 * dependencies don't match. In this case we return false to
+	 * signal that this is not supported.
+	 * TODO: don't hardcode this here, read it from a yaml file or similar
+	 */
+	minimumVersion := make(map[string]string)
```
harche, line 781: May be it's worth considering https://github.com/coreos/go-semver to handle version checking?

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -694,3 +715,95 @@ func (r *ReconcileKataConfigOpenShift) monitorKataConfigInstallation() (reconcil
...
+	imageslice := strings.Split(sourceImage, ":")
+	if len(imageslice) <= 1 {
+		return "", fmt.Errorf("bad format of sourceImage")
+	}
+	mytag := imageslice[len(imageslice)-1]
+
+	return mytag, nil
+}
+
+func getClientSet() (*kubernetes.Clientset, error) {
```
harche, line 758: Although it's used only in OpenShift reconciler, it's not really OpenShift specific code per say. I would suggest that leave at the common location. i.e. `kubeconfig_controller.go`. It could get useful in future for other controllers if they want to fetch the clientset to interact with API objects.

deploy/operator.yaml
```diff
@@ -34,8 +34,7 @@ spec:
       containers:
         - name: kata-operator
           # Replace this with the built image name
-          #image: quay.io/isolatedcontainers/kata-operator:v1.0
-          image: quay.io/isolatedcontainers/kata-operator:v1.0
+          image: quay.io/isolatedcontainers/kata-operator:v2.0
```
harche, line 37: Should we add a config map to specify that image instead of hard-coding it in a source code?

deploy/crds/kataconfiguration.openshift.io_v1alpha1_kataconfig_cr.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: kataconfiguration.openshift.io/v1alpha1
 kind: KataConfig
 metadata:
   name: example-kataconfig 
+spec:
+  config:
+     sourceImage: quay.io/isolatedcontainers/kata-operator-payload:4.5.4
```
harche, line 7: This was supposed to be vanilla k8s support feature only. We do not want users to specify the source image in OpenShift. Instead in OpenShift it should infer the correct image to use from the cluster version.

deploy/operator.yaml
jensfr: yes, but I'd do it as a follow-on PR?

deploy/crds/kataconfiguration.openshift.io_v1alpha1_kataconfig_cr.yaml
jensfr: It's right, we should just infer the right version from the OCP version. But: I thought about this, I still think it is a useful feature even for the OpenShift version for testing. If a user chooses this in production I'd just consider it not supported. Which is something we should document. What do you think?

pkg/controller/kataconfig/openshift_reconciler.go
jensfr: Right, I'll move it to a common file

pkg/controller/kataconfig/openshift_reconciler.go
jensfr: I didn't know about it. I'll switch to that. Thanks!

jensfr:
closing this PR and will send a new one

PR #16, opened 2020-10-15
add /host prefix to directories to be cleaned up

jensfr:
The preStop hook command is not run in a chroot environment, so we
need to delete /host/<dir> instead of just /<dir>

### DIFF
--- a/pkg/controller/kataconfig/openshift_reconciler.go
+++ b/pkg/controller/kataconfig/openshift_reconciler.go
@@ -210,7 +210,7 @@ func (r *ReconcileKataConfigOpenShift) processDaemonsetForCR(operation DaemonOpe
 							Lifecycle: &corev1.Lifecycle{
 								PreStop: &corev1.Handler{
 									Exec: &corev1.ExecAction{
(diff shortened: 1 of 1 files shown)


pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -210,7 +210,7 @@ func (r *ReconcileKataConfigOpenShift) processDaemonsetForCR(operation DaemonOpe
 							Lifecycle: &corev1.Lifecycle{
 								PreStop: &corev1.Handler{
 									Exec: &corev1.ExecAction{
-										Command: []string{"rm", "-rf", "/opt/kata-install", "/usr/local/kata/"},
+										Command: []string{"/bin/sh", "-c", "rm -rf /host/opt/kata-install /host/usr/local/kata/ ; touch /opt/jens.txt"},
```
fidencio, line 213: The `touch /opt/jens.txt` seems like a debug leftover?

jensfr: yes! I'll remove it

PR #17, opened 2020-10-19
add support for master/worker combined nodes

jensfr:
In some compact clusters variants nodes have both the master and worker
role.  However, the machine config pools that are used in
these configurations have all nodes in the master pool and zero nodes in
the worker pool.

Zero nodes in the worker pool breaks our current practice of installing
kata on all nodes in the worker pool. The solution is to install on all
nodes in the master pool instead. For that we fetch the worker pool
information and check if it has nodes in it. If it has a machine count
of zero we use the master pool instead.

This has an important implication:
in this master-only configuration we can't support installing on a
subset of nodes by setting a label on the selected nodes the user wants
to install kata on. Why? This feature depends on creating a custom pool
that inherits from the worker pool. MCP only supports inheriting from
the worker pool but not from the master pool. See
https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md

This is a trade-off for being able to support these compact cluster
types.

### DIFF
--- a/pkg/controller/kataconfig/openshift_reconciler.go
+++ b/pkg/controller/kataconfig/openshift_reconciler.go
@@ -274,7 +274,7 @@ func (r *ReconcileKataConfigOpenShift) newMCPforCR() *mcfgv1.MachineConfigPool {
 	return mcp
 }
 
-func (r *ReconcileKataConfigOpenShift) newMCForCR() (*mcfgv1.MachineConfig, error) {
+func (r *ReconcileKataConfigOpenShift) newMCForCR(machinePool string) (*mcfgv1.MachineConfig, error) {
 	isenabled := true
 	name := "kata-osbuilder-generate.service"
 	content := `
@@ -289,12 +289,18 @@ ExecRestart=/usr/libexec/kata-containers/osbuilder/kata-osbuilder.sh
 WantedBy=multi-user.target
 `
 
-	var workerRole string
+	kataOC, err := r.kataOcExists()
+	if err != nil {
+		return nil, err
+	}
 
-	if _, ok := r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels["node-role.kubernetes.io/worker"]; ok {
-		workerRole = "worker"
+	if kataOC {
+		machinePool = "kata-oc"
+	} else if _, ok := r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels["node-role.kubernetes.io/"+machinePool]; ok {
+		r.kataLogger.Info("in newMCforCR machinePool" + machinePool)
 	} else {
-		workerRole = "kata-oc"
+		r.kataLogger.Error(err, "no valid role for mc found")
+		return nil, err
 	}
 
 	mc := *&mcfgv1.MachineConfig{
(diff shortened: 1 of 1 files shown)


jensfr:
somethiing wrong with this, checking...

harche:
@jensfr have you tested this on a regular 6 nodes OC cluster?

jensfr:
@harche Yes, I've tested it also on regular 6 node clusters in the cloud and on 5-node  configuration on my bare-metal cluster.

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -405,22 +412,61 @@ func (r *ReconcileKataConfigOpenShift) listKataPods() error {
 	return nil
 }
 
+func kataOcExists(r *ReconcileKataConfigOpenShift) (bool, error) {
```
harche, line 415: if you change the method signature to `func ((r *ReconcileKataConfigOpenShift) kataOcExists()` you don't have to pass a pointer to the object as an argument.

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -405,22 +412,61 @@ func (r *ReconcileKataConfigOpenShift) listKataPods() error {
...
+	kataOcMcp := &mcfgv1.MachineConfigPool{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: "kata-oc"}, kataOcMcp)
+	if err != nil && errors.IsNotFound(err) {
+		r.kataLogger.Info("No kata-oc machine config pool found!")
```
harche, line 419: I haven't checked, but I was wondering does this logger not have `Error` or do we must use `Info` to convey the error message?

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -405,22 +412,61 @@ func (r *ReconcileKataConfigOpenShift) listKataPods() error {
...
+		return false, nil
+	} else if err != nil {
+		r.kataLogger.Info("Something went wrong trying to get the kata-oc machine config pool!")
+		return false, err
+	}
+
+	return true, nil
+}
+
+func workerOrMaster(r *ReconcileKataConfigOpenShift) (string, error) {
```
harche, line 429: same as, https://github.com/openshift/kata-operator/pull/17/files#r509025406

harche (commented):
minor comments, otherwise LGTM

pkg/controller/kataconfig/openshift_reconciler.go
jensfr: good point :-) I'll do that, thx

pkg/controller/kataconfig/openshift_reconciler.go
jensfr: It does have Error, I changed it here and in a few other places added in this PR. Thanks!

pkg/controller/kataconfig/openshift_reconciler.go
jensfr: fixed, thanks!

PR #18, opened 2020-10-22
Bug 1890509: openshift: handle errors when updating kataconfig

jensfr:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1890509

### DIFF
--- a/pkg/controller/kataconfig/openshift_reconciler.go
+++ b/pkg/controller/kataconfig/openshift_reconciler.go
@@ -403,7 +403,10 @@ func (r *ReconcileKataConfigOpenShift) addFinalizer() error {
 
 	// Update CR
 	err := r.client.Update(context.TODO(), r.kataConfig)
-	if err != nil {
+	if err != nil && errors.IsConflict(err) {
+		r.kataLogger.Info("Updating kataconfig finalizer failed, will retry")
+		return err
+	} else if err != nil {
 		r.kataLogger.Error(err, "Failed to update KataConfig with finalizer")
 		return err
 	}
@@ -432,7 +435,7 @@ func (r *ReconcileKataConfigOpenShift) kataOcExists() (bool, error) {
 	kataOcMcp := &mcfgv1.MachineConfigPool{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: "kata-oc"}, kataOcMcp)
 	if err != nil && errors.IsNotFound(err) {
-		r.kataLogger.Error(err, "No kata-oc machine config pool found!")
+		r.kataLogger.Info("No kata-oc machine config pool found!")
 		return false, nil
 	} else if err != nil {
 		r.kataLogger.Error(err, "Could not get the kata-oc machine config pool!")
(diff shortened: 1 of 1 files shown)


pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -16,6 +16,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	nodeapi "k8s.io/api/node/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apimachineryErrors "k8s.io/apimachinery/pkg/api/errors"
```
harche, line 19: `k8s.io/apimachinery/pkg/api/errors` is getting imported here and the line above too.

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -414,8 +418,8 @@ func (r *ReconcileKataConfigOpenShift) listKataPods() error {
 func (r *ReconcileKataConfigOpenShift) kataOcExists() (bool, error) {
 	kataOcMcp := &mcfgv1.MachineConfigPool{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: "kata-oc"}, kataOcMcp)
-	if err != nil && errors.IsNotFound(err) {
-		r.kataLogger.Error(err, "No kata-oc machine config pool found!")
+	if err != nil && apimachineryErrors.IsNotFound(err) {
+		r.kataLogger.Info("No kata-oc machine config pool found!")
```
harche, line 422: It's not an error? We moved from Error to Info, so just checking.

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -500,7 +504,10 @@ func (r *ReconcileKataConfigOpenShift) processKataConfigInstallRequest() (reconc
 		if err != nil && errors.IsNotFound(err) {
 			r.kataLogger.Info("Creating a new installation Daemonset", "ds.Namespace", ds.Namespace, "ds.Name", ds.Name)
 			err = r.client.Create(context.TODO(), ds)
-			if err != nil {
+			/* Retry when there was a conflict */
+			if err != nil && apimachineryErrors.IsConflict(err) {
```
harche, line 508: Nice. :)

harche (commented):
minor nits

cpmeadors:
It looks like this addresses all three errors I saw in the logs.  I was going to try to test it quickly, but realized I think I need a new image.  Is there a quick way for me to build an image to test against?  Otherwise I will test once merged it makes it way through the build pipeline.

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -500,7 +504,10 @@ func (r *ReconcileKataConfigOpenShift) processKataConfigInstallRequest() (reconc
...
+				return reconcile.Result{}, nil
```
harche, line 509: If you get a conflict, shouldn't you reschedule another reconciliation? Since it's a conflict, your object has not been modified. I think you will have to request the reconciliation.

harche: the comment you added `/* Retry when there was a conflict */` means you intend to retry, but if you return `reconcile.Result{}, nil` the controller will not reschedule another reconciliation.

jensfr: You're right. It should be reconcile.Result{Requeue: true,...} or err has to be != nil. Thanks!

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -500,7 +504,10 @@ func (r *ReconcileKataConfigOpenShift) processKataConfigInstallRequest() (reconc
...
+			} else if err != nil {
```
zanetworker, line 510: WDYT about:

```go	
		if err != nil {
				if apimachineryErrors.IsConflict(err) {
					return reconcile.Result{}, nil
				}
				return reconcile.Result{}, err
			}
```

pkg/controller/kataconfig/openshift_reconciler.go
jensfr: will fix it!

pkg/controller/kataconfig/openshift_reconciler.go
jensfr: I changed it to Info on puropose. We don't expect it to exist, we're just checking but both is fine. NotFound is the good case of all error reasons. Thanks!

harche (approved):
/lgtm

cpmeadors (commented):
Changes look good, but I forgot how I reproduced this.  Any way to functionally test and verify?  Otherwise regression testing is all we have,

PR #22, opened 2020-10-28
openshift: allow custom payload images for development

jensfr:
Sometimes we need to test new builds of RPMs, for example to
verify a bug in QEMU, kata-runtime or other packages.

For this purpose the user can now add a configmap with a single data
entry that specifies a URL to a container image, e.g.
quay.io/someuser/repository:tag.

The ConfigMap will be added to the Pod spec of the installer daemon pods as an
environment variable. If the variable is set the daemon will use it. If
not the default mechanism for choosing the payload image is used.

The purpose of this feature is for development purposes only. When the
ConfigMap is used the operator will print a warning that using
self-built RPMs taints the kataconfig installation.

Also add a description in docs/HACKING.md on how to use the ConfigMap
and how to build a custom payload image.

### DIFF
--- a/deploy/configmap_payload.yaml
+++ b/deploy/configmap_payload.yaml
@@ -1,0 +1,9 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  creationTimestamp: 2016-02-18T19:14:38Z
+  name: payload-config
+  namespace: kata-operator
+data:
+  # change to your custom payload repository:tag value
+  daemon.payload: quay.io/user/repository:test
--- a/images/daemon/pkg/daemon/kata_openshift.go
+++ b/images/daemon/pkg/daemon/kata_openshift.go
@@ -406,9 +406,18 @@ func installRPMs(k *KataOpenShift) (err error) {
 		fmt.Println(err)
 	}
 
-	srcRef, err := alltransports.ParseImageName("docker://quay.io/isolatedcontainers/kata-operator-payload:" + k.PayloadTag)
+	payloadImage := os.Getenv("KATA_PAYLOAD_IMAGE")
+	if payloadImage == "" {
+		payloadImage = "docker://quay.io/isolatedcontainers/kata-operator-payload:" + k.PayloadTag
+	} else {
+		log.Println("WARNING: kataconfig installation is tainted")
+		log.Println("Using env variable KATA_PAYLOAD_IMAGE " + payloadImage)
+		payloadImage = "docker://" + payloadImage
+	}
+
+	srcRef, err := alltransports.ParseImageName(payloadImage)
 	if err != nil {
-		fmt.Println("Invalid source name")
+		fmt.Println("Invalid source name of payload container image: " + payloadImage)
 		return err
 	}
 
--- a/pkg/controller/kataconfig/openshift_reconciler.go
+++ b/pkg/controller/kataconfig/openshift_reconciler.go
@@ -161,8 +161,11 @@ func (r *ReconcileKataConfigOpenShift) isOldestCR() (bool, error) {
 }
 
 func (r *ReconcileKataConfigOpenShift) processDaemonsetForCR(operation DaemonOperation) *appsv1.DaemonSet {
-	runPrivileged := true
-	var runAsUser int64 = 0
+	var (
+		runPrivileged = true
+		configmapOptional = true
+		runAsUser int64 = 0
+	)
 
 	dsName := "kata-operator-daemon-" + string(operation)
 	labels := map[string]string{
@@ -219,6 +222,20 @@ func (r *ReconcileKataConfigOpenShift) processDaemonsetForCR(operation DaemonOpe
 								{
 									Name:      "hostroot",
 									MountPath: "/host",
+								},
+							},
+							Env: []corev1.EnvVar{
+								{
+									Name: "KATA_PAYLOAD_IMAGE",
+									ValueFrom: &corev1.EnvVarSource{
+										ConfigMapKeyRef: &corev1.ConfigMapKeySelector{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: "payload-config",
+											},
+											Key:      "daemon.payload",
+											Optional: &configmapOptional,
+										},
+									},
 								},
 							},
 						},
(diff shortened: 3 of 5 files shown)


jensfr:
@zanetworker

harche:
@jensfr nice :)

doc/HACKING.md
```diff
@@ -0,0 +1,27 @@
+# Hacking on the kata-operator
```
zanetworker, line 1: I thinking Hacking is cool in the content of the doc, but for the title (also the first doc we add :D) maybe `local-development` / `development` or something along would fit more? WDYT?

doc/HACKING.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Using a custom kata-operator-payload image
+
```
zanetworker, line 4: Do you think it makes sense to add the "why" (probably copied from the issue description) we would want to set a custom image?

jensfr:
Just realized the daemon is also part of this repo now and I can include the change in this PR. Will update the branch soon.

pkg/controller/kataconfig/openshift_reconciler.go
```diff
@@ -162,6 +162,7 @@ func (r *ReconcileKataConfigOpenShift) isOldestCR() (bool, error) {
 
 func (r *ReconcileKataConfigOpenShift) processDaemonsetForCR(operation DaemonOperation) *appsv1.DaemonSet {
 	runPrivileged := true
+	configmapOptional := true
```
zanetworker, line 165: maybe group all the vars under a var block: 

example: 
```go
var (
	runPrivileged = true
	configmapOptional = true
	var runAsUser int64 = 0
)
```

deploy/configmap_payload.yaml
```diff
@@ -0,0 +1,8 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  creationTimestamp: 2016-02-18T19:14:38Z
+  name: payload-config
+  namespace: kata-operator
+data:
+  daemon.payload: quay.io/jensfr/kata-operator-payload:test
```
zanetworker, line 8: maybe add an indication that the value is an image tag?

doc/HACKING.md
```diff
@@ -0,0 +1,27 @@
...
+Note: it is not possible to add additonal RPMs this way
+
+An example:
+
+1. skopeo copy docker://quay.io/jensfr/kata-operator-payload:4.6.0 oci:/tmp/kata-operator-payload:4.6.0
+2. oci-image-tool unpack --ref name=4.6.0  /tmp/kata-operator-payload kata-operator-payload-unpacked-4.6.0
+3. cp -r /tmp/kata-operator-payload-unpacked-4.6.0/packages $KATA_OPERATOR_REPO/images/payload
+4. cd $KATA_OPERATOR_REPO/images/payload
+5. replace RPMs in packages/ with custom RPMs
+6. podman build --no-cache -f Dockerfile.custom quay.io/<username>/mykatapayload:mytag
```
zanetworker, line 24: can it also be built with docker?

zanetworker:
One other thing, are you also planning to cherry pick this master?

images/daemon/pkg/daemon/kata_openshift.go
```diff
@@ -406,9 +406,18 @@ func installRPMs(k *KataOpenShift) (err error) {
 		fmt.Println(err)
 	}
 
-	srcRef, err := alltransports.ParseImageName("docker://quay.io/isolatedcontainers/kata-operator-payload:" + k.PayloadTag)
+	payloadImage := os.Getenv("KATA_PAYLOAD_IMAGE")
+	if payloadImage == "" {
+		payloadImage = "docker://quay.io/isolatedcontainers/kata-operator-payload:" + k.PayloadTag
+	} else {
+		log.Println("WARNING: kataconfig installation is tainted")
```
zanetworker, line 413: What is meant by tainted here?

images/daemon/pkg/daemon/kata_openshift.go
```diff
@@ -406,9 +406,18 @@ func installRPMs(k *KataOpenShift) (err error) {
...
+		log.Println("Using env variable KATA_PAYLOAD_IMAGE " + payloadImage)
+		payloadImage = "docker://" + payloadImage
+	}
+
+	srcRef, err := alltransports.ParseImageName(payloadImage)
 	if err != nil {
-		fmt.Println("Invalid source name")
+		fmt.Println("Invalid source name of payload container image: " + payloadImage)
```
zanetworker, line 420:
```suggestion
		fmt.Printf("Invalid source name of payload container image: %q", payloadImage)
```

doc/HACKING.md
jensfr: I thought the name was common in related projects. machine-config-operator has the same file, opencontaimer/image-spec, cluster-etc-operator all have a HACKING.md. I wanted not to deviate from that. But that's the only reason. I have no strong opinion either way, so I'll just name it DEVELOPMENT. Ok?

doc/HACKING.md
jensfr: It sure doesn't hurt :-) :+1:

doc/HACKING.md
jensfr: what's docker? ;-) Yes, I  see no reason why not.

images/daemon/pkg/daemon/kata_openshift.go
jensfr: Same meaning it has when you load a proprietary kernel module. You're on your own when problems occur. 
https://www.kernel.org/doc/html/latest//admin-guide/tainted-kernels.html

pkg/controller/kataconfig/openshift_reconciler.go
jensfr: :+1:

doc/HACKING.md
zanetworker: Sure, the reason I commented is that other OS projects have a dedicated folder usually called `hack` that contains scripts for many things. But it is up to you, no strong opinions on my side as well :)

jensfr: ah, I did not get that from your comment. Sure, I think we add a helper script in hack/ to build a custom payload but I'd prefer to do that in another PR.

jensfr:
> One other thing, are you also planning to cherry-pick this to master?

Yes, I'll send a PR for master as well but it requires some small changes.

doc/HACKING.md
zanetworker: It looks good as it is now :)

Issue #24, opened 2020-11-04
Fix command to clean up after installer pods finished


jensfr:
Add /host prefix to directories which need to be deleted as this command is not running in chroot

PR #25, opened 2020-11-04
fix command run at preStop event

jensfr:
Add /host to directories in rm command because it is not run in chroot
environment.

Fixes: #24 


jensfr:
This was fixed in release-4.6 by 

commit 62a4b8ef5ce4072385a41f8553501b3a991e8450
Author: Jens Freimann <jfreimann@redhat.com>
Date:   Thu Oct 15 15:39:54 2020 +0200

    add /host prefix to directories to be cleaned up
    
    The preStop hook command is not run in a chroot environment, so we
    need to delete /host/<dir> instead of just /<dir>
    
fidencio:
/approve
/lgtm

Thanks for the forward port, @jensfr!

jensfr:
Thanks for the review @fidencio!

PR #28, opened 2020-11-11
Fix example fedora yaml runtimeclass

jensfr:
We've switched to 'kata' as runtimeclass because it is what upstream 
kata-deploy uses as well and users got confused about the -oc.

The changes is already in the release-4.6 branch but on the master branch
we are still missing this change. 

Fix the example pod file accordingly.


harche (approved):
/lgtm

PR #21, opened 2020-10-28
fix URLs in openshift container image description

jensfr:
The URLs are pointing to 404 pages, let's fix it.

PR #32, opened 2020-11-20
change kata-operator-daemon container image

jensfr:
When code for the new operator was merged first we missed that
there is still a private test container image referenced in the
code.

This patch fixes it and changes the container image to
quay.io/isolatedcontainers/kata-operator-daemon:4.7

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -153,9 +153,8 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForCR(operation DaemonOp
 					NodeSelector:       nodeSelector,
 					Containers: []corev1.Container{
 						{
-							Name: "kata-install-pod",
-							// Image:           "quay.io/isolatedcontainers/kata-operator-daemon:v1.0",
-							Image:           "docker.io/pharshal/kata-install-daemon:0.1.1",
+							Name:            "kata-install-pod",
+							Image:           "quay.io/isolatedcontainers/kata-operator-daemon:4.7",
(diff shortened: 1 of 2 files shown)


harche (approved):
/lgtm

zanetworker (commented):
/lgtm

zanetworker (approved):

zanetworker:
Maybe also relevant in this PR. Noticed that the links for the `kata-operator-daemon` does not work. 

image

controllers/openshift_controller.go
```diff
@@ -153,9 +153,8 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForCR(operation DaemonOp
 					NodeSelector:       nodeSelector,
 					Containers: []corev1.Container{
 						{
-							Name: "kata-install-pod",
-							// Image:           "quay.io/isolatedcontainers/kata-operator-daemon:v1.0",
-							Image:           "docker.io/pharshal/kata-install-daemon:0.1.1",
+							Name:            "kata-install-pod",
```
zanetworker, line 156: In different places, I found `kata-install-daemon`, `kata-operator-daemon`, `kata-install-pod`. This can be confusing a bit, I think consistency here is relevant.

Also, here it is called `kata-install-pod` but its really just a container in a pod in that context.

jensfr: Yes, I'll fix the README. I was about to do it in pr #21 but closed it for some reason. Anyway thanks for the hint!
And about the different namings, I also agree and will open an issue for it.

PR #34, opened 2020-11-24
runtimeclass: set pod overhead 

jensfr:
Upstream kata-deploy has the overhead of a Kata Pod
set as

overhead:
    podFixed:
        memory: "160Mi"
        cpu: "250m"

(https://github.com/kata-containers/packaging/blob/master/kata-deploy/k8s-1.18/kata-runtimeClasses.yaml#L7)

So far we had no Pod Overhead set at all. Let's use the same values
as upstream uses to take the Pod resource overhead into account when the
'kata' runtime class is used.

Tested by checking for Overhead in runtimeclass yaml after it is created and found:
Overhead:                                                                                                                                                                                                                                      
  Pod Fixed:                                                                                                           
    Cpu:     250m                                          
    Memory:  160Mi

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -33,6 +33,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	nodeapi "k8s.io/api/node/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,9 +98,9 @@ The kata-operator uses three containers:
 
 Container image name | Description | Repository
 ---------------| ----------- | ----------
- _kata-operator_ |  It contains the controller part of the operator that watches and manages the kataconfig custom resource. It runs as a cluster scoped container. The operator itself is build with operator-sdk. | https://github.com/isolatedcontainers/kata-operator
- _kata-operator-daemon_ | The daemon part of the operator that runs on the nodes and performs the actual installation. It pulls down the container kata-operator-payload image. | https://github.com/isolatedcontainers/kata-operator-daemon
- _kata-operator-payload_ | The payload that is used by the daemon to install the kata binaries and dependencies (like e.g. QEMU). It's a container image with (currently) RPMs in it that will be installed on the chosen worker nodes by the daemon. | https://github.com/isolatedcontaineres/kata-operator-payload
+ _kata-operator_ |  It contains the controller part of the operator that watches and manages the kataconfig custom resource. It runs as a cluster scoped container. The operator itself is build with operator-sdk. | https://github.com/openshift/kata-operator/tree/release-4.6/
```
harche, line 101: Can we handle this change in a separate PR? It has nothing to do with pod overhead fix.

jensfr: yes, that should be a separate PR

jensfr: actually it's also wrong

controllers/openshift_controller.go
```diff
@@ -447,6 +448,12 @@ func (r *KataConfigOpenShiftReconciler) setRuntimeClass() (ctrl.Result, error) {
 				Name: runtimeClassName,
 			},
 			Handler: runtimeClassName,
+			Overhead: &nodeapi.Overhead{
+				PodFixed: corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("250m"),
```
harche, line 453: Maybe it would be a good idea to add a comment here explaining where these values came from to help future readers understand the rationale behind setting those values. You have already mentioned that you got those values from, https://github.com/kata-containers/packaging/blob/master/kata-deploy/k8s-1.18/kata-runtimeClasses.yaml#L7, in the PR description, so I would suggest add that info here as in the form a comment too.

jensfr: Yes, it's a good idea. I've done it and pushed

harche (approved):
/lgtm

PR #35, opened 2020-11-24
Fix description and URLs of containers in README

jensfr:
To make it more clear I change the column "Repository" to "Container repository" and also add a sentence in the description as to where the Dockerfile can be found. 
Also fix the URls from github.com to quay.io.


harche (approved):
/lgtm

PR #39, opened 2020-11-30
Custom payload images

jensfr:
    
    Sometimes we need to test new builds of RPMs, for example to
    verify a bug in QEMU, kata-runtime or other packages.
    
**- What I did**

    For this purpose the user can now add a configmap with a single data
    entry that specifies a URL to a container image, e.g.
    quay.io/someuser/repository:tag.
    
    The ConfigMap will be added to the Pod spec of the installer daemon pods as an
    environment variable. If the variable is set the daemon will use it. If
    not the default mechanism for choosing the payload image is used.
    
    The purpose of this feature is for development purposes only. When the
    ConfigMap is used the operator will print a warning that using
    self-built RPMs taints the kataconfig installation.
    
    Also add a description in docs/DEVELOPMENT.md on how to use the ConfigMap
    and how to build a custom payload image.

**- How to verify it**

    Create a configMap and apply it.  See docs/DEVELOPMENT.md. Then check the log of kata-operator-daemon to see if it is using the custom payload 
    image specified in the configMap.

   The user can now add a configmap with a single data
    entry that specifies a URL to a container image

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -114,8 +114,11 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(req ctrl.Request) (ctrl.Result
 }
 
 func (r *KataConfigOpenShiftReconciler) processDaemonsetForCR(operation DaemonOperation) *appsv1.DaemonSet {
-	runPrivileged := true
-	var runAsUser int64 = 0
+	var (
+		runPrivileged           = true
+		configmapOptional       = true
+		runAsUser         int64 = 0
+	)
 
 	dsName := "kata-operator-daemon-" + string(operation)
 	labels := map[string]string{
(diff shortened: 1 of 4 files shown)


controllers/openshift_controller.go
```diff
@@ -174,6 +177,20 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForCR(operation DaemonOp
 									MountPath: "/host",
 								},
 							},
+							Env: []corev1.EnvVar{
```
zanetworker, line 180: qq: what happens if the config map value changes at runtime?

jensfr: It's ignored. It's evaluated only once before the daemon downloads the payload image.

zanetworker: If that's the desired behavior then it's fine. I was asking because if that is for the purpose of development, a developer would probably want to test with multiple images. To do so, they would be required to restart the controller.

PR #40, opened 2020-12-01
Prevent multiple KataConfig CRs

harche:
Signed-off-by: Harshal Patil <harpatil@redhat.com>

Prevent multiple KataConfig CRs from getting processed by the controller. This is a similar solution to what we did in 4.6, https://github.com/openshift/kata-operator/pull/11 

**- What I did**
Prevent the controller from processing the KataConfig newer than the first KataConfig

**- How to verify it**
If the second KataConfig is created, the failed node count in the status of that CR should have been set to -1. 

Prevent multiple KataConfig CRs from getting processed by the controller

### DIFF
--- a/api/v1/kataconfig_types.go
+++ b/api/v1/kataconfig_types.go
@@ -26,6 +26,7 @@ type KataConfigSpec struct {
 	// KataConfigPoolSelector is used to filer the worker nodes
 	// if not specified, all worker nodes are selected
 	// +optional
+	// +nullable
 	KataConfigPoolSelector *metav1.LabelSelector `json:"kataConfigPoolSelector"`
 
 	// +optional
@@ -68,6 +69,7 @@ type KataConfig struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	// +optional
+	// +nullable
 	Spec   KataConfigSpec   `json:"spec,omitempty"`
 	Status KataConfigStatus `json:"status,omitempty"`
 }
--- a/config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
+++ b/config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
@@ -1,10 +1,10 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.0
+    controller-gen.kubebuilder.io/version: v0.4.0
   creationTimestamp: null
   name: kataconfigs.kataconfiguration.openshift.io
 spec:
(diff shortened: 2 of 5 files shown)


zanetworker:
To me this raises the question on whether a CR is the right mapping here?

harche:
This is what MCO had been doing with CRs like KubeletConfig in absence of the a webhook.

harche:
Also, CRs marked duplicate aren't re-queued, https://github.com/openshift/kata-operator/pull/40/files#diff-fbc9237b068d36a3e13bd2aa40352cfacfd5a642fc85ddd2f8651fb15fe69c63R91

zanetworker:
>This is what MCO had been doing with CRs like KubeletConfig in absence of the a webhook. 

Interesting, so this is a temporary approach till we get the webhook integration?

harche:
> > This is what MCO had been doing with CRs like KubeletConfig in absence of the a webhook.
> 
> Interesting, so this is a temporary approach till we get the webhook integration?

Yes

api/v1/kataconfig_types.go
```diff
@@ -26,6 +26,7 @@ type KataConfigSpec struct {
 	// KataConfigPoolSelector is used to filer the worker nodes
 	// if not specified, all worker nodes are selected
 	// +optional
+	// +nullable
```
jensfr, line 29: what does this do?

config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
```diff
@@ -15,229 +15,232 @@ spec:
...
-              description: KataConfigPoolSelector is used to filer the worker nodes
-                if not specified, all worker nodes are selected
-              properties:
-                matchExpressions:
-                  description: matchExpressions is a list of label selector requirements.
-                    The requirements are ANDed.
-                  items:
-                    description: A label selector requirement is a selector that contains
-                      values, a key, and an operator that relates the key and values.
+  versions:
```
jensfr, line 18: This part is hard to review for me just by looking at the diff. What are the important changes here to prevent multiple CRs?

controllers/openshift_controller.go
```diff
@@ -735,3 +743,63 @@ func (r *KataConfigOpenShiftReconciler) SetupWithManager(mgr ctrl.Manager) error
 		For(&kataconfigurationv1.KataConfig{}).
 		Complete(r)
 }
+
```
jensfr, line 746: this is exactly what we have on the release-4.6 branch, right?

jensfr (commented):
This looks good to me. Thanks Harshal!

api/v1/kataconfig_types.go
harche: The validation now checks if it's nullable or not.

config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
harche: Autogenerated :)

controllers/openshift_controller.go
harche: yup.

PR #42, opened 2020-12-14
increase memory limits and requests for manager container

jensfr:
On OCP 4.7 clusters we have repeatedly seen the operator pod run into
OOM errors. The manager container is usually using around 40M and the limit was
30M.

**- What I did**
 This patch increases the memory limit for the manager container in
the pod to 60M and the request value to 45M.

**- How to verify it**
Install Kata with the operator and uninstall, let it run for half an hour. There should be no restarts
of the operator pod and no OOM errors.

Increase memory limit and requests values for the manager container to limit 60M and request 45M

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -49,8 +49,8 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 30Mi
+            memory: 50Mi
           requests:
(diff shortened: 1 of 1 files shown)


harche (approved):
/lgtm

jensfr:
Thank you for the review Harshal!

PR #43, opened 2020-12-15
forward-port from release-4.6: add support for master/worker combined nodes

jensfr:

In some compact clusters variants nodes have both the master and worker
role.  However, the machine config pools that are used in these
configurations have all nodes in the master pool and zero nodes in the
worker pool.

**- What I did**

Zero nodes in the worker pool breaks our current practice of installing
kata on all nodes in the worker pool. The solution is to install on all
nodes in the master pool instead. For that we fetch the worker pool
information and check if it has nodes in it. If it has a machine count
of zero we use the master pool instead.

This has an important implication: in this master-only configuration we
can't support installing on a subset of nodes by setting a label on the
selected nodes the user wants to install kata on. Why? This feature
depends on creating a custom pool that inherits from the worker pool.
MCP only supports inheriting from the worker pool but not from the
master pool. See
https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md

This is a trade-off for being able to support these compact cluster
types.

**- How to verify it**

On a cluster with 3 (actually the number doesn't matter) combined master/worker nodes run the operator, create/delete a kataconfig CR to install and uninstall. It should complete without errors.

Add support for clusters where nodes have both the master and worker role

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -253,7 +253,7 @@ func (r *KataConfigOpenShiftReconciler) newMCPforCR() *mcfgv1.MachineConfigPool 
 	return mcp
 }
 
-func (r *KataConfigOpenShiftReconciler) newMCForCR() (*mcfgv1.MachineConfig, error) {
+func (r *KataConfigOpenShiftReconciler) newMCForCR(machinePool string) (*mcfgv1.MachineConfig, error) {
 	isenabled := true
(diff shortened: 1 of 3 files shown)


harche (commented):
Would it be possible to have a test case supporting this change?

harche (approved):
/lgtm

PR #45, opened 2020-12-21
add kata-operator-payload dockerfiles

jensfr:
It builds a container image with the kata-* and QEMU (plus
dependencies) RPMs in it. The kata-operator-daemon when running
on the worker nodes downloads this image and then installs
the RPMs it contains.

**- What I did**
This is a 1:1 copy from the release-4.6 branch
(commit 4765c70)

**- How to verify it**
The dockerfile can be used to build the container with podman build.
The build should finish succesfully. The payload image can be tested by
using the payload configmap to make the operator use it.

### DIFF
--- a/images/payload/README.md
+++ b/images/payload/README.md
@@ -1,0 +1,5 @@
+# kata-operator-payload
+
+A container image used by kata-operator to install Kata and dependencies (like QEMU...) on worker nodes.
+During podman build the rpms are downloaded and a local repo is created. The resulting container image includes the rpms.
+The kata-operat-daemon pulls down this image unpacks it and installs it on the node where it is running. 
--- a/images/payload/README.md
+++ b/images/payload/README.md
@@ -1,5 +1,5 @@
 # kata-operator-payload
 
 A container image used by kata-operator to install Kata and dependencies (like QEMU...) on worker nodes.
-During podman build the rpms are downloaded and a local repo is created. The resulting container image includes the rpms.
+During container build (e.g., podman, docker, etc), the RPMs are downloaded and a local repo is created. The resulting container image includes the RPMs.
 The kata-operat-daemon pulls down this image unpacks it and installs it on the node where it is running. 
--- a/images/payload/hack/deploy.sh
+++ b/images/payload/hack/deploy.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -xe
+
+IMAGE_NAME="quay.io/isolatedcontainers/kata-operator-payload"
+
+if "${TRAVIS_BRANCH}" == "master"; then
+	TAG="latest"
+else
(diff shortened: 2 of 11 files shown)


bpradipt:
@jensfr looks like the related release-4.6 commit is `4765c70`

jensfr:
@bpradipt @zanetworker could you please ack this?

images/payload/hack/prepare.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
zanetworker, line 1: Where is `hack/prepare` used?

images/payload/hack/prepare.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -xe
+
+# Install the latest version of Docker
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
zanetworker, line 6: Here you are assuming that the build will happen on linux, can you mention it somewhere?

images/payload/README.md
```diff
@@ -0,0 +1,5 @@
+# kata-operator-payload
+
+A container image used by kata-operator to install Kata and dependencies (like QEMU...) on worker nodes.
+During podman build the rpms are downloaded and a local repo is created. The resulting container image includes the rpms.
+The kata-operat-daemon pulls down this image unpacks it and installs it on the node where it is running.
```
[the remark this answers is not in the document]
zanetworker: Also, I still see instances where `kata-install-daemon` is used, is this intended?

zanetworker (commented):
I see you added a README which is awesome, what I saw missing though is how / where would one use the scripts introduced in this PR (e.g., prepare.sh, cleanup.sh, ...).

images/payload/hack/prepare.sh
lysliu: Where does `prepare.sh` referenced?

images/payload/hack/deploy.sh
```diff
@@ -0,0 +1,22 @@
...
+if "${TRAVIS_BRANCH}" == "master"; then
+	TAG="latest"
+else
+	TAG="$(echo ${TRAVIS_BRANCH} | awk '{print $2}' FS='-')"
+fi
+
+docker login \
+	--username="${QUAY_USERNAME}" \
+	--password="${QUAY_PASSWORD}" \
+	quay.io
```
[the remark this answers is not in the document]
jensfr: No, I changed it to REGISTRY_USERNAME and _PASSWORD. Thank you!

images/payload/hack/prepare.sh
jensfr: I'm not sure. @fidencio ? Is prepare.sh called manually? Or is it somehow used by Travis?

jensfr:
This is not required anymore because we are not using the payload image any more.

PR #48, opened 2021-01-11
controllers: Prevent host devices from being passed to privileged containers

fidencio:
`privileged_without_host_devices` is an option from CRI-O configuration
files that controls whether the runtime handler will prevent host
devices from being passed to privileged containers.  Its default value
is `false` and we're switching it to `true` in order to avoid exposing
the host's `/dev`.

**- What I did**
Enforced `privileged_without host_devices = true` for our installation.

**- How to verify it**
When using `privileged: true`, check the content of `/dev` in the container.

Prevent host devices from being passed to privileged containers

**- Notes**
@bpradipt, @dgibson, I'm not exactly sure how it'd break (or not) SR-IOV and would be really nice if you guys could give it a try.  You don't need the operator or anything like that, just changing the CRI-O configuration file in your node (take a look at the changes provided here) and restarting CRI-O should be enough.

@jensfr, @zanetworker, we can't blindly merged this and we may need to find a solution based on different use cases. Anyhow, I'm opening the PR so we have a place to discuss, to figure out the limitations, and to gather ideas about how to proceed with this.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -345,6 +345,7 @@ func generateDropinConfig(handlerName string) (string, error) {
   runtime_path = "/usr/bin/containerd-shim-kata-v2"
   runtime_type = "vm"
   runtime_root = "/run/vc"
+  privileged_without_host_devices = true
   
 [crio.runtime.runtimes.runc]
   runtime_path = ""


controllers/openshift_controller.go
```diff
@@ -345,6 +345,7 @@ func generateDropinConfig(handlerName string) (string, error) {
   runtime_path = "/usr/bin/containerd-shim-kata-v2"
   runtime_type = "vm"
   runtime_root = "/run/vc"
+  privileged_without_host_devices = true
```
zanetworker, line 348: We might want to make this configurable on the `KataConfig` level and default this to `true` by default. WDYT?

fidencio: @jensfr :-)

jensfr: I guess the use case for privileged_without_host_devices=false will be SR-IOV and possibly other devices down the road. If we need it for those use cases I would make it configurable.

zanetworker: We will probably need it for those.  @bpradipt @dgibson WDYT?

bpradipt: Hmm, for SRIOV I don't think this change will have an impact. However I would like to test it once. I'll update this thread.

snir911:
Note that currently tying to run a privileged kata pod on Openshift with the default **privileged_without_host_devices=false** is failing due to:
` kubelet Error: CreateContainer failed: Timeout reached after 10s waiting for device 0:0:0:0/block: unknown`

fidencio:
ping?

jensfr:
@bpradipt did you have a chance to test this?

bpradipt:
@fidencio @jensfr didn't get chance to test yet. I finally got the SRIOV setup in a good state and hopefully should be able to update by tomorrow

bpradipt:
@fidencio verified successfully that this change is not having a negative impact to Kata SRIOV support with OpenShift. Was able to create privileged Kata POD with SRIOV (vfio) device without any issue. 
The setup is still available for any additional verification if required.

controllers/openshift_controller.go
jensfr: lgtm, we can make it configurable in Kataconfig later on. I'll create an issue for it

jensfr (approved):
thanks Fabiano!

bpradipt (approved):
/lgtm

controllers/openshift_controller.go
bpradipt: yeah we can make it configurable later on.

PR #51, opened 2021-01-19
Daemon authenticate for payload image download

jensfr:
 When a payload image is stored in a private repository the daemon
    needs to authenticate with the registry to be able to download it.

**- What I did**
Two sets of changes, one in the controller the other one in the daemon.
controller:

>When using a private repository for the payload image the
>daemon can now use the authentification credentials from a
>Kubernetes secret that is provided by the user.
>     
>This patch adds two environment variables for username and password
>to the daemon pod specification. The values are taken from a
>secret that the user needs to create before the pods are created.
>The environment variables are declared as optional so that no error
>occurs without the secret.
>

daemon:

>When a payload image is stored in a private repository the daemon
>needs to authenticate with the registry to be able to download it.
>     
>There are two environment variables defined in the daemons pod
>specification. These variables are populated by a Kubernetes secret that
>the user can create. It needs to be created before the daemons start to
>run.
>The variables and the secret are optional and will only be used when
>a configmap for a custom payload image is used. See the documentation
>for instructions for how to use the configmap and the secret for custom
>payload containers.

**- How to verify it**
See description in docs/DEVELOPMENT.md

Allow using payload containers in private repositories by providing credentials with a secret.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -127,6 +127,7 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForCR(operation DaemonOp
 		runPrivileged           = true
 		configmapOptional       = true
 		runAsUser         int64 = 0
+		trueValue               = true
 	)
 
 	dsName := "kata-operator-daemon-" + string(operation)
@@ -187,6 +188,30 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForCR(operation DaemonOp
 								},
 							},
 							Env: []corev1.EnvVar{
+								{
+									Name: "PAYLOAD_REGISTRY_USERNAME",
+									ValueFrom: &corev1.EnvVarSource{
+										SecretKeyRef: &corev1.SecretKeySelector{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: "payload-secret",
+											},
+											Key:      "username",
+											Optional: &trueValue,
+										},
+									},
+								},
+								{
+									Name: "PAYLOAD_REGISTRY_PASSWORD",
+									ValueFrom: &corev1.EnvVarSource{
+										SecretKeyRef: &corev1.SecretKeySelector{
+											LocalObjectReference: corev1.LocalObjectReference{
+												Name: "payload-secret",
+											},
+											Key:      "password",
+											Optional: &trueValue,
+										},
+									},
+								},
 								{
 									Name: "KATA_PAYLOAD_IMAGE",
 									ValueFrom: &corev1.EnvVarSource{
--- a/images/daemon/pkg/daemon/kata_openshift.go
+++ b/images/daemon/pkg/daemon/kata_openshift.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containers/image/v5/copy"
 	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/transports/alltransports"
+	"github.com/containers/image/v5/types"
 	"github.com/coreos/go-semver/semver"
 	"github.com/opencontainers/image-tools/image"
 	confv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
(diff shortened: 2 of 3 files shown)


images/daemon/pkg/daemon/kata_openshift.go
```diff
@@ -354,6 +355,11 @@ func uninstallRPMs(k *KataOpenShift) error {
 }
 
 func installRPMs(k *KataOpenShift) error {
+	var (
+		sourceCtx 	*types.SystemContext
+		payloadImage 	string
+	)
```
zanetworker, line 361: Since these are not used till line 389, maybe you could skip the definition here, and define them where they are used?

images/daemon/pkg/daemon/kata_openshift.go
```diff
@@ -400,7 +417,18 @@ func installRPMs(k *KataOpenShift) error {
 		return err
 	}
 
-	_, err = copy.Image(context.Background(), policyContext, destRef, srcRef, &copy.Options{})
+	_, err = copy.Image(context.Background(), policyContext, destRef, srcRef,
+		&copy.Options{SourceCtx: sourceCtx})
+
+	if err != nil {
+		fmt.Println("Error occured when downloading payload image:")
+		fmt.Println(err)
```
zanetworker, line 425: Just for consistency with what's above. 

```suggestion
		log.Printf("Error occured when downloading payload image: %v", err)
```

images/daemon/pkg/daemon/kata_openshift.go
```diff
@@ -400,7 +417,18 @@ func installRPMs(k *KataOpenShift) error {
...
+		if os.Getenv("PAYLOAD_SECRET_USERNAME_ENV_VAR") != "" {
+			fmt.Println("payload secret env vars are set and used. Please check the credentials used?")
```
zanetworker, line 427:
```suggestion
			log.Println("payload secret env vars are set and used. Please check the credentials used?")
```

images/daemon/pkg/daemon/kata_openshift.go
jensfr: I'm used to defining variables at the beginning of a function. Is it common go coding style to do it differently?

zanetworker: My comment was about the location but rather the duplication. For example, you are redefining the variable with ":=" [here](https://github.com/openshift/kata-operator/pull/51/files#diff-2c9b50b98430081b2bae1f8eed978d7b21c1d24e1c30cd754063c209a7a5c405R390) again.

jensfr: now I get it :-) yes that makes sense then. Thanks!

controllers/openshift_controller.go
```diff
@@ -187,6 +188,30 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForCR(operation DaemonOp
 								},
 							},
 							Env: []corev1.EnvVar{
+								{
+									Name: "PAYLOAD_SECRET_USERNAME_ENV_VAR",
```
bpradipt, line 192: Is `PAYLOAD_REGSITRY_USERNAME` and `PAYLOAD_REGSITRY_PASSWORD` a better name ?

images/daemon/pkg/daemon/kata_openshift.go
bpradipt: > I'm used to defining variables at the beginning of a function. Is it common go coding style to do it differently?

I'm not sure about it.  However in this func since you are using the two vars much later, it might make sense to use the short var declaration approach for better readability. Wdyt ?

controllers/openshift_controller.go
jensfr: Yes, it's shorter and precise. I'll change it. Thanks!

images/daemon/pkg/daemon/kata_openshift.go
jensfr: got rid of the var block and declared the variables directly with :=

bpradipt (approved):
/lgtm

PR #55, opened 2021-01-28
doc: re-format secret yaml

jensfr:
Fix formatting of an example secret definition in developers documentation


jensfr:
already merged pr #59 with same content, closing this one.

PR #58, opened 2021-02-02
make sure images are pushed as schema v2

jensfr:
On the second push to a repository in quay.io all images are converted
to schema v1. 

**- What I did**
To work around this bug we build the container image
with --no-cache --squash-all

**- How to verify it**
Run 'make docker-build IMG=' and push it. Then push it again. Inspect the image
with 'skopeo inspect' and verify it is still in schema v2 format.

Make sure container image is stored in schema v2 format.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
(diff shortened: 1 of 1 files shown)


pmores (approved):
/lgtm

PR #62, opened 2021-02-08
Add additional unit tests for KataConfig CR

bpradipt:
This patch series adds additional unit tests to increase the overall coverage from 10% to 10.4%

**- What I did**
Added additional unit tests for openshift_controller

**- How to verify it**
Running `make tests` 

Add additional unit tests for KataConfig CR

### DIFF
--- a/controllers/openshift_controller_test.go
+++ b/controllers/openshift_controller_test.go
@@ -57,6 +57,31 @@ var _ = Describe("OpenShift KataConfig Controller", func() {
 
 			By("Creating and marking the second KataConfig CR correctly")
 			Expect(kataconfig2.Status.InstallationStatus.Failed.FailedNodesCount).Should(Equal(-1))
+
+			// Delete
+			By("Deleting KataConfig CR successfully")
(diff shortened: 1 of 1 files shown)


zanetworker:
@pradeepitm12 these are more like smoke/integration tests as they are running against a real cluster right?. Do we have something we could run just to validate and increase coverage (e.g., mocks)?

bpradipt:
> @pradeepitm12 these are more like smoke/integration tests as they are running against a real cluster right?. Do we have something we could run just to validate and increase coverage (e.g., mocks)?

@zanetworker you are right. The tests are against a controller that gets created. Not really a real cluster but something in-between a mock and real cluster. I followed the same pattern that is already existing in the code. 

@jensfr any comments here ?

jensfr (approved):
Thank you @bpradipt! It looks good to me. 

I noticed when running the test against a real cluster it leaves two kataconfigs in place. Which is odd for two reasons, 1. it shouldn't be possible to create a second kataconfig and 2. it should be cleaned up after the tests ran.

But this problem existed before your PR and should be fixed indepently.

Issue #23, opened 2020-10-30
Migration to Ignition spec 3


travier:
The [openshift controller](https://github.com/openshift/kata-operator/blob/master/controllers/openshift_controller.go) is currently using Ignition spec 2
https://github.com/openshift/kata-operator/blob/d36a9ddde10e31e9dad2cb7b3b803cc79b7f3b8b/controllers/openshift_controller.go#L28
https://github.com/openshift/kata-operator/blob/d36a9ddde10e31e9dad2cb7b3b803cc79b7f3b8b/go.mod#L7
which is still supported on 4.6 but not under development anymore. Ignition spec 3 support is recommended from 4.6 onward.

Let me know if you need help to convert the code to spec 3. Note that Ignition spec 3 support is only available since 4.6.

travier:
As far as I gathered from a quick look at the code, this is still a valid RFE. Feel free to reach out if you need help for the migration.
/remove-lifecycle stale

travier:
Closing following comments in https://github.com/openshift/sandboxed-containers-operator/pull/64

PR #64, opened 2021-02-12
controller: update ignition version

jensfr:
The controller is still using ignition v2 which is deprecated.

**- What I did**
Switch to ignition v2.9.0 and spec 3.2. This required minor changes as
data structures in ignition have changed.

**- How to verify it**
Install the operator and make sure the machine config is created and places the
crio config file correctly on the worker nodes.

use ignition v3 instead of deprecated v2

Fixes: #23

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -25,7 +25,7 @@ import (
 	"text/template"
 	"time"
 
-	ignTypes "github.com/coreos/ignition/config/v2_2/types"
+	ignTypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/go-logr/logr"
 	kataconfigurationv1 "github.com/openshift/kata-operator/api/v1"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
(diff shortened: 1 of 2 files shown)


controllers/openshift_controller.go
```diff
@@ -307,27 +307,27 @@ WantedBy=multi-user.target
...
+	c := file.Contents
 
 	dropinConf, err := generateDropinConfig(r.kataConfig.Status.RuntimeClass)
 	if err != nil {
 		return nil, err
 	}
 
-	c.Source = "data:text/plain;charset=utf-8;base64," + dropinConf
+	dropinFile := "data:text/plain;charset=utf-8;base64," + dropinConf
+	c.Source = &dropinFile
```
bpradipt, line 318: can we simply use `file.Contents.Source = &dropinFile` and avoid using `c` ?

jensfr: yes, we can! Thanks @bpradipt !

travier:
If this will be backported to 4.6, please note that 4.6 only supports Ignition 3.1.

travier:
Looks fine otherwise!

jensfr:
we have made this change as part of the "enable sandboxed-containers extension" PR which was already merged and is part of the release-4.8 and master branch. Because of this I'm closing this PR.

PR #76, opened 2021-03-17
Rename operator to sandboxed-containers-operator

bpradipt:
Rename kata-operator to sandboxed-containers-operator
**- What I did**
Renamed kata-operator to sandboxed-containers-operator and bumped up the version to 4.8
Renamed the namespace to sandboxed-containers-operator-system
**- How to verify it**
Deploy the operator by following the instructions

Rename kata-operator to sandboxed-containers-operator

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ jobs:
 
   # Build and push the image for amd64
   - stage: deploy
-    name: Build image for kata-operator-daemon (amd64)
+    name: Build image for sandboxed-containers-operator-daemon (amd64)
     arch: amd64
     deploy:
       provider: script
@@ -58,7 +58,7 @@ jobs:
 
   # Build and push the image for ppc64le
   - stage: deploy
-    name: Build image for kata-operator-daemon (ppc64le)
+    name: Build image for sandboxed-containers-operator-daemon (ppc64le)
     arch: ppc64le
     deploy:
       provider: script
@@ -70,7 +70,7 @@ jobs:
 
   # Build and push the image for s390x
   - stage: deploy
-    name: Build image for kata-operator-daemon (s390x)
+    name: Build image for sandboxed-containers-operator-daemon (s390x)
     arch: s390x
     deploy:
       provider: script
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,6 +1,6 @@
 <!--
 Before reporting an issue, check our backlog of
-[open issues](https://github.com/openshift/kata-operator/issues)
+[open issues](https://github.com/openshift/sandboxed-containers-operator/issues)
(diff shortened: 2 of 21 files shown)


bpradipt:
tested on OCP 4.7 and OCP 4.8 but will be good to do few additional tests to rule out issues

jensfr:
It looks good to me. I'm running a few tests and will approve. 

Looking at this reminded me of the question, should we also rename the CRD? It's currently called kataconfig. Should it be scconfig or something like that? Would be good to change it early before we have to support upgrades. @zanetworker ? In a new PR of course.

zanetworker:
> Would be good to change it early before we have to support upgrades. @zanetworker ? In a new PR of course.

I think this one is debatable, and I have no strong opinion on this. On the one hand, this is a resource we expose, on the other hand,  it is descriptive of the runtime we are using. 


The questions are if we do support other runtimes, do we want to have a `SandboxedContainersConfig` and have `kata` as a field in the spec, for example: 

```yaml
apiVersion: ....
kind: SandboxedOperatorConfig
spec: 
   runtimeRef: 
       apiVersion: kata.sandboxedContainers.io
       kind: KataConfig
       name: kata-test 
   ....
```

or do we want to have different configs i.e., RuntimeOneConfig, RuntimeTwoConfig,... I tend to be in favor of the first but I have no strong opinion. @jensfr and @bpradipt WDYT?

jensfr (approved):
lgtm, thanks Pradipta!

jensfr:
> or do we want to have different configs i.e., RuntimeOneConfig, RuntimeTwoConfig,... I tend to be in favor of the first but I have no strong opinion. @jensfr and @bpradipt WDYT?

I like your idea with an umbrella sandboxedcontainersconfig that contains a sub for every runtime we support, but I'll have to play with it a bit to form an opinion. Thank you!

bpradipt:
> > Would be good to change it early before we have to support upgrades. @zanetworker ? In a new PR of course.
> 
> I think this one is debatable, and I have no strong opinion on this. On the one hand, this is a resource we expose, on the other hand, it is descriptive of the runtime we are using.
> 
> The questions are if we do support other runtimes, do we want to have a `SandboxedContainersConfig` and have `kata` as a field in the spec, for example:
> 
> ```yaml
> apiVersion: sandboxedContainers.io/v1alpha1
> kind: SandboxedContainersConfig
> spec: 
>    runtimeRef: 
>        apiVersion: kata.sandboxedContainers.io/v1alpha1
>        kind: KataConfig
>        name: kata-test 
>    ....
> ```
> 
> or do we want to have different configs i.e., RuntimeOneConfig, RuntimeTwoConfig,... I tend to be in favor of the first but I have no strong opinion. @jensfr and @bpradipt WDYT?

+1 for SandboxedContainersConfig .

PR #79, opened 2021-03-22
Makefile: `make docker-push` should use podman

fidencio:
Although the subject line may be confused, we should aim for fully using
podman or docker, consistently. :-)

In order to do so, let's go for "docker" rather than "podman" and do the
switch of `make docker-push` to do a `podman push` rather than `docker
push`.

**- What I did**

**- How to verify it**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,10 @@ docker-build: test
 docker-push:
 	docker push ${IMG}
 
+# Build the docker image, using podman
+podman-build: test
+	podman build --squash-all --no-cache . -t ${IMG}
+
 # find or download controller-gen
 # download controller-gen if necessary
 controller-gen:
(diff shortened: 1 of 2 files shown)


bpradipt (approved):
/lgtm
makes sense to keep it consistent

zanetworker:
IMO this is a bit unintuitive,  an alternative could be to add two extra steps (more or less if required) for podman: 

```
make docker-build -> does docker build 
make docker-push -> does docker push 
make podman-build -> does podman build 
make podman-push -> does podman push
```

fidencio:
I've re-written the series.

Now, `make docker-build` is probably broken (and I have **zero** time to test it).

jensfr (approved):
For the record, I'd be fine with droppin docker-build and -push entirely. Also the reason I added podman at all was because we need --squash-all and --no-cache to work around a bug in quay.io that turns images into schema v1 images at the second push. 

This change is fine. We don't test with docker and could either remove it in the future or once the above bug is fixed count on users using podman-docker.

PR #80, opened 2021-03-22
controller: The service should be enabled by the OS

fidencio:

**- What I did**

**- How to verify it**


fidencio:
The needed bits were merged as part of openshift OS (https://github.com/openshift/os/pull/524) and are part of the kata-containers-2.1.0-1.alpha2.el8 RPM.

fidencio:
Ah, by the way, this should be tested in an altogether build of the operator + the latest extensions containing the RPM mentioned in the previous comment.  Only after the test, this one should be merged,

jensfr (approved):
/lgtm

jensfr (approved):
/lgtm

PR #81, opened 2021-03-22
controller: The CRI-O drop-in file should be provided by the RPM

fidencio:
Instead of having the logic of the CRI-O drop-in file in the operator,
let's just rely on it being distributed by the `kata-containers` RPM.

This is done because that file only makes sense in a system where
`kata-containers` RPM is installed, because that file is only used in a
system where `kata-containers` RPM is installed.  So, leave it for the
RPM to take care of that and let's make our lives simpler on the
operator side.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -17,12 +17,9 @@ limitations under the License.
 package controllers
 
 import (
-	"bytes"
 	"context"
-	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
-	"html/template"
 	"k8s.io/apimachinery/pkg/labels"
 	"time"
 
@@ -158,28 +155,11 @@ func (r *KataConfigOpenShiftReconciler) newMCForCR(machinePool string) (*mcfgv1.
 		r.Log.Error(err, "no valid role for mc found")
 	}
 
-	file := ignTypes.File{}
-	c := file.Contents
-
-	dropinConf, err := generateDropinConfig(r.kataConfig.Status.RuntimeClass)
-	if err != nil {
-		return nil, err
-	}
-
-	dropinFile := "data:text/plain;charset=utf-8;base64," + dropinConf
-	c.Source = &dropinFile
-
-	file.Contents = c
-	m := 420
-	file.Mode = &m
-	file.Path = "/etc/crio/crio.conf.d/50-kata.conf"
-
 	ic := ignTypes.Config{
 		Ignition: ignTypes.Ignition{
 			Version: "3.2.0",
 		},
 	}
-	ic.Storage.Files = []ignTypes.File{file}
 
 	icb, err := json.Marshal(ic)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
/lgtm
This looks much simpler. 
Thanks @fidencio

zanetworker:
IMO this is a bit unintuitive,  an alternative could be to add two extra steps (more or less if required) for podman: 

```
make docker-build -> does docker build 
make docker-push -> does docker push 
make podman-build -> does podman build 
make podman-push -> does podman push
```

zanetworker:
@fidencio This is simpler indeed, my only concern is that it makes it harder to control CRI-O config for Kata without publishing an RPM that includes that change. Good thing we have independent versioning in-place but still that does take away some of the flexibility we had with this in place (which we could have potentially exposed as fields in the CRD).

fidencio:
@zanetworker, I sincerely fail to understand how this change would limit us in the future, in case we want to do something in that field.

We have a drop-in file coming from kata-containers RPM and we:
1. could re-access that based on the ideas, when those are implemented (we **own** the release cadence of the RPM);
2. could still drop-in a file using the operator, which would have a higher priority and take the precedence, as mentioned in https://github.com/cri-o/cri-o/commit/b74ec1c3c6d54c4dc392d3db32c4e134d75fbb06:
"""
    Add support for crio drop-in config files

    We now support partial drop-in configuration files which will be
    assembled together on CRI-O startup.

    A default (or user specified via the CLI) directory will be recursively
    iterated, whereas each file gets applied to the configuration in their
    processing order. This means that a configuration file named
    `/etc/crio/crio.conf.d/00-default` has a lower priority than a file
    `/etc/crio/crio.conf.d/01-my-config`. The file provided via
    `--config,-c` always has a higher priority than those inside the
    `--config-dir,-d`, whereas the command line parameters always have the
    highest configuration priority at all.
"""

So, right now, what we're doing is providing a default configuration via the RPM and just by doing that, we leave the complexity of a conciliation (which, by the way, has caused issues like https://issues.redhat.com/browse/KATA-553) for the moment when we actually need that.  Right now, we don't need that and I sincerely don't think we should be adding that complication to the equation.

Again, nothing changes with regarding adding value to the operator, adding configuration options, or anything related to that in the future.

Does this sound reasonable?
If not, I'm super fine on dropping the PR entirely.

zanetworker:
@fidencio thank you. 

> could still drop-in a file using the operator, which would have a higher priority and take the precedence, as mentioned in cri-o/cri-o@b74ec1c:

That addressed my concern.

jensfr (approved):
lgtm, thank you Fabiano!

c3d (approved):
Looks like a nice simplification. Thanks @fidencio . Always nice to remove the word "Hacky" from the code anyway.

fidencio:
I've rebased the code atop of git master and this PR includes the content of #80 (which should be merged first).  The needed changes on the RPM side are present kata-containers-2.1.0-1.alpha2.el8 and should be tested altogether before having this one merged.

fidencio:
I had to manually remove the "needs-rebase" label.

PR #82, opened 2021-03-24
add preinstallation checking whether node can even run kata

pmores:
Aims to fix a problem where kata daemon will happily install kata even on nodes that fundamentally cannot run kata.

### DIFF
--- a/images/daemon/pkg/daemon/kata_check_node.go
+++ b/images/daemon/pkg/daemon/kata_check_node.go
@@ -1,0 +1,180 @@
+
+package daemon
+
+import (
+    "io/ioutil"
+    "strings"
+    "regexp"
+    "fmt"
+    "path/filepath"
+    "os"
+    "os/exec"
+)
+
+const (
+	procCPUInfo  = "/proc/cpuinfo"
+	sysModuleDir = "/sys/module"
+	moduleParamDir = "parameters"
+	modProbeCmd  = "modprobe"
+
+    cpuFlagsTag    = "flags"      // nolint: varcheck, unused, deadcode
+    cpuVendorField = "vendor_id"  // nolint: varcheck, unused, deadcode
+    cpuModelField  = "model name" // nolint: varcheck, unused, deadcode
+)
+
+var (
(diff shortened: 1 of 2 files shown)


images/daemon/pkg/daemon/kata_check_node.go
```diff
@@ -0,0 +1,180 @@
...
+        return true
+    }
+
+    // try loading the module (as kata-daemon we're root by definition
+    // so permissions shouldn't be a concern)
+    cmd := exec.Command("modprobe", moduleName)
+    if output, err := cmd.CombinedOutput(); err != nil {
+        fmt.Printf("'modprobe %s' failed: %s", moduleName, output)
+        return false
+    }
```
fidencio, line 82: To me, this is quite an intrusive thing to be done as part of the check.
Consider we can load the module, how do we ensure the module will loaded next time we boot the node up?

Does the operator do something to ensure the module is up and will ever be while we have kata installed? If not, I'd just error out and say, "hey, sorry, no can do."

fidencio (commented):
I have left one comment and, please, take my review with a grain of salt as I'm not familiar with the operator code base. :-)

bpradipt:
@pmores it will be good to explore usage of node feature discovery for this requirement.
Details
https://kubernetes-sigs.github.io/node-feature-discovery/stable/get-started/index.html
https://kubernetes-sigs.github.io/node-feature-discovery/v0.8/get-started/features.html#cpu
https://github.com/openshift/node-feature-discovery

images/daemon/pkg/daemon/kata_check_node.go
pmores: @fidencio This is what upstream kata-check does.  I'm pointing this out just as a rationale of this PR's current form, I don't have a strong opinion of this and I'm open to doing it either way.

jensfr:
We don't have a daemon any more and we (Pavel, I and others) are exploring other options to do this kind of check without it. For example the node feature discovery operator. So I'm closing this one for now , thank you for your work Pavel!

PR #83, opened 2021-03-25
added pre-req for kataconfig deletion

pruan-rht:
The current uninstall kata-runtime is missing the requirement that all existing pod in the cluster that use kata must be removed.

**- What I did**
added instruction about the pods removal requirement

The current uninstall kata-runtime is missing the requirement that all existing pod in the cluster that use kata must be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,10 @@ oc apply -f config/samples/example-fedora.yaml
 ## Uninstall
 
 ### Openshift
+Prior to deletion of your kataconfig, it is expected that all existing pods that use kata runtime have been removed.
+The `oc delete` operation will fail if there are pods in the cluster that are still referencing the kata runtime.
+
+
 ```
 oc delete kataconfig <KataConfig_CR_Name>
 ```


README.md
```diff
@@ -93,6 +93,10 @@ oc apply -f config/samples/example-fedora.yaml
 ## Uninstall
 
 ### Openshift
+Prior to deletion of your kataconfig, it is expected that all existing pods that use kata runtime have been removed.
```
c3d, line 96: I would write `kataconfig` (with backticks).

c3d (approved):
Thanks! Good addition. A minor nit about formatting.

README.md
```diff
@@ -93,6 +93,10 @@ oc apply -f config/samples/example-fedora.yaml
...
+The `oc delete` operation will fail if there are pods in the cluster that are still referencing the kata runtime.
```
jensfr, line 97: But it won't fail in the sense that the command is aborted, it will sit there waiting for the pods to be deleted. An error message will only show up in the log. Maybe it's worth making that more clear in the README until we fix it? A fix could be to put the error into the kataconfig status, but that's for another PR.

jensfr: @pruan-rht could you change that last sentence to say something like "The 'oc delete' operatorion will block if there are pods in the cluster that are still using the kata runtime until those pods have been deleted"

c3d:
@jensfr What do we do with this one?

jensfr:
@c3d It definitely shouldn't rot away. We need a fix like this, but it needs to be slightly changed to match what actually happens, i.e. it shows an error in the kataconfig description (not just in the logs like before). And it should say that the 'oc delete' command blocks until all the pods have been stopped. I'll take care of this.

c3d:
@pruan-rht Can you address @jensfr's comments?

bpradipt:
@pruan-rht the README for this project is completely revamped and we recommend users to refer to the official OpenShift sandboxed containers docs. This ensures the doc is maintained in one place and provides upto date information.  For example the 4.10 doc mentions this prereq 
https://docs.openshift.com/container-platform/4.10/sandboxed_containers/uninstalling-sandboxed-containers.html#sandboxed-containers-deleting-operator-deployment-cr_uninstalling-sandboxed-containers

I'm closing this PR. If there is something specific that requires addition to the README please re-open this.

Issue #75, opened 2021-03-16
Adjust podOverhead values for kata-containers pods


fidencio:
We currently use the very same podOverhead values gotten from upstream, which is:
```
overhead:
  podFixed:
    cpu: 250m
    memory: 160Mi
```

This, unfortunately, is not accurate, and should be properly calculated for our use case, considering our restrictions (such as the kernel img we load).

Past measurements show we have a memory overhead of ~300Mi when spawning sleep pods.  This value should be used instead of the 160Mi currently present there.

About the CPU, we'd need some help from @RobertKrawitz to get to a more accurate number for our use cases.

PR #84, opened 2021-03-31
Pod Overhead: increase memory to 350Mi

fgiudici:
Current Pod overhead memory is set to 160Mi.
Creating a simple Pod with and without the kata runtimeClass will show an accounted memory difference of 300Mi in the OpenShift console. As the memory usage due to the VM operations is dynamic and may easily increase (e.g., for fs caching) a sane value for the Pod overhead memory should be a bit more than that:  let's set it to 350Mi.


bpradipt (approved):
/lgtm
Thanks @fgiudici

bpradipt:
@fgiudici can you add this issue in the commit message - https://github.com/openshift/sandboxed-containers-operator/issues/75

jensfr (approved):

PR #85, opened 2021-04-13
enable sandboxed-containers extension

jensfr:
Until now we used a separate image containing RPMs that were installed
by a daemon running on each node. This was error prone and complicated.
With this patch we can simplify the operator overall because we don't
need a daemon and payload image any more.

This patch changes the way Kata, QEMU and dependencies are installed in
the cluster. Instead of installing RPMs directly the operator now
enables a RHCOS extension using the Machine Config Operator. This is
achieved by creating a machine config object that enables the 'sandboxed
containers' extension.

This simplifies the installation of the runtime quite a lot but it also
brings some necessary changes with it. Before this patch the daemon was
updating the Kataconfig CR with status information about the
installation. This has to change now because there is no daemon any
more. Instead these status information is filled in by the controller.
The required information is gathered from the Machine Config Pool and
Node objects and is displayed in a very similar manner as before in the
description of the Kataconfig CR.

This brings no changes in how the operator itself is deployed and the
installation of the Kata runtime is triggered. By creating an instance
of the kataconfig CR the installation is triggered. The most important
part is the creation of the machine config object that enables the
sandboxed-containeres RHCOS extension. The controller then monitors the
progress by watching the data in the description of the machine config
pool. Once all nodes are updated with the new machine config state the
final step is to create the runtime class named 'kata'.

Uninstallation of the Kata runtime works in the same way. The machine
config entry is removed and this triggers MCO to disable the RHCOS
extension and eventually remove the kata runtime class.

Support for 3-node clusters stays the same as well as the possibility to
install the runtime on a chosen subset of the worker nodes by adding a
label to each node.

This patch also removes all files to build the daemon and payload image.


jensfr:
Once we have a working nightly build again I can test with the new kata rpm in the extension. Then we can also merge PR #80 and PR #81

controllers/openshift_controller.go
```diff
@@ -595,212 +390,162 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
-			} else if err != nil {
-				return ctrl.Result{}, err
-			}
+		if workreMcp.Status.ReadyMachineCount != workreMcp.Status.MachineCount {
+			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
+		}
+	} else {
+		if len(r.kataConfig.Status.UnInstallationStatus.InProgress.BinariesUnInstalledNodesList) > 0 {
+			r.Log.Info("Pausing for a minute to make sure parent mcp has started syncing up")
+			time.Sleep(60 * time.Second)
```
fidencio, line 462: Yet another sleep that worries me a little bit.

fidencio (commented):
@jensfr, there are some **minor** comments, mostly from the point of view of someone trying to understand the code (rather than a proper reviewer).

One thing that I missed (and I'm quite convinced it's there) is where we disable the sandboxed-containers extension during the uninstall.  Do you mind to point me to that piece of code? :-)

controllers/openshift_controller.go
```diff
@@ -595,212 +390,162 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
+		if !isMcDeleted {
+			err = r.Client.Delete(context.TODO(), mc)
+			if err != nil {
+				// error during removing mc, don't block the uninstall. Just log the error and move on.
+				r.Log.Info("Error found deleting machine config. If the machine config exists after installation it can be safely deleted manually.",
+					"mc", mc.Name, "error", err)
 			}
+			// Sleep for MCP to reflect the changes
+			r.Log.Info("Pausing for a minute to make sure worker mcp has started syncing up")
+			time.Sleep(60 * time.Second)
```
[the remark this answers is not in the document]
jensfr: It's a good pint and yes, I think it could be replaced with e.g. a Watch on the machine config pool. But since this has been used since a looong time and is not new in this PR I would prefer to do it in another PR. Do you agree?

controllers/openshift_controller.go
jensfr: yep, same as above. I think this can be done in a more k8s native way, but would rather not change it in this PR.

controllers/openshift_controller.go
```diff
@@ -89,161 +89,30 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(req ctrl.Request) (ctrl.Result
 		if !oldest && err != nil {
 			return reconcile.Result{Requeue: true}, err
 		} else if !oldest && err == nil {
+			r.Log.Info("not oldest CR")
```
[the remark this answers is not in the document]
jensfr: We only allow one instance of the kataconfig CR (what would be the point of two?). But this requirement is rather unusual because with most other application level operators you can have many, think of a database or something. 

"not oldest CR" means this reconcile loop is operating on a second CR. We consider the kataconfig that was created first to be the legit one. So any CR that is not the oldest would be not supported and we write a log to announce that and also set an error in the kataconfig status description.

I think Harshal made an attempt to solve it differently but it wasn't possible at the time. We can revisit for 4.9 if there's another way like a webhook or a singleton operator feature that was discussed in the past.

README.md
```diff
@@ -56,7 +55,7 @@ oc describe kataconfig example-kataconfig
 and look at the field 'Completed nodes' in the status. If the value matches the number of worker nodes the installation is completed.
 
 #### Runtime Class
-Once the kata runtime binaries are successfully installed on the intended workers, the sandboxed containers operator will create a [runtime class](https://kubernetes.io/docs/concepts/containers/runtime-class/) `kata`. This runtime class can be used to deploy the pods that will use the Kata Runtime.
+Once the sandboxed-containers extension is enabled successfully on the intended workers, the sandboxed containers operator will create a [runtime class](https://kubernetes.io/docs/concepts/containers/runtime-class/) `kata`. This runtime class can be used to deploy the pods that will use the Kata Runtime.
```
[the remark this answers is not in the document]
jensfr: fixed, thanks!

controllers/openshift_controller.go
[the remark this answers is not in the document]
fidencio: Ack.  Feel free to open an issue to track this one, if you think that's needed.  I'm fine with the code as it's now, it was just not so easy to understand without the background.

Thanks for the explanation.

controllers/openshift_controller.go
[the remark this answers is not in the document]
fidencio: I've opened https://github.com/openshift/sandboxed-containers-operator/issues/87 to track such cases, and I'm totally fine on having this as a different effort / different PR.

controllers/openshift_controller.go
fidencio: same answer as above. :-)

fidencio:
LGTM, @jensfr, thanks!

controllers/openshift_controller.go
```diff
@@ -595,212 +390,162 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
-						node.SetLabels(nodeLabels)
-						_, err = r.clientset.CoreV1().Nodes().Update(context.Background(), node, metav1.UpdateOptions{})
-
-						if err != nil {
-							return ctrl.Result{}, err
-						}
-					}
-				}
+		if !isMcDeleted {
+			err = r.Client.Delete(context.TODO(), mc)
```
jensfr, line 430: @fidencio this is where we delete the machine config

controllers/openshift_controller.go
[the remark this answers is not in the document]
pmores: From my - admittedly very limited - experience I think this is a common pattern in controller programming - if something turns out not to be ready just try again in x seconds.  Not as elegant or efficient as an async solution but dead simple, and since the wait periods tend to be in (tens of) seconds it might not be worth the added complexity to sweat the efficiency...

controllers/openshift_controller.go
```diff
@@ -300,34 +168,33 @@ WantedBy=multi-user.target
...
-	} else {
+	} else if _, ok := r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels["node-role.kubernetes.io/"+machinePool]; !ok {
 		r.Log.Error(err, "no valid role for mc found")
 	}
 
 	file := ignTypes.File{}
-	c := ignTypes.FileContents{}
+	c := file.Contents
 
 	dropinConf, err := generateDropinConfig(r.kataConfig.Status.RuntimeClass)
```
bpradipt, line 178: Do we still need the drop-in config given the config will come from kata rpm ?

controllers/openshift_controller.go
```diff
@@ -300,34 +168,33 @@ WantedBy=multi-user.target
...
 	if err != nil {
 		return nil, err
 	}
 
-	c.Source = "data:text/plain;charset=utf-8;base64," + dropinConf
+	dropinFile := "data:text/plain;charset=utf-8;base64," + dropinConf
+	c.Source = &dropinFile
+
```
bpradipt, line 185: We should be able to directly use `file.Contents.Source = &dropinFile`

controllers/openshift_controller.go
```diff
@@ -595,212 +388,162 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
+			// Sleep for MCP to reflect the changes
+			r.Log.Info("Pausing for a minute to make sure worker mcp has started syncing up")
+			time.Sleep(60 * time.Second)
 		}
 
-		r.Log.Info("Making sure parent MCP is synced properly, KataNodeRole=" + machinePool)
-		if _, ok := r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels["node-role.kubernetes.io/"+machinePool]; ok {
-			mc, err := r.newMCForCR(machinePool)
-			var isMcDeleted bool
+		workreMcp := &mcfgv1.MachineConfigPool{}
```
bpradipt, line 439: curious why it's `workreMcp` and not `workerMcp`?

controllers/openshift_controller.go
```diff
@@ -595,212 +388,162 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
-	foundDs := &appsv1.DaemonSet{}
-	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: ds.Name, Namespace: ds.Namespace}, foundDs)
-	if err != nil && errors.IsNotFound(err) {
-		// DaemonSet not found, nothing to delete, ignore the request.
-		return nil
-	} else if err != nil {
-		return err
+	r.kataConfig.Status.UnInstallationStatus.InProgress.IsInProgress = corev1.ConditionFalse
+	_, result, err2, done := r.updateStatus(machinePool)
+	r.clearInstallStatus()
```
bpradipt, line 499: r.clearUnInstallStatus() ?

controllers/openshift_controller.go
jensfr: no, but the rpm is not yet in a nightly build that passes the tests. So we can't test it in combination with the extensions. I would suggest we merge this first and when we can test the new RPM we will merge Fabianos PRs #80 and PR #81 which remove all of the dropin file related code

controllers/openshift_controller.go
jensfr: I think it's a typo we have never fixed. The patch is already huge and I tried to avoid fixing all the stuff I find along the way :-)

controllers/openshift_controller.go
jensfr: This is intentional to clear the status of the last installation now that the uninstallation is done as it is not relevant any more.

controllers/openshift_controller.go
jensfr: yes, I was going to fix that because I think you already suggested that in another PR. But then I realized that we will delete this code very soon anyway. As soon as the new kata-containers rpm is available that creates the dropin file.

jensfr:
Merging this as there were no comments that should block this and we have one lgtm. Thank you for the reviews @fidencio and @bpradipt

PR #88, opened 2021-04-19
inform user if kata uninstallation is blocked by existing kata-based …

pmores:
…pods

If user requests uninstallation of Kata by deleting their KataConfig
instance, the request cannot be fulfilled if there still are pods that use
Kata as runtime.  So far, the controller logged a message on detecting this
condition but since the message went to the controller log it wasn't very
well visible or discoverable.

This change creates a new field in KataConfig status to communicate the
error to the user more effectively.

**- What I did**

**- How to verify it**


jensfr (approved):
Thank you @pmores !

Code looks good. I tested it and it works as expected.

PR #91, opened 2021-04-23
[backport] controller: The service should ne enabled by the OS

jensfr:

**- What I did**

**- How to verify it**


fidencio (approved):
/lgtm

thanks for backporting this one, @jensfr!

PR #92, opened 2021-04-26
Clean up and renames

jensfr:
- use new namespace name and refer to new quay.io repository
    We should adhere to the naming convention in OpenShift and use
    the default namespace openshift-sandboxed-containers. The same goes for
    the quay.io user/repository we use. Use the new
    quay.io/openshift_sandboxed_containers instead of
    quay.io/isolatedcontainres.
    
- delete example configmap for payload. We don't have a payload any more.

**- What I did**

This cleans up README and example files to use the namespace
openshift-sandboxed-containers instead of sandboxed-containers-operator-system.

**- How to verify it**

   Follow the README descripton to install/uninstall and make sure all steps work.

 Use new namespace openshift-sandboxed-containers and refer to new quay.io organisation quay.io/openshift_sandboxed_containers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Current Operator version
-VERSION ?= 4.8.0
+VERSION ?= 1.0.0
 # Default bundle image tag
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
 # Options for 'bundle-build'
@@ -12,7 +12,7 @@ endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # Image URL to use all building/pushing image targets
-IMG ?= quay.io/isolatedcontainers/sandboxed-containers-operator:4.8.0
+IMG ?= quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:$(VERSION)
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
--- a/README.md
+++ b/README.md
@@ -16,15 +16,15 @@ An operator to perform lifecycle management (install/upgrade/uninstall) of [Kata
 
 
    ```
-   git clone https://github.com/openshift/sandboxed-containers-operator 
+   git clone https://github.com/openshift/sandboxed-containers-operator
    git checkout -b master --track origin/master
    ```
 3. Install the sandboxed containers operator on the cluster,
 
    ```
-   make install && make deploy IMG=quay.io/isolatedcontainers/sandboxed-containers-operator:4.8
+   make install && make deploy IMG=quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:latest
    ```
-4. To begin the installation of the kata runtime on the cluster,
+4. To begin the installation of the kata runtime on the cluster
 
    ```
    oc create -f config/samples/kataconfiguration_v1_kataconfig.yaml
@@ -105,7 +105,7 @@ oc delete kataconfig example-kataconfig
 ### Openshift
 1. During the installation you can watch the values of the kataconfig CR. Do `watch oc describe kataconfig example-kataconfig`.
 2. To check if the nodes in the machine config pool are going through a config update watch the machine config pool resource. For this do `watch oc get mcp kata-oc`
-3. Check the logs of the sandboxed containers operator controller pod to see detailled messages about what the steps it is executing. To find out the name of the controller pod, `oc get pods -n sandboxed-containers-operator-system | grep sandboxed-containers-operator-controller-manager` and then monitor the logs of the container `manager` in that pod. 
+3. Check the logs of the sandboxed containers operator controller pod to see detailled messages about what steps it is executing. To find out the name of the controller pod, `oc get pods -n openshift-sandboxed-containers-operator | grep controller-manager` and then monitor the logs of the container `manager` in that pod.
 
 ## Components
 
@@ -114,7 +114,7 @@ The sandboxed containers operator uses three containers:
 
 Container image name | Description | Container repository
 ---------------| ----------- | ----------
- _sandboxed-containers-operator_ |  It contains the controller part of the operator that watches and manages the kataconfig custom resource. It runs as a cluster scoped container. The operator itself is build with operator-sdk. | https://quay.io/isolatedcontainers/sandboxed-containers-operator
+ _openshift-sandboxed-containers-operator_ |  It contains the controller part of the operator that watches and manages the kataconfig custom resource. It runs as a cluster scoped container. The operator itself is build with operator-sdk. | https://quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator
 
 # Build from source
 
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-oc apply -f https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/master/deploy/deploy.yaml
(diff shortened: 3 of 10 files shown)


deploy/deploy.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-oc apply -f https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/master/deploy/deploy.yaml
+oc apply -f https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/release-4.8/deploy/deploy.sh
```
c3d, line 3: This part, with explicit `4.8`, is a bit inconsistent with `latest` above.

deploy/deploy.yaml
```diff
@@ -146,6 +146,8 @@ spec:
                         - name
                         type: object
                       type: array
+                    failedNodesReason:
```
c3d, line 149: This commit seems to have a mix of "pure rename" and "additional stuff". If it's not too much work, you may want to split the two using good old `git add -p` ;-)

deploy/deploy.yaml
```diff
@@ -159,11 +161,18 @@ spec:
                       description: InProgressNodesCount reflects the number of nodes
                         that are in the process of kata installation
                       type: integer
+                    isInProgress:
+                      description: IsInProgress reflects if installation is still
```
c3d, line 165: Just out of curiosity, why is the capitalisation consistently different between the name of the field (e.g. here `isInProgress`) and the name in the documentation (`IsInProgress`)? It's so consistent that it's clearly intentional, but now I won't sleep until I know why ;-)

deploy/deploy.yaml
```diff
@@ -491,22 +505,12 @@ spec:
...
-        ports:
-        - containerPort: 8443
-          name: https
       - args:
         - --metrics-addr=127.0.0.1:8080
         - --enable-leader-election
         command:
         - /manager
-        image: quay.io/isolatedcontainers/sandboxed-containers-operator:4.8
+        image: quay.io/openshift_sandboxed_containers/openshift-openshift-sandboxed-containers:4.8.0-1
```
c3d, line 513: It's not entirely clear to me why we have `$(VERSION)`, `4.8.0` and `4.8.0-1` in the same commit. That looks confusing. I believe that having separate commits that explain the rationale each time would be somewhat helpful if someone were to take over and upgrade to 4.9.

Makefile
```diff
@@ -12,7 +12,7 @@ endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # Image URL to use all building/pushing image targets
-IMG ?= quay.io/isolatedcontainers/sandboxed-containers-operator:4.8.0
+IMG ?= quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:$(VERSION)
```
c3d, line 15: Is it an OpenShift convention that the component name uses `_` and the namespace / container names contain `-`?

If I search for `openshift_` on quay.io, yours is the only org that shows up. If I search for `openshift-`, there are quite a few more, and at least that proves that `openshift-sandboxed-containers` would be a valid org name.

c3d (commented):
Many questions and a few suggestions ;-)

README.md
```diff
@@ -16,13 +16,13 @@ An operator to perform lifecycle management (install/upgrade/uninstall) of [Kata
...
    ```
-   git clone https://github.com/openshift/sandboxed-containers-operator 
+   git clone https://github.com/openshift/sandboxed-containers-operator
    git checkout -b master --track origin/master
    ```
 3. Install the sandboxed containers operator on the cluster,
 
    ```
-   make install && make deploy IMG=quay.io/isolatedcontainers/sandboxed-containers-operator:4.8
+   make install && make deploy IMG=quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:latest
```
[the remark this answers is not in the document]
jensfr: thinking about it latest it probably a bad idea anyway, it should always be the latest of the 4.8.x release

README.md
```diff
@@ -105,7 +105,7 @@ oc delete kataconfig example-kataconfig
 ### Openshift
 1. During the installation you can watch the values of the kataconfig CR. Do `watch oc describe kataconfig example-kataconfig`.
 2. To check if the nodes in the machine config pool are going through a config update watch the machine config pool resource. For this do `watch oc get mcp kata-oc`
-3. Check the logs of the sandboxed containers operator controller pod to see detailled messages about what the steps it is executing. To find out the name of the controller pod, `oc get pods -n sandboxed-containers-operator-system | grep sandboxed-containers-operator-controller-manager` and then monitor the logs of the container `manager` in that pod. 
+3. Check the logs of the sandboxed containers operator controller pod to see detailled messages about what the steps it is executing. To find out the name of the controller pod, `oc get pods -n openshift-sandboxed-containers | grep openshift-sandboxed-containers-controller-manager` and then monitor the logs of the container `manager` in that pod.
```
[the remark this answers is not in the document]
jensfr: yes, that sounds wrong. will fix it, thx

deploy/deploy.sh
jensfr: agreed

deploy/deploy.yaml
jensfr: yes, I can do that

deploy/deploy.yaml
jensfr: it is some convention, but tbh I just followed the example here :-)

deploy/deploy.yaml
jensfr: $(VERSION) is a variable available in the Makefile and I thought it's good to make use of it. Where we don't have it or some other version variable it should be 4.8-latest I think to avoid having to change it everywhere any time we push a new image. Instead we will always tag the latest image with both 4.8.y-n and 4.8.y-latest

Makefile
jensfr: It only worked with underscores when I created it, maybe a bug. It would be nice to change the org name but we can't do it ourselves we have to open a ticket with quay.io.

deploy/deploy.sh
jensfr: thinking about it. the release-4.8 is just the branch name. I think that's okay, the yaml file on that branch will always use 4.8-latest. But I introduced a bug here changing deploy.yaml to deploy.sh, which is nonsense :-)

jensfr: fixed now, thank you @c3d!

deploy/deploy.yaml
jensfr: using $VERSION now in the Makefile and set it to 4.8-latest. everywhere else we're now using 4.8-latest

deploy/deploy.yaml
jensfr: done :-)

README.md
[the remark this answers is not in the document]
jensfr: fixed in latest push

jensfr:
I pushed another update. The biggest change is that we're now using openshift-sandboxed-containers**-operator** as the namespace. It matches the other openshift operators.

README.md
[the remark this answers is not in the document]
jensfr: also fixed in latest push

deploy/deploy.yaml
jensfr: done

c3d (approved):
Looking good now.

bpradipt (approved):
/lgtm

bpradipt (approved):
/lgtm

Minor nit: Can the VERSION be defaulted to `latest` ?

c3d (approved):
/lgtm

Thanks for splitting version and latest.

c3d:
Apparently I'm not a collaborator. Sill approving, for what it's worth ;-)

PR #93, opened 2021-04-26
use completedNodesCount field

jensfr:
Increment the completedNotesCount field in kataconfig during
install/uninstall.

Other than in 4.7 with the 4.8 operator we don't update the completedNodesCount field
in kataconfig anymore. A regression introduced when we removed the daemon.

**- What I did**
Add code to update the field during install/uninstall.

**- How to verify it**
Install/uninstall and watch the value being set as nodes are completed

Fix regression that led to the updatedNodesCount field not being updated


pmores:
Looks good to me!

pmores (approved):

PR #94, opened 2021-04-28
controller: The CRI-O drop-in file should be provided by the RPM

jensfr:
Instead of having the logic of the CRI-O drop-in file in the operator,
let's just rely on it being distributed by the `kata-containers` RPM.

This is done because that file only makes sense in a system where
`kata-containers` RPM is installed, because that file is only used in a
system where `kata-containers` RPM is installed.  So, leave it for the
RPM to take care of that and let's make our lives simpler on the
operator side.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -17,12 +17,9 @@ limitations under the License.
 package controllers
 
 import (
-	"bytes"
 	"context"
-	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
-	"html/template"
 	"k8s.io/apimachinery/pkg/labels"
 	"time"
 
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -210,14 +190,6 @@ WantedBy=multi-user.target
...
 			Kind:       "MachineConfig",
 		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "50-enable-sandboxed-containers-extension",
-			Labels: map[string]string{
-				"machineconfiguration.openshift.io/role": machinePool,
-				"app":                                    r.kataConfig.Name,
-			},
-			Namespace: "sandboxed-containers-operator",
-		},
```
fidencio, line 220: Wasn't this part that I mistakenly removed?  Shouldn't we also backport the patch re-adding it?

jensfr: no that was the name of the machine config that we added back, not the namespace

fidencio: Perfect then!

fidencio (approved):
/lgtm

thanks, @jensfr!

controllers/openshift_controller.go
jensfr: now I see this removes indeed the entire thing not just the namespace. so you were right and we need to backport my fix as well.

fidencio: ack, I'm removing the lgtm for now and will re-add as soon as this one is updated. Thanks!

fidencio:
/lgtm
thanks @jensfr!

PR #97, opened 2021-04-28
Clean up and renames 4.8

jensfr:
- Description of the problem which is fixed/What is the use case

    use new namespace name and refer to new quay.io repository
    We should adhere to the naming convention in OpenShift and use
    the default namespace openshift-sandboxed-containers. The same goes for
    the quay.io user/repository we use. Use the new
    quay.io/openshift_sandboxed_containers instead of
    quay.io/isolatedcontainres.

    delete example configmap for payload. We don't have a payload any more.

- What I did

This cleans up README and example files to use the namespace
openshift-sandboxed-containers instead of sandboxed-containers-operator-system.

- How to verify it

Follow the README descripton to install/uninstall and make sure all steps work.

- Description for the changelog
Use new namespace openshift-sandboxed-containers and refer to new quay.io organisation quay.io/openshift_sandboxed_containers
@jensfr

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Current Operator version
-VERSION ?= 4.8.0
+VERSION ?= 1.0.0
 # Default bundle image tag
 BUNDLE_IMG ?= controller-bundle:$(VERSION)
 # Options for 'bundle-build'
@@ -12,7 +12,7 @@ endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
 # Image URL to use all building/pushing image targets
-IMG ?= quay.io/isolatedcontainers/sandboxed-containers-operator:4.8.0
+IMG ?= quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:$(VERSION)
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
--- a/README.md
+++ b/README.md
@@ -16,15 +16,15 @@ An operator to perform lifecycle management (install/upgrade/uninstall) of [Kata
 
 
    ```
-   git clone https://github.com/openshift/sandboxed-containers-operator 
+   git clone https://github.com/openshift/sandboxed-containers-operator
    git checkout -b master --track origin/master
    ```
 3. Install the sandboxed containers operator on the cluster,
 
    ```
-   make install && make deploy IMG=quay.io/isolatedcontainers/sandboxed-containers-operator:4.8
+   make install && make deploy IMG=quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:latest
    ```
-4. To begin the installation of the kata runtime on the cluster,
+4. To begin the installation of the kata runtime on the cluster
 
    ```
    oc create -f config/samples/kataconfiguration_v1_kataconfig.yaml
@@ -105,7 +105,7 @@ oc delete kataconfig example-kataconfig
 ### Openshift
 1. During the installation you can watch the values of the kataconfig CR. Do `watch oc describe kataconfig example-kataconfig`.
 2. To check if the nodes in the machine config pool are going through a config update watch the machine config pool resource. For this do `watch oc get mcp kata-oc`
-3. Check the logs of the sandboxed containers operator controller pod to see detailled messages about what the steps it is executing. To find out the name of the controller pod, `oc get pods -n sandboxed-containers-operator-system | grep sandboxed-containers-operator-controller-manager` and then monitor the logs of the container `manager` in that pod. 
+3. Check the logs of the sandboxed containers operator controller pod to see detailled messages about what steps it is executing. To find out the name of the controller pod, `oc get pods -n openshift-sandboxed-containers-operator | grep controller-manager` and then monitor the logs of the container `manager` in that pod.
 
 ## Components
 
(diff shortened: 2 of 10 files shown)


bpradipt (approved):
/lgtm

Same comment as for #92 . Can VERSION be defaulted to `latest` ?

c3d:
> Same comment as for #92 . Can VERSION be defaulted to latest ?

@bpradipt Do you mean replacing `1.0-lastest` with just `latest`? That seems consistent with usage for most containers.

Makefile
```diff
@@ -1,5 +1,5 @@
 # Current Operator version
-VERSION ?= 4.8.0
+VERSION ?= 1.0-latest
```
c3d, line 2: We may want to split the `VERSION` for the operator, which I would code to `1.0` at the moment, from the version of the _image_, which as @bpradipt suggested, we might want to set to `latest` for every build.

README.md
```diff
@@ -114,7 +114,7 @@ The sandboxed containers operator uses three containers:
 
 Container image name | Description | Container repository
 ---------------| ----------- | ----------
- _sandboxed-containers-operator_ |  It contains the controller part of the operator that watches and manages the kataconfig custom resource. It runs as a cluster scoped container. The operator itself is build with operator-sdk. | https://quay.io/isolatedcontainers/sandboxed-containers-operator
+ _openshift-sandboxed-containers-operator_ |  It contains the controller part of the operator that watches and manages the kataconfig custom resource. It runs as a cluster scoped container. The operator itself is build with operator-sdk. | https://quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator
```
c3d, line 117: Still annoyed like for #92 that the quay.io name does not match the spelling for the namespace. Maybe see if you could _also_ get `openshift-sandboxed-containers`? If not, I predict that 100% of our users will get it wrong at least once.

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/isolatedcontainers/sandboxed-containers-operator
-  newTag: 4.8.0
+  newName: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator
+  newTag: 1.0-latest
```
c3d, line 8: Is there a way to have multiple tags? Here, I might want to have both `1.0` and `latest`, where the next release would have `1.0.1` and `latest` (moving the `latest` tag away from the earlier build).

deploy/deploy.yaml
```diff
@@ -421,50 +421,50 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: sandboxed-containers-operator-leader-election-rolebinding
-  namespace: sandboxed-containers-operator-system
+  name: openshift-sandboxed-containers-leader-election-rolebinding
```
c3d, line 438: Just curious if it is usual to keep a very long name even when there is a namespace?

In other words, do we want the whole name to be `openshift-sandboxed-containers-operator/openshift-sandboxed-containers-leader-election-rolebinding`, or the simpler `openshift-sandboxed-containers-operator/leader-election-rolebinding`?

Checking with pods in the various operator namespaces, I see no consistent pattern, but I don't see often `openshift` in the rolebinding names. Specifically, if I do `oc get rolebindings --all-namespaces`, this one stands out.

c3d (approved):
Approving, but like @bpradipt wondering about the version number, and also wondering if we could not shorten the namespaced names.

config/manager/kustomization.yaml
bpradipt: > > Same comment as for #92 . Can VERSION be defaulted to latest ?
> 
> @bpradipt Do you mean replacing `1.0-lastest` with just `latest`? That seems consistent with usage for most containers.
@c3d yes

jensfr: why I choose 1.0-latest is: we have users of the release-4.6 and release-4.7 branch. we have release-4.8 which we should soon rename to release-1.0.0. and we have master. Would :latest relate to a build of the latest 4.6, 4.7, 4.8(1.0.0) or master?
latest build of master would make most sense but is also not intuitive in my opinion and 1.0-latest, 4.6-latest etc is more clear.
What do you think?

deploy/deploy.yaml
jensfr: The result is operator-sdk prefixing all those resources by default. One can disable this prefixing but that has other disadvanteges. Since nobody should have to type these things manually and in the webconsole they typically show up without the namespace in front I left the prefixing in.  Does that make sense?

README.md
jensfr: I made a request with customer service but it might take a while :-)

jensfr: it's a paid account and getting it was quite a process so I think requesting another one is worse then asking for a rename

Makefile
jensfr: see my other comment regarding _latest_. I think it only makes sense when everyone knows it's latest of master or latest of the 4.7 release etc. But if I'm the only one who thinks that way I'll be happy to change it :-)

jensfr: Hmm, thinking about it maybe it's best to use `latest` for the latest build of the latest release-x.y branch? and 4.7-latest, 4.6-latest for older release branches. Yes, I'll go with that unless there are other opinions?

jensfr: I'm going to leave VERSION in the Makefile at 1.0.0 because it's also used for the csv version not just the image tag. So to be clear the latest image should be tagged latest, but the version number of the operator and the ClusterServiceVersion is 1.0.0.

PR #99, opened 2021-04-30
fix function name spelling of clearUninstallStatus

jensfr:
We had two functions that were spelled almost the same and we removed
the wrong one. 

**- What I did**
This commit is renaming the remaining function
to make it compile again.

**- How to verify it**
make docker-build

Fix compilation error introduced when we removed one of two functions that were spelled almost the same


pmores (approved):
My apologies for not having caught this in #98! :-(

PR #101, opened 2021-05-05
Update README.md to use files from release-4.7 branch

jensfr:
Right now it uses files from the master branch. This is wrong, we always want to use the files from the release-4.7 branch.


snir911 (approved):
LGTM thanks!

snir911 (approved):

snir911:
@jensfr really nit, would it make sense to change links to use sandboxed-containers-operator instead of kata-operator?

jensfr:
> @jensfr really nit, would it make sense to change links to use sandboxed-containers-operator instead of kata-operator? (also in #102 )

yes, it would. I mean both works, but we should update it anyway. I'll change it. Thanks!

PR #102, opened 2021-05-05
Update install.sh, use files from release-4.7 branch

jensfr:
Right now it uses files from the master branch. This is wrong, we always want to use the files from the release-4.7 branch.


snir911 (approved):
LGTM

snir911 (approved):

PR #103, opened 2021-05-06
deploy.yaml: Fix operator repo name

jensfr:
the repo name of the operator is wrong

**- What I did**
use the correct name

**- How to verify it**
Deploy the operator via oc apply -f deploy/deploy.yaml and check the image field in the pod spec of the controller

Fix deploy.yaml to use correct operator repository name


jensfr:
thank you @fidencio

PR #104, opened 2021-05-14
Update kataconfig CR status on scaling of nodes

bpradipt:
Output of kataconfig CR does not update when nodes are scaled up

**- What I did**
Watch for changes to MachineConfigPool and trigger reconcile 

**- How to verify it**
Add a new node to the existing cluster. Run `oc describe kataconfig example-kataconfig` to display the status of 
the CR

Update kataconfig CR status on scaling of nodes

### DIFF
--- a/config/extension-crds/machineconfig.crd.yaml
+++ b/config/extension-crds/machineconfig.crd.yaml
@@ -1,0 +1,341 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  # name must match the spec fields below, and be in the form: <plural>.<group>
+  name: machineconfigs.machineconfiguration.openshift.io
+  labels:
+    "openshift.io/operator-managed": ""
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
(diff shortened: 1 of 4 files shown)


config/extension-crds/machineconfig.crd.yaml
```diff
@@ -0,0 +1,341 @@
+apiVersion: apiextensions.k8s.io/v1
```
jensfr, line 1: I'm surprised we have to keep a copy of the the CRD. Is this done automatically by operator-sdk? What happens if MCO changes something on one of the CRDs? Do we have to check the version of the CRD or something like that?

bpradipt: Using the OpenShift CRDs with `envtest` looks a bit convoluted. 
It took me fair amount of time to figure out a working scenario. I'm sure there will be a better way.

Please have a look at the following issues - 
https://github.com/kubernetes-sigs/controller-runtime/issues/1191
https://github.com/operator-framework/operator-sdk/issues/4434
https://github.com/kubernetes-sigs/controller-runtime/pull/1393

We do run the risk of test cases breaking or giving false positives if MCO changes the CRD and we don't update the same for the operator tests.

jensfr: I see. It's a known problem and this is the best we can do right now. Thanks!

jensfr:
lgtm

jensfr (approved):

PR #105, opened 2021-05-17
add bundle.Dockerfile

jensfr:
To enable the OpenShift CI to build the bundle image and test
installation of an Operator we need to add the Dockerfile to build
the bundle container image.

This was generated with 'make bundle'


wainersm (approved):
We are still working on the ci-operator configuration for this project. This bundle.Dockerfile is required. At this point I don't know if it will need more changes on this file, let's merge it to see how the CI uses it.

PR #106, opened 2021-05-18
add bundle data

jensfr:
Openshift CI is able to build the operator bundle and index image and
test installation via OLM. For that it needs the files like the
clusterserviceversion, the crd and other metadata.

This patch adds these files which are a copy of the downstream product
version.


wainersm (approved):
I think that with those files the bundle will be built. Thanks @jensfr !

Issue #113, opened 2021-06-02
Kata deployment stalls with invalid extensions error


benemon:
**Description**

**Steps to reproduce the issue:**
1. oc label node <node name> runtime=kata
2. update the CRD as described by the docs to target a specific node

```
kind: KataConfig
metadata:
  name: demolab-kataconfig
spec:
  kataConfigPoolSelector:
    matchLabels:
       runtime: kata
```

3. run the deployment steps as described by the docs

**Describe the results you received:**
Kata operator is deployed successfully. 

Rollout of kata runtime stalls after creation of CR with
```
Status:
  Installation Status:
    Is In Progress:  True
    Completed:
    Failed:
      Failed Nodes Reason:  Node skylake-fn6tt-worker-5mzzg is reporting: "invalid extensions found: [sandboxed-containers]"
```
This condition is never reconciled.

**Describe the results you expected:**
Succesful deployment of kata runtime

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `oc describe kataconfig <your-kataconfig>`:**
```
Name:         demolab-kataconfig
Namespace:    
Labels:       <none>
Annotations:  <none>
API Version:  kataconfiguration.openshift.io/v1
Kind:         KataConfig
Metadata:
  Creation Timestamp:  2021-06-02T11:20:01Z
  Finalizers:
    finalizer.kataconfiguration.openshift.io
  Generation:  2
  Managed Fields:
    API Version:  kataconfiguration.openshift.io/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:spec:
        .:
        f:kataConfigPoolSelector:
          .:
          f:matchLabels:
            .:
            f:runtime:
    Manager:      kubectl-create
    Operation:    Update
    Time:         2021-06-02T11:20:01Z
    API Version:  kataconfiguration.openshift.io/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:finalizers:
          .:
          v:"finalizer.kataconfiguration.openshift.io":
      f:spec:
        f:config:
          .:
          f:sourceImage:
      f:status:
        .:
        f:installationStatus:
          .:
          f:IsInProgress:
          f:completed:
          f:failed:
            .:
            f:failedNodesReason:
          f:inprogress:
        f:kataImage:
        f:prevMcpGeneration:
        f:runtimeClass:
        f:totalNodesCount:
        f:unInstallationStatus:
          .:
          f:completed:
[...]

PR #114, opened 2021-06-02
Fix instructions in 'with git checkout' section

jensfr:
Users shouldn't check out the master branch but the release-4.7 branch. Fix this in the README. 

Fixes: #113


bpradipt (approved):
/lgtm

benemon (commented):
Looks good to me - have run into another issue, but I don't think it's related.

PR #116, opened 2021-06-09
Update logging

bpradipt:
Add new log statements and update log statements with correct labels

Certain log statements were using incorrect labels (eg. INFO instead of ERROR).
Further, some specific paths didn't had logging statements making it harder to debug issues

**- What I did**
Added new log statements, removed few redundant log statements and updated labels 

**- How to verify it**
Get the logs of the operator POD using the `oc` command

Update Logging


jensfr (approved):
looks good to me! Thank you Pradipta!

PR #117, opened 2021-06-10
(wip/do not merge) Upgradeable condition

jensfr:
This adds a Condition to the kataconfig called Upgradeable. When set to false it signals to OLM that the Operator can't be upgraded at the moment. We can set this for example while an installation is going on.

(description will be improved soon)


jensfr:
I created a new pull request for this. Closing.

Issue #118, opened 2021-06-11
NodeSelector is not being set for RuntimeClass when using MatchExpressions in KataConfigPoolSelector


bpradipt:
**Description**
NodeSelector is not being set for RuntimeClass when using MatchExpressions in KataConfigPoolSelector

**Steps to reproduce the issue:**
1. Create KataConfig from OpenShift UI which uses `matchExpressions` or use the following yaml
```
apiVersion: kataconfiguration.openshift.io/v1
kind: KataConfig
metadata:
  name: example-kataconfig
spec:
  kataConfigPoolSelector:
     matchExpressions:
       - key: custom-kata1
         operator: In
         values:
         - "test"
```

**Describe the results you received:**

Kata runtime will be successfully installed on the nodes labelled with `custom-kata1=test`, however NodeSelector settings in RuntimeClass will be empty
`oc get runtimeclass kata -o yaml`

```
scheduling:
  nodeSelector: {}
```

**Describe the results you expected:**
`oc get runtimeclass kata -o yaml`

```
scheduling:
    nodeSelector:
        custom-kata1: test
```

PR #119, opened 2021-06-11
Set NodeSelector correctly when using MatchExpressions in KataConfigPoolSelector

bpradipt:
Fixes: #118

Using MatchExpressions in KataConfigPoolSelector resulted in empty NodeSelector in RuntimeClass.
This is because RuntimeClass expects only MatchLabels and not MatchExpressions and Operator only
handled MatchLabels

**- What I did**
Set the NodeSelector properly irrespective of using MatchLabels or MatchExpressions in KataConfigPoolSelector

**- How to verify it**
Create KataConfig from OpenShift UI or use the following yaml
```
apiVersion: kataconfiguration.openshift.io/v1
kind: KataConfig
metadata:
  name: example-kataconfig
spec:
  kataConfigPoolSelector:
    matchExpressions:
       - key: custom-kata1
         operator: In
         values:
         - "test"
```

Set NodeSelector correctly when using MatchExpressions in KataConfigPoolSelector

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -20,8 +20,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"k8s.io/apimachinery/pkg/labels"
 	"time"
+
+	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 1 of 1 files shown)


jensfr (approved):
Hi @bpradipt this looks great. Thank you!

Issue #121, opened 2021-06-16
Removed unused code 


bpradipt:
**Description**
With the switch to RHCOS extension quite a few code paths are unused.
For example  KataInstallConfig is no longer used, payload image is not needed since daemonset is no longer used and
kubernetes controller is obsolete.

PR #122, opened 2021-06-16
Remove unused code

bpradipt:
Remove unused code
Fixes: #121 

**- What I did**
Removed the code files and objects that are not in use.

**- How to verify it**
Deploy the operator either via CLI or UI and you shouldn't see the KataInstallConfig object in the KataConfig CRD

Remove unused code and object


jensfr (approved):
Looks good to me. Thank you Pradipta!

bpradipt:
The failure appears to be environment related

PR #123, opened 2021-06-30
KATA-798: openshift controller: use upgradeable condition

jensfr:
There are situations when we don't want OLM to initiate an
update/upgrade of the Operator:
 - during an installation of Kata Containers
 - during an uninstallation of Kata Containers

In these situations the state of Kata on the worker nodes is
in-flight and an update/upgrade could result in inconsistent state

**- What I did**
OLM has built-in support for this and checks if the Operator has a
condition called 'Upgradeable'. OLM will not attempt an update when the
Operator's Upgradeable condition is set to false.
https://olm.operatorframework.io/docs/advanced-tasks/communicating-operator-conditions-to-olm/#upgradeable

We set the condition to *false* when
 - the kataconfig CR was created and we initiated the installation
 - the kataconfig Cr was deleted and we initiated the uninstllation

We set the condition to *true* when
 - an installation has finished
 - an uninstallation has finished

The condition is initialized the first time the reconcile loop is run
and a new kataconfig CR instance was found.

The Upgradeable condition can be overriden by an administrator:
https://olm.operatorframework.io/docs/advanced-tasks/communicating-operator-conditions-to-olm/#overriding-operatorconditions

**- How to verify it**
1. Check that conditions are shown during install/uninstall
2. create a bundle and an index image. Install the operator. add bundle to index image and update operator while installation is ongoing. Update should be denied by OLM.
3. create a bundle and an index image. Install the operator. add bundle to index image and update operator while uninstallation is ongoing. Update should be denied by OLM.

Add Upgradeable condition to preven OLM to do updates while install/uninstall is ongoing.

Fixes: https://issues.redhat.com/browse/KATA-798


controllers/openshift_controller.go
```diff
@@ -24,6 +24,9 @@ import (
 
 	"k8s.io/apimachinery/pkg/labels"
 
+	operatorv1 "github.com/openshift/api/operator/v1"
+	cov1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
+
```
bpradipt, line 29: Is there any guidelines to use the following project to handle Conditions in OpenShift operators - https://github.com/openshift/custom-resource-status ?

controllers/openshift_controller.go
```diff
@@ -416,6 +439,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 		return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
 	}
 
+	r.setUpgradeableTrue()
```
bpradipt, line 442: Any specific reason to not have this statement before the successful return from delete method - https://github.com/openshift/sandboxed-containers-operator/blob/master/controllers/openshift_controller.go#L439 ?

jensfr: in L439 we return because not all machines are updated yet and the uninstall is still in progress. I think this is the right spot to set the upgradeable condition to true.

jensfr: I discussed this with Pradipta offline. The call to r.setUpgradeableTrue() is in this place because it should be done before we update the KataConfig. But that means we should set it to false in case the updates fail. I'll update the code accordingly. Thank you Pradipta!

bpradipt (approved):
/lgtm

cpmeadors:
Any chance of getting some unit tests for these new functions?

jensfr:
> Any chance of getting some unit tests for these new functions?

@cpmeadors yes, of course. I'll add some.

PR #127, opened 2021-07-16
add operator-related data to must-gather collection

pmores:
Please comment on paths, filenames, sanity checks, completeness etc.  I'm submitting this PR mostly to get early feedback rather than to claim this is ready to be merged!

### DIFF
--- a/must-gather/collection-scripts/gather_sandboxed_containers_logs
+++ b/must-gather/collection-scripts/gather_sandboxed_containers_logs
@@ -4,11 +4,24 @@ IFS=$'\n'
 
 BASE_COLLECTION_PATH="/must-gather"
 
-# kataconfig
 OSC_PATH=${BASE_COLLECTION_PATH}/sandboxed-containers
 mkdir -p ${OSC_PATH}
 
-oc describe kataconfig > "${OSC_PATH}/kataconfig_description"
+operator_resources=()
+operator_resources+=(kataconfig)
+operator_resources+=(clusterserviceversion)
+operator_resources+=(subscription)
+operator_resources+=(services)
+
+for operator_resource in "${operator_resources[@]}"; do
+	oc describe ${operator_resource} -n openshift-sandboxed-containers-operator > "${OSC_PATH}/${operator_resource}_description"
+done
+
+# NOTE Jul 16, 2021: a single one is expected as of this writing, with more
+# possibly being added in the future
+KATA_WEBHOOK_CONFIGS=$(oc get validatingwebhookconfigurations -A -o'custom-columns=name:metadata.name' --no-headers | grep kata)
+oc describe validatingwebhookconfigurations ${KATA_WEBHOOK_CONFIGS} > "${OSC_PATH}/validatingwebhookconfigurations_description"
+
 
 # operators
 operators=()
@@ -71,4 +84,18 @@ mkdir -p ${LIFECYCLE_MANAGER_NS}
 for pod in $(oc get pods -n openshift-operator-lifecycle-manager -o'custom-columns=name:metadata.name' --no-headers); do
     oc logs "${pod}" --all-containers -n openshift-operator-lifecycle-manager > "${LIFECYCLE_MANAGER_NS}/${pod}_logs"
     oc describe pod "${pod}" -n openshift-operator-lifecycle-manager > "${LIFECYCLE_MANAGER_NS}/${pod}_description"
+done
+
+# sandboxed containers operator deployment resources
+
+DEPLOYMENT_UNIT_PATH=${NAMESPACE_PATH}/openshift-sandboxed-containers-operator/deployments
+mkdir -p ${DEPLOYMENT_UNIT_PATH}
+
+deploymentUnits=(pods deployments statefulsets deploymentconfigs)
+
+for deploymentUnit in "${deploymentUnits[@]}"; do
+	echo "$deploymentUnit using kata runtime class still running:"
+
+	oc get $deploymentUnit -A -o json | jq -r '.items[] | select(.spec.runtimeClassName | test("kata")).metadata.name' 2>/dev/null > ${DEPLOYMENT_UNIT_PATH}/${deploymentUnit}
+
 done


must-gather/collection-scripts/gather
```diff
@@ -39,4 +39,7 @@ done
 # Collect sandboxed-containers details
 /usr/bin/gather_sandboxed_containers_logs
 
+# Collect operator details
+/usr/bin/gather_operator
```
fidencio, line 43: @pmores, any specific reason for creating a `gather_operator` file instead of re-using the `gather_sandboxed_containers_logs`?

I'd prefer keeping all the "operator related & operator itself" logs as part of the `gather_sandboxed_containers_logs`, at least for now.

fidencio (changes requested):
@pmores, in general it looks good, but I think it could be fit together with the `gather_sandboxed_containers_logs` script.

Would you mind changing it to fit there?

Nice work!

must-gather/collection-scripts/gather
pmores: @fidencio Sure, I didn't understand `gather_sandboxed_containers_logs` was the right place for my stuff, sorry.  I'll remove the new script and move its contents to the existing file.

pmores: @fidencio I'm implementing your request.  While it's clear now where to put code, I'm struggling a bit trying to find a reasonable path to put the gathered *data* to. The script `gather_sandboxed_containers_logs` generally writes to `/must-gather/sandboxed-containers` where it creates subdirs most of the time.

I could dump the operator data directly into the directory but that might litter it considerably, especially with time.

Or I could create a subdir too, however one with the obvious name (`openshift-sandboxed-containers-operator`) already exists in `/must-gather/sandboxed-containers/namespaces` so it'd be rather confusing to create same-named one in `/must-gather/sandboxed-containers`...

What's the idea behind the data directory layout and where does the data generated by this PR fit?

must-gather/collection-scripts/gather_operator
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+BASE_COLLECTION_PATH="/must-gather"
+OSC_PATH=${BASE_COLLECTION_PATH}/sandboxed-containers/operator
+mkdir -p ${OSC_PATH}
+
+/usr/bin/oc describe csv -n openshift-sandboxed-containers-operator > ${OSC_PATH}/csv
```
fidencio, line 7: This looks very similar to https://github.com/openshift/sandboxed-containers-operator/blob/master/must-gather/collection-scripts/gather_sandboxed_containers_logs#L11

Can you simply adopt that pattern and add it to the exactly same location as that? (with a different file name).

must-gather/collection-scripts/gather_operator
```diff
@@ -0,0 +1,29 @@
...
+/usr/bin/oc describe subscription -n openshift-sandboxed-containers-operator > ${OSC_PATH}/subscription
```
fidencio, line 8: This also looks very similar to https://github.com/openshift/sandboxed-containers-operator/blob/master/must-gather/collection-scripts/gather_sandboxed_containers_logs#L11

Can you simply adopt that pattern and add it to the exactly same location as that? (with a different file name).

This makes me think a loop would benefit the code, making it cleaner.

must-gather/collection-scripts/gather_operator
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+BASE_COLLECTION_PATH="/must-gather"
+OSC_PATH=${BASE_COLLECTION_PATH}/sandboxed-containers/operator
```
fidencio, line 4: Rather than using this, you should take advantage of what already exists in the code, as in https://github.com/openshift/sandboxed-containers-operator/blob/0c4e7aaab1df9ae92b66b28ceb1528eab9c7bd23/must-gather/collection-scripts/gather_sandboxed_containers_logs#L23

There we already create the namespace for logs we collect from the operators that are interesting for us.

must-gather/collection-scripts/gather_operator
```diff
@@ -0,0 +1,29 @@
...
+/usr/bin/oc describe subscription -n openshift-sandboxed-containers-operator > ${OSC_PATH}/subscription
+
+deploymentUnits=(pods deployments statefulsets deploymentconfigs)
+
+for deploymentUnit in "${deploymentUnits[@]}"; do
+	echo "$deploymentUnit using kata runtime class still running:"
+
+	/usr/bin/oc get $deploymentUnit -A -o json | jq -r '.items[] | select(.spec.runtimeClassName | test("kata")).metadata.name' 2>/dev/null > ${OSC_PATH}/${deploymentUnit}
+
+done
```
fidencio, line 17: could this whole thing be under a `deployments` directory, under https://github.com/openshift/sandboxed-containers-operator/blob/0c4e7aaab1df9ae92b66b28ceb1528eab9c7bd23/must-gather/collection-scripts/gather_sandboxed_containers_logs#L23 ?

must-gather/collection-scripts/gather_operator
```diff
@@ -0,0 +1,29 @@
...
+
+# NOTE Jul 16, 2021: mostly for webhook debugging. There are no services other
+# than the webhook one expected in the namespace as of this writing.  If any
+# are added later on it surely won't hurt to have them recorded as well.
+/usr/bin/oc get services -n openshift-sandboxed-containers-operator -o yaml > ${OSC_PATH}/services
```
fidencio, line 22: This feels more similar to https://github.com/openshift/sandboxed-containers-operator/blob/master/must-gather/collection-scripts/gather_sandboxed_containers_logs#L11, therefore should be in the same directory?

must-gather/collection-scripts/gather_operator
```diff
@@ -0,0 +1,29 @@
...
+
+# NOTE Jul 16, 2021: as with services above, a single one is expected as of
+# this writing, with more possibly being added in the future
+KATA_WEBHOOK_CONFIGS=$(/usr/bin/oc get validatingwebhookconfigurations -A -o'custom-columns=name:metadata.name' --no-headers | grep kata)
+/usr/bin/oc get validatingwebhookconfigurations ${KATA_WEBHOOK_CONFIGS} -o yaml > ${OSC_PATH}/validatingwebhookconfigurations
```
fidencio, line 27: Not sure about this one.

fidencio (changes requested):
@pmores, please, check my suggestions.

must-gather/collection-scripts/gather
fidencio: I replied as a review.

fidencio:
Another thing, Pavel, I know this is coming from a downstream request.  This downstream request is nowhere accessible to the external world, but it could be.

I'd like to **suggest** to open an issue here, describe exactly what we're trying to solve there, and associate the issue to the PR, as this can help others using the operator / reviewing the code.

pmores:
@fidencio I just pushed a reworked version that I hope addresses your comments.

As for opening the issue - can you point me to the downstream request you mention to use as a template?  I myself know hardly anything about must-gather apart from what's in its README.

must-gather/collection-scripts/gather_sandboxed_containers_logs
```diff
@@ -4,11 +4,24 @@ IFS=$'\n'
...
+operator_resources+=(services)
+
+for operator_resource in "${operator_resources[@]}"; do
+	oc describe ${operator_resource} -n openshift-sandboxed-containers-operator > "${OSC_PATH}/${operator_resource}_description"
+done
+
+# NOTE Jul 16, 2021: a single one is expected as of this writing, with more
+# possibly being added in the future
+KATA_WEBHOOK_CONFIGS=$(/usr/bin/oc get validatingwebhookconfigurations -A -o'custom-columns=name:metadata.name' --no-headers | grep kata)
+/usr/bin/oc describe validatingwebhookconfigurations ${KATA_WEBHOOK_CONFIGS} > "${OSC_PATH}/validatingwebhookconfigurations_description"
```
fidencio, line 23: I think we're super fine just using `oc` rather than `/usr/bin/oc`, and the usage of `oc` (merely) is consistent throughout the whole file.

fidencio (approved):
@pmores, there's one minor thing that you could address, but I'm fine with the patch as it's now.
Thanks for working on this.

fidencio:
> @fidencio I just pushed a reworked version that I hope addresses your comments.
> 
> As for opening the issue - can you point me to the downstream request you mention to use as a template? I myself know hardly anything about must-gather apart from what's in its README.

For instance, this is the downstream issue https://issues.redhat.com/browse/KATA-867 you're working on.
Anyone, apart from Red Hatters, have access to that.  My suggestion is for you to open a GitHub issue, with basically the same information there, so when someone looks for this we have a way to track, publicly.

Does this make sense?

must-gather/collection-scripts/gather_sandboxed_containers_logs
pmores: @fidencio Just pushed a fixed version.  Good catch, the absolute paths were just an oversight, they were supposed to be all removed at this point - my bad!

pmores:
@fidencio I've filed the issue you requested: https://github.com/openshift/sandboxed-containers-operator/issues/129

fidencio:
Thanks, @pmores
/lgtm

PR #130, opened 2021-08-03
Fix unit tests to handle webhook

bpradipt:
With addition of webhook, the related unit tests need to be fixed to handle the same

**- What I did**
Updated the code to install webhook in envtest and modify the test cases

**- How to verify it**
`make test` 

Fix unit tests to handle webhook

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.13 as builder
+FROM quay.io/bitnami/golang:1.13 as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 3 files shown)


pmores (approved):
@bpradipt Great job,  thanks!  You'll have to tell me sometime a story of how you figured it out! :-)

(Nit: I'd consider making the commits cleaner by removing the parts where the first one adds some commented-out lines and the other removes them again.)

bpradipt:
> @bpradipt Great job, thanks! You'll have to tell me sometime a story of how you figured it out! :-)
> 
> (Nit: I'd consider making the commits cleaner by removing the parts where the first one adds some commented-out lines and the other removes them again.)

Ah, I didn't realise it. Thanks for catching it. Will make changes and re-send

jensfr (approved):
lgtm, thank you @bpradipt

PR #131, opened 2021-08-17
add initial monitor daemonset support

jensfr:
To collect metrics from Kata, we need to run kata-monitor on all nodes where we
can run Kata containers. 

**- What I did**

This adds support for deploying a Daemonset with kata-monitor
on nodes. It will run on the same nodes that the runtime is installed
on by the operators and scale up and down as the machine pool size is
changed. The same label we use for marking nodes for installing the
runtime is used to run the DaemonSet on.

The container image used for the DaemonSet is specified in an environment
variable.

We create and destroy the DaemonSet as part of the Reconcile loop. It
keeps running as long as the KataConfig exists. We creat it after the
Runtimeclass was created and destroy it after the Runtimeclass was
removed.

**- How to verify it**
Install operator and create a KataConfig instance. Check if a Daemonset called
'openshift-sandboxed-containers-monitor' exists. Now delete the kataconfig and verify
that the Daemonset was destroyed. 
Verification of the actually functionality of metrics being collected etc should be done
as part of verifying the Daemonset itself.

- add support for create/destroy of the monitor daemonset

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	appsv1 "k8s.io/api/apps/v1"
+	"os"
 	"time"
 
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


controllers/openshift_controller.go
```diff
@@ -428,6 +545,19 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 		return ctrl.Result{}, err
 	}
 
+	ds := r.processDaemonsetForMonitor()
+	if ds == nil {
+		r.Log.Error(err, "error deleting monitor Daemonset")
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 15}, nil
```
bpradipt, line 551: Two semantic questions
1. Should the error message be more aligned to DS being nil ? May be something like "received empty Daemonset"?
2. Should this be requeue with error ?

controllers/openshift_controller.go
```diff
@@ -111,10 +113,125 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(req ctrl.Request) (ctrl.Result
...
+		if ds != nil {
+			r.Log.Info("daemonset is not nil")
+			if err := controllerutil.SetControllerReference(r.kataConfig, ds, r.Scheme); err != nil {
+				r.Log.Error(err, "setcontrollerreference failed")
+				return ctrl.Result{}, err
+			}
+			r.Log.Info("setcontrollerreference successful")
+		} else {
+			r.Log.Info("ds returned was nil")
+			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
```
bpradipt, line 127: Same comment as https://github.com/openshift/sandboxed-containers-operator/pull/131/files#r720001824

controllers/openshift_controller.go
```diff
@@ -111,10 +113,125 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(req ctrl.Request) (ctrl.Result
...
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: ds.Name, Namespace: ds.Namespace}, foundDs)
+		if err != nil && k8serrors.IsNotFound(err) {
+			r.Log.Info("Creating a new installation monitor daemonset", "ds.Namespace", ds.Namespace, "ds.Name", ds.Name)
+			err = r.Client.Create(context.TODO(), ds)
+			if err != nil {
+				r.Log.Error(err, "error when creating monitor daemonset")
+				res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
+			}
+		} else if err != nil {
+			r.Log.Info("could not get monitor daemonset, try again", err)
```
fgiudici, line 139: guess better a r.log.Error here

nit: what about changing the msg error to something like "could not check if the monitor daemonset is running, try again" ?

controllers/openshift_controller.go
fgiudici: Wondering if this can happen (processDaemonsetForMonitor() returning nil) or is a check to make the code more robust for the future.

fgiudici (commented):
Thanks Jens for adding the daemonset integration!
Found a log.Info that may be a log.Error.

bpradipt:
Closing this in favour of https://github.com/openshift/sandboxed-containers-operator/pull/149
The changes are incorporated into #149

PR #132, opened 2021-08-26
a couple of follow-up changes to webhook addition

pmores:
This commit removes the old guard against handling duplicate KataConfig
instances from the operator code which was made obsolete by deploying
a webhook to prevent duplicates in the first place.

It also upgrades cert-manager resources versions to 'v1'.  Frankly it's
not quite clear if this change is necessary at this point, however given
both 'v1alpha2' and 'v1' work it seems that going with a released version
is reasonable and more future-proof.

**- What I did**

**- How to verify it**


jensfr (approved):
lgtm, Pavel! Thank you!

bpradipt (approved):
/lgtm

bpradipt:
Re-running the e2e test to view latest artifacts

bpradipt (approved):
/lgtm

PR #134, opened 2021-09-13
Patch Series to handle upgrading to latest version of operator-sdk

bpradipt:
This patch series makes necessary changes to the code base to align with latest  operator-sdk (v1.11.0) , controller-runtime (0.9.2) and kubernetes version (1.21.2)

**- What I did**
Updated code, dependencies etc

**- How to verify it**
Building and installing the cluster by following the developer documentation

Upgrade to handle latest operator-sdk (v1.11.0) , controller-runtime (0.9.2) and kubernetes version (1.21.2)


bpradipt:
Seems the CI is failing due to change of operator name (OO_PACKAGE)

pmores:
I won't pretend I'm able to review properly all of this :-) so I'll have just two questions:
- with all the changes in `Makefile`, is the idea that its interface should still work the same?  I mean, `make podman-build IMG=<...>` and `podman-push` and also `make bundle-build BUNDLE_IMG=<...>`.
- there are [changes in the series](https://github.com/openshift/sandboxed-containers-operator/pull/134/commits/4c3900eba717f4b1a5cb58a2fe0a4f78b3d98e59) that would break the validating webhook (well unless the new SDK works radically differently from the previous one).  However there's also some back-and-forth in the series about changing these files or not, so while I *think* the final state is good I'd like to ask whether you verified that the webhook still works (by just attempting to create a duplicate KataConfig)?

Thanks!

bpradipt:
> I won't pretend I'm able to review properly all of this :-) so I'll have just two questions:
> 
> * with all the changes in `Makefile`, is the idea that its interface should still work the same?  I mean, `make podman-build IMG=<...>` and `podman-push` and also `make bundle-build BUNDLE_IMG=<...>`.

Installing `podman-docker` will ensure no changes to default Makefile targets. Otherwise for every target an equivalent `podman-*` will be needed. The development doc mentions `podman-docker` as a pre-req

> * there are [changes in the series](https://github.com/openshift/sandboxed-containers-operator/pull/134/commits/4c3900eba717f4b1a5cb58a2fe0a4f78b3d98e59) that would break the validating webhook (well unless the new SDK works radically differently from the previous one).  However there's also some back-and-forth in the series about changing these files or not, so while I _think_ the final state is good I'd like to ask whether you verified that the webhook still works (by just attempting to create a duplicate KataConfig)?

I tested webhook on a real cluster and it prevents multiple KataConfig from being created. I would also suspect that any issues will get covered in unit tests itself as we do have webhook based tests.


> 
> Thanks!

jensfr (approved):
lgtm, thank you @bpradipt !

Issue #135, opened 2021-09-20
fix retrieving list of nodes


jensfr:
**Description**
When the cluster has nodes in the master pool but none in the worker pool the operator
will not update status information in the KataConfig CR. This is due to a bug in the way we
update the nodes, we have a hardcoded node/roles=worker in there.

**Steps to reproduce the issue:**
This is not super easy to reproduce, you have to change the machine config pools.

Create a cluster, install the operator

1. take all nodes out of the worker machine config pool, remove the worker label from the nodes
2. create a KataConfig, dont select a custom node selector
3. watch during the installation that the status in the KataConfig is not updated, however the installation will still go through, but the runtime will be automatically installed on the maser nodes

**Describe the results you received:**
Status in KataConfig was not updated during installation/uninstallation.

**Describe the results you expected:**
An updates Status in the KataConfig (nodes in progress, completed etc.)

**Additional information you deem important (e.g. issue happens only occasionally):**
This should be ported to release-1.1 and the master branch

PR #136, opened 2021-09-20
KATA-1011: update Status of nodes with selected label

jensfr:
We used to select nodes labelled as workers only. This is wrong
when the user selected a custom label for nodes where the runtime
should be installed. Another case where this doesn't work is on
compact clusters because there we use the master machine config pool.

**- What I did**
This can be fixed by using the label that is specified in the
KataConfigPoolSelector.

**- How to verify it**
Remove worker label from nodes so that the worker MCP is empty. Then
create a kataconfig (empty pool selector) and check that the status is updated
during the installation of the runtime.

Correct status update when installing on non-worker nodes (for example nodes labelled differently)

Fixes: #135

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -639,9 +639,23 @@ func (r *KataConfigOpenShiftReconciler) getMcp() (*mcfgv1.MachineConfigPool, err
 	return foundMcp, nil
 }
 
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
/lgtm

pmores (commented):
/lgtm

PR #142, opened 2021-10-07
Remove deprecated k8s apis

bpradipt:
Remove deprecated k8s apis

**- What I did**
Removed v1beta1 apis with v1

**- How to verify it**
1. Usual deploy/install workflow 
2. By looking at the runtimeClass object

Remove deprecated k8s apis


pmores (approved):
Looks good!

PR #149, opened 2021-11-05
KATA-1136: Metrics prometheus setup

littlejawa:
**- Description**
Implements [KATA-1136](https://issues.redhat.com/browse/KATA-1136)

This pull request is meant to make the Operator install everything needed to have kata metrics visible in Prometheus.
This includes:
- a Daemonset, running the kata-monitor binary, to generate kata metrics from each node
- a Service, to expose the kata metrics
- a ServiceMonitor, allowing Prometheus to find the service
- several RBAC rules, allowing prometheus to access all the above
- a label added to our namespace (openshift-sandboxed-containers-operator), so that Prometheus actually looks for the ServiceMonitor in our namespace
- an additional RBAC (to be defined), allowing the operator to create the DaemonSet with privileged status (required to access the socket from shimV2 on the node)

**- What I did**
- cherry-picked Jens' work for the DaemonSet (see #131) - we may want to finish reviewing that separately, or make the review here, as is easier
- added YAML files for Service/ServiceMonitor/RBAC (as we created them when prototyping)
- modified the YAML files (names and labels) to match the name and labels used on the DaemonSet

**- What is missing**
- additional rights must be given to the operator to be allowed to create the DaemonSet. As a workaround, after deployment, and before you create the kataconfig, you can run:
`oc adm policy add-scc-to-user privileged -z default -n openshift-sandboxed-containers-operator`
- a label need to be added to the namespace for Prometheus to actually start scraping data. This can be done after successful creation of the kataconfig:
`oc label namespaces openshift-sandboxed-containers-operator openshift.io/cluster-monitoring=true`

**- How to verify it**
- deploy the operator
  - I have a yaml file that can be used to deploy all this, using development builds for the images.
  - this PR can of course be checked out and built locally, for deployment from your own repositories

At this point, there is no difference from previous versions of the operator. You can check that the controller-manager pod is running properly, using the development image and not the official one.

[WIP] run `oc adm policy add-scc-to-user privileged -z default -n openshift-sandboxed-containers-operator`

- create the kataconfig
`oc apply -f https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/master/config/samples/kataconfiguration_v1_kataconfig.yaml
`
At this point, you can check that additional elements have been installed in the openshift-sandboxed-containers-operator namespace:
  - a DamonSet, named "openshift-sandboxed-containers-monitor", running on all worker nodes, with the development image from fgiudici's repository on quay.io
  - a Service, as below
```Name:              metrics
Namespace:         openshift-sandboxed-containers-operator
Labels:            name=openshift-sandboxed-containers-monitor
Annotations:       <none>
Selector:          name=openshift-sandboxed-containers-monitor
Type:              ClusterIP
IP Family Policy:  SingleStack
IP Families:       IPv4
IP:                172.30.93.177
IPs:               172.30.93.177
Port:              metrics  8090/TCP
TargetPort:        8090/TCP
Endpoints:         10.132.2.4:8090,10.132.4.4:8090,10.133.2.3:8090 + 3 more...
Session Affinity:  None
Events:            <none>
```
  - a ServiceMonitor, as below
```Name:         openshift-sandboxed-containers-monitor
Namespace:    openshift-sandboxed-containers-operator
Labels:       <none>
Annotations:  <none>
API Version:  monitoring.coreos.com/v1
Kind:         ServiceMonitor
Metadata:
  Creation Timestamp:  2021-11-04T17:19:46Z
  Generation:          1
  Managed Fields:
    API Version:  monitoring.coreos.com/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
      f:spec:
        .:
        f:endpoints:
        f:namespaceSelector:
          .:
          f:matchNames:
        f:selector:
          .:
          f:matchLabels:
            .:
            f:name:
    Manager:         kubectl-client-side-apply
    Operation:       Update
    Time:            2021-11-04T17:19:46Z
  Resource Version:  1233875
  UID:               aa518ee6-3a1e-4641-8aad-e8a4470209b5
Spec:
  Endpoints:
    Port:  metrics
  Namespace Selector:
    Match Names:
      openshift-sandboxed-containers-operator
  Selector:
    Match Labels:
      Name:  openshift-sandboxed-containers-monitor
Events:      <none>
```

[WIP] If all of that is verified, you can add the label that will make Prometheus get this information
`oc label namespaces openshift-sandboxed-containers-operator openshift.io/cluster-monitoring=true`

- Now you can go to the Web console
  - in the "Observe/Dashboard" section, use the "Prometheus/Overview" dashboard, and look at the "TargetSync" graph
    The list of targets scraped by prometheus appears there.
    You will see a new target appear the next time Prometheus gathers data (look for "serviceMonitor/openshift-sandboxed-containers-operator/openshift-sandboxed-containers-monitor/0")
    (NOTE: this can take some time, as Prometheus is polling for data every once in a while. The graph actually helps you guess when your data should be appearing, depending on when you created the label compared to where there is "scrape activity" shown on the graph)
  - when the target appears, you can go to the "Observe/Metrics" section, and enter "kata" in the search box. This should give you the list of available metrics with "kata" in their name. If they show up, it means the metrics is being gathered by Prometheus.

MORE:
- run a pod under kata (see sample "example-fedora" for instance), to actually generate metrics from kata
- run custom queries from the "metrics" page to verify metrics are consistent with the pods being run

Enable sandboxed-containers vm- and pod- specific metrics gathering from Prometheus

### DIFF
--- a/config/rbac/katamonitor.yaml
+++ b/config/rbac/katamonitor.yaml
@@ -1,0 +1,31 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: monitor-role
+  namespace: openshift-sandboxed-containers-operator
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - endpoints
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: monitor-rolebinding
+  namespace: openshift-sandboxed-containers-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: monitor-role
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -142,7 +142,7 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
 
 	if monitorImage == "" {
 		r.Log.Info("OSC_MONITOR_IMAGE is not set, using default image")
-		monitorImage = "quay.io/openshift_sandboxed_monitors/openshift-sandboxed-containers-monitor:latest"
+		monitorImage = "quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-monitor:latest"
 	}
 	dsName := "openshift-sandboxed-containers-monitor"
 	dsLabels := map[string]string{
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -126,7 +126,7 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 				res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
 			}
 		} else if err != nil {
-			r.Log.Info("could not get monitor daemonset, try again", err)
+			r.Log.Error(err, "could not get monitor daemonset, try again")
 			res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
 		}
 
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -19,6 +19,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	appsv1 "k8s.io/api/apps/v1"
+	"os"
 	"time"
 
 	"k8s.io/apimachinery/pkg/labels"
@@ -101,8 +103,123 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 			return ctrl.Result{}, updateErr
 		}
 
+		ds := r.processDaemonsetForMonitor()
+		// Set KataConfig instance as the owner and controller
+		if ds != nil {
+			r.Log.Info("daemonset is not nil")
+			if err := controllerutil.SetControllerReference(r.kataConfig, ds, r.Scheme); err != nil {
+				r.Log.Error(err, "setcontrollerreference failed")
+				return ctrl.Result{}, err
+			}
+			r.Log.Info("setcontrollerreference successful")
+		} else {
+			r.Log.Info("ds returned was nil")
+			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
+		}
+		foundDs := &appsv1.DaemonSet{}
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: ds.Name, Namespace: ds.Namespace}, foundDs)
+		if err != nil && k8serrors.IsNotFound(err) {
+			r.Log.Info("Creating a new installation monitor daemonset", "ds.Namespace", ds.Namespace, "ds.Name", ds.Name)
+			err = r.Client.Create(context.TODO(), ds)
+			if err != nil {
+				r.Log.Error(err, "error when creating monitor daemonset")
+				res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
+			}
+		} else if err != nil {
+			r.Log.Info("could not get monitor daemonset, try again", err)
+			res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
+		}
+
 		return res, err
 	}()
+}
+
+func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.DaemonSet {
+	var (
+		runPrivileged = true
+		monitorImage  = os.Getenv("OSC_MONITOR_IMAGE")
+	)
+
+	if monitorImage == "" {
+		r.Log.Info("OSC_MONITOR_IMAGE is not set, using default image")
+		monitorImage = "quay.io/openshift_sandboxed_monitors/openshift-sandboxed-containers-monitor:latest"
+	}
+	dsName := "openshift-sandboxed-containers-monitor"
+	dsLabels := map[string]string{
+		"name": dsName,
+	}
+
+	var nodeSelector map[string]string
+	if r.kataConfig.Spec.KataConfigPoolSelector != nil {
+		nodeSelector = r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels
+	} else {
+		nodeSelector = map[string]string{
+			"node-role.kubernetes.io/worker": "",
+		}
+	}
+
+	return &appsv1.DaemonSet{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "apps/v1",
+			Kind:       "DaemonSet",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      dsName,
+			Namespace: "openshift-sandboxed-containers-operator",
+		},
+		Spec: appsv1.DaemonSetSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: dsLabels,
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: dsLabels,
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: "default",
+					NodeSelector:       nodeSelector,
+					Containers: []corev1.Container{
+						{
+							Name:            "kata-monitor",
+							Image:           monitorImage,
+							ImagePullPolicy: "Always",
+							SecurityContext: &corev1.SecurityContext{
+								Privileged: &runPrivileged,
+							},
+							Command: []string{"/usr/bin/kata-monitor", "--log-level=debug", "--runtime-endpoint=/run/crio/crio.sock"},
+							VolumeMounts: []corev1.VolumeMount{
+								{
+									Name:      "crio-sock",
+									MountPath: "/run/crio/",
+								},
+								{
+									Name:      "sbs",
+									MountPath: "/run/vc/sbs/",
+								}},
+						},
+					},
+					Volumes: []corev1.Volume{
+						{
+							Name: "crio-sock",
+							VolumeSource: corev1.VolumeSource{
+								HostPath: &corev1.HostPathVolumeSource{
+									Path: "/run/crio/",
+								},
+							},
+						},
+						{
+							Name: "sbs",
+							VolumeSource: corev1.VolumeSource{
+								HostPath: &corev1.HostPathVolumeSource{
+									Path: "/run/vc/sbs/",
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
 }
 
 func (r *KataConfigOpenShiftReconciler) newMCPforCR() (*mcfgv1.MachineConfigPool, error) {
@@ -508,6 +625,19 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 	if err != nil {
 		r.Log.Error(err, "Unable to update KataConfig status")
 		return ctrl.Result{}, err
+	}
+
+	ds := r.processDaemonsetForMonitor()
+	if ds == nil {
+		r.Log.Error(err, "error deleting monitor Daemonset")
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 15}, nil
+	}
+	err = r.Client.Delete(context.TODO(), ds)
+	if err != nil && k8serrors.IsNotFound(err) {
+		r.Log.Info("monitor daemonset was already deleted")
+	} else if err != nil {
+		r.Log.Error(err, "error when deleting monitor Daemonset, try again")
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 15}, err
 	}
 
 	r.Log.Info("Uninstallation completed. Proceeding with the KataConfig deletion")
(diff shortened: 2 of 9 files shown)


littlejawa:
/cc @fgiudici @bpradipt @cpmeadors 

Still work in progress - need to review/discuss/improve

fgiudici (commented):
@littlejawa , great work!
Not sure if we can end-up with the official img name for the Daemonset soon (guess we need the CDN), we can anyway use the internal RH repo :-) 
It looks good to me, but of course we need @bpradipt to review this :-)

config/rbac/katamonitor.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s
```
bpradipt, line 22: Usually we add a `binding` suffix to to distinguish with the role name. For example `prometheus-k8s-binding`.
Having said that, how about having the role name as `monitor-role` and rolebinding name as `monitor-rolebinding` ?
Will be in sync with the other roles and rolebindings .

controllers/openshift_controller.go
```diff
@@ -143,7 +143,7 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
 
 	if monitorImage == "" {
 		r.Log.Info("OSC_MONITOR_IMAGE is not set, using default image")
-		monitorImage = "quay.io/openshift_sandboxed_monitors/openshift-sandboxed-containers-monitor:latest"
+		monitorImage = "quay.io/fgiudici/kata-monitor:latest"
```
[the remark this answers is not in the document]
bpradipt: Better to change the default image to `quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-monitor`. For anyone deploying from upstream, we'll have all the images under `quay.io/openshift_sandboxed_containers/`. This is true for all existing images - operator, bundle, must-gather.

controllers/openshift_controller.go
```diff
@@ -102,10 +104,125 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: ds.Name, Namespace: ds.Namespace}, foundDs)
+		if err != nil && k8serrors.IsNotFound(err) {
+			r.Log.Info("Creating a new installation monitor daemonset", "ds.Namespace", ds.Namespace, "ds.Name", ds.Name)
+			err = r.Client.Create(context.TODO(), ds)
+			if err != nil {
+				r.Log.Error(err, "error when creating monitor daemonset")
+				res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
+			}
+		} else if err != nil {
+			r.Log.Info("could not get monitor daemonset, try again", err)
```
bpradipt, line 130: this should be `r.Log.Error`

controllers/openshift_controller.go
```diff
@@ -102,10 +104,125 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
...
+			res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
+		}
+
 		return res, err
 	}()
 }
 
+func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.DaemonSet {
+	var (
+		runPrivileged = true
```
bpradipt, line 140: Is there a possibility to know the minimum capabilities needed for the monitor daemon instead of all privileges ?

bpradipt:
@littlejawa thanks for the PR. I think we can close https://github.com/openshift/sandboxed-containers-operator/pull/131.
Also, it will be good to add the Dockerfile and other artifacts for the kata-monitor image as a sub-folder like what's done for must-gather.

config/rbac/katamonitor.yaml
littlejawa: Sure, I will change that.

controllers/openshift_controller.go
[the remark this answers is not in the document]
littlejawa: We will definitely make it use an official repo at some point, but we needed a way to make it testable quickly while WIP, and I couldn't think of a better way to do it without publishing untested bits.

controllers/openshift_controller.go
littlejawa: My understanding is that this is needed to access the sockets used by the shim running on the node. They are available in a folder, so we need special access to that folder, in order to find new sockets being created there.
I am not sure how that translate into specific capabilities.
@fgiudici maybe?

controllers/openshift_controller.go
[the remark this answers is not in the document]
bpradipt: For testing, you can use the environment variable - `OSC_MONITOR_IMAGE`

cpmeadors:
I was able to get this working on a nightly OCP 4.10 build, by installing the 1.1 operator (which installed kata-containers-2.3.0-0.1.alpha2.el8.x86_64 rpms).  Then deploying the daemonset and service bits manually.  I would like to try with a new operator build, but so far, looks good.

controllers/openshift_controller.go
bpradipt: Let's go ahead with this for now and we can create an issue to specifically investigate on the minimum capabilities needed.

bpradipt:
@littlejawa can you also include the jira ticket in the description ?

controllers/openshift_controller.go
```diff
@@ -102,10 +104,125 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
...
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: ds.Name, Namespace: ds.Namespace}, foundDs)
+		if err != nil && k8serrors.IsNotFound(err) {
+			r.Log.Info("Creating a new installation monitor daemonset", "ds.Namespace", ds.Namespace, "ds.Name", ds.Name)
+			err = r.Client.Create(context.TODO(), ds)
+			if err != nil {
+				r.Log.Error(err, "error when creating monitor daemonset")
+				res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
+			}
+		} else if err != nil {
+			r.Log.Error("could not get monitor daemonset, try again", err)
```
bpradipt, line 130: This should be r.Log.Error(err, "could....")

littlejawa:
> @littlejawa can you also include the jira ticket in the description ?

Done :)

controllers/openshift_controller.go
```diff
@@ -102,10 +104,125 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 			return ctrl.Result{}, updateErr
 		}
 
+		ds := r.processDaemonsetForMonitor()
+		// Set KataConfig instance as the owner and controller
+		if ds != nil {
+			r.Log.Info("daemonset is not nil")
```
c3d, line 110: "daemonset exists", maybe even pass a bit of useful data to the log?

("not nil" is internal jargon)

controllers/openshift_controller.go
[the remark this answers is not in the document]
littlejawa: As discussed with @bpradipt : modified the default image name. The target has been copied from fgiudici's repository.

controllers/openshift_controller.go
c3d: @littlejawa You marked the suggestion as "resolved" without indicating how you plan to resolve it. So let me add the comment here: it's fixed in the next commit :-)

bpradipt:
Thanks @littlejawa. Let's wait for CI to be green and we are done :-)

controllers/openshift_controller.go
c3d: I agree that this should be handled with regular permissions (a group would be enough).

controllers/openshift_controller.go
```diff
@@ -102,10 +104,125 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
...
 }
 
+func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.DaemonSet {
+	var (
+		runPrivileged = true
+		monitorImage  = os.Getenv("OSC_MONITOR_IMAGE")
+	)
+
+	if monitorImage == "" {
+		r.Log.Info("OSC_MONITOR_IMAGE is not set, using default image")
```
c3d, line 145: What about having the log always record which image is used?

controllers/openshift_controller.go
```diff
@@ -102,10 +104,125 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
...
+					Containers: []corev1.Container{
+						{
+							Name:            "kata-monitor",
+							Image:           monitorImage,
+							ImagePullPolicy: "Always",
+							SecurityContext: &corev1.SecurityContext{
+								Privileged: &runPrivileged,
+							},
+							Command: []string{"/usr/bin/kata-monitor", "--log-level=debug", "--runtime-endpoint=/run/crio/crio.sock"},
+							VolumeMounts: []corev1.VolumeMount{
```
c3d, line 191: Shouldn't we read that info from a configuration file (like you did for the service and other objects two commits down)? Or is it standard practice to hard-code paths and command-line options for daemonsets for some reason I don't know about?

controllers/openshift_controller.go
```diff
@@ -419,6 +536,19 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 		return ctrl.Result{}, err
 	}
 
+	ds := r.processDaemonsetForMonitor()
+	if ds == nil {
+		r.Log.Error(err, "error deleting monitor Daemonset")
+		return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 15}, nil
+	}
+	err = r.Client.Delete(context.TODO(), ds)
+	if err != nil && k8serrors.IsNotFound(err) {
```
c3d, line 545: MInor nit: you could test `err != nil` once and have the second test `k8serrors.isNotFound(err)` inside that if.

controllers/openshift_controller_test.go
```diff
@@ -345,6 +346,15 @@ var _ = Describe("OpenShift KataConfig Controller", func() {
 				return k8sClient.Get(context.Background(), types.NamespacedName{Name: "kata"}, rc)
 			}, timeout, interval).Should(Succeed())
 
+			time.Sleep(10 * time.Second)
```
c3d, line 349: Why is this delay necessary? (Add comment).

controllers/openshift_controller.go
littlejawa: I don't know to be honest.
The go code installing the daemonset was made by Jens. The rest (YAML files) is mine. That's why there is a difference in method. I'm not sure what is standard practice, or preferable :-(

controllers/openshift_controller.go
littlejawa: I guess we can.
In practice, you will be quicker seeing what image is being used by looking at the daemonset itself (oc describe [...]). That's how I've been troubleshooting, even before looking at the logs.

controllers/openshift_controller.go
littlejawa: Right, sorry for that :-)

controllers/openshift_controller.go
c3d: I prefer the approach of using a yaml file, I find it much more readable. Personal preference.

controllers/openshift_controller.go
littlejawa: Another related point that @fgiudici asked in the original PR: can r.processDaemonsetForMonitor() actually return "nil", and in what condition ? I guess it can, as Jens made sure to requeue if it does end up being nil (which I think means we will just retry later). But I'm not sure in what conditions.
So in addition to more data in the "success" case here, we probably also want to get more info on the "fail" part below, in case somehow we end up in a retry loop.

controllers/openshift_controller.go
littlejawa: Right. I will modify that.

controllers/openshift_controller.go
littlejawa: I have tried to address that, and the other "info" strings in the same area.
Unfortunately, I am not sure what additional information I could put in there. Any suggestion is welcome.

controllers/openshift_controller.go
littlejawa: Added a line of log in all cases, showing what image will be used.

controllers/openshift_controller.go
littlejawa: I have done just that :-)

controllers/openshift_controller.go
bpradipt: I have seen both the approaches in use with operators. And as @c3d mentions, it's a matter of personal preference.
When using yaml, it needs to be rendered to a go object before being used with the Create function. In the existing case, the object is directly created with the specific values. I think we can stay with the current approach for the time being if there is no strong objection.

littlejawa:
@bpradipt : FYI - I rebased, and modified the namespace name in parts of the code that you added (in openshift_controller_test.go), for consistency with the other changes from Jens.

@c3d : I think I addressed your comments, can you doublecheck?

c3d:
> @c3d : I think I addressed your comments, can you doublecheck?

@littlejawa Looks good to me. Thanks.

bpradipt (approved):
/lgtm
Thanks @littlejawa

PR #151, opened 2021-11-16
Update manifest files

bpradipt:
Add daemonset file
  Update channel to candidate


littlejawa:
Hi @bpradipt,
I missed those in my commit - thanks for fixing it :-/

/lgtm

PR #152, opened 2021-11-17
Patch series to fix issues with monitor daemonset and CheckNodeEligibility

bpradipt:
When CheckNodeEligibility is selected, operator unable to find labelled nodes and uses incorrect nodeSelector
for monitor DaemonSet

**- What I did**
Fixed the label handling when searching for eligible nodes and used the right nodeSelector

**- How to verify it**
Creating KataConfig and observing the status updates

Fix issues with monitor daemonset and CheckNodeEligibility

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -152,14 +152,7 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
 		"name": dsName,
 	}
 
-	var nodeSelector map[string]string
-	if r.kataConfig.Spec.KataConfigPoolSelector != nil {
-		nodeSelector = r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels
-	} else {
-		nodeSelector = map[string]string{
-			"node-role.kubernetes.io/worker": "",
-		}
-	}
+	nodeSelector := r.getNodeSelectorAsMap()
 
 	return &appsv1.DaemonSet{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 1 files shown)


valtair1891:
Verified on 4.9 and 4.10

Kataconfig
status:
  installationStatus:
    IsInProgress: 'false'
    completed:
      completedNodesCount: 3
      completedNodesList:
        - valiev-kata-410-pnhbc-worker-a-ck7rc.c.openshift-qe.internal
        - valiev-kata-410-pnhbc-worker-b-p8q7p.c.openshift-qe.internal
        - valiev-kata-410-pnhbc-worker-c-jct4r.c.openshift-qe.internal
    failed: {}
    inprogress: {}
  prevMcpGeneration: 2
  runtimeClass: kata
  totalNodesCount: 3
  unInstallationStatus:
    completed: {}

Pod is running

  - name: default-dockercfg-vqkch
  priority: 0
  schedulerName: default-scheduler
  enableServiceLinks: true
  terminationGracePeriodSeconds: 30
  preemptionPolicy: PreemptLowerPriority
  runtimeClassName: kata
  nodeName: valiev-kata-49-ci-rngps-worker-b-jgmn5.c.openshift-qe.internal

zanetworker:
QQ: Does this consider the intersection of both eligible nodes and selected nodes?

bpradipt:
> QQ: Does this consider the intersection of both eligible nodes and selected nodes?

Yes.

littlejawa:
Looks good to me

controllers/openshift_controller.go
```diff
@@ -406,6 +399,40 @@ func (r *KataConfigOpenShiftReconciler) checkNodeEligibility() error {
 	return nil
 }
 
+func (r *KataConfigOpenShiftReconciler) getNodeSelectorAsMap() map[string]string {
```
pmores, line 402: Great that you factored this code out to its own function! There's one more place - in newMCPforCR() - that builds a very similar selector, could the relationship between the two selector be used to simplify the code even more?

I realise that the two selectors aren't precisely the same (the one in newMCPforCR() is used to label nodes with 'kata-oc' whereas this one uses 'kata-oc') but still, the amount of code is considerable and the similarities close enough to be worth considering a further refactor - what do you think?

bpradipt: You are right. I was thinking about refactoring the code in newMCPforCR as well, however newMCPforCR expects labelselectors. So the method will be something like get NodeSelectorAsSelector(). Hence left it as-is to tackle it later on. Need some investigation to find out existing helper methods for conversion and any limitations.

pmores (approved):

PR #153, opened 2021-11-23
operator-sdk upgrade and Dockerfile update

bpradipt:
Upgrade SDK and Dockerfile

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=sandboxed-containers-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=candidate
 LABEL operators.operatorframework.io.bundle.channel.default.v1=candidate
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.13.0+git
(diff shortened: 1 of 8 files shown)


bundle.Dockerfile
```diff
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=sandboxed-containers-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=candidate
 LABEL operators.operatorframework.io.bundle.channel.default.v1=candidate
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.13.0+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.14.0+git
```
pmores, line 10: @bpradipt The commit message says the SDK is being upgraded to 1.15.0 but here it looks like it's actually 1.14.0 - or am I misunderstanding something?

bpradipt: @pmores Yeah, I fail to understand why is it so :-( 
`operator-sdk version` returns `operator-sdk version: "v1.15.0"`, but when generating the manifests, it mentions v1.14.0. I'll create a bug in operator-sdk to check or manually edit the value

bpradipt: @pmores updated the version string

pmores (approved):

PR #154, opened 2021-11-24
RBAC: add privileged role to the default user

fgiudici:
In order to deploy the privileged monitor daemonset we need to allow the
default user to deploy privileged pods.
This is needed as the monitor pod will need to access reserved hostpaths
in the worker nodes (/run/crio/crio.sock and /run/vc/sbs dir).

Fixes: https://issues.redhat.com/browse/KATA-1189

### DIFF
--- a/config/rbac/katamonitor.yaml
+++ b/config/rbac/katamonitor.yaml
@@ -29,3 +29,17 @@ subjects:
 - kind: ServiceAccount
   name: prometheus-k8s
   namespace: openshift-monitoring
+---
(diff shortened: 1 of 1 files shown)


bpradipt:
@fgiudici marking it hold for now to have further discussions on the least privilege that's needed.

bpradipt:
Closing this after discussion with @fgiudici. The correct fix to deploy the rbac and other related configs for kata-monitor is addressed via https://github.com/openshift/sandboxed-containers-operator/pull/165

Issue #156, opened 2021-11-30
README from `master` doesn't point to 4.8 nor 4.9 README


fidencio:
**Description**

README from the `master` branch points to 4.7 and 4.6 installation instructions, but doesn't mentioned 4.8 one.

**Steps to reproduce the issue:**
1. Open a browser
2. Go to https://github.com/openshift/sandboxed-containers-operator
3. Read the README

**Describe the results you received:**

The first thing you'll see is two warnings saying that "if you're using OCP 4.x please follow ...", for both OpenShift 4.6 and 4.7.

**Describe the results you expected:**

The same info presented for 4.6, and 4.7. to be also present for OpenShift 4.8 and OpenShift 4.9

**Additional information you deem important (e.g. issue happens only occasionally):**

Not relevant

**Output of `oc describe kataconfig <your-kataconfig>`:**

```
Not relevant
```

**Additional environment details (platform, options, etc.):**

Not relevant

Issue #159, opened 2021-11-30
release-1.1: deployment breaks due to `flag provided but not defined: -metrics-addr`


fidencio:
**Description**

It's absolutely impossible to deploy the sandboxed-containers operator following instructions from the release-1.1 branch.
After manually editing the installation script to point to the content of the release-1.1 branch, due to https://github.com/openshift/sandboxed-containers-operator/issues/158, the deployment simply breaks due to:
```
[fidencio@kundera ~]$ oc -n openshift-sandboxed-containers-operator logs openshift-sandboxed-containers-controller-manager-7df78c78mn4fw                                             
flag provided but not defined: -metrics-addr
Usage of /manager:        
  -kubeconfig string                                                                                                                                                                 
        Paths to a kubeconfig. Only required if out-of-cluster.
  -leader-elect                                                                                                                                                                      
        Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.                                                        
  -metrics-bind-address string
```

**Steps to reproduce the issue:**
1. Open your web browser
2. Go to https://github.com/openshift/sandboxed-containers-operator/tree/release-1.1#without-a-git-repo-checkout
3. Work around https://github.com/openshift/sandboxed-containers-operator/issues/158 by:
    3.1. Manually getting https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/master/deploy/install.sh
    3.2. Change references from master to release-1.1
4. Run the script
5. After some time check the state of the controller manager pod (`oc get pods -n openshift-sandboxed-containers-operator | grep controller-manager`), and verify it's in CrashLoopBackOff
6. Check its logs (

**Describe the results you received:**

```
[fidencio@kundera ~]$ oc get pods -n openshift-sandboxed-containers-operator | grep controller-manager                                                                               
openshift-sandboxed-containers-controller-manager-7df78c78mn4fw   0/1     CrashLoopBackOff   10         28m                                                                          
[fidencio@kundera ~]$ oc -n openshift-sandboxed-containers-operator logs openshift-sandboxed-containers-controller-manager-7df78c78mn4fw                                             
flag provided but not defined: -metrics-addr
Usage of /manager:        
  -kubeconfig string                                                                                                                                                                 
        Paths to a kubeconfig. Only required if out-of-cluster.
  -leader-elect                                                                                                                                                                      
        Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.                                                        
  -metrics-bind-address string
[...]

Issue #158, opened 2021-11-30
release-1.1 deployment instructions is pointing to wrong links


fidencio:
**Description**

https://github.com/openshift/sandboxed-containers-operator/tree/release-1.1#without-a-git-repo-checkout points to a non-existent file.

The document mentions:
> To deploy the operator and create a custom resource (which installs Kata on all worker nodes), run curl https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/master/deploy/install.sh | bash

However, https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/master/deploy/install.sh is a non-existent file.
Most likely it should point to https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/release-1.1/deploy/install.sh

**Steps to reproduce the issue:**
1. Open your web browser
2. Go to https://github.com/openshift/sandboxed-containers-operator/tree/release-1.1#without-a-git-repo-checkout
3. Try to follow the instructions provided

**Describe the results you received:**
```
[fidencio@kundera ocp]$ curl https://raw.githubusercontent.com/openshift/sandboxed-containers-operator/master/deploy/install.sh | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    14  100    14    0     0     57      0 --:--:-- --:--:-- --:--:--    57
bash: line 1: 404:: command not found
```
**Describe the results you expected:**

The deployment would start.

**Additional information you deem important (e.g. issue happens only occasionally):**

Not relevant

**Output of `oc describe kataconfig <your-kataconfig>`:**

```
Not relevant
```

**Additional environment details (platform, options, etc.):**

Not relevant

fidencio:
Note that the script itself also has to be updated to point to the release branch.

bpradipt:
Fixed via https://github.com/openshift/sandboxed-containers-operator/pull/160

Issue #157, opened 2021-11-30
Create and publish a release-4.9 branch


fidencio:
**Description**

There's no release-4.9 branch, even with OpenShift 4.9 being released  on Oct, 18th 2021.

**Steps to reproduce the issue:**
1. Open a browser
2. Go to  https://github.com/openshift/sandboxed-containers-operator
3. Check the branches

**Describe the results you received:**

The newest release-x.y branch is 4.8.

**Describe the results you expected:**

A release-4.9 branch should be present and published, considering `master` is being used for the "under development OCP 4.10".

**Additional information you deem important (e.g. issue happens only occasionally):**

Not relevant

**Output of `oc describe kataconfig <your-kataconfig>`:**

```
Not relevant
```

**Additional environment details (platform, options, etc.):**

Not relevant

bpradipt:
Hi @fidencio, thanks for raising the issue. 
Starting from 4.9, switched to release-x.y based on operator version instead of OpenShift versions.
First one is release-1.1 which should work for 4.9/4.10..

However I see that deploy.yaml is not existing for release-1.1 branch. This is something that needs fixing.

Also I think README should be updated to reflect the release branch name change. Sorry about it. Will create a PR soon.

fidencio:
> Also I think README should be updated to reflect the release branch name change.

Oh yeah, anyone who's outside of the internal team meetings will not be able to associate that the release-1.1 should be used with OCP 4.9 / 4.10, mainly with the existing release-4.6, and release-4.7 branches there.

> Sorry about it. Will create a PR soon.

No problemo, thanks for the quick reply.

bpradipt:
Updated readme reflects the branch name change now. Closing this issue.

PR #160, opened 2021-12-01
Update README for 1.1 release

bpradipt:
The README was dated containing incorrect installation instructions.
Fixes: #156, #157 , #158 , #159 
 
**- What I did**
Added instructions to deploy 1.1 on OpenShift 4.9

**- How to verify it**
By following the instructions to deploy the operator

Fix deployment instructions for release-1.1 on OCP4.9

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,8 +92,25 @@ and look at the field `Completed nodes` in the status. If the value matches the 
 Once the sandboxed-containers extension is enabled successfully on the intended workers, the sandboxed containers operator will create a [runtime class](https://kubernetes.io/docs/concepts/containers/runtime-class/) `kata`. This runtime class can be used to deploy the pods that will use the Kata Runtime.
 
 #### Run an Example Pod using the Kata Runtime
-```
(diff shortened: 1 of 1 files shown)


bpradipt:
@fidenco kindly check the updated deployment instructions and let me know.

fidencio:
LGTM, @bpradipt.  Although I don't think it actually fixes one of problems I hit as part of #159, which you may prefer treating it in a different issue (see: https://github.com/openshift/sandboxed-containers-operator/issues/159#issuecomment-983353183)

Last but not least, the README from the `master` branch is still wrong, so I don't think #156 should be closed, but rather have the main branch updated to point to the release-1.1 if folks are using the OCP 4.9.

bpradipt:
> LGTM, @bpradipt. Although I don't think it actually fixes one of problems I hit as part of #159, which you may prefer treating it in a different issue (see: [#159 (comment)](https://github.com/openshift/sandboxed-containers-operator/issues/159#issuecomment-983353183))
> 
> Last but not least, the README from the `master` branch is still wrong, so I don't think #156 should be closed, but rather have the main branch updated to point to the release-1.1 if folks are using the OCP 4.9.

Right, for 156 will send a new PR against the main branch.

PR #161, opened 2021-12-01
Update deployment instructions for OCP 4.8

bpradipt:
The install instructions were incorrect
Partially Fixes: #156 

**- What I did**
Updated with correct install instructions
**- How to verify it**
Install the operator by following the instructions in the README

Fix install instruction.


fidencio (approved):
lgtm, thanks @bpradipt!

PR #167, opened 2021-12-15
KATA-1145: Introduce metrics dashboard in OCP console.

littlejawa:
**What is the use case**
Have the kata-monitor metrics visible as a graph under the "Dashboard" section of OpenShift

**- What I did**
The content of the dashboard is defined as a ConfigMap, under the operator's
namespace.
When kataconfig is deployed, this ConfigMap is copied and installed under
the appropriate namespace (openshift-config-managed).
The resulting dashboard is accessible under the "Observe/Dashboard" section
in the OpenShift console.

**- How to verify it**
Install the operator, and deploy kataconfig as usual.
Once correctly setup, open the web console, move to "Observe/Dashboard".
Open the "Dashboard" dropdown field at the top. A new dashboard must be available, named "Sandboxed Containers".
This Dashboard shows simple graphs on the pods running with sandboxed containers.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -58,6 +58,11 @@ type KataConfigOpenShiftReconciler struct {
 	kataConfig *kataconfigurationv1.KataConfig
 }
 
+const (
+	dashboard_configmap_name      = "grafana-dashboard-sandboxed-containers"
+	dashboard_configmap_namespace = "openshift-config-managed"
+)
+
 // +kubebuilder:rbac:groups=kataconfiguration.openshift.io,resources=kataconfigs;kataconfigs/finalizers,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=kataconfiguration.openshift.io,resources=kataconfigs/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets;replicasets;statefulsets,verbs=get;list;watch;create;update;patch;delete
@@ -242,15 +247,13 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
 func (r *KataConfigOpenShiftReconciler) processDashboardConfigMap() *corev1.ConfigMap {
 
 	r.Log.Info("Creating sandboxed containers dashboard in the OpenShift console")
-	cmName := "grafana-dashboard-sandboxed-containers"
 	cmLabels := map[string]string{
 		"console.openshift.io/dashboard": "true",
 	}
-	cmNamespace := "openshift-config-managed"
 
 	// retrieve content of the dashboard from our own namespace
 	foundCm := &corev1.ConfigMap{}
-	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: "openshift-sandboxed-containers-operator"}, foundCm)
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: dashboard_configmap_name, Namespace: "openshift-sandboxed-containers-operator"}, foundCm)
 	if err != nil {
 		r.Log.Error(err, "could not get dashboard data")
 		return nil
@@ -262,8 +265,8 @@ func (r *KataConfigOpenShiftReconciler) processDashboardConfigMap() *corev1.Conf
 			Kind:       "ConfigMap",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      cmName,
-			Namespace: cmNamespace,
+			Name:      dashboard_configmap_name,
+			Namespace: dashboard_configmap_namespace,
 			Labels:    cmLabels,
 		},
 		Data: foundCm.Data,
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -19,8 +19,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	appsv1 "k8s.io/api/apps/v1"
 	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
 
 	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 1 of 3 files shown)


controllers/openshift_controller.go
```diff
@@ -133,6 +134,26 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 			}
 		}
 
+		cMap := r.processDashboardConfigMap()
+		if ds == nil {
```
pmores, line 138: Should this `ds` be `cMap` really?

controllers/openshift_controller.go
```diff
@@ -218,6 +239,37 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
...
+	}
+
+	return &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "apps/v1",
+			Kind:       "ConfigMap",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cmName,
+			Namespace: cmNamespace,
```
pmores, line 266: So no additional RBAC permission necessary to be able to do this after all?  There is much to learn... ;-)

pmores (changes requested):

controllers/openshift_controller.go
littlejawa: Yes, absolutely. Will fix it - thanks !

controllers/openshift_controller.go
littlejawa: I was surprised too. I launched it for testing, thinking "ok, I'll have an error", and it just worked :-)
Now I didn't check what permissions we would need, compared to what we already have. Maybe we actually had it already for another reason?
For instance, we have get/update permissions on the "namespaces" resources... Would that cover it?

controllers/openshift_controller.go
```diff
@@ -218,6 +239,37 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
...
+		"console.openshift.io/dashboard": "true",
+	}
+	cmNamespace := "openshift-config-managed"
+
+	// retrieve content of the dashboard from our own namespace
+	foundCm := &corev1.ConfigMap{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: "openshift-sandboxed-containers-operator"}, foundCm)
+	if err != nil {
+		r.Log.Error(err, "could not get dashboard data")
+		//return nil
```
littlejawa, line 256: This line must not be commented. I did it for troubleshooting, and did not remove it.
If we leave it, having an error here would result in creating an empty ConfigMap, which actually breaks the "Observe / Dashboard" panel in the OCP console.

controllers/openshift_controller.go
littlejawa: ok, hold on : I realize that in order to install the daemonset, I'm still using the workaround to give "privileged" scc to the operator... when we actually use the required (limited) privileges there, maybe this will just stop working?

controllers/openshift_controller.go
```diff
@@ -218,6 +239,37 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
...
 }
 
+func (r *KataConfigOpenShiftReconciler) processDashboardConfigMap() *corev1.ConfigMap {
+
+	r.Log.Info("Creating sandboxed containers dashboard in the OpenShift console")
+	cmName := "grafana-dashboard-sandboxed-containers"
+	cmLabels := map[string]string{
+		"console.openshift.io/dashboard": "true",
+	}
+	cmNamespace := "openshift-config-managed"
```
bpradipt, line 249: @littlejawa can you please add `openshift-config-managed` and `grafana-dashboard-sandboxed-containers` as constants at the top of the file? We need to do that for other such existing strings as well. But lets start with your code ;)

controllers/openshift_controller.go
bpradipt: I think you could do this because of the privileged SCC. The latest code does have the RBAC rules for namespace updates. But I'm not sure if there is a specific RBAC rule required for configMap creation in different namespace.
Can you try with the latest operator code which sets the SCC ? You don't need to manually add the SCC to the service account.

pmores: IIUC the syntax of ClusterRole `rules:` stanza doesn't allow you to specify a particular namespace for which a permission is valid (although it does have syntax for particular `resourceNames`).  Looking at the current contents of `config/rbac/role.yaml` though, it seems that the permission to create, delete, get, list, patch, update and watch configmaps (among other resource types) is already granted which might explain why you didn't have to add it (should have verified this earlier, sorry!).

Or it's the SCC as Pradipta says. :-)

pmores (approved):

controllers/openshift_controller.go
littlejawa: Ok, so I can confirm it works WITHOUT manually adding the scc.
I used to do that out of habit during my test process, but since I rebased from master, I don't need it anymore - the daemonset and configmap are installed properly, without touching anything...
... it just works ! :-)

controllers/openshift_controller.go
littlejawa: Done !

bpradipt (approved):
/lgtm

bpradipt:
@littlejawa can you please run `make bundle` and check if there are any changes to the bundle manifest? Especially check if the the osc-dashboard.yaml is copied to the `bundle/manifest` directory ?

littlejawa:
> @littlejawa can you please run `make bundle` and check if there are any changes to the bundle manifest? Especially check if the the osc-dashboard.yaml is copied to the `bundle/manifest` directory ?

Hey @bpradipt,
Yes, the manifest is there. Should I have added it as part of this PR ?

PR #168, opened 2022-01-06
KATA-1190: always update mcp during reconcile

jensfr:
We don't update the node selector in the machine config pool
object when a user changes the selector in the KataConfig CR spec.

**- What I did**
What we should do is to make sure that in every reconcile 
the specified node selector matches what we set in the machine
config pool. With this fixed the users settings are applied
to the cluster.

**- How to verify it**
1. label a node with custom-kata=test
2.  create a kataconfig with a node selector kata=test (it shouldn't match the label of the node in step 1)
3. verify that the kata-oc pool is empty
4. edit the kataconfig and set the node selector to kata=test (so that it matches the label from step 1)
5. open the log and check that you see a message "Updating machine config pool"
6. check that the node selector in the spec of the kata-oc machine config pool matches with the one
    from the kataconfig
7. delete the kataconfig and verify everything is removed/uninstalled

update node selector in the custom machine config pool when the node selector in the KataConfig CR is changed

Fixes: https://issues.redhat.com/browse/KATA-1190

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -773,6 +773,17 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 			return ctrl.Result{}, err
 		}
 
+		// Update node selector in machine config pool with value from kataconfig instance
+		r.Log.Info("Updating machine config pool")
+		if foundMcp != nil {
+			*foundMcp.Spec.NodeSelector = *r.kataConfig.Spec.KataConfigPoolSelector
+			err = r.Client.Update(context.TODO(), foundMcp)
+			if err != nil {
+				r.Log.Error(err, "Error when updating MachineConfigPool")
+				return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+			}
+		}
+
 		// Wait till MCP is ready
 		if foundMcp.Status.MachineCount == 0 {
 			r.Log.Info("Waiting till MachineConfigPool is initialized ", "machinePool", machinePool)


controllers/openshift_controller.go
```diff
@@ -773,6 +773,17 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 			return ctrl.Result{}, err
 		}
 
+		// Update node selector in machine config pool with value from kataconfig instance
+		r.Log.Info("Updating machine config pool")
+		if foundMcp != nil {
+			*foundMcp.Spec.NodeSelector = *r.kataConfig.Spec.KataConfigPoolSelector
+			err = r.Client.Update(context.TODO(), foundMcp)
```
bpradipt, line 780: Interesting approach to handle the case of kataconfig update post installation.
I'm wondering if it would make sense to watch for changes to kataconfig in general and take care of updating mutable fields in kataconfig ? Or it's over kill at this moment ?

bpradipt:
Hitting quota issue with cluster creation.

controllers/openshift_controller.go
pmores: @bpradipt I'd say if a KataConfig field is mutable, in principle we should handle its change. Otherwise we're bound to have clients changing fields and finding out they do nothing - not very good long-terms... :-)

@jensfr Would it make sense to check if the selector has _actually_ changed before calling `r.Client.Update()`? Updating an MCP isn't a lightweight operation and unless updating an MCP without actually changing it is optimised (=ignored) by MCO itself we might consider doing it on our side I suppose.

bpradipt: @pmores valid point. Just that I'm not sure if we want all the fields to be mutable and the related consequences.
Regarding your question on checking the selector, afaik it's not needed as the update operation will be a no-op if there are no changes to the object.

jensfr: I was going for the smallest change possible for this fix, but I agree that we should take a closer look at what other changes in kataconfig we should act on. 

Good point @pmores, I was counting on MCO not doing anything if there are no changes in the applied config. Similar to how it says 'no changes' when you apply the same yaml twice from the command-line.

jensfr:
setting to do-not-merge until I figured out the uninstall part

jensfr:
We also don't update when the label is removed from a node. Fixing this properly is more than I originally aimed for with this PR. 
I suggest we merge this one because it does fix the problem of the operator not reconciling when the node selector is changed in the kataconfig. 
For the other issues #170

bpradipt:
> We also don't update when the label is removed from a node. Fixing this properly is more than I originally aimed for with this PR. I suggest we merge this one because it does fix the problem of the operator not reconciling when the node selector is changed in the kataconfig. For the other issues #170

Yeah, lets go ahead with this one for now and work on other scenarios that will require reconciliation on kataconfig changes

bpradipt (approved):
/lgtm
Thanks @jensfr

PR #171, opened 2022-01-19
KATA-1200: use osc_monitor policy for metrics daemonset

jensfr:
To make the daemonset that runs kata-metrics unprivileged
we change the SELinux type to a custom policy that is installed
via the kata-containers rpm.

See https://issues.redhat.com/browse/KATA-1189 for more information
on the new policy.

Fixes: https://issues.redhat.com/browse/KATA-1200


bpradipt (approved):
/lgtm
Thanks @jensfr

PR #175, opened 2022-01-28
enable changing CRI-O log level on kata workers by changing KataConfig

pmores:
To debug some kinds of problems with kata-enabled workers it's helpful
or even necessary to make CRI-O on the workers log in debug mode.  So
far, this was in principle achievable by a user but some knowledge of
internals of Sandboxed Containers Operator and OpenShift itself was
required.

This commit introduces a new 'logLevel' field in KataConfig.spec which
can be used to set easily a corresponding log level on nodes.
Internally, it works by creating or updating a ContainerRuntimeConfig
instance which requests MCO to set the desired log level on relevant
nodes.  The relevant nodes are all masters (more specifically, all
members of the master Machine Config Pool) in case of a converged
cluster, or nodes defined by the OSC-internal 'kata-oc' MCP otherwise.

To support this implentation, this commit gives the OSC operator a
permission to manipulate ContainerRuntimeConfig resources and labels the
'kata-oc' MCP to enable its selection in
ContainerRuntimeConfig.spec.machineConfigPoolSelector.

**- What I did**

**- How to verify it**

### DIFF
--- a/bundle/manifests/kataconfiguration.openshift.io_kataconfigs.yaml
+++ b/bundle/manifests/kataconfiguration.openshift.io_kataconfigs.yaml
@@ -111,6 +111,10 @@ spec:
                 description: KataMonitorImage is used to specify the container image
                   used for kata-monitor
                 type: string
+              logLevel:
+                default: info
+                description: Sets log level on kata-equipped nodes.  Valid values are the same as for `crio --log-level`.
+                type: string
             required:
             - checkNodeEligibility
             - kataMonitorImage
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -61,6 +61,7 @@ type KataConfigOpenShiftReconciler struct {
 const (
 	dashboard_configmap_name      = "grafana-dashboard-sandboxed-containers"
 	dashboard_configmap_namespace = "openshift-config-managed"
+	container_runtime_config_name = "kata-crio-config"
 )
 
 // +kubebuilder:rbac:groups=kataconfiguration.openshift.io,resources=kataconfigs;kataconfigs/finalizers,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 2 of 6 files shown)


controllers/openshift_controller.go
```diff
@@ -314,6 +395,9 @@ func (r *KataConfigOpenShiftReconciler) newMCPforCR() (*mcfgv1.MachineConfigPool
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "kata-oc",
+			Labels: map[string]string{
+				"crio-conf": "",
```
bpradipt, line 399: @pmores two requests:
1. Can you please add a comment mentioning the need for MCP label as ContainerRuntimeConfig expects an MCP label selector
2. How about using the MCP name itself as label - `kata-oc` instead of `crio-conf`. Better make the MCP name as constant and replace it for the entire code.

pmores: @bpradipt 
ad 1) sure!
ad 2) can do. My original thought was that labels are cheap so why not call the label to express what it's for (it's another thing though whether 'crio-conf' is the best way to do that ;-)).  That way, even if, say, we end up with multiple MCPs in the future we'd just label the right one with no other changes required.  But if we don't think we'll be handling multiple MCPs then using the MCP name makes sense, and there's certain consistency in doing so.

bpradipt: @pmores I'm thinking from the admin pov and feel keeping the labels consistent with existing patterns will be useful.  How about we just use this `"pools.operator.machineconfiguration.openshift.io/kata-oc: ""`.
Similar to existing MCP labels, but it's not mandatory. I'll leave it to your choice :-)

bundle/manifests/kataconfiguration.openshift.io_kataconfigs.yaml
```diff
@@ -111,9 +111,14 @@ spec:
...
                   used for kata-monitor
                 type: string
+              logLevel:
+                default: info
+                description: Sets log level on kata-equipped nodes.  Valid values are the same as for `crio --log-level`.
+                type: string
             required:
             - checkNodeEligibility
             - kataMonitorImage
+            - logLevel
```
jensfr, line 121: when we make it required we should bump up the version of kataconfig

controllers/openshift_controller.go
pmores: @bpradipt Alright then, I'll fix it!

bundle/manifests/kataconfiguration.openshift.io_kataconfigs.yaml
pmores: @jensfr Are we fine bumping up the version, are there risks? I don't really have a strong opinion about making it required, I only did it because I saw `checkNodeEligibility` was required as well and because it slightly simplifies handling of the field in code.  For the user it doesn't seem to make much difference - if a required field has a default they don't have to specify it anyway.

bpradipt: @pmores, we learnt it the hard way. Having it `required` will fail migration.  We'll have to make checkNodeEligibility and kataMonitorImage also optional to sync with downstream.  Or we bump up the version.

pmores:
@bpradipt @jensfr I updated this PR with requested changes:
- `KataConfig.spec.logLevel` is no longer required
- the label on 'kata-oc' allowing to select it with a label selector is now called `pools.operator.machineconfiguration.openshift.io/kata-oc`.

bpradipt (approved):
/lgtm

PR #176, opened 2022-02-02
KATA-1288: update kataMonitorImage during reconcile

jensfr:
changes to spec.kataMonitorImage had no effect. This fixes it and
applies whatever is in the spec to the osc-monitor daemon set.

Fixes: https://issues.redhat.com/browse/KATA-1288


bpradipt (approved):
/lgtm
Thanks @jensfr

PR #178, opened 2022-02-04
backport release-1.2: KATA-1190: always update mcp during reconcile

jensfr:
We don't update the node selector in the machine config pool
object when a user changes the selector in the KataConfig CR spec.
    
What we should do is to make sure in every reconcile that
the specified node selector matches what we set in the machine
config pool. With this fixed the users settings are applied
to the cluster.


pmores (approved):

bpradipt (approved):
/lgtm

PR #181, opened 2022-02-10
 KATA-1340: allow re-install on all workers

jensfr:
When we first install with KataConfigPoolSelector set on specific nodes
and later on remove that selector from kataconfig, re-installation on
all worker nodes is never going to happen.

The reason is that the kata-oc machine config pool is never updated
correctly, the old matchLabel key-value pair is never removed.

This patch fixes the problem by adding all matchLabel entries when
kataConfigPoolSelector is non-empty in every reconcile run.  When the
selector is deleted from kataconfig we create a new kata-oc machine
config pool with only our default matchLabels for the worker and kata-oc
role set. The kata-oc entry is required because we always create this
role even when we install on all worker nodes.

The deletetion-and-reinstall also works when a matchExpression is used
in kataconfig instead of matchLabels. This is because machine config
pools nodeSelector always works with matchLabels only.

Fixes: https://issues.redhat.com/browse/KATA-1340

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -883,11 +883,18 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		}
 
 		// Update node selector in machine config pool with value from kataconfig instance
-		r.Log.Info("Updating machine config pool")
+		r.Log.Info("Updating machine config pool name ", "found Mcp name", foundMcp.Name)
 		if foundMcp != nil {
-			if foundMcp.Spec.NodeSelector != nil {
-				foundMcp.Spec.NodeSelector = r.kataConfig.Spec.KataConfigPoolSelector.DeepCopy()
+			foundMcp.Spec.NodeSelector.MatchLabels = make(map[string]string)
+			if r.kataConfig.Spec.KataConfigPoolSelector != nil {
+				for key, value := range r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels {
+					foundMcp.Spec.NodeSelector.MatchLabels[key] = value
+				}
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -782,7 +782,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		// Update node selector in machine config pool with value from kataconfig instance
 		r.Log.Info("Updating machine config pool")
 		if foundMcp != nil {
-			if foundMcp.Spec.NodeSelector != nil {
+			if foundMcp.Spec.NodeSelector != nil && r.kataConfig.Spec.KataConfigPoolSelector != nil {
```
pmores, line 785: @jensfr Could you remind me why we test for `foundMcp.Spec.NodeSelector` being non-nil? Looking at it now, it seems to me that if `r.kataConfig.Spec.KataConfigPoolSelector` is non-nil we should replace `foundMcp.Spec.NodeSelector`'s value no matter the previous value. (?) Also, what if a user had a KataConfigPoolSelector previously but now they want kata on all of their workers so they just delete the selector - is that case handled?

bpradipt (approved):
/lgtm

controllers/openshift_controller.go
jensfr: @pmores I think you're right,the only check neccessary is if KataConfigPoolSelector is not nil. 

The other case is covered because we always create a kata-oc mcp and the default label used is `nodeSelector := &metav1.LabelSelector{MatchLabels: map[string]string{"node-role.kubernetes.io/worker": ""}}`. So when the nodeselector is removed from kataconfig.spec we will use this as default and the kata-oc mcp now contains all workers.

bpradipt (approved):

jensfr:
verified that this is the exact same code as in #182

pmores (approved):

PR #182, opened 2022-02-10
KATA-1340: allow re-install on all workers

jensfr:
[KATA-1340](https://issues.redhat.com/browse/KATA-1340): allow re-install on all workers

When we first install with KataConfigPoolSelector set on specific nodes
and later on remove that selector from kataconfig, re-installation on
all worker nodes is never going to happen.

The reason is that the kata-oc machine config pool is never updated
correctly, the old matchLabel key-value pair is never removed.

This patch fixes the problem by adding all matchLabel entries when
kataConfigPoolSelector is non-empty in every reconcile run.  When the
selector is deleted from kataconfig we create a new kata-oc machine
config pool with only our default matchLabels for the worker and kata-oc
role set. The kata-oc entry is required because we always create this
role even when we install on all worker nodes.

The deletetion-and-reinstall also works when a matchExpression is used
in kataconfig instead of matchLabels. This is because machine config
pools nodeSelector always works with matchLabels only.

Fixes: https://issues.redhat.com/browse/KATA-1340
Suggested-by: Pradipta Banerjee <bpradipt@redhat.com>

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -780,11 +780,18 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		}
 
 		// Update node selector in machine config pool with value from kataconfig instance
-		r.Log.Info("Updating machine config pool")
+		r.Log.Info("Updating machine config pool name ", "found Mcp name", foundMcp.Name)
 		if foundMcp != nil {
-			if foundMcp.Spec.NodeSelector != nil {
-				foundMcp.Spec.NodeSelector = r.kataConfig.Spec.KataConfigPoolSelector.DeepCopy()
+			foundMcp.Spec.NodeSelector.MatchLabels = make(map[string]string)
+			if r.kataConfig.Spec.KataConfigPoolSelector != nil {
+				for key, value := range r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels {
+					foundMcp.Spec.NodeSelector.MatchLabels[key] = value
+				}
+				foundMcp.Spec.NodeSelector.MatchExpressions = r.kataConfig.Spec.KataConfigPoolSelector.MatchExpressions
 			}
+			foundMcp.Spec.NodeSelector.MatchLabels["node-role.kubernetes.io/worker"] = ""
+			foundMcp.Spec.NodeSelector.MatchLabels["node-role.kubernetes.io/kata-oc"] = ""
+
 			err = r.Client.Update(context.TODO(), foundMcp)
 			if err != nil {
 				r.Log.Error(err, "Error when updating MachineConfigPool")


bpradipt (approved):
/lgtm

controllers/openshift_controller.go
```diff
@@ -780,11 +780,22 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
+		// When user deletes kataConfigPoolSelector, reset matchLabels in mcp to our defaults
+		// node-role.kubernetes.io/worker and kata-oc
+		r.Log.Info("Updating machine config pool name ", "found Mcp name", foundMcp.Name)
 		if foundMcp != nil {
-			if foundMcp.Spec.NodeSelector != nil {
-				foundMcp.Spec.NodeSelector = r.kataConfig.Spec.KataConfigPoolSelector.DeepCopy()
+			if r.kataConfig.Spec.KataConfigPoolSelector != nil {
+				for key,val := range r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels {
+					foundMcp.Spec.NodeSelector.MatchLabels[key] = val
+				}
```
pmores, line 790: @jensfr What happens if `foundMcp.Spec.NodeSelector.MatchLabels` already contains a key that's not in the current`r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels`, don't we end up with a stale key still hanging around? Shouldn't we clear `foundMcp.Spec.NodeSelector.MatchLabels` first, like we do in the other branch (IOW shouldn't we move the line 793 above the branching on line 787?)?

jensfr: @pmores yes, I think you're right. I think we can get rid of the else branch and always create a new MatchLabels, initialize from scratch and if KataConfigPoolSelector is not empty we add those key-value pairs as well.

pmores: @jensfr Yeah.  In fact, if we can rely on Kubernetes/MCO to do nothing when applying a resource that hasn't actually changed (I know we can in general but I'm not sure if this still holds even when e.g. the label set remains the same but is reordered) the simplest way might be just to ignore the current contents of `foundMcp.Spec.NodeSelector` altogether and just
- copy `r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels` to `foundMcp.Spec.NodeSelector.MatchLabels`
- copy `r.kataConfig.Spec.KataConfigPoolSelector.MatchExpressions` to `foundMcp.Spec.NodeSelector.MatchExpressions`
- if both `foundMcp.Spec.NodeSelector.MatchLabels` and `foundMcp.Spec.NodeSelector.MatchExpressions` end up empty at this point, run lines 794 and 795 of the current patch.

Am I understanding correctly?

controllers/openshift_controller.go
```diff
@@ -780,11 +780,18 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
-				foundMcp.Spec.NodeSelector = r.kataConfig.Spec.KataConfigPoolSelector.DeepCopy()
+			foundMcp.Spec.NodeSelector.MatchLabels = make(map[string]string)
+			if r.kataConfig.Spec.KataConfigPoolSelector != nil {
+				for key, value := range r.kataConfig.Spec.KataConfigPoolSelector.MatchLabels {
+					foundMcp.Spec.NodeSelector.MatchLabels[key] = value
+				}
+				foundMcp.Spec.NodeSelector.MatchExpressions = r.kataConfig.Spec.KataConfigPoolSelector.MatchExpressions
 			}
+			foundMcp.Spec.NodeSelector.MatchLabels["node-role.kubernetes.io/worker"] = ""
+			foundMcp.Spec.NodeSelector.MatchLabels["node-role.kubernetes.io/kata-oc"] = ""
```
pmores, line 793: @jensfr Alright, so I was wrong about these two in the final bullet point above.  We add them always so no matter what the user does in `kataConfigPoolSelector`, we won't ever install kata on nodes that aren't labelled with both `worker` and `kata-oc` in the first place.  Good to know!

PR #187, opened 2022-04-29
KATA-1321: add versions of kata and qemu to must-gather output

pmores:
The version information is collected separately from individual nodes and
stored at the root of each node's respective subtree of must-gather output.


bpradipt (approved):
Much needed :-)

pmores:
Thanks! How do we go about reviewing & merging while Jens is gone?

PR #189, opened 2022-05-05
Adding a PrometheusRule for kata monitor

littlejawa:
We want to provide some information back through telemetry, but there is limitation on how much data can be sent.
As kata-monitor report information per node, any metric we want to expose would be multiplied by the number of nodes in the cluster, making us go above the limit in some situations.

We need to sum information up, and expose total numbers at the cluster level, and not individual per-node information.

**- What I did**
Adding a PrometheusRule to have prometheus take the number of running shim per node, make the total of it, and expose it as a new metric.
This rule can be extended in the future to add additional metrics as we need.

### DIFF
--- a/config/kata-monitor/kata-monitor-prometheus-rules.yaml
+++ b/config/kata-monitor/kata-monitor-prometheus-rules.yaml
@@ -1,0 +1,11 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: prometheus-sandboxed-containers-rules
+  namespace: openshift-sandboxed-containers-operator
+spec:
+  groups:
+    - name: kata_monitor_rules
+      rules:
+      - record: cluster:kata_monitor_running_shim_count:sum
+        expr: sum(kata_monitor_running_shim_count)
--- a/config/kata-monitor/kustomization.yaml
+++ b/config/kata-monitor/kustomization.yaml
@@ -2,3 +2,4 @@ resources:
 - kata-monitor-service.yaml
 - kata-monitor-servicemonitor.yaml
 - osc-dashboard.yaml
+- kata-monitor-prometheus-rules.yaml


bundle/manifests/prometheus-sandboxed-containers-rules_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: prometheus-sandboxed-containers-rules
+spec:
+  groups:
+  - name: kata_monitor_rules
+    rules:
+    - expr: sum(kata_monitor_running_shim_count)
+      record: cluster:kata_monitor_running_shim_count:sum
```
bpradipt, line 10: Hi @littlejawa, can you please confirm if this yaml was generated by running `make bundle` after adding the `config/kata-monitor/kata-monitor-prometheus-rules.yaml` ?

littlejawa: I'm not sure which command generated it, but I found it in my "bundle" folder after building / testing the operator.

Is it wrong?

bpradipt: No it's not wrong, but I expected the bundle manifest and the config manifest to be exactly the same. So I was curious :-)

simonpasquier (commented):
:wave: I'm from @openshift/openshift-team-monitoring 
My remark is a bit off-topic but `kata_monitor_running_shim_count` doesn't follow Prometheus good practices: it looks like it's a gauge metric while the `_count` suffix is reserved for histograms and summaries (see https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations). It's not a big issue but it would be nice if it could be fixed upstream :)

littlejawa:
> wave I'm from @openshift/openshift-team-monitoring My remark is a bit off-topic but `kata_monitor_running_shim_count` doesn't follow Prometheus good practices: it looks like it's a gauge metric while the `_count` suffix is reserved for histograms and summaries (see https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations). It's not a big issue but it would be nice if it could be fixed upstream :)

Thanks @simonpasquier - you're right, it's probably a gauge.
This metric comes from a (long) list exported by one of our components. I suspect other items in there would benefit from a rename. I doubt this can be done quickly though - I expect pushbacks if other users of the upstream component are already relying on them.
I'll keep that in mind though.

As this metric is bound to be exported via telemetry, do you think we should rename the record rule (`cluster:kata_monitor_running_shim_count:sum`)? I feel it's supposed to keep the name of the underlying metrics for consistency, but at least it would appear correct when used in telemetry.
What do you think?

bundle/manifests/prometheus-sandboxed-containers-rules_monitoring.coreos.com_v1_prometheusrule.yaml
littlejawa: Maybe what it is trying to tell me is that I should fix the indentation in the config manifest :-)
Also: it removes the namespace... As it's installed by the operator, maybe I don't need to specify it in the config manifest either? 

Note that I did the same (have the namespace specified) in the ServiceMonitor config manifest, and it was removed from the bundle too.

bpradipt:
> > wave I'm from @openshift/openshift-team-monitoring My remark is a bit off-topic but `kata_monitor_running_shim_count` doesn't follow Prometheus good practices: it looks like it's a gauge metric while the `_count` suffix is reserved for histograms and summaries (see https://prometheus.io/docs/practices/histograms/#count-and-sum-of-observations). It's not a big issue but it would be nice if it could be fixed upstream :)
> 
> Thanks @simonpasquier - you're right, it's probably a gauge. This metric comes from a (long) list exported by one of our components. I suspect other items in there would benefit from a rename. I doubt this can be done quickly though - I expect pushbacks if other users of the upstream component are already relying on them. I'll keep that in mind though.
> 
> As this metric is bound to be exported via telemetry, do you think we should rename the record rule (`cluster:kata_monitor_running_shim_count:sum`)? I feel it's supposed to keep the name of the underlying metrics for consistency, but at least it would appear correct when used in telemetry. What do you think?

@simonpasquier any comments on the above question from @littlejawa ?

simonpasquier:
sorry for the lag :) 

`cluster:kata_monitor_running_shim_count:sum` is fine by me.

snir911 (approved):

bpradipt (approved):

PR #191, opened 2022-05-13
Remove scc on uninstall

gkurz:
Let's make sure that the controller removes the SCC it has created when it exits.
This ensures that the SCC doesn't stay around when OSC is uninstalled or when
installation fails.

Fixes: https://issues.redhat.com/browse/KATA-156

### DIFF
--- a/main.go
+++ b/main.go
@@ -104,6 +104,7 @@ func runManager() int {
 			setupLog.Error(err, "unable to create SCC")
 			return 1
 		}
+		defer deleteScc(context.TODO(), mgr)
 
 		setupLog.Info("created SCC")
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -152,6 +153,13 @@ func createScc(ctx context.Context, mgr manager.Manager) error {
 	return err
 }
 
+func deleteScc(ctx context.Context, mgr manager.Manager) error {
```
c3d, line 156: Notice that this function returns an `error` that is ignored.

c3d (commented):
Looks like a necessary first step, but as usual, problems with nested errors (e.g. error in a defer)

main.go
```diff
@@ -146,6 +153,13 @@ func createScc(ctx context.Context, mgr manager.Manager) error {
...
+
```
bpradipt, line 157: I think we should check for the presence of existing KataConfig and proceed with delete when no KataConfig exist.

PR #195, opened 2022-05-25
Rename setRuntimeClass to createRuntimeClass and change return values

bpradipt:
This is a cleanup patch to rename method name to better reflect the functionality. Additionally only returning error.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -130,7 +130,7 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 			//The DaemonSet (DS) should be ideally created after the required SeLinux policy is installed on the
 			//node. One of the ways to ensure this is to check for the existence of "kata" runtimeclass before
 			//creating the DS
-			//Alternatively we can create the DS post execution of setRuntimeClass()
+			//Alternatively we can create the DS post execution of createRuntimeClass()
 			if k8serrors.IsNotFound(err) && r.kataConfig.Status.RuntimeClass == "kata" {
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -936,13 +936,17 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 	if mcfgv1.IsMachineConfigPoolConditionTrue(foundMcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdated) &&
 		foundMcp.Status.ObservedGeneration > r.kataConfig.Status.BaseMcpGeneration &&
 		foundMcp.Status.UpdatedMachineCount == foundMcp.Status.MachineCount {
-		r.Log.Info("set runtime class")
+		r.Log.Info("create runtime class")
 		r.kataConfig.Status.InstallationStatus.IsInProgress = "false"
-		return r.setRuntimeClass()
+		err := r.createRuntimeClass()
+		if err != nil {
+			return reconcile.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
```
gkurz, line 943: @bpradipt Sorry but I'm still confused here... this seems to change the behavior of `processKataConfigInstallRequest()` since it would have return `ctrl.Result{}, err` before this patch, right ? If so, this patch seems to be more than just changing the signature or I'm missing something(TM) ?

bpradipt: Ah I see what you mean. I'll not add the "Requeue" params. I wanted to avoid an immediate reconciliation. I'll change this in a follow-on PR later on.

gkurz (approved):
LGTM. THanks @bpradipt !

PR #197, opened 2022-06-03
Fix some error paths in the controller

gkurz:
The recently introduced `createScc()` is missing an error path. Same issue exists in `createRuntimeClass()`. Fix both.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -632,7 +632,11 @@ func (r *KataConfigOpenShiftReconciler) createScc() error {
 
 	foundScc := &secv1.SecurityContextConstraints{}
 	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: scc.Name}, foundScc)
-	if err != nil && k8serrors.IsNotFound(err) {
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			return err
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -686,12 +686,14 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass() error {
...
-			return err
-		}
+	if err != nil && ! k8serrors.IsNotFound(err) {
+		return err
+	}
+
+	r.Log.Info("Creating a new RuntimeClass", "rc.Name", rc.Name)
+	err = r.Client.Create(context.TODO(), rc)
+	if err != nil {
+		return err
```
pmores, line 696: @gkurz Won't this however create a new SCC even if one was found?

gkurz: oops of course...

pmores: Then the other commit might have the same problem I suspect! :-)

gkurz: Exactcly :-D I've fixed both and repushed.

pmores (approved):
Thanks @gkurz !

bpradipt (approved):
/lgtm
Thanks @gkurz

PR #199, opened 2022-06-21
Update doc links to latest OCP version

gkurz:
The operator's manifest and documentation still has references to OCP 4.9. We could possibly bump to 4.10 them since OCP 4.10 is GA, but it looks simpler that the master branch HEAD always points to the `latest` (as already done is some other places).


pmores (approved):
Thanks @gkurz !

bpradipt (approved):

PR #200, opened 2022-06-22
update doc links to OCP 4.11

pmores:
OCP 4.11 is the version that corresponds to OSC 1.3.


gkurz:
@pmores Also, OSC 1.3 will be documented in OCP 4.11, not 4.10.

pmores:
@gkurz As for the `kataconfig_types.go` reference, I'm not sure about that - it seems to be more about a general concept than a version-specific procedure. The other two are handled by this PR I believe, am I overlooking something?

gkurz:
> @gkurz As for the `kataconfig_types.go` reference, I'm not sure about that - it seems to be more about a general concept than a version-specific procedure. 

+1

>The other two are handled by this PR I believe, am I overlooking something?

I'm still seeing them even with this PR applied.

pmores:
Oh you mean the CRD, not CSV. Right, we haven't talked about those and frankly, if they are to be changed as well (which I'm not sure about if I take context into account but it does make some sense) then I'd rather remove the links altogether as a fix than chase them manually every release.

gkurz:
> Oh you mean the CRD, not CSV. Right, we haven't talked about those and frankly, if they are to be changed as well (which I'm not sure about if I take context into account but it does make some sense) then I'd rather remove the links altogether as a fix than chase them manually every release.

Please go ahead :smile:

pmores:
> Oh you mean the CRD, not CSV. Right, we haven't talked about those and frankly, if they are to be changed as well (which I'm not sure about if I take context into account but it does make some sense) then I'd rather remove the links altogether as a fix than chase them manually every release.

@bpradipt Are you fine that as well, what do you think?

bpradipt:
> Oh you mean the CRD, not CSV. Right, we haven't talked about those and frankly, if they are to be changed as well (which I'm not sure about if I take context into account but it does make some sense) then I'd rather remove the links altogether as a fix than chase them manually every release.

@pmores are you referring to If you Are you referring to https://github.com/openshift/sandboxed-containers-operator/blob/master/bundle/manifests/kataconfiguration.openshift.io_kataconfigs.yaml#L61 ?
Btw this was added to help the user understand the NFD requirements when creating the Kataconfig from the UI. Since the NFD installation is a separate activity. And the instructions can change over time, so we'll need to keep updating the link for every release.

To avoid effort on our side, we can look at automating these changes via Makefile.

bpradipt (approved):

PR #201, opened 2022-06-24
Fix SCC in OSC 1.2

gkurz:
This brings the fix of PR #196 to the 1.2 release, along with an extra patch to an already created SCC.
This will simplify OCP upgrades.


bpradipt (approved):
@gkurz Thanks for tackling this issue.
While I was thinking about a way to trigger reconcile and then make the changes to SCC in the reconcile loop, I feel given the current 1.2.x code this is the simplest approach.
Sadly we don't have CI setup for release-* branches. So we will have to test it manually for now.

pmores (approved):

PR #202, opened 2022-06-28
Fix 1.2-style SCC

gkurz:
This patches an existing SCC so that it no longer breaks MCO, i.e. applying
the same changes as described in MR #196. This will allow seamless upgrade
of OCP (extra manual steps are required otherwise).


bpradipt (approved):
/lgtm
Thanks @gkurz

Issue #209, opened 2022-08-17
Modified files after building


cpmeadors:
After running docker-build, bundle-build, catalog-build  There are a number of modified files.  Are these safe if they accidentally get checked in?  Should they be getting updated? Should they be updated in the repo?

bundle.Dockerfile
config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
config/rbac/role.yaml
config/webhook/manifests.yaml
bundle/manifests/kataconfiguration.openshift.io_kataconfigs.yaml
bundle/metadata/annotations.yaml

bpradipt:
Files under `config/` typically get updated if there are changes to the API, sdk, or when building your own operator image.
The changes due to API or sdk are core code related changes and hence needs to be tracked. 

Files under `bundle/` are generated from `config/` on execution of `make bundle`. The reason for checking this is to get a repeatable build of the bundle and catalog when using the upstream code without regenerating the bundle. We can think about completely removing the `bundle` directory.

One option is to provide git instructions to not track the local changes if needed by the developer.

cpmeadors:
Based on the operator-sdk docs, "make generate" creates:

api/v1/zz_generated.deepcopy.go

"make manifests" creates: 

config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
config/rbac/role.yaml
config/webhook/manifests.yaml

"make bundle" generates

bundle.Dockerfile
bundle/metadata/annotations.yaml
bundle

Why should any of these files be tracked?  The files that define the generation of them are tracked.

PR #211, opened 2022-08-30
Remove generated files and update docs to generate files as they were

cpmeadors:
There are files being tracked that are completely generated by make targets.  Fewer files makes maintaining the code easier. 

**- What I did**
Based on the operator-sdk docs [1], I have deleted the generated files from the repo.  One file required an additional command line change to generate the file exactly as it was before deleting.  That change was added to the DEVELOPMENT.md file.  I also add the relevant entries in the .gitignore to keep the files from being added back.

[1] https://sdk.operatorframework.io/docs/building-operators/golang/tutorial/

This fixes: #209 and #208 

**- How to verify it**

If you execute the make file targets all files that were deleted should be recreated exactly the same:

make docker-build
make bundle
make bundle-build CHANNELS=candidate

Generated files have been deleted from the repo


docs/DEVELOPMENT.md
```diff
@@ -48,7 +48,7 @@ make docker-push
 
 ## Building Operator bundle image
 ```
-make bundle
+make bundle CHANNELS=candidate
```
jensfr, line 51: candidate is also what we use in our example subscription in config/samples/deploy, so I think this is a good change

jensfr (approved):
Looks good to me! Thank you Cameron!

pmores (approved):
Thanks @cpmeadors !

pmores:
Merging is blocked though because CI is failing which seems to be due to missing `DeepCopyObject()` methods in `KataConfig`, `KataConfigList` and possibly others.  I suspect that our current build sequence is missing something to regenerate `zz_generated.deepcopy.go` which is removed by this PR.

pmores:
What's missing seems to be a `make generate` invocation. Unfortunately I've no idea ATM how CI works but from looking at the `Makefile`, the `generate` target is also invoked by `test`, `build` and `run` targets as a prerequisite. So it looks like none of these targets are invoked by CI.

PR #216, opened 2022-09-14
Remove kata monitor image

pmores:
This PR removes spec.kataMonitorImage from the KataConfig CR.  kata-monitor container image URL is basically a technical implementation detail of OSC and as such has no good place in KataConfig.

The OSC controller now gets the kata-monitor image URL from an environment variable KATA_MONITOR_IMAGE instead of a KataConfig instance.  The default value is now part of the manager container template in OSC controller-manager Deployment.

### DIFF
--- a/api/v1/kataconfig_types.go
+++ b/api/v1/kataconfig_types.go
@@ -35,10 +35,6 @@ type KataConfigSpec struct {
 	// +kubebuilder:default:=false
 	CheckNodeEligibility bool `json:"checkNodeEligibility"`
 
-	// KataMonitorImage is used to specify the container image used for kata-monitor
-	// +kubebuilder:default:="quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-monitor:latest"
-	KataMonitorImage string `json:"kataMonitorImage"`
-
 	// Sets log level on kata-equipped nodes.  Valid values are the same as for `crio --log-level`.
 	// +kubebuilder:default:="info"
 	LogLevel string `json:"logLevel,omitempty"`
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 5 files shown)


pmores:
This is core functionality of the change.  It works but please comment on any additional formal aspects if appropriate.  In particular, CRD version change might be required.  I haven't included it in this PR as submitted since I'm not at all clear if and how to handle that.

controllers/openshift_controller.go
```diff
@@ -277,7 +278,18 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
...
-	r.Log.Info("Creating monitor DaemonSet with image file: " + r.kataConfig.Spec.KataMonitorImage)
+	kataMonitorImage := os.Getenv("KATA_MONITOR_IMAGE")
+	if len(kataMonitorImage) == 0 {
+		// kata-monitor image URL is generally impossible to verify or sanitise,
+		// with the empty value being pretty much the only exception where it's
+		// fairly clear what good it is.  If we can only detect a single one
+		// out of an infinite number of bad values, we choose not to return an
+		// error here (giving an impression that we can actually detect errors)
+		// but just log this incident and plow ahead.
+		r.Log.Info("KATA_MONITOR_IMAGE env var is unset or empty, kata-monitor pods will not run")
```
bpradipt, line 289: love this explanation :-)

api/v1/kataconfig_types.go
```diff
@@ -35,10 +35,6 @@ type KataConfigSpec struct {
 	// +kubebuilder:default:=false
 	CheckNodeEligibility bool `json:"checkNodeEligibility"`
 
-	// KataMonitorImage is used to specify the container image used for kata-monitor
-	// +kubebuilder:default:="quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-monitor:latest"
-	KataMonitorImage string `json:"kataMonitorImage"`
-
```
bpradipt, line 41: Since this modifies the CRD, anything required to handle upgrades from a previously created  CR having this value ?

gkurz: `KataMonitorImage` should also be dropped from the controller test here:

https://github.com/openshift/sandboxed-containers-operator/blob/master/controllers/openshift_controller_test.go#L281

controllers/openshift_controller.go
pmores: Well you can see I really wanted to find a way to return an error that made sense but couldn't. ;-)

The decision isn't clear cut so to prevent someone (probably myself) thinking three months down the line, "what the hell did they think they were doing?", I thought I'd document it.

api/v1/kataconfig_types.go
pmores: @gkurz good catch, thanks!

pmores:
Pushed a new version that includes corresponding changes to tests.

api/v1/kataconfig_types.go
pmores: Does anybody know how to address @bpradipt 's concern above? @jensfr perhaps?

gkurz: > Does anybody know how to address @bpradipt 's concern above? @jensfr perhaps?

It seems that the usual process when changing the CRD is to bump the api version and setup conversion webhooks to be able to expose an older object through the new api. This being said, I don't know if we really need all that in this case. If we have OSC 1.3 and a kataconfig already deployed, we want the upgrade to deploy the new monitor images (see KATA-1650) and we thus don't really care for any pre-existing value in the kataconfig.

bpradipt (approved):
/lgtm
Thanks @pmores

gkurz:
This will need rebase since PR #211 is already merged, but anyway:

/lgtm

Thanks @pmores !

gkurz (approved):

api/v1/kataconfig_types.go
jensfr: so what happens when you try to create an 'old' kataconfig with monitorImage set after this change to the CRD is merged?

jensfr:
/override  ci/prow/4.9-ci-index-sandboxed-containers-operator-bundle  ci/prow/4.9-images     ci/prow/4.9-sandboxed-containers-operator-e2e     ci/prow/ci-index-sandboxed-containers-operator-bundle     ci/prow/images     ci/prow/sandboxed-containers-operator-e2e     pull-ci-openshift-sandboxed-containers-operator-master-4.9-ci-index-sandboxed-containers-operator-bundle

PR #219, opened 2022-09-22
generate files in dockerfiles

cpmeadors:
These changes add multistage docker files for both the operator and the operator bundle images. The builder stage uses the make targets to generate the necessary files to build the two images.  This should fix the Prow CI image build issues and it should allow the copy of the bundle files in the midstream repo to be removed.  Ultimately it means the same process is being use to build the images no matter what is building them.

Outstanding issues are the need for "go mod vendor" and the hard coding of the VERSION for the bundle.


bundle.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+FROM scratch
+
+# Core bundle labels.
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=sandboxed-containers-operator
+LABEL operators.operatorframework.io.bundle.channels.v1=alpha
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.19.0+git
```
jensfr, line 33: we're using operator-sdk 1.23 above to build. I don't think it has any effect but we could update it here to 1.23 as well.

bundle.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN VERSION=1.3.1 make bundle
+
+FROM scratch
+
+# Core bundle labels.
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=sandboxed-containers-operator
+LABEL operators.operatorframework.io.bundle.channels.v1=alpha
```
jensfr, line 32: should be stable-1.3?

jensfr (approved):
lgtm, except for two small nits

Dockerfile
```diff
@@ -3,17 +3,12 @@ FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.10 AS
...
+RUN go mod vendor
 
-# Copy the go source
-COPY main.go main.go
-COPY api/ api/
-COPY controllers/ controllers/
+RUN make generate
 
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -mod=mod -o manager main.go
```
pmores, line 14: I'd prefer replacing this with simply `RUN make build` so that we don't have two places to maintain where the `manager` binary is built.

Note that doing so also requires adjusting the `COPY` command below to `COPY --from=builder /workspace/bin/manager .` This would also let us get rid of the `RUN make generate` layer above since `generate` is a prerequisite of `build`.

Dockerfile
```diff
@@ -3,17 +3,12 @@ FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.10 AS
 
 WORKDIR /workspace
 
-# Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
-# cache deps before building and copying source so that we don't need to re-download as much
-# and so that source changes don't invalidate our downloaded layer
+COPY ./ ./
```
pmores, line 6: I'm not sure about this change. I'm afraid it will greatly and unnecessarily increase the size of the resulting layer which might not be an issue on build machines but can be an impediment for development in my experience.  I know that even without this change, when I worked on this Dockerfile it accumulated over a hundred of gigs in layer cache during just a couple of hour of work so I had to purge periodically, with the obvious bad effects on the next build time.

pmores:
Thanks @cpmeadors and kudos for figuring this out! I left a bunch of comments but also overall, I'd like to make sure that the efficiency of the originals (esp. in case of `Dockerfile`) in terms of layering and disk space requirements is preserved.

Dockerfile
cpmeadors: Yes, you are correct.  I will fix this.

Dockerfile
cpmeadors: I think this is more efficient than doing multiple COPY to most of the dirs.  Fewer commands in a dockerfile is generally a good thing.  I find I have to purge docker caches/images a lot when I do any dockerfile development

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: openshift-sandboxed-containers-operator
```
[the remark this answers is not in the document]
jensfr: I don't think it matters much as it is being overwritten typically when you run make IMG=xxx deploy.  Maybe setting it to quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:latest would make sense as the default?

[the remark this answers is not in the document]
cpmeadors: OK. this is again controlled by VARS in the makefile.  According to the operator-sdk docs and makefile comments, we should set IMAGE_TAG_BASE to repo+imagename and then set IMG ?= $(IMAGE_TAG_BASE):$(VERSION). 
This should represent our default operator repo, image name, and image version.  The question is what should that be.  I feel like it should be the production build.  If you need a local build, you should set the IMG as mentioned in our docs.  Does this logic make sense?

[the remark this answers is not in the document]
cpmeadors: The idea is that if you do a make docker-build or bundle-build with no options you would get the production version.

Makefile
```diff
@@ -231,3 +233,31 @@ catalog-build: opm ## Build a catalog image.
 .PHONY: catalog-push
 catalog-push: ## Push a catalog image.
 	$(MAKE) docker-push IMG=$(CATALOG_IMG)
+
```
jensfr, line 236: this chunk should go away when you rebase against master

.gitignore
```diff
@@ -85,5 +85,4 @@ config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
 config/rbac/role.yaml
 config/webhook/manifests.yaml
 # make bundle files
-bundle.Dockerfile
 bundle/
```
gkurz, line 88: Missing newline since commit 846b467294adb3aa0bd23294389339835d7162a4. Maybe fix this as well while here ?

Makefile
gkurz: > this chunk should go away when you rebase against master

Yes but there is more to it. See below.

Makefile
```diff
@@ -231,3 +233,31 @@ catalog-build: opm ## Build a catalog image.
...
+	$(RM) api/v1/zz_generated.deepcopy.go
+
+.PHONY: test-clean
+test-clean: ## Clean generated test files
+	$(RM) cover.out
+
+.PHONY: bundle-clean
+bundle-clean: ## Clean generated bundle files
+	$(RM) -r bundle
+	$(RM) bundle.Dockerfile
```
gkurz, line 256: AFAICT bundle.Dockerfile is being re-added to the repo by f3329b7d5cc3651d7596288537dddc755c9656bb. The `bundle-clean` target should stop removing it.

bundle.Dockerfile
```diff
@@ -0,0 +1,52 @@
+# Use OpenShift golang builder image
```
gkurz, line 1: I'm a bit confused actually. With this PR checked out I'm seeing :

```
greg@bahia:[generate-files-in-dockerfile]sandboxed-containers-operator$ git grep bundle.\*Dockerfile
Makefile:       docker build -f bundle-custom.Dockerfile -t $(BUNDLE_IMG) .
Makefile:       $(RM) bundle.Dockerfile
```

Is `bundle.Dockerfile` being used at all ?

bundle.Dockerfile
gkurz: Depending on the outcome of [my earlier comment](https://github.com/openshift/sandboxed-containers-operator/pull/219/files/3821f2a206321b806f4179cf6ae323cbb13daf7b#r989812670) this change might be made on `bundle-custom.Dockerfile` instead.

pmores:
Thanks, but why do we now have 21 commits? If this is not a target branch mess up I think this needs some clean-up and squashing.

jensfr:
I think bundle.Dockerfile can be removed as the Makefile is now using bundle-custom.Dockerfile

Makefile
```diff
@@ -29,7 +31,7 @@ BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 #
 # For example, running 'make bundle-build bundle-push catalog-build catalog-push' will build and push both
 # example.com/memcached-operator-bundle:$VERSION and example.com/memcached-operator-catalog:$VERSION.
-IMAGE_TAG_BASE ?= quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator
+IMAGE_TAG_BASE ?= registry-proxy.engineering.redhat.com/rh-osbs/openshift-sandboxed-containers-operator
```
jensfr, line 34: why not leave it at quay.io/... and just override IMAGE_TAG_BASE=registry-proxy.... downstream?

gkurz:
> Thanks, but why do we now have 21 commits? If this is not a target branch mess up I think this needs some clean-up and squashing.

Agreed. This PR should also be rebased on current main since https://github.com/openshift/sandboxed-containers-operator/pull/221 got merged.

> I think bundle.Dockerfile can be removed as the Makefile is now using bundle-custom.Dockerfile. And then the bundle-clean target also needs to remove bundle-custom.Dockerfile

AFAICS `bundle.Dockerfile` is still generated by `make bundle`:

```
cd config/manager && /home/greg/Work/openshift/sandboxed-containers-operator/bin/kustomize edit set image controller=quay.io/rhgkurz/openshift-sandboxed-containers-operator
/home/greg/Work/openshift/sandboxed-containers-operator/bin/kustomize build config/manifests | operator-sdk generate bundle -q --overwrite --version 1.3.1 --channels="stable-1.3" --default-channel="stable-1.3"
INFO[0000] Creating bundle.Dockerfile                   
```

Just for my understanding : we're deliberately ignoring the generated `bundle.Dockerfile` and we're only to use the custom one from git instead from now on, right ? Not sure to fully understand the rationale for this change. Some explainations in the commit changelog would be appreciated.

And the `bundle-clean` target should also continue to remove `bundle.Dockerfile` no matter what, not `bundle-custom.Dockerfile` since it belongs to the repo.

jensfr (changes requested):
see comments

bundle-custom.Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v3
+
+# Labels for testing.
+LABEL operators.operatorframework.io.test.mediatype.v1=scorecard+v1
+LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
+
+# Copy files to locations specified by labels.
+COPY --from=builder /workspace/bundle/manifests /manifests/
+COPY --from=builder /workspace/bundle/metadata /metadata/
+COPY --from=builder /workspace/bundle/tests/scorecard /tests/scorecard/
```
gkurz, line 52: Trailing newline here also please.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: registry-proxy.engineering.redhat.com/rh-osbs/openshift-sandboxed-containers-operator
```
gkurz, line 15: Like @jensfr commented elsewhere, exposing `registry-proxy.engineering.redhat.com/rh-osbs/openshift-sandboxed-containers-operator` looks like a downstream change.

Makefile
pmores: @gkurz hmm, `bundle.Dockerfile` has always been in the repo AFAIK. It shouldn't be removed by clean, I agree.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
...
+  newTag: 1.3.1
```
gkurz, line 16: So we need to bump the version here as well, not just in `VERSION` ?

pmores:
> Just for my understanding : we're deliberately ignoring the generated `bundle.Dockerfile` and we're only to use the custom one from git instead from now on, right ? Not sure to fully understand the rationale for this change. Some explainations in the commit changelog would be appreciated.

That's true, and until that's added I think I can explain at least some of the rationale (I haven't realised that the original `bundle.Dockerfile` was generated until now which shows that my understanding isn't quite complete either).

The generated `bundle.Dockerfile` is basically just a couple of `LABEL`s and a few `COPY` commands just to get stuff inside `bundle/` into the bundle image.  That stopped being adequate the moment we removed generated files (that includes `bundle/*`) from the repository. Since they don't come from the repo anymore, the contents of `bundle/` have to be generated first by `bundle.Dockerfile`.

I believe/assume that's the reason for `bundle-custom.Dockerfile` - essentially a `bundle.Dockerfile` with the added ability to generate generated files first.
 
> And the `bundle-clean` target should also continue to remove `bundle.Dockerfile` no matter what, not `bundle-custom.Dockerfile` since it belongs to the repo.

Yes, you're right, now I understand, `bundle.Dockerfile` should indeed be cleaned up (or, better yet, never generated in the first place if operator-sdk offers such an option) as the custom Dockerfile is always used to build the bundle image.

Dockerfile
```diff
@@ -3,7 +3,16 @@ FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.10 AS
...
-COPY ./ ./
+COPY Makefile Makefile
+COPY hack hack/
+COPY PROJECT PROJECT
+COPY main.go main.go
+COPY api api/
+COPY config config/
+COPY controllers controllers/
+COPY go.mod go.mod
+COPY go.sum go.sum
```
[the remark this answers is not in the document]
cpmeadors: Agreed.  Will move go.mod and go.sum above api in all docker files

Makefile
cpmeadors: Due to build tooling downstream we need defaults to represent product build.  We consider this a temporary solution while we investigate how to build with env vars downstream

bundle.Dockerfile
cpmeadors: It is only being used by Prow CI job.  We will point it to bundle-custom.Dockerfile and remove bundle.Dockerfile after that.

gkurz (approved):
LGTM. Thanks @cpmeadors !

jensfr (approved):
lgtm, thanks!

PR #222, opened 2022-10-04
update makefile VERSION to 1.3.1

cpmeadors:
CSV still references 1.3.0.  In the Makefile there is a VERSION variable that was set to 1.3.0.  I have changed that to 1.3.1.  After running make bundle the CSV is update with 1.3.1 in places where it used to say 1.3.0.

Once multistage builds dockerfiles are in place, this would update everything appropriately at build time.  Until then, we also need to manually update the midstream copy of the bundle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-VERSION ?= 1.3.0
+VERSION ?= 1.3.1
 
 # CHANNELS define the bundle channels used in the bundle.
 # Add a new line here if you would like to change its default config. (E.g CHANNELS = "candidate,fast,stable")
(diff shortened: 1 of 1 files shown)


beraldoleal:
@cpmeadors I'm not 100% sure ... but should this be ported to release-1.3 branch as well?

cpmeadors:
@beraldoleal You might be right.  @jensfr should I just change the target branch to release-1.3? if we go single stream shouldn't we just release from master and tag?  Do we still need release branches?

jensfr:
> @beraldoleal You might be right. @jensfr should I just change the target branch to release-1.3?
To keep master and release-1.3 in sync until we switch over to master (or main?) I would merge this to master and backport to release-1.3

>  if we go single stream shouldn't we just release from master and tag? Do we still need release branches?
Yes, when we use a single stream we don't need release branches any more.

jensfr (approved):
lgtm

pmores:
It would make sense to me to make this change both on master & release-1.3, in case we need to make another release in the 1.3 series for any reason.

pmores (approved):

cpmeadors:
everything is in #219.  I am closing this.

PR #225, opened 2022-10-27
Use numerical uid/gid in Dockerfile

gkurz:
Older versions of OCP are unhappy with the non-numeric ids and the controller fails to start with the following error:

container has runAsNonRoot and image has non-numeric user (nonroot),
          cannot verify user is non-root

The nonroot stuff was generated by an old operator sdk. Recent versions now put numerical ids. Do the same.

Fixes: https://issues.redhat.com/browse/KATA-1824
Fixes: rhjira#[KATA-1824](https://issues.redhat.com//browse/KATA-1824)

**- What I did**

**- How to verify it**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,5 +27,5 @@ COPY --from=builder /workspace/bin/manager .
 RUN useradd  -r -u 499 nonroot
 RUN getent group nonroot || groupadd -o -g 499 nonroot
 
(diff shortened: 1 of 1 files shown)


pmores (approved):
Thanks @gkurz !

cpmeadors:
@gkurz if you have a build for this can you share that so it can be tested.

PR #226, opened 2022-10-28
use generated bundle.dockerfile; remove bundle from .dockerignore

cpmeadors:
When building manually (using make targets and setting env vars) the output is a bundle that still points to the production image for the operator.  This is because the bundle-build uses the bundle-custom.Dockerfile which has the production values hard coded.

**- What I did**
Change the bundle-build target back to using the generated bundle.Dockerfile so it respects set env vars and includes the respective operator image.  Also had to remove "bundle" from .dockerignore so that the docker build can succeed.

**- How to verify it**
make bundle-build
install the operator using that bundle
Verify it is using the expected operator image for the controller pods

Closes: https://issues.redhat.com/browse/KATA-1828

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,4 +8,3 @@ api/v1/zz_generated.deepcopy.go
 config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
 config/rbac/role.yaml
 config/webhook/manifests.yaml
-#bundle
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,4 +8,4 @@ api/v1/zz_generated.deepcopy.go
 config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
 config/rbac/role.yaml
 config/webhook/manifests.yaml
-bundle
+#bundle
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,7 @@ bundle: manifests kustomize ## Generate bundle manifests and metadata, then vali
 
 .PHONY: bundle-build
 bundle-build: ## Build the bundle image.
-	docker build -f bundle-custom.Dockerfile -t $(BUNDLE_IMG) .
+	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
 
 .PHONY: bundle-push
 bundle-push: ## Push the bundle image.


pmores (approved):
lgtm thanks @cpmeadors !

jensfr (approved):
LGTM

.dockerignore
```diff
@@ -8,4 +8,4 @@ api/v1/zz_generated.deepcopy.go
 config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
 config/rbac/role.yaml
 config/webhook/manifests.yaml
-bundle
+#bundle
```
gkurz, line 11: Why not simply dropping the line ?

Makefile
```diff
@@ -188,7 +188,7 @@ bundle: manifests kustomize ## Generate bundle manifests and metadata, then vali
 
 .PHONY: bundle-build
 bundle-build: ## Build the bundle image.
-	docker build -f bundle-custom.Dockerfile -t $(BUNDLE_IMG) .
+	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
```
gkurz, line 191: This was the sole user of `bundle-custom.Dockerfile`. It should be removed in this commit as well IMHO.

gkurz:
`bundle.Dockerfile` is a generated file, but it is also tracked in git. Running `make clean` on a pristine tree removes it and you end up with a non-empty `git diff` which is quite confusing... what is the rationale with this file ?

.dockerignore
cpmeadors: You are right. Will do that

Makefile
cpmeadors: Prow CI will use bundl-custom.Dockerfile.  It has to stay.

cpmeadors:
I need to point the prow job to bundle-custome.Dockerfile before this gets merged.  I will create a PR for that.

cpmeadors:
PR for prow https://github.com/openshift/release/pull/33681

Makefile
gkurz: > Prow CI will use bundl-custom.Dockerfile. It has to stay.

I realized/remembered about that when you provided your status during the daily standup. This should be documented somewhere in the repo.

PR #227, opened 2022-11-02
remove bundle.Dockerfile

cpmeadors:
Nothing expects this auto-generated file to exist so we don't need to track it

Closes: KATA-1835


gkurz:
@cpmeadors shouldn't `bundle.Dockerfile` be added back to `.gitignore` as well ?

cpmeadors:
I restarted the 4.9 bundle build, but I think it is obsolete and we should just waive it if it doesn't pass, which it will since I would need to change it to point to the custom bundle dockerfile as well. I plan on removing it when I update the master file to 4.12 configuration.

pmores:
> @cpmeadors shouldn't `bundle.Dockerfile` be added back to `.gitignore` as well ?

Yes, that  sounds reasonable to me.

pmores:
> I restarted the 4.9 bundle build, but I think it is obsolete and we should just waive it if it doesn't pass, which it will since I would need to change it to point to the custom bundle dockerfile as well. I plan on removing it when I update the master file to 4.12 configuration.

This seems to be part of a bigger problem which is that the CI as a whole appears effectively unmaintained.

gkurz (approved):
LGTM. Thanks @cpmeadors !

gkurz:
/override ci/prow/4.9-ci-index-sandboxed-containers-operator-bundle

bpradipt (approved):
/lgtm
Thanks @cpmeadors

PR #231, opened 2022-11-15
README: s/master/main/

jensfr:
After we renamed the branch, let's also change it in the README 
I didn't change .travis.yml because it's not used any more and we should get rid of it in a separate PR.


gkurz (approved):
LGTM. Thanks @jensfr !

PR #232, opened 2022-11-23
Add controller-tools to go.mod for downstream build

cpmeadors:
Our downstream build process of the operator image has not network access so we need to use the go module caching feature to run controller-gen directly rather than downloading the binary

Required for [KATA-1800](https://issues.redhat.com//browse/KATA-1800)

**- What I did**
`$ go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0`

**- How to verify it**
` $ go mod download`
` $ go run sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."`
THis should generate the same files as running make generate does.


jensfr (approved):
lgtm! Thank you @cpmeadors

PR #233, opened 2022-11-24
Update to operator-sdk 1.21.0

littlejawa:
Following steps from https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.21.0/
- Makefile: replace "go get" calls with "go install" calls (copy/paste of a code snippet)
- manager_auth_proxy_path.yaml: update kube-rbac-proxy version
- go.mod/go.sum changes according to the steps + go mod tidy + additional changes to fix build errors

Dockerfile:
- update to go 1.18 to fix build issues
- as there is no 1.18 image for 4.10, use 4.11 image for the builder, and change the base image accordingly
- empty the GOFLAGS variable when calling "make build": the new "go install" commands were failing because of the default "-mod=vendor" flag. The actual build in the makefile is explicitely using "-mod=mod" flag anyway.

Misc:
- update the sdk version in bundle-custom.Dockerfile too
- update the DEVELOPMENT.md file with new versions


littlejawa:
/hold
I have an issue with the webhook when testing this - working on it

littlejawa:
/unhold
The failure was due to my environment. With a clean setup, I can install the operator and run kata containers.

littlejawa:
I have a question: we will need to upgrade to 1.23 at least. Should I make separate PRs building one on top of the other, or a single PR, with one commit for each upgrade?

jensfr:
> I have a question: we will need to upgrade to 1.23 at least. Should I make separate PRs building one on top of the other, or a single PR, with one commit for each upgrade?

I think one PR for every upgrade will be easier to review

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy@sha256:db06cc4c084dd0253134f156dddaaf53ef1c3fb3cc809e5d81711baa4029ea4c
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.11.0
```
jensfr, line 13: I think we use the digest instead of tag here so that it can work in disconnected environments

jensfr (approved):
lgtm, thanks @littlejawa 
@cpmeadors, @beraldoleal  installing kustomize has changed, that may affect our downstream scripts

config/default/manager_auth_proxy_patch.yaml
littlejawa: Fixed. Thanks!

cpmeadors (approved):
All changes look good

PR #234, opened 2022-11-29
set version to 1.3.2

jensfr:
This sets VERSION in the Makefile to 1.3.2.
It also updates the csv version and skipRange in config/manifests/base because we don't generate that part of the CSV yet and have to update it manually.


gkurz (commented):
@jensfr shouldn't the following be updated as well ?

```
$ git grep ' 1\.3\.1'
...
config/manager/kustomization.yaml:  newTag: 1.3.1
```

pmores (approved):
lgtm, thanks @jensfr !

gkurz (approved):
LGTM. Thanks @jensfr !

PR #240, opened 2022-12-08
[WIP] Upgrade operator sdk 125

littlejawa:
Following steps from https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.25.0/

This PR will require more time, mostly because I'm not sure about the consequences of the changes I'm making.

### Enable multi-arch build support
I have added the code related to that in the `Makefile`, but my understanding is that this is just enabling it if we ever want to support more archs. It requires additional changes to actually build anything.
I'm not sure if/how we can test that change without the appropriate support, and I don't know if we actually want/can support other archs... so I'm leaving it here for discussion.

### Update ginkgo to v2
Updating the dependency in `go.mod` is reverted by the next `go mod tidy` call if we don't actually use that version of ginkgo in our tests.
Moving to ginkgo v2 is not a simple step. It has its own migration guide to follow :
https://onsi.github.io/ginkgo/MIGRATING_TO_V2

According to this, the v1 is deprecated so it really seems we have to update.

Doing it reveals two issues in the suite_test.go file:
- `BeforeSuite` doesn't use the `done` parameter anymore, nor a timeout value, effectively making it synchronous.
  This is on purpose according to https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-async-testing
  The migration guide provides guidance on how to make it async again if we really need to, using different methods, but I didn't go that far.
  I don't think we actually have parallel tests running anyway, and the execution time of `make test` on `main` and on this PR branch are the same.
  I feel we can leave it that way, but tell me if you think otherwise, and we can try to change it.

- `RunSpecsWithDefaultAndCustomReporters` is deprecated
  See https://onsi.github.io/ginkgo/MIGRATING_TO_V2#removed-custom-reporters
  The call is then made without the `NewlineReporter`... but I admit I have no idea what the consequences are.
  Where can I check for the output ? I see nothing from `make test` on the main branch already.
  The migration guide says you should use an additional parameter to ginkgo to get some reports, but I don't see ginkgo being called anywhere actually (we just use "go test"). Is that where we should add it? Do we need it?
  I tried to use the `ReportAfterSuite` method from the migration guide, but the `NewlineReporter` is not accepted there, and I can't really test it if I don't know where to look for the output. Some guidance would be appreciated.

I kept the ginkgo changes in a separate commit so that we can modify this part separately if needed.

### misc
- the chmod call in the makefile was added because the modification to KUBEBUILDER_ASSETS creates a folder under bin/k8s/ that cannot be deleted by `make clean` otherwise. I kept the commit separated in case a better method is suggested for fixing it.
- I updated all the way up to 1.25.3, because there is no change for 1.25.1, 1.25.2 and 1.25.3.


cpmeadors (approved):
Multiarch support looks additive and shouldn't change affect our downstream builds.  ginkgov2 changes look good and nice to see.  Everything else looks good

jensfr (approved):
lgtm, thank you @littlejawa

PR #241, opened 2022-12-08
manifests: add back valid-subscription annotation

jensfr:
This was removed earlier but needs to be added back in order to pass validation.

Fixes rhjira:[KATA-1876](https://issues.redhat.com//browse/KATA-1876)

Please provide the following information:
This annotation was removed by accident. It is required to show users which subscription of OCP is required and also to pass a validation test downstream

**- What I did**
add the annotation to the CSV

**- How to verify it**
The CVP tests should pass and it should also be displayed in Operatorhub


cpmeadors (approved):
This matches what is in midstream and that is passing.  Looks good.

littlejawa (approved):
lgtm
Thanks!

PR #243, opened 2022-12-13
increase timeout for peer pods to 600 seconds

jensfr:
This increases the remote_hypervisor_timeout value to 600 seconds. The change is done in the machine config that creates the config file. I'm also adding the kata-remote.conf file in clear text. We should update it whenever we update the machine config so that it's easier to find out what the configuration is.


bpradipt (approved):
/lgtm
Thanks @jensfr

PR #244, opened 2022-12-21
Add columns for uninstall progress

etrunko:

Delete kataconfig, CLI output did not show uninstall progress, only the YAML view in web UI.

**- What I did**

Added new columns to the output so that uninstall progress is reported.

**- How to verify it**

Previous output

```
 $ oc get kataconfig
 NAME                 INPROGRESS   COMPLETED   TOTAL   AGE
 example-kataconfig   false        2           2       17m

```
With patch applied

```
 $ oc get kataconfig
 NAME                 INSTALLINPROGRESS   INSTALLCOMPLETE   TOTAL   AGE   UNINSTALLINPROGRESS   UNINSTALLCOMPLETE
 example-kataconfig   false               2                 2       16m   True                  1

```

Added columns to indicate uninstall progress to CLI output.


etrunko:
After working on this issue for long time, I wonder if it would look better something like `STATUS` and `NODES`.

Nodes would always display the number of nodes kataconfig has been installed, so when it finishes installing, its value will be incremented, and when it gets uninstalled from a given node, the value is decremented.

Status could be some kind of enumerations with defined values:

- `Installing`
- `Installed`
- `Uninstalling`
- `Uninstalled`

Or maybe:

- `Install,InProgress`
- `Install,Complete`
- `Uninstall,InProgress`
- `Uninstall,Complete`

Example:

```
   $ oc get kataconfig
  NAME                 STATUS                NODES   AGE
  example-kataconfig   Install,InProgress    2       17m

  $ oc get kataconfig
  NAME                 STATUS                NODES   AGE
  example-kataconfig   Install,Complete      2       17m

  $ oc get kataconfig
  NAME                 STATUS                NODES   AGE
  example-kataconfig   Uninstall,InProgress  2       17m

  $ oc get kataconfig
  NAME                 STATUS                NODES   AGE
  example-kataconfig   UnInstall,Complete    2       17m
```

littlejawa:
> After working on this issue for long time, I wonder if it would look better something like `STATUS` and `NODES`.
> 
> Nodes would always display the number of nodes kataconfig has been installed, so when it finishes installing, its value will be incremented, and when it gets uninstalled from a given node, the value is decremented.
> 
> Status could be some kind of enumerations with defined values:
> 
> * `Installing`
> * `Installed`
> * `Uninstalling`
> * `Uninstalled`
> 
> Or maybe:
> 
> * `Install,InProgress`
> * `Install,Complete`
> * `Uninstall,InProgress`
> * `Uninstall,Complete`
> 
> Example:
> 
> ```
> ...
> ```

I think I like it better, but that's a personal taste :-)

I usually look at "oc describe" to get the status of uninstallation, but it's definitely useful to have this status on the "get" command too.

api/v1/kataconfig_types.go
```diff
@@ -70,10 +70,12 @@ type KataConfigStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=kataconfigs,scope=Cluster
-// +kubebuilder:printcolumn:name="InProgress",type=string,JSONPath=".status.installationStatus.IsInProgress",description="Status of Kata runtime installation"
-// +kubebuilder:printcolumn:name="Completed",type=integer,JSONPath=".status.installationStatus.completed.completedNodesCount",description="Number of nodes with Kata runtime installed"
+// +kubebuilder:printcolumn:name="InstallInProgress",type=string,JSONPath=".status.installationStatus.IsInProgress",description="Status of Kata runtime installation"
```
bpradipt, line 73: Long back @gkurz suggested to try this but I never got to try a conditional - `status.installationStatus.IsInProgress | status.unInstallationStatus.inProgress.status`
Can you try this ?

etrunko: Verified this does not work. With patch applied, the output is as follows:

```
$ oc get kataconfig
NAME
example-kataconfig

```

etrunko: > I think I like it better, but that's a personal taste :-)
> 

So, I think we could proceed with the proposed patch, and I will work on trying to improve the output, as it requires some investigation. I don't really know how properly do it yet.

> I usually look at "oc describe" to get the status of uninstallation, but it's definitely useful to have this status on the "get" command too.

I will check the output with oc describe, thanks. I still need some more fluency on the `oc` commands.

etrunko: > > I usually look at "oc describe" to get the status of uninstallation, but it's definitely useful to have this status on the "get" command too.
> 
> I will check the output with oc describe, thanks. I still need some more fluency on the `oc` commands.

In the end,  `oc describe` is equivalent to the YAML view in the web interface. TIL.

Makefile
```diff
@@ -189,7 +189,7 @@ bundle: manifests kustomize ## Generate bundle manifests and metadata, then vali
 	operator-sdk bundle validate ./bundle
 
 .PHONY: bundle-build
-bundle-build: ## Build the bundle image.
+bundle-build: bundle ## Build the bundle image.
```
jensfr, line 192: This seems like an unrelated change?

jensfr:
If you could remove the unrelated change in the Makefile, I think we can merge this and optimize in follow-on patches.

Makefile
etrunko: Kind of unrelated to the $TOPIC, but I added it as a separate commit.

etrunko: It seemed too simple of a change to create a new MR for it, but if you like it can be done separately.

api/v1/kataconfig_types.go
```diff
@@ -100,7 +100,7 @@ func init() {
 // KataInstallationStatus reflects the status of the ongoing kata installation
 type KataInstallationStatus struct {
 	// InProgress reflects the status of nodes that are in the process of kata installation
-	InProgress KataInstallationInProgressStatus `json:"inprogress,omitempty"`
+	InProgress KataInstallationInProgressStatus `json:"inProgress,omitempty"`
```
gkurz, line 105: Is it safe to change this with respect to operator upgrade ?

jensfr: that's a good point, for this PR in general. We are changing the API. Do we need to create a new api version? Does it need a webhook that converts from the old to the new version?

etrunko: I didn't know it was changing API. I only thought about consistency with other fields following camel case. As it is such a small change I can revert it back.

Makefile
etrunko: I'll be sending this change with a new MR then. Thanks

etrunko:
Tried with my cluster setup, and it does not install the new version.

```
error validating existing CRs against new CRD's schema for "kataconfigs.kataconfiguration.openshift.io": error validating custom resource against new schema for KataConfig /example-kataconfig: [].status.unInstallationStatus.IsInProgress: Required value
```
I'll be working on a fix

PR #248, opened 2022-12-22
webhook: Ensure webhook gets called for pods created in all user namespaces

jensfr:
change similar to but adapted to openshift namespaces (list provided by @bpradipt)


bpradipt (approved):
/lgtm
Thanks @jensfr

PR #257, opened 2023-01-27
Makefile: Target bundle-build requires bundle

etrunko:

If one calls make bundle-build without a make bundle beforehand, it will error out:

```
$ make bundle-build
docker build -f bundle.Dockerfile -t openshift-sandboxed-containers-operator-bundle:v1.3.2 .
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat sandboxed-containers-operator/bundle.Dockerfile: no such file or directory

```

**- What I did**

Added `bundle` as dependency of `bundle-build`

**- How to verify it**

Run `make bundle-build` without calling `make bundle` beforehand.

Makefile: Target bundle-build requires bundle

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,7 @@ bundle: manifests kustomize ## Generate bundle manifests and metadata, then vali
 	operator-sdk bundle validate ./bundle
 
 .PHONY: bundle-build
-bundle-build: ## Build the bundle image.
+bundle-build: bundle ## Build the bundle image.
 	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) .
 
 .PHONY: bundle-push


cpmeadors (commented):
I checked out your branch and make sure the make bundle-build worked after a make clean.  Build succeeded, but I noticed that the tag was still 1.3.2.  Not sure where that is coming from but maybe a rebase is needed.  Otherwise it looks good.

etrunko:
Rebased to latest main, it should get 1.3.3 version now.

littlejawa (approved):
lgtm

gkurz (approved):
LGTM. Thanks @etrunko !

PR #261, opened 2023-02-09
Encode log timestamps as RFC3339

gkurz:
Timestamps are currently displayed as number of seconds since the Epoch using scientific notation, i.e. 1.6759526089944808e+09.

This is definitely not human friendly. Newer versions of the zap package now render timestamps according to RFC3339, i.e. 2023-02-09T14:38:20Z.

We're not ready to bump the controller-runtime package that brings us zap. Let's configure zap to explicitly use RFC3339.

**- What I did**

**- How to verify it**


pmores (approved):

cpmeadors (approved):
looks good

PR #266, opened 2023-02-28
Setup os dependent machine config extensions.

### LINKED ISSUE #267: OKD: Choosing the right machine config extension on FCOS
**Description**

The machine config operator handles RHCOS and FCOS differently when effectuating machine config extensions.

On RHCOS it uses a hardcoded list which is then translated into the actual packages. In our case, sandboxed-containers are translated into kata-containers and then rpm-ostree is called.

On FCOS it just pipes the extensions directly into rpm-ostree, which means that in the current implementation it tries to install sandboxed-containers which doesn't exist in FCOS yum repo.

**Steps to reproduce the issue:**
1. Install OKD.
2. Enable the fcos.repo yum repo.
3. Install sandboxed-containers and create a kataconfig

**Describe the results you received:**

On FCOS, the worker node complains about there is no rpm package sandboxed-containers.

**Describe the results you expected:**

The package kata-containers would be installed.

**Additional environment details (platform, options, etc.):**
OKD 4.12 with FCOS.

smuda:

The machine config operator handles RHCOS and FCOS differently when effectuating machine config extensions.

On RHCOS it uses a hardcoded list which is then translated into the actual packages. In our case, `sandboxed-containers` are translated into `kata-containers` and then rpm-ostree is called.

On FCOS it just pipes the extensions directly into rpm-ostree, which means that in the current implementation it tries to install `sandboxed-containers` which doesn't exist in FCOS yum repo.

Fixes #267 

**- What I did**

I added an OS detection mechanism and if it's RHCOS I use `sandboxed-containers` as extension, otherwise I use `kata-containers`. 

In theory it's unfortunate that it's the OS of the operator that I use for guessing what OS the actual worker node has, but I've never heard of an OCP with FCOS or OKD with RHCOS so I think it's a reasonable assumption.

**- How to verify it**

1. Install OKD.
2. Enable the fcos.repo yum repo
3. Build from source according to [DEVELOPMENT.md](https://github.com/openshift/sandboxed-containers-operator/blob/main/docs/DEVELOPMENT.md).
4. Install from OKD console.

Support for choosing the right RPM package on OKD/FCOS.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -58,6 +58,7 @@ type KataConfigOpenShiftReconciler struct {
 
 	clientset  kubernetes.Interface
 	kataConfig *kataconfigurationv1.KataConfig
+	Os         OperatingSystem
 }
 
 const (
@@ -473,6 +474,21 @@ func (r *KataConfigOpenShiftReconciler) newMCForCR(machinePool string) (*mcfgv1.
 		return nil, err
 	}
 
+	// RHCOS uses "sandboxed-containers" as thats resolved/translated in the machine-config-operator to "kata-containers"
+	// FCOS however does not get any translation in the machine-config-operator so we need to
+	// send in "kata-containers".
+	// Both are later send to rpm-ostree for installation.
+	//
+	// As RHCOS is rather special variant, use "kata-containers" by default.
+	var extensions = []string{"kata-containers"}
+
+	if r.Os.IsEL() {
+		extensions = []string{"sandboxed-containers"}
+	}
+	if r.Os.IsFCOS() {
+		extensions = []string{"kata-containers"}
+	}
+
 	mc := mcfgv1.MachineConfig{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "machineconfiguration.openshift.io/v1",
@@ -487,7 +503,7 @@ func (r *KataConfigOpenShiftReconciler) newMCForCR(machinePool string) (*mcfgv1.
 			Namespace: "openshift-sandboxed-containers-operator",
 		},
 		Spec: mcfgv1.MachineConfigSpec{
-			Extensions: []string{"sandboxed-containers"},
+			Extensions: extensions,
 			Config: runtime.RawExtension{
 				Raw: icb,
 			},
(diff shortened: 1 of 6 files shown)


controllers/openshift_controller.go
```diff
@@ -473,6 +474,21 @@ func (r *KataConfigOpenShiftReconciler) newMCForCR(machinePool string) (*mcfgv1.
...
+	// send in "kata-containers".
+	// Both are later send to rpm-ostree for installation.
+	//
+	// As RHCOS is rather special variant, use "kata-containers" by default.
+	var extensions = []string{"kata-containers"}
+
+	if r.Os.IsEL() {
+		extensions = []string{"sandboxed-containers"}
+	}
+	if r.Os.IsFCOS() {
```
bpradipt, line 488: Is this condition needed since default for extensions is "kata-containers" ?

controllers/osrelease.go
```diff
@@ -0,0 +1,81 @@
...
+// OperatingSystem is a wrapper around a subset of the os-release fields
+// and also tracks whether ostree is in use.
+type OperatingSystem struct {
+	// id is the ID field from the os-release
+	id string
+	// variantID is the VARIANT_ID field from the os-release
+	variantID string
+	// version is the VERSION, RHEL_VERSION, or VERSION_ID field from the os-release
+	version string
+	// osrelease is the underlying struct from github.com/ashcrow/osrelease
```
bpradipt, line 31: I was bit sceptic about using the osrelease project as it has not been updated since last 6 years. However found that MCO is also using  it in `machine-config-operator/pkg/daemon/osrelease.go`.  So should be ok :-)

controllers/openshift_controller.go
smuda: Not really, just wanted to be explicit. I'll remove it.

bpradipt: You can add a comment mentioning that the default applies to FCOS.

pmores (approved):
Looks good to me, thanks @smuda !

controllers/openshift_controller.go
smuda: I've removed the if-clause and added a comment that it applies to FCOS.

smuda: @bpradipt Do you think we could resolve this conversation?

bpradipt (approved):
/lgtm
Thanks @smuda

controllers/openshift_controller.go
bpradipt: Yeah. Thanks @smuda

PR #270, opened 2023-03-02
set version to 1.4.0

jensfr:
In preparation for the next release, set version in Makefile and the CSV to 1.4.0. Not having to touch the CSV manually is still a todo that we're working on, at the moment it is still required though.


bpradipt (approved):
/lgtm
Thanks @jensfr

cpmeadors (approved):

PR #272, opened 2023-03-02
update controller-tools

jensfr:
update controller-tools


cpmeadors:
We need version 0.9.2 which goes with operator-sdk 1.26.  This can get specified with "go get sigs.k8s.io/controller-tools@v0.9.2".  Clarification on why we need this:  it is for downstream builds.  We need the controller-tools cached and adding them as a go dep is the only way we know how to do it.

cpmeadors (changes requested):
needs to be version 0.9.2

PR #274, opened 2023-03-09
Optionally use public images

### LINKED ISSUE #268: Images in Dockerfile not acessible to public
**Description**

When building a docker image with `make docker-build` it uses docker images which are not accessable publicly. There is an instruction in  [DEVELOPMENT.md](https://github.com/openshift/sandboxed-containers-operator/blob/main/docs/DEVELOPMENT.md) to login to one of Redhat clusters and get a token from there. However, it seems the clusters requires RH internal authentication.

Looking at [other issues](https://github.com/okd-project/okd/issues/514#issuecomment-1090233047) it seems that there are other images that are available.

What do you think about these alternative:
1. Update the Dockerfile with public images
2. Parametrizing the Dockerfile and adding a new make target `docker-build-public` which uses the public images to allow contributers to build
3. Updating the instructions to have these images as alternative.

smuda:
Current build assumes the developer is a RH employee and uses images not publicly available.

Closes #268 

**- What I did**
- Updated Dockerfile to take images as arguments
- Updated current make target to include (the same) images as arguments when building docker images.
- Added documentation about building docker images from public images.

**- How to verify it**
Build with existing make targets:
```
make docker-build
make docker-buildx
```

Build with public images:
```
export BUILDER_IMAGE=registry.ci.openshift.org/openshift/release:golang-1.18
export TARGET_IMAGE=registry.ci.openshift.org/origin/4.10:base
make docker-build
make docker-buildx
```

Added support for overriding images in Makefile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 # Use OpenShift golang builder image
+# These images needs to be synced with the images in the Makefile.
 ARG BUILDER_IMAGE=${BUILDER_IMAGE:-registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11}
 ARG TARGET_IMAGE=${TARGET_IMAGE:-registry.ci.openshift.org/ocp/4.11:base}
 FROM ${BUILDER_IMAGE} AS builder
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 # Use OpenShift golang builder image
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11 AS builder
+ARG BUILDER_IMAGE=${BUILDER_IMAGE:-registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11}
+ARG TARGET_IMAGE=${TARGET_IMAGE:-registry.ci.openshift.org/ocp/4.11:base}
+FROM ${BUILDER_IMAGE} AS builder
 
 WORKDIR /workspace
 
@@ -20,7 +22,7 @@ RUN go mod vendor
 RUN GOFLAGS="" make build
 
 # Use OpenShift base image
-FROM registry.ci.openshift.org/ocp/4.11:base
+FROM ${TARGET_IMAGE}
 WORKDIR /
 COPY --from=builder /workspace/bin/manager .
 
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,7 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# These images needs to be synced with the default values in the Dockerfile.
 BUILDER_IMAGE ?= registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11
 TARGET_IMAGE  ?= registry.ci.openshift.org/ocp/4.11:base
 
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,9 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+BUILDER_IMAGE ?= registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -60,6 +60,11 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+BUILDER_IMAGE_PRIVATE = registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11
+BUILDER_IMAGE_PUBLIC  = registry.ci.openshift.org/openshift/release:golang-1.18
+TARGET_IMAGE_PRIVATE  = registry.ci.openshift.org/ocp/4.11:base
+TARGET_IMAGE_PUBLIC   = registry.ci.openshift.org/origin/4.10:base
+
```
bpradipt, line 67: Can you please change the assignments to use `?=`. This will allows us to set these externally if required.
Also do you mind adding few lines in the doc to use the public targets .

bpradipt:
@cpmeadors @jensfr ptal

Makefile
smuda: Sure!

smuda: Fixed.

bpradipt (approved):
/lgtm
Thanks @smuda

cpmeadors:
I am concerned with these changes.  I understand the desire to have everything build-able for the community.  No one thinks we should have private RH images required.  The problem I see is that this is working around a problem that cause by internal RH build processes.  The proper solution is to use all publicly available resources here and fix downstream.

smuda:
I agree this is a workaround and it would be much better to have public images that is always used. Could you give me URL's for such images? I'd be happy to adjust the PR accordingly.

smuda:
Yes, I might have over-engineered a bit. 😀 I've simplified Makefile changes to simply allow overriding the images and added documentation to DEVELOPMENT.md with the appropriate public images.

The PR still does not address @cpmeadors [comment](https://github.com/openshift/sandboxed-containers-operator/pull/274#issuecomment-1470408450) about always having public images, but perhaps it's a step in the right direction?

bpradipt:
@gkurz ptal

gkurz:
> @gkurz ptal

Ack @bpradipt . Sorry @smuda  for the delay...

docs/DEVELOPMENT.md
```diff
@@ -22,6 +22,17 @@ In summary:
 - log in from the command line with the provided command
 - use "oc registry login" to save the token locally
 
+### Using public images
+
+If you cannot login to registry.ci.openshift.org, a temporary solution is to use
```
gkurz, line 27: Nit: trailing space at end of line.

Dockerfile
```diff
@@ -1,5 +1,7 @@
 # Use OpenShift golang builder image
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11 AS builder
+ARG BUILDER_IMAGE=${BUILDER_IMAGE:-registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11}
+ARG TARGET_IMAGE=${TARGET_IMAGE:-registry.ci.openshift.org/ocp/4.11:base}
```
gkurz, line 3: Do we really need to duplicate the default values here ? AFAICT you'll get the default values from the Makefile.

@cpmeadors can you think of a case where this Dockerfile would be used standalone ?

smuda: If my memory serves me right, it's required during automatic tests.

gkurz: @smuda  That rings a bell indeed. Let's add a comment in both places stating that changes should be propagated to the other file as well.

smuda: I've added a comment about keeping them in sync.

docs/DEVELOPMENT.md
smuda: I've removed trailing whitespace.

gkurz (approved):
Thanks @smuda !

PR #276, opened 2023-03-15
Fix corner case node selector handling

pmores:
This PR fixes problems in kata node selector handling introduced by recent changes.  They include
- an empty kata node selector can be represented by nil as opposed to an empty `metav1.LabelSelector` instance
- nil and empty LabelSelector selectors are allowed but handled radically differently by apimachinery (an empty instance matches everything, nil matches nothing)
- default k8s syntax checking of `matchExpressions` doesn't catch invalid `matchExpressions.operator` values.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -688,10 +688,15 @@ func (r *KataConfigOpenShiftReconciler) getKataConfigNodeSelectorAsLabelSelector
 		return &metav1.LabelSelector { MatchLabels: map[string]string{ "node-role.kubernetes.io/master": "" }}
 	}
 
-	nodeSelector := r.kataConfig.Spec.KataConfigPoolSelector.DeepCopy()
+	nodeSelector := &metav1.LabelSelector{}
+	if r.kataConfig.Spec.KataConfigPoolSelector != nil {
+		nodeSelector = r.kataConfig.Spec.KataConfigPoolSelector.DeepCopy()
+	}
+
 	if r.kataConfig.Spec.CheckNodeEligibility {
 		nodeSelector = labelsutil.AddLabelToSelector(nodeSelector, "feature.node.kubernetes.io/runtime.kata", "true")
 	}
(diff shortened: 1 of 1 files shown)


pmores:
@cpmeadors Thanks! Please keep in mind though that even if this PR fixes problem that it aims to fix the controller is still broken until #266 is fixed. So there's no point in testing just yet until @gkurz greenlights it.

controllers/openshift_controller.go
```diff
@@ -100,6 +100,19 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
...
+
+		// k8s resource correctness checking on creation/modification
+		// isn't fully reliable for matchExpressions.  Specifically,
+		// it doesn't catch an invalid value of matchExpressions.operator.
+		// With this work-around we check early if our kata node selector
+		// is workable and bail out before making any changes to the
+		// cluster if it turns out it isn't.
+		_, err := r.getKataConfigNodeSelectorAsSelector()
+		if err != nil {
+			r.Log.Info("Invalid KataConfig.spec.kataConfigPoolSelector - please fix your KataConfig", "err", err)
```
gkurz, line 112: Isn't the problem serious enough to qualify for `r.Log.Error()` ?

controllers/openshift_controller.go
```diff
@@ -1413,7 +1426,11 @@ func (r *KataConfigOpenShiftReconciler) updateNodeLabels() (err error) {
 		return err
 	}
 
-	kataNodeSelector, _ := r.getKataConfigNodeSelectorAsSelector()
+	kataNodeSelector, err := r.getKataConfigNodeSelectorAsSelector()
+	if err != nil {
+		r.Log.Info("Couldn't getKataConfigNodeSelectorAsSelector()", "err", err)
```
gkurz, line 1431: Same question here.

gkurz (approved):
Talked with @pmores on slack. Errors bloat the logs with huge unwanted stack traces. Let's go with info then.

LGTM. Thanks @pmores !

PR #277, opened 2023-03-17
Fix machine config extension

gkurz:
**- Description of the problem which is fixed/What is the use case**

Recent PR #266 broke the deployment of kata-containers with RHCOS/OpenShift (see details in first commit).

**- What I did**

Basically revert the OS detection logic from #266 and replace it with a `SANDBOXED_CONTAINERS_EXTENSION` environment variable that can be configured in the manifest (see `config/manager/manager.yaml`). If not set, the default value `kata-containers` is used to preserve existing behavior with FCOS.

**- How to verify it**

This should be verified with OpenShift. A pre-requisite is to set the value of  `SANDBOXED_CONTAINERS_EXTENSION` 
to `sandboxed-containers` in the manifest of the OSC operator bundle as shown below:

```
          template:
            metadata:
              labels:
                control-plane: controller-manager
            spec:
              containers:
              - args:
                - --metrics-bind-address=127.0.0.1:8080
                - --leader-elect
                command:
                - /manager
                env:
                - name: KATA_MONITOR_IMAGE
                  value: registry.redhat.io/openshift-sandboxed-containers/osc-monitor-rhel8:1.3.4
                - name: SANDBOXED_CONTAINERS_EXTENSION
          ====>   value: sandboxed-containers   <=== 
                image: registry-proxy.engineering.redhat.com/rh-osbs/openshift-sandboxed-containers-operator:1.3.4
                imagePullPolicy: Always
                name: manager
```

Deploy this in an OpenShift cluster and create a kataconfig.

**- Description for the changelog**

Configure the name of the CoreOS extension with a `SANDBOXED_CONTAINERS_EXTENSION` environment variable

Fixes: https://issues.redhat.com/browse/KATA-2079

PR #278, opened 2023-03-22
Peer pods rebase

gkurz:
Let's rebase to get all recent fixes from main, especially #277.

Fixes: https://issues.redhat.com/browse/KATA-2079


cpmeadors (approved):
Everything looks familiar.  I didn't read every line, but I think we have to move forward.

PR #279, opened 2023-03-24
controller: fix wrong amout of cpu request in RuntimeClass

jensfr:
In commit 7dae22183ecfa781c5fb50e4017a337d577bac78 and df06ae971b52df90e336457cd489788a9463f6b4 I turned '250m' into '250M'. This means a pod will request 250 full CPUs instead of a quarter CPU.

Fixes: rhjira#[KATA-2069](https://issues.redhat.com//browse/KATA-2069)


jensfr:
the main branch does not need this fix, the wrong values are only in this branch

gkurz (approved):
Another option would be to pass `0.25` : it is explicit enough and less error prone, but this can be done later on. Let's go forward with this fix.

PR #281, opened 2023-03-30
bundle: add peerpodconfig crd file

jensfr:
We need the PeerPodConfig CRD file in the bundle so that it is deployed with the Operator.

**- What I did**
add the peerpodconfig crd to the PROJECTS file, so that it's included as an owned resource in the CSV and add
the full CRD yaml file to config/manifests/extension-crds/

**- How to verify it**
Deploy the operator and check the controller log, if there is no error message saying that it can't find
the peerpodconfig CRD, then the problem is fixed.

Fixes: rhjira#[KATA-2098](https://issues.redhat.com//browse/KATA-2098)

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -7,6 +7,10 @@ plugins:
 projectName: sandboxed-containers-operator
 repo: github.com/openshift/sandboxed-containers-operator
 resources:
+- group: confidentialcontainers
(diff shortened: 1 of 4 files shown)


cpmeadors (approved):
looks good

PR #282, opened 2023-04-04
controller: use new naming peerpodconfig-ctrl

snir911:
instead of peer-pod-controller

PR #283, opened 2023-04-04
peerpod ctrl

snir911:
This is integration of [peerpod-ctrl](https://github.com/confidential-containers/cloud-api-adaptor/tree/staging/peerpod-ctrl) in this controller 

<del>
* This is rebased on top of #282
* currently supports only aws (waiting for azure upstream merge)

To install:
1. build and install controller normally, and KataConfig with `enablePeerPods: true`
2. make sure caa pods are up and running
3. run `kubectl apply -f hack/caa-rbac.yaml` workaround to set rbac for the caa pods (I'm not sure if they require to be restarted)
4. change the caa pod images from `quay.io/confidential-containers/cloud-api-adaptor-aws` to `quay.io/confidential-containers/cloud-api-adaptor` to make sure the peerpod-ctrl caa changes included (related also to #284)
5. then you should be able to run a (peer)pod and see PeerPod object created (it owned by the pod) and managed by the controller in case of deletion failure

TODOs:
* make caa pod to be deployed with the correct rbac permissions
* controller rbac for PeerPod modified manually, generate it using kubebuilder instead
* remove hack patch
* make sure it's rebased correctly once
</del>

This is all fixed, just build and test
* azure is enabled but not tested


snir911:
Removing WIP, I added a commit to keep up with https://github.com/openshift/sandboxed-containers-operator/pull/286 however, I'm not sure it's correct as locally i'm failing to build after this change
cc @bpradipt @jensfr @cpmeadors

Makefile
```diff
@@ -66,6 +66,12 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+BUILTIN_CLOUD_PROVIDERS ?= aws
```
bpradipt: azure ?

config/manager/manager.yaml
```diff
@@ -30,6 +30,7 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      hostNetwork: true
```
bpradipt, line 33: Is `hostNetwork=true` due to peerpod-ctrl and is this really needed ?

Makefile
snir911: That's odd, i added this, i'm not sure why you didn't see it on your end

config/manager/manager.yaml
snir911: yes, AFAIR for it's needed in order to use Instance Metadata Service

cpmeadors:
The PR description is really confusing.  What does this PR actually implement/change?  Is there a jira issue for this?

snir911:
This is the downstream integration part of the dangling resources controller in the sandboxed-container-operator, IIRC it's referred under the same Jira.
Also updated the commit msg to point to the upstream controller doc to make it clearer

Dockerfile
```diff
@@ -2,6 +2,7 @@
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11 AS builder
 
 WORKDIR /workspace
+ARG GOFLAGS
```
bpradipt, line 5: Note: I think a downstream change will be needed to incorporate the GOFLAGS into the build
@beraldoleal @jensfr

main.go
```diff
@@ -73,13 +78,31 @@ func init() {
...
 
 func SetTimeEncoderToRfc3339() zap.Opts {
 	return func(o *zap.Options) {
 		o.TimeEncoder = zapcore.RFC3339TimeEncoder
 	}
+
+}
+
+func setProvider() (cloud.Provider, error) {
+	cloudName := os.Getenv("CLOUD_PROVIDER")
```
bpradipt, line 94: Is `CLOUD_PROVIDER` env variable available for the operator controller-manager ?

snir911: no, it assumes it's being set with the CM/Secret which is attached to the pod, now you mentioning it sound like a bad idea, I'll think about it

snir911: fixed, we fetch CM/Secret at reconcile now

controllers/openshift_controller.go
```diff
@@ -1717,8 +1720,8 @@ func (r *KataConfigOpenShiftReconciler) updateFailedNodes(node *corev1.Node,
 	if err != nil {
 		return err, failedList
 	}
-	if (mcfgv1.IsMachineConfigPoolConditionTrue(foundMcp.Status.Conditions, mcfgv1.MachineConfigPoolNodeDegraded) ||
-		mcfgv1.IsMachineConfigPoolConditionTrue(foundMcp.Status.Conditions, mcfgv1.MachineConfigPoolDegraded)) {
+	if mcfgv1.IsMachineConfigPoolConditionTrue(foundMcp.Status.Conditions, mcfgv1.MachineConfigPoolNodeDegraded) ||
+		mcfgv1.IsMachineConfigPoolConditionTrue(foundMcp.Status.Conditions, mcfgv1.MachineConfigPoolDegraded) {
```
bpradipt, line 1724: this looks to be a gofmt change.

go.mod
```diff
@@ -3,88 +3,149 @@ module github.com/openshift/sandboxed-containers-operator
...
 	go.uber.org/zap v1.24.0
-	k8s.io/api v0.26.0
-	k8s.io/apimachinery v0.26.0
-	k8s.io/client-go v0.26.0
-	k8s.io/kubernetes v1.26.0
-	sigs.k8s.io/controller-runtime v0.14.0
+	k8s.io/api v0.27.1
+	k8s.io/apimachinery v0.27.1
+	k8s.io/client-go v1.5.2
+	k8s.io/kubernetes v1.27.1
```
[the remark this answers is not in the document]
snir911: i think it's either because of  peerpod-ctrl, or peerpodconfig-ctrl

controllers/openshift_controller.go
gkurz: Yes it is. I was pretty sure I had cleaned up everything with #290 but apparently not.. :disappointed:

gkurz:
I don't feel like I'm qualified to do a thorough review of this PR, but I did not noticed anything that looked obviously wrong.

bpradipt (approved):
/lgtm

Dockerfile
snir911: removed

jensfr (approved):
lgtm, thank you @snir911

PR #285, opened 2023-04-04
docs: small fix on displayName

beraldoleal:
DisplayName is an unknown key and currently CatalogSource will be created without the displayName.

**- What I did**
Just renamed DisplayName to displayName.
**- How to verify it**


littlejawa (approved):
lgtm
thanks @beraldoleal

bpradipt (approved):
/lgtm

PR #286, opened 2023-04-05
Add controller-tools mods back for downstream builds

cpmeadors:
Downstream Red Hat builds require controller-tools to be in the go.mod so they can be cached due to network limitations.  They can easily be removed accidentally with go mod tidy because they are not actually dependencies.

**- What I did**
go get sigs.k8s.io/controller-tools@v0.9.2
go get sigs.k8s.io/controller-tools/pkg/genall/help/pretty@v0.9.2
go get sigs.k8s.io/controller-tools/pkg/crd@v0.9.2

**- How to verify it**
go run sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./...""


jensfr (approved):
lgtm, thanks @cpmeadors

PR #287, opened 2023-04-06
controller: checkNodeEligibility vs peerpods: noop

beraldoleal:

When PeerPods is enabled checkNodeEligibility should not take into consideration since PeerPods doesn’t require KVM/qemu.

**- What I did**

I'm checking if enablePeerPods is enabled during checkNodeEligibility().

**- How to verify it**

Enable both checkNodeEligibility and enablePeerPods in you kata-config and check controller logs.

Configuration: Ignoring checkNodeEligibility when peer pods is enabled.

### DIFF
--- a/controllers/openshift_controller_test.go
+++ b/controllers/openshift_controller_test.go
@@ -631,6 +631,19 @@ var _ = Describe("OpenShift KataConfig Controller", func() {
 			By("Creating the KataConfig CR successfully")
 			Expect(k8sClient.Create(context.Background(), kataConfig)).Should(Succeed())
 
+			// Delete
+			By("Deleting KataConfig CR successfully")
+			kataConfigKey := types.NamespacedName{Name: kataConfig.Name}
+			Eventually(func() error {
+				k8sClient.Get(context.Background(), kataConfigKey, kataConfig)
+				return k8sClient.Delete(context.Background(), kataConfig)
+			}, timeout, interval).Should(Succeed())
+
+			By("Expecting to delete KataConfig CR successfully")
+			Eventually(func() error {
+				return k8sClient.Get(context.Background(), kataConfigKey, kataConfig)
+			}, timeout, interval).ShouldNot(Succeed())
+
 		})
 	})
 })
--- a/controllers/openshift_controller_test.go
+++ b/controllers/openshift_controller_test.go
@@ -646,4 +646,39 @@ var _ = Describe("OpenShift KataConfig Controller", func() {
 
 		})
 	})
+	Context("Custom KataConfig with CheckNodeEligibility and PeerPods enabled", func() {
+		It("Should ignore CheckNodeEligibility when PeerPods is enabled", func() {
+
+			kataConfig := &kataconfigurationv1.KataConfig{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: "kataconfiguration.openshift.io/v1",
+					Kind:       "KataConfig",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: name + "-with-peerpods",
+				},
+				Spec: kataconfigurationv1.KataConfigSpec{
+					CheckNodeEligibility: true,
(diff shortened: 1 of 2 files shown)


cpmeadors:
@beraldoleal any change you could write a unit test for this?

beraldoleal:
> @beraldoleal any change you could write a unit test for this?

@cpmeadors just did it with a force-push.

controllers/openshift_controller_test.go
```diff
@@ -631,6 +631,42 @@ var _ = Describe("OpenShift KataConfig Controller", func() {
...
+				k8sClient.Get(context.Background(), kataConfigKey, kataconfig)
+				return k8sClient.Delete(context.Background(), kataconfig)
+			}, timeout, interval).Should(Succeed())
+
+			By("Expecting to delete KataConfig CR successfully")
+			Eventually(func() error {
+				return k8sClient.Get(context.Background(), kataConfigKey, kataconfig)
+			}, timeout, interval).ShouldNot(Succeed())
+
+		})
```
gkurz, line 647: It is certainly a good thing to clean things up but I'd make this a separate fix for the `Custom KataConfig with CheckNodeEligibility create` test.

controllers/openshift_controller_test.go
```diff
@@ -631,6 +631,42 @@ var _ = Describe("OpenShift KataConfig Controller", func() {
...
+	Context("Custom KataConfig with CheckNodeEligibility and PeerPods enabled", func() {
+		It("Should ignore CheckNodeEligibility when PeerPods is enabled", func() {
+
+			kataConfig := &kataconfigurationv1.KataConfig{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: "kataconfiguration.openshift.io/v1",
+					Kind:       "KataConfig",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: name + "with-peerpods",
```
gkurz, line 658:
```suggestion
					Name: name + "-with-peerpods",
```
This will produce a prettier `example-kataconfig-with-peerpods` name :smiley:

controllers/openshift_controller_test.go
```diff
@@ -631,6 +631,42 @@ var _ = Describe("OpenShift KataConfig Controller", func() {
...
+				},
+				Spec: kataconfigurationv1.KataConfigSpec{
+					CheckNodeEligibility: true,
+					EnablePeerPods:       true,
+				},
+			}
+
+			By("Creating the KataConfig CR successfully")
+			Expect(k8sClient.Create(context.Background(), kataConfig)).Should(Succeed())
+
```
gkurz, line 668: Please delete the KataConfig CR as well. This will provide a clean state to start from when someone wants to add a new test.

controllers/openshift_controller_test.go
```diff
@@ -631,6 +631,42 @@ var _ = Describe("OpenShift KataConfig Controller", func() {
...
 		})
 	})
+
```
gkurz, line 671:
```suggestion
````
Empty line nit.

controllers/openshift_controller_test.go
beraldoleal: Hi @gkurz, I'm not sure if I got your suggestion.

This section, that is deleting KataConfig after the test, is already in the test you mentioned. It is also duplicated on 3 other tests, I just copy and pasted as it was being done in the current file. Do you mind expand a little bit, please?


Note: I was planning to remove all those duplicated codes, maybe using ` AfterEach(func()`, but for some reason was not working. So I would like to address this in a different PR latter.

controllers/openshift_controller_test.go
beraldoleal: Fixed.

beraldoleal:
Hi @gkurz, thank you for the comments. Just sent a force-push with the changes.

cpmeadors (approved):

PR #288, opened 2023-04-06
(wip, don't merge) read machineConfig files for peer pod configuration

jensfr:
To configure Kata and cri-o for the peer pods runtime class we need
to create two MachineConfig files. We can't add them to the operator
bundle, because OLM only accepts certain kinds of manifests and we don't
want to hard code the config files in the controller code.
    
Include the machine config files in the container image and read them
during runtime when we create other resources for peer pods as well.
    
Fixes: [KATA-2130](https://issues.redhat.com//browse/KATA-2130)


bpradipt:
Closing this in favour of https://github.com/openshift/sandboxed-containers-operator/pull/292

PR #289, opened 2023-04-11
controller: automatically create a Secret containing the cluster's pull-secret

esposem:
```
Instead of letting the user manually do the following: 

oc create secret generic auth-json-secret --from-literal=auth.json=\ 
$(oc get -n openshift-config secret/pull-secret -ojson | jq -r '.data.".dockerconfigjson" | @base64d') \
-n openshift-sandboxed-containers-operator

do the same programmatically in enablePeerPods().

Fixes: KATA-1998
```

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1728,6 +1728,43 @@ func (r *KataConfigOpenShiftReconciler) clearFailedStatus(status kataconfigurati
 	return status
 }
 
+func (r *KataConfigOpenShiftReconciler) createAuthJsonSecret() error {
+	var err error = nil
+
+	pullSecret := &corev1.Secret{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: "pull-secret", Namespace: "openshift-config"}, pullSecret)
+	if err != nil {
+		r.Log.Info("Error fetching pull-secret", "err", err)
+		return err
+	}
+
+	authJsonSecret := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "auth-json-secret",
+			Namespace: "openshift-sandboxed-containers-operator",
+		},
+		Data: map[string][]byte{
+			"auth.json": pullSecret.Data[".dockerconfigjson"],
+		},
+		Type: corev1.SecretTypeOpaque,
+	}
+
+	err = r.Client.Create(context.TODO(), &authJsonSecret)
+	if err != nil {
+		if k8serrors.IsAlreadyExists(err) {
+			err = r.Client.Update(context.TODO(), &authJsonSecret)
+			if err != nil {
+				r.Log.Info("Error UPDATING auth-json-secret", "err", err)
+				return err
+			}
+		}
+		r.Log.Info("Error creating auth-json-secret", "err", err)
+		return err
+	}
+
+	return err
+}
+
 func (r *KataConfigOpenShiftReconciler) enablePeerPods() error {
 	peerPodConfig := v1alpha1.PeerPodConfig{
 		TypeMeta: metav1.TypeMeta{},
@@ -1745,6 +1782,13 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPods() error {
 	err := r.Client.Create(context.TODO(), &peerPodConfig)
 	if err != nil && !k8serrors.IsAlreadyExists(err) {
 		r.Log.Info("Error in creating peerpodconfig", "err", err)
+		return err
+	}
+
+	//Get pull-secret from openshift-config ns and save it as auth-json-secret in our ns
+	err = r.createAuthJsonSecret()
+	if err != nil {
+		r.Log.Info("Error in creating auth-json-secret", "err", err)
 		return err
 	}
 


snir911:
Thanks! have you tested it? it didn't require adding RBAC roles?

bpradipt:
@esposem can you please check if an rbac entry to create secrets will be required or not here - https://github.com/openshift/sandboxed-containers-operator/blob/main/controllers/openshift_controller.go#L72-L80 ?

controllers/openshift_controller.go
```diff
@@ -1731,6 +1731,33 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPods() error {
...
+		Data: map[string][]byte{
+			"auth.json": pullSecret.Data[".dockerconfigjson"],
+		},
+		Type: corev1.SecretTypeOpaque,
+	}
+
+	err = r.Client.Create(context.TODO(), &authJsonSecret)
+	if err != nil {
+		if k8serrors.IsAlreadyExists(err) {
+			return nil
```
snir911, line 1755: Should it return if already exist? I'm not sure if pull secret may change at some point but i think anyway it would be better to continue than return

controllers/openshift_controller.go
```diff
@@ -1731,6 +1731,33 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPods() error {
 		return err
 	}
 
+	pullSecret := &corev1.Secret{}
```
snir911, line 1734: Consider encapsulate the everything in a function (to keep it clean with the latest mc functions that has been added)

snir911 (commented):
tested on tech-preview branch, seems to be working well, please rebase.
I also added two minor comments,
Thanks @esposem !

esposem:
@snir911 @bpradipt yes, it looks like there is no need to add a new RBAC role. I think there is already one that covers this:
```
// +kubebuilder:rbac:groups="";machineconfiguration.openshift.io,resources=nodes;machineconfigs;machineconfigpools;containerruntimeconfigs;pods;services;services/finalizers;endpoints;persistentvolumeclaims;events;configmaps;secrets,verbs=get;list;watch;create;update;patch;delete
```

Anyways, thanks for testing! I rebased and applied all of your suggestions :+1:

snir911 (approved):
LGTM, thanks @esposem !

controllers/openshift_controller.go
```diff
@@ -1728,6 +1728,39 @@ func (r *KataConfigOpenShiftReconciler) clearFailedStatus(status kataconfigurati
 	return status
 }
 
+func (r *KataConfigOpenShiftReconciler) createAuthJsonSecret() error {
+	var err error = nil
+
+	pullSecret := &corev1.Secret{}
+	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: "pull-secret", Namespace: "openshift-config"}, pullSecret)
+	if err != nil {
+		r.Log.Error(err, "Error fetching pull-secret")
```
bpradipt, line 1737: Based on suggestions in other PRs from @gkurz and @pmores, can you change the `r.Log.Error` usage to `r.Log.Info`. 
Rest looks good to me.

esposem:
Updated with @bpradipt feedbacks

controllers/openshift_controller.go
```diff
@@ -1728,6 +1728,39 @@ func (r *KataConfigOpenShiftReconciler) clearFailedStatus(status kataconfigurati
...
+		},
+		Data: map[string][]byte{
+			"auth.json": pullSecret.Data[".dockerconfigjson"],
+		},
+		Type: corev1.SecretTypeOpaque,
+	}
+
+	err = r.Client.Create(context.TODO(), &authJsonSecret)
+	if err != nil {
+		if k8serrors.IsAlreadyExists(err) {
```
gkurz, line 1754: Shouldn't we update in case the original pull secret changed ?

esposem:
@gkurz updated, I just remove the old auth-json-secret and add always a new one. Is it what you had in mind? I thought comparing byte slices was not ideal from the performance point of view.
I still have to test this though

gkurz:
> @gkurz updated, I just remove the old auth-json-secret and add always a new one. Is it what you had in mind? I thought comparing byte slices was not ideal from the performance point of view. I still have to test this though

@esposem I was rather thinking about calling `Client.Update()` if the object already exists, i.e. at this line in your previous patch https://github.com/openshift/sandboxed-containers-operator/commit/3d77a529d0b0ef69d6f753889219de8eca9000ad#diff-fbc9237b068d36a3e13bd2aa40352cfacfd5a642fc85ddd2f8651fb15fe69c63R1755 .

gkurz (approved):
LGTM. Thanks @esposem !

jensfr:
should we remove the do-not-merge label now and merge? @gkurz

esposem:
@jensfr  I was not able to test this yet (not that I expect anything special). Maybe @snir911  can re-test it while I am still figuring how to compile peer pods on the peer-pods-tech-preview branch?

gkurz:
> @jensfr I was not able to test this yet (not that I expect anything special). Maybe @snir911 can re-test it while I am still figuring how to compile peer pods on the peer-pods-tech-preview branch?

Ok. Let me put the hold label back then. Even if the change looks good, it should be compiled and smoke tested before being merged.

snir911:
tested, seem to work @gkurz i think you may remove hold

PR #291, opened 2023-04-18
Improve installation steps

pmores:
This PR reworks the installation flow with two main objectives in mind: rearrange major steps (e.g. MC and MCP creation) so that they follow the most straightforward and natural succession possible, and remove arbitrary waits throughout the procedure, replacing them with a small number of better-founded ones.

A number of smaller clean-ups are also done in the initial commits of this series.

Fixes: [KATA-1969](https://issues.redhat.com//browse/KATA-1969)
Fixes: [KATA-2143](https://issues.redhat.com//browse/KATA-2143)

### DIFF
--- a/api/v1/kataconfig_types.go
+++ b/api/v1/kataconfig_types.go
@@ -59,6 +59,11 @@ type KataConfigStatus struct {
 	// Upgradestatus reflects the status of the ongoing kata upgrade
 	// +optional
 	Upgradestatus KataUpgradeStatus `json:"upgradeStatus,omitempty"`
+
+	// Used internally to persist state between reconciliations
+	// +optional
+	// +kubebuilder:default:=false
+	WaitingForMcoToStart bool `json:"waitingForMcoToStart,omitempty"`
 }
 
 // +genclient
--- a/api/v1/kataconfig_types.go
+++ b/api/v1/kataconfig_types.go
@@ -59,8 +59,6 @@ type KataConfigStatus struct {
 	// Upgradestatus reflects the status of the ongoing kata upgrade
 	// +optional
 	Upgradestatus KataUpgradeStatus `json:"upgradeStatus,omitempty"`
-
-	BaseMcpGeneration int64 `json:"prevMcpGeneration"`
 }
 
 // +genclient
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -960,18 +960,28 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 	wasMcJustCreated, err := r.createExtensionMc(machinePool)
 	if err != nil {
 		return ctrl.Result{Requeue: true}, nil
-	} else if wasMcJustCreated {
-		return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, nil
 	}
 
 	// Create kata-oc MCP only if it's not a converged cluster
 	if !isConvergedCluster {
-		_, err = r.updateNodeLabels()
+		labelingChanged, err := r.updateNodeLabels()
 		if err != nil {
 			if k8serrors.IsConflict(err) {
 				return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
 			} else {
 				return ctrl.Result{Requeue: true}, nil
+			}
+		}
+		if labelingChanged {
+			r.Log.Info("node labels updated")
+			// IsInProgress == ConditionFalse would be more direct and clear but it doesn't handle
+			// the possibility that IsInProgress == "" which actually comes up right after
+			// KataConfig is created.
+			if r.kataConfig.Status.InstallationStatus.IsInProgress != corev1.ConditionTrue {
+				r.Log.Info("Starting to wait for MCO to start")
+				r.kataConfig.Status.WaitingForMcoToStart = true
+			} else {
+				r.Log.Info("installation already in progress", "IsInProgress", r.kataConfig.Status.InstallationStatus.IsInProgress)
 			}
 		}
 
@@ -986,18 +996,68 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 				r.Log.Error(err, "Error in creating new MachineConfigPool ", "machinePool", machinePool)
 				return ctrl.Result{}, err
 			}
-			// mcp created successfully - requeue to check the status later
-			return ctrl.Result{Requeue: true, RequeueAfter: 20 * time.Second}, nil
+			// Don't requeue - the MCP creation will send us
+			// a reconcile request via our MCP watching so we should be
+			// guaranteed to run again at due time even without requeueing.
 		} else if err != nil {
 			r.Log.Error(err, "Error in retreiving MachineConfigPool ", "machinePool", machinePool)
 			return ctrl.Result{}, err
 		}
-
-		// Wait till MCP is ready
-		if mcp.Status.MachineCount == 0 {
-			r.Log.Info("Waiting till MachineConfigPool is initialized ", "machinePool", machinePool)
-			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
+	} else {
+		if wasMcJustCreated {
+			r.kataConfig.Status.WaitingForMcoToStart = true
 		}
+	}
+
+	isMcpUpdating := func(mcpName string) bool {
+		mcp := &mcfgv1.MachineConfigPool{}
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: mcpName}, mcp)
+		if err != nil {
+			r.Log.Info("Getting MachineConfigPool failed ", "machinePool", mcpName, "err", err)
+			return false
+		}
+		return mcfgv1.IsMachineConfigPoolConditionTrue(mcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdating)
+	}
+
+	isKataMcpUpdating := isMcpUpdating(machinePool)
+	r.Log.Info("MCP updating state", "MCP name", machinePool, "is updating", isKataMcpUpdating)
+	if isKataMcpUpdating {
+		r.kataConfig.Status.InstallationStatus.IsInProgress = corev1.ConditionTrue
+	}
+	isMcoUpdating := isKataMcpUpdating
+	if !isConvergedCluster {
+		isWorkerUpdating := isMcpUpdating("worker")
+		r.Log.Info("MCP updating state", "MCP name", "worker", "is updating", isWorkerUpdating)
+		if isWorkerUpdating {
+			r.kataConfig.Status.InstallationStatus.IsInProgress = corev1.ConditionTrue
+		}
+		isMcoUpdating = isKataMcpUpdating || isWorkerUpdating
+	}
+
+	// This condition might look tricky so here's a quick rundown of
+	// what each possible state means:
+	// - isMcoUpdating && WaitingForMcoToStart:
+	//     We've just finished waiting for the MCO to start updating.
+	//     The MCO is updating already but "Waiting" is still 'true'
+	//     (it will be set to 'false' shortly).
+	// - !isMcoUpdating && WaitingForMcoToStart:
+	//     We're waiting for the MCO to pick up our recent changes and
+	//     start updating.
+	// - isMcoUpdating && !WaitingForMcoToStart:
+	//     The MCO is updating (it hasn't yet finished processing our
+	//     recent changes).
+	// - !isMcoUpdating && !WaitingForMcoToStart:
+	//     The MCO isn't updating nor do we think it should be.  This is
+	//     the case e.g. when we're reconciliating a KataConfig change
+	//     that doesn't affect kata installation on cluster.
+	if !isMcoUpdating && r.kataConfig.Status.WaitingForMcoToStart == true {
+		r.Log.Info("Waiting for MCO to start updating.")
+		// We don't requeue, an MCP going Updated->Updating will
+		// trigger reconciliation by itself thanks to our watching MCPs.
+		return reconcile.Result{}, nil
+	} else {
+		r.Log.Info("No need to wait for MCO to start updating.", "isMcoUpdating", isMcoUpdating, "Status.WaitingForMcoToStart", r.kataConfig.Status.WaitingForMcoToStart)
+		r.kataConfig.Status.WaitingForMcoToStart = false
 	}
 
 	foundMcp, doReconcile, err, done := r.updateStatus(machinePool)
@@ -1007,16 +1067,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 
 	r.kataConfig.Status.TotalNodesCount = int(foundMcp.Status.MachineCount)
 
-	if mcfgv1.IsMachineConfigPoolConditionTrue(foundMcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdating) &&
-		r.kataConfig.Status.InstallationStatus.IsInProgress == "false" &&
-		r.kataConfig.Status.RuntimeClass == "kata" {
-		r.Log.Info("New node being added to existing cluster")
-		r.kataConfig.Status.InstallationStatus.IsInProgress = corev1.ConditionTrue
-		return reconcile.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
-	}
-
-	if mcfgv1.IsMachineConfigPoolConditionTrue(foundMcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdated) &&
-		foundMcp.Status.UpdatedMachineCount == foundMcp.Status.MachineCount {
+	if !isMcoUpdating {
 		r.Log.Info("create runtime class")
 		r.kataConfig.Status.InstallationStatus.IsInProgress = corev1.ConditionFalse
 		err := r.createRuntimeClass()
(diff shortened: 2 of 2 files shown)


controllers/openshift_controller.go
```diff
@@ -942,9 +955,15 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
+	wasMcJustCreated, err := r.createExtensionMc(machinePool)
+	if err != nil {
+		return ctrl.Result{Requeue: true}, nil
+	} else if wasMcJustCreated {
+		return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, nil
+	}
 
 	// Create kata-oc MCP only if it's not a converged cluster
-	if machinePool != "master" {
+	if machinePool == "kata-oc" {
```
gkurz: Not sure to understand the rationale behind this change, especially since it is somehow reverted in the next commit.

controllers/openshift_controller.go
```diff
@@ -963,7 +965,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 	}
 
 	// Create kata-oc MCP only if it's not a converged cluster
-	if machinePool == "kata-oc" {
+	if !isConvergedCluster {
```
gkurz, line 966: Is this the bug fix ? Can you elaborate ?

controllers/openshift_controller.go
```diff
@@ -1022,7 +1024,16 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
+// If the first return value is 'true' it means that the MC was just created
+// by this call, 'false' means that it's already existed.  As usual, the first
+// return value is only valid if the second one is nil.
+func (r *KataConfigOpenShiftReconciler) createExtensionMc(machinePool string) (bool, error) {
+
+	// In case we're returning an error we want to make it explicit that
+	// the first return value is "not care".  Unfortunately golang seems
+	// to lack syntax for creating an expression with default bool value
+	// hence this work-around.
+	var dummy bool
```
gkurz, line 1103: There might be some functions where this could happen as well. What about making this a a function ?

```
// To be used in errors paths of functions that return (bool, error). This
// makes it explicit that the bool value is irrelevant in this case.
func dummyBool() bool {
	return false
}
```

controllers/openshift_controller.go
```diff
@@ -1445,13 +1445,13 @@ func (r *KataConfigOpenShiftReconciler) updateNodeLabels() (err error) {
 
 	if err := r.Client.List(context.TODO(), workerNodeList, listOpts...); err != nil {
 		r.Log.Error(err, "Getting list of nodes failed")
-		return err
+		return false, err
```
gkurz, line 1504: This could be a user for `dummyBool()` as suggested earlier.

api/v1/kataconfig_types.go
```diff
@@ -59,8 +59,6 @@ type KataConfigStatus struct {
 	// Upgradestatus reflects the status of the ongoing kata upgrade
 	// +optional
 	Upgradestatus KataUpgradeStatus `json:"upgradeStatus,omitempty"`
-
-	BaseMcpGeneration int64 `json:"prevMcpGeneration"`
```
gkurz, line 63: This is an API change. Will this fly if the server has an old KataConfig with the generation field ?

controllers/openshift_controller.go
pmores: The original `getMcpName()` could return "kata-oc" or "" for a regular cluster or "master" for a converged cluster so the meaning of the conditional was "if this is a regular cluster" (as the comment points out).  The new function is more straightforward, it will only return "kata-oc" or "master" for regular or converged clusters respectively.  Taking this into account, if we want to ask "is this cluster regular" I consider it more straightforward to ask that question directly instead of "is this cluster not non-regular", hence this change.

As for this line being further changed in a subsequent commit, I tried to take a piecemeal approach for this series, gradually improving the code step-by-step to the extent possible at every given stage.  I recognise that this is not the only possible approach but I feared that going from the original code state straight to the final desired state would make for a series with just a couple of fairly complicated commits.

controllers/openshift_controller.go
pmores: Not really, I don't think.  The semantics of the `bool` return value is to tell the caller whether any "kata-oc" node-role labels were added or removed (as suggested by the name of the boolean).  When getting the worker list fails then the function is obviously unable to change any labels so it returns `false` to the caller to indicate this.

controllers/openshift_controller.go
gkurz: Thanks for the clarification.

controllers/openshift_controller.go
gkurz: Bug was maybe a bit excessive then :wink: but the change is good.

If you happen to re-push, you could maybe drop the comment on the way.

controllers/openshift_controller.go
pmores: That's a good point but I'd strongly prefer doing this as a separate PR, if for no other reason then because I'm not aware of any other such place in the current code (though I agree that there probably are some).

BTW there are other similar refactors in my queue that I'd like to do (reconciliation return value, logging errors, getting rid of the "create resource manifest before we check if it's even going to be needed" pattern and possibly others) but I think it would be confusing to do those in a single PR, especially considering that they aren't prerequisites for what this PR tries to achieve.

controllers/openshift_controller.go
gkurz: Ok, I get it now : unlike with the other commit, the bool return value is meaningful even if an error is returned.

controllers/openshift_controller.go
gkurz: Thinking again about this trick. Since the bool return value is undefined in case of error, why not making that very explicit in the function and most of all in the callers by changing its signature to (*bool, error) ?

controllers/openshift_controller.go
pmores: The bug fix is the final chunk in this commit, not this one - sorry, apparently it's not as clear as I thought. ;-)

gkurz: Oh my... I really misread the changelog ! :astonished:

controllers/openshift_controller.go
```diff
@@ -1016,7 +1018,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 	if mcfgv1.IsMachineConfigPoolConditionTrue(foundMcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdated) &&
 		foundMcp.Status.UpdatedMachineCount == foundMcp.Status.MachineCount {
 		r.Log.Info("create runtime class")
-		r.kataConfig.Status.InstallationStatus.IsInProgress = "false"
+		r.kataConfig.Status.InstallationStatus.IsInProgress = corev1.ConditionFalse
```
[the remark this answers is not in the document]
pmores: True... I didn't do that though as I knew I was going to remove the line altogether (removing conditionals like the one you point to is one of the main goals of this PR :-)).

Anyway, I suspect the real fix would be to change the type of `r.kataConfig.Status.InstallationStatus.IsInProgress` (and likely others in `KataConfig.status`) to boolean.  `corev1.ConditionStatus` is really a tri-state (True, False and Unknown IIRC), probably primarily meant for k8s Conditions where k8s might legimately not know if a condition is true or false.

Our fields like `IsInProgress` cannot really benefit from this and a simple boolean feels therefore more appropriate.  Moreover, `corev1.ConditionStatus` is really just another name for `string` so any mistyped string can be compared or assigned to it without the compiler noticing so that would be in my view another reason to use it only when necessary.

controllers/openshift_controller.go
pmores: Hm... that would make every caller branch both on non-nil error and nil bool, wouldn't it?  I'd consider that unfortunate, given that Go code is already 4-5 lines of error handling for each line that does something.  I think I like the `dummyBool()` idea better...

gkurz: > Hm... that would make every caller branch both on non-nil error and nil bool, wouldn't it?

No. The explicit contract would still be `As usual, the first return value is only valid if the second one is nil.`
i.e. if err is non-nil you go down the error path, else you can just dereference the bool pointer because it is guaranteed to be non-nil. No extra branch but an obvious crash if some caller doesn't honor the terms of the contract. This could be detected by a working CI.

controllers/openshift_controller.go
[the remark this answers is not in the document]
gkurz: > True... I didn't do that though as I knew I was going to remove the line altogether (removing conditionals like the one you point to is one of the main goals of this PR :-)).
> 

Well... this obviously breaks bisect, which has always been considered a capital sin in my previous assignment :wink: . This being said, since I haven't had the opportunity to bisect on the operator so far, maybe it isn't a big deal.

> Anyway, I suspect the real fix would be to change the type of `r.kataConfig.Status.InstallationStatus.IsInProgress` (and likely others in `KataConfig.status`) to boolean. `corev1.ConditionStatus` is really a tri-state (True, False and Unknown IIRC), probably primarily meant for k8s Conditions where k8s might legimately not know if a condition is true or false.
> 
> Our fields like `IsInProgress` cannot really benefit from this and a simple boolean feels therefore more appropriate. Moreover, `corev1.ConditionStatus` is really just another name for `string` so any mistyped string can be compared or assigned to it without the compiler noticing so that would be in my view another reason to use it only when necessary.

I tend to agree that it doesn't make sense to have an unknown state for something that is controlled by the operator and thus always known.

api/v1/kataconfig_types.go
pmores: The controller should be fine since it just never references the field anymore.  The control plane ignores unknown fields by default, and AFAIK  it only does this checking on committing resource creation & update.  From our previous experience it appears that when the updated controller checks a KataConfig instance back into database unknown fields (wrt to the new CRD schema) are dropped (we had this situation e.g. with the `.spec.kataMonitorImage` already).

controllers/openshift_controller.go
pmores: I believe such a simple protocol would be too strict to be generally useful.  Take the `updateNodeLabels()` function as an example whose return value is being dealt with in another commit in this PR.  Say it updates labelling of a bunch of nodes and then fails to do so for another node, maybe because updating its Node resource fails for any reason.  It will return an error but the first return value is still valid (it will be `true` because the function did update some nodes' labelling).  The caller will want to check the first value, error or not, since if any nodes' labelling was changed, it needs to follow up, plus it can reasonably assume that the Node resource update was transient so the node labelling job will likely be finished on the next reconciliation.

Also, using a contract that asks for dereferencing pointers without checking them first seems like asking for trouble with static code analysis.  I'm not sure about the current Golang but e.g. Rust would most likely complain, and if Golang doesn't at the moment it probably will in the future.  This is one of my personal gripes with static code analysis but people do seem to love linters.

api/v1/kataconfig_types.go
```diff
@@ -59,6 +59,10 @@ type KataConfigStatus struct {
 	// Upgradestatus reflects the status of the ongoing kata upgrade
 	// +optional
 	Upgradestatus KataUpgradeStatus `json:"upgradeStatus,omitempty"`
+
+	// Used internally to persist state between reconciliations
+	// +optional
+	WaitingForMcoToStart bool `json:"waitingForMcoToStart,omitempty"`
```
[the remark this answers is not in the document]
pmores: It's still a problem though as it turned out, thanks for pointing this out.  The controller logs

`KubeAPIWarningLogger    unknown field "status.waitingForMcoToStart"`

when it tries to process a legacy KataConfig instance without the new field.

I've pushed a fix that assigns a default value to it which makes the problem go away.

controllers/openshift_controller.go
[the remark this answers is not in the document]
pmores: > Well... this obviously breaks bisect, which has always been considered a capital sin in my previous assignment wink . This being said, since I haven't had the opportunity to bisect on the operator so far, maybe it isn't a big deal.

I'm not sure if I follow correctly but it would seem to me that it doesn't really _break_ it at least in the sense that it doesn't break it more than it already was.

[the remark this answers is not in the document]
pmores: You're right in general, I agree.  In this specific case though, I think that there still would be no breakage since if you take a look at the block of code guarded by that conditional, it only
- sets `IsInProgress` to `true` which is superfluous at that point (it was done in `createExtensionMc()` already)
- requeues for 15 seconds which is superflous too. Let me explain: if the `IsInProgress == "false"` test was the _only_ reason the conditional we're discussing wouldn't trigger (which appears implicitly assumed - otherwise there would be nothing to discuss ;-)) then the MCP in question is `Updating`.  In that case however, the conditional right after this one will _not_ trigger as it requires the MCP in `Updated` so its `else` branch would be executed which... requeues for 15 seconds. ;-)

So I think that while I cannot prove that there's no effect on the check, assuming the above is correct I might be able to prove that there's no effect on code execution anyway.

I do agree though that it was kind of reckless of me.  BTW I think we should discuss the standards required of PRs explicitly at some point.  I've been trying to stick to the standard you use (functionality shouldn't be harmed by any commit) but I'm noticing that if we relaxed the requirement to e.g. every commit has just to be compilable it would make it possible to split some of the bigger commits (like the final one in this series) into a bunch of much more understandable & reviewable chunks.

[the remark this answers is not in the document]
pmores: `IsInProgress` is in `KataConfig.status`, not in an MC, but I get your point.

Well the original controller, before I set out to change it, only handled initial installation and subsequent uninstallation, pretty much no effort was made to handle updating an existing installation (by adding or removing kata-enabled nodes).  This is kind of a breakage (by omission) by itself but if we accept it it makes sense only to set `InstalllationStatus.IsInProgress` when the extension MC is created.  So yes, if the MC already existed `IsInProgress` wouldn't be set - but if the MC already existed (at the beginning of an installation reconciliation, as opposed to in the middle of the initial installation when `IsInProgress` is already set from the time of the MC creation) that means that we're performing an update.  But update was already broken in the context of the original controller so no additional breakage here.

Now, since my recent changes which fixed and enabled installation updating, this could have mattered since updating is not broken anymore, except that `IsInProgress` is actually only used by status reporting (`updateStatus()` and its callees). But while the original support for updating was rudimentary, the corresponding support for update progress reporting was outright missing, non-existent, stubbed-out, and unlike the updating proper which has been fixed recently, status reporting hasn't changed yet.

So all in all, wrong handling of `IsInProgress` in the context of installation update basically breaks code that doesn't even exist yet.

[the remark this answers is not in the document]
gkurz: Oops sorry about the confusion with MC but fortunately you understood :wink:

Thanks for the detailed explanation, it helps.

controllers/openshift_controller.go
```diff
@@ -941,47 +957,107 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
-	if isMcCreated {
-		return doReconcile, err
+	isKataMcpUpdating := isMcpUpdating(machinePool)
+	r.Log.Info("MCP updating state", "MCP name", machinePool, "is updating", isKataMcpUpdating)
+	if isKataMcpUpdating {
+		r.kataConfig.Status.InstallationStatus.IsInProgress = corev1.ConditionTrue
+	}
+	isMcoUpdating := isKataMcpUpdating
+	if !isConvergedCluster {
+		isWorkerUpdating := isMcpUpdating("worker")
```
gkurz, line 1029: I don't quite get why we care for the MCP being updating.

pmores: Remember how when workers migrate from a source to a target pool, the source pool is drained instantly (without ever entering the Updating state) but the target pool goes into Updating and is joined by the migrating workers one by one over time?  From this it follows that if we want to know when the MCO is finished we need to keep an eye on the _target_ pool.

Now the target pool might be "kata-oc" for nodes that are having kata installed on them and are thus migrating from "worker" to "kata-oc", or "worker" for nodes that are having kata uninstalled from them and are migrating back to "worker".  So we know the MCO is done only when both potential target MCPs, "kata-oc" and "worker", are settled.

gkurz: @pmores What if some unrelated operation causes "worker" to update, e.g. something else creating machine configs ? Could this prevent to reach the MCO is done state ?

pmores: BTW the original code didn't need to wait for "worker" because it only handled the initial installation where nodes can only go from "worker" to "kata-oc" by definition.  However, since we've been supporting a subsequent update of kata installation on cluster, we need to handle nodes returning back to "worker" as well.

pmores: We cannot figure out _why_ an MCP is Updating, at least not easily if at all.  But we can still assume that the MCP should settle, finally.

So, let's say that user creates a KataConfig to put kata on a half of their workers, and they create another MCP and put the other half of their nodes in it.  Now if they if they remove some node(s) from the "kata-oc" pool and while this is running, they, say, delete their other custom MCP, nodes from both MCPs will be rejoining "worker" which will not settle until all nodes migrating to it are updated.  It will still settle although it's true that we might waiting a bit longer than necessary.

If that becomes a problem I believe we could just make the `isMcpUpdating()` function more precise (like, iterating over nodes belonging to that MCP and checking for each individually if it's Updated at the right rendered-* MC).  I also believe though that doing so at this point would smell like overengineering.

gkurz: Please add a comment to state that : `isMcpUpdating()` might catch unrelated changes, improve its precision if it becomes a problem.

gkurz (approved):
I gave a try and it does improve things. Especially, when creating a kataconfig with log level set to debug, we don't get the double reboot anymore : the MCO is generating a single rendered-kata-oc MC which applies both the extension and the cri-o dropin.

Let's go forward. Thanks @pmores !

pmores:
As for the double reboot - with the new install flow there's about 5-10 seconds between the `kata-oc` being created and starting updating.  This is plenty of time for the `logLevel` machinery to be set up and in place already when the MCP controller starts processing the new MCP.

So I think we can explain why this PR works wrt the double reboot issue.  What I'm not sure about frankly is why exactly this wasn't the case previously.

pmores:
/override ci/prow/check ci/prow/sandboxed-containers-operator-e2e

PR #292, opened 2023-04-20
Apply peerpod related MachineConfigs

bpradipt:
Peer pod enablement requires creating a new kata runtime class and providing the configuration.toml. In OpenShift we need to create MachineConfig objects to create these files.
The required MachineConfig yaml files are bundled within the controller-manager image and the manager process reads the yaml and creates the required MachineConfig objects. This gives us the flexibility to keep updating the yaml as needed without hardcoding the struct in the code

Fixes: #[Kata-2130](https://issues.redhat.com//browse/Kata-2130)

PeerPod enablement requires creating configuration for a new runtimeclass and a new kata configuration.toml.
This PR enables creation of the required configs via the operator
**- What I did**
Added the required MachineConfig yamls to the controller-manager image and used the same in the code
to create the required configurations

**- How to verify it**
1. Create a KataConfig with `enablePeerPods: true`
2. Wait for KataConfig to be created successfully and `kata-oc` MCP to be in Updated state. 
3. Login the worker node and check for `/etc/crio/crio.conf.d/50-kata-remote-cc` and `/opt/kata/configuration-remote.toml`
4. These files are created by the MachineConfigs

Add peerpod related MachineConfigs


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,4 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator
-  newTag: 1.3.1
+  newName: quay.io/bpradipt/openshift-sandboxed-containers-operator
```
pmores, line 15: This looks like a local modification, I guess it shouldn't be part of this PR?

config/manager/manager.yaml
```diff
@@ -62,7 +62,7 @@ spec:
         - name: KATA_MONITOR_IMAGE
           value: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-monitor:latest
         - name: SANDBOXED_CONTAINERS_EXTENSION
-          value: kata-containers
+          value: sandboxed-containers
```
pmores, line 65: This too looks like a local modification, I guess it shouldn't be part of this PR?

config/peerpods/mc-50-crio-config.yaml
```diff
@@ -2,7 +2,8 @@ apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfig
 metadata:
   labels:
-    machineconfiguration.openshift.io/role: worker
+    machineconfiguration.openshift.io/role: kata-oc
+    app: example-kataconfig
```
pmores, line 6: Does this belong upstream? (A honest question, admittedly I'm not sure how the `app` label is used.)

controllers/openshift_controller.go
```diff
@@ -936,6 +952,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 	}
 
 	// peer pod enablement
+	// This should be done after kata-oc creation and kata installation
```
pmores, line 955: Wouldn't this be executed _before_ kata-oc creation on the first reconciliation after a KataConfig instance is created?

controllers/openshift_controller.go
```diff
@@ -1727,14 +1742,119 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPods() error {
...
+		return err
+	}
+	return nil
+}
+
+func (r *KataConfigOpenShiftReconciler) disablePeerPods() error {
+	peerpodconf := v1alpha1.PeerPodConfig{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "peerpodconfig-example",
```
pmores, line 1779: Just a nitpick but we might want to use another name in published code than "-example"...

config/manager/kustomization.yaml
bpradipt: Right.

config/peerpods/mc-50-crio-config.yaml
bpradipt: It's a placeholder. This is not used by the MCO. The `app` label is also set for sandboxed extension - https://github.com/openshift/sandboxed-containers-operator/blob/main/controllers/openshift_controller.go#L483 
I think I should change it to the peerpodconfig name

controllers/openshift_controller.go
bpradipt: This is an incorrect comment. I'll remove it.

controllers/openshift_controller.go
bpradipt: Sure, I'll change it

config/manager/manager.yaml
bpradipt: Right

pmores (approved):
Looks good to me, thanks Pradipta!

controllers/openshift_controller.go
```diff
@@ -1727,14 +1742,119 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPods() error {
...
+			"mc", mc.Name)
+	}
+
+	return nil
+}
+
+func (r *KataConfigOpenShiftReconciler) createMcFromFile(mcFileName string) error {
+	yamlData, err := readMachineConfigYAML(mcFileName)
+	if err != nil {
+		r.Log.Error(err, "Unable to read MachineConfigYaml ", "mcFileName", mcFileName)
```
snir911, line 1829: nit: generally speaking, why not returning formatted error instead

snir911 (approved):
didn't try but LGTM overall, thanks!

jensfr (approved):
Lgtm, thank you pradipta!

controllers/openshift_controller.go
```diff
@@ -1727,14 +1742,119 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPods() error {
 		},
 	}
 
-	err = r.Client.Create(context.TODO(), &peerpodconf)
+	err := r.Client.Create(context.TODO(), &peerpodconf)
 	if k8serrors.IsAlreadyExists(err) {
```
gkurz, line 1746: Does this mean that the two MCs and the runtime class are assumed to be created already if the peer pod config is already there ? I'm asking because if a previous call to this function created the peer pod config but not all the other objects it looks like it will never try to create them again.

controllers/openshift_controller.go
bpradipt: The `err` is already formatted. Just logging it here for convenience.

controllers/openshift_controller.go
bpradipt: Right, it won't create the objects if peerpodconfig already exists.  Do you think it's too restrictive and better to handle this case?

gkurz: I'd tend to say yes but since this code predates the PR I guess it can be addressed separately.

bpradipt: @gkurz I went ahead and handled the case of existing peerpodconfig.

bpradipt:
In this version addressed the comments from @gkurz, especially switching to Log.Info instead of Log.Error and handling the case when peerpodconfig might exist

gkurz (approved):
LGTM. Thanks @bpradipt !

PR #295, opened 2023-04-24
Include a direct dependency to controller-tools

bpradipt:
controller-tools is required for manifest generation and for downstream builds we need to cache all the modules. Consequently all the required modules should be in go.mod

**- What I did**

Called the Version() method from controller-tools

**- How to verify it**
Running `go mod tidy` shouldn't remove the controller-tools

Add explicit dependency to controller-tools


pmores (approved):
Thanks @bpradipt !

gkurz (approved):
This will definitely help. Thanks @bpradipt !

Let's wait for @jensfr's ack before merging.

jensfr (approved):
lgtm, thank you @bpradipt

Issue #294, opened 2023-04-21
Add a git hook to ensure controller-tools is not removed from go.mod


bpradipt:
Downstream builds require controller-tools to be present in go.mod to allow for caching. However it gets automatically removed during updating deps (eg go mod tidy)
Adding a git pre-commit hook will make it easier to implement this.

PR #296, opened 2023-04-25
Include a direct dependency to controller-tools.

bpradipt:
This is backporting the fix for #294 from main to peer-pods-tech-preview branch


jensfr (approved):

gkurz (approved):
Thanks @bpradipt !

PR #298, opened 2023-04-25
Add unused references to controller-gen and deps

bpradipt:
Adding unused references to go modules in the import section.  This is to ensure that the modules are in go.mod and hence can be cached by downstream builds.  This also ensures that running `go mod tidy` will not remove these modules as these are not used in the code. This improves upon the PR-294 fix as it was an incomplete fix and didn't include all the required deps.


bpradipt:
Will close this PR and target it for peer-pods-tech-preview branch

PR #299, opened 2023-04-26
Add unused references to go modules in the import section.

bpradipt:
This is to ensure that the modules are in go.mod and hence can be cached by downstream builds.  This also ensures that running `go mod tidy` will not remove these modules as these are not used in the code. This improves upon the PR-294 fix as it was an incomplete fix and didn't include all the required deps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,7 +184,7 @@ ENVTEST ?= $(LOCALBIN)/setup-envtest
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v4.5.5
-CONTROLLER_TOOLS_VERSION ?= v0.9.2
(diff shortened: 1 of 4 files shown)


gkurz (approved):
LGTM and `go mod tidy` doesn't remove anything (while it does remove cobra and controller-tools if I drop the changes in main.go). So it seems to be working as expected.

Thanks @bpradipt !

cpmeadors (approved):
Tested with:
go clean -modcache
go mod download
go run sigs.k8s.io/controller-tools/cmd/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."

No additional mod were downloaded and command was successful.  Versions looks good to based on all the slack conversations.

PR #300, opened 2023-04-27
Improve uninstallation flow

pmores:
This PR is an uninstallation couterpart of PR #291.  The core of the change is migrating uninstallation to the exact same flow logic that installation has benefited from with a side benefit of a consistent idea behind both.

Apart from the tracked issues listed below, this PR also fixes a leakage of `logLevel` ContainerRuntimeConfig resource and removes spurious error messages at the end of each uninstallation that have been confusing people routinely.

Fixes: [KATA-2139](https://issues.redhat.com//browse/KATA-2139)
Fixes: [KATA-1926](https://issues.redhat.com//browse/KATA-1926)
Fixes: [KATA-1972](https://issues.redhat.com//browse/KATA-1972)

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -781,12 +781,6 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 		return reconcile.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
 	}
 
-	foundMcp := &mcfgv1.MachineConfigPool{}
-	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: machinePool}, foundMcp)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
-
 	if contains(r.kataConfig.GetFinalizers(), kataConfigFinalizer) {
 		// Get the list of pods that might be running using kata runtime
 		err := r.listKataPods()
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -774,6 +774,16 @@ func (r *KataConfigOpenShiftReconciler) getNodeSelectorAsLabelSelector() *metav1
 	return &metav1.LabelSelector{MatchLabels: r.getNodeSelectorAsMap()}
 }
 
+func (r *KataConfigOpenShiftReconciler) isMcpUpdating(mcpName string) bool {
+	mcp := &mcfgv1.MachineConfigPool{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: mcpName}, mcp)
+	if err != nil {
+		r.Log.Info("Getting MachineConfigPool failed ", "machinePool", mcpName, "err", err)
+		return false
+	}
+	return mcfgv1.IsMachineConfigPoolConditionTrue(mcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdating)
+}
+
 func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.Result, error) {
 	r.Log.Info("KataConfig deletion in progress: ")
 	machinePool, err := r.getMcpNameIfMcpExists()
@@ -1038,24 +1048,14 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		}
 	}
 
-	isMcpUpdating := func(mcpName string) bool {
-		mcp := &mcfgv1.MachineConfigPool{}
-		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: mcpName}, mcp)
-		if err != nil {
-			r.Log.Info("Getting MachineConfigPool failed ", "machinePool", mcpName, "err", err)
-			return false
-		}
-		return mcfgv1.IsMachineConfigPoolConditionTrue(mcp.Status.Conditions, mcfgv1.MachineConfigPoolUpdating)
-	}
-
-	isKataMcpUpdating := isMcpUpdating(machinePool)
+	isKataMcpUpdating := r.isMcpUpdating(machinePool)
 	r.Log.Info("MCP updating state", "MCP name", machinePool, "is updating", isKataMcpUpdating)
 	if isKataMcpUpdating {
 		r.kataConfig.Status.InstallationStatus.IsInProgress = corev1.ConditionTrue
 	}
 	isMcoUpdating := isKataMcpUpdating
 	if !isConvergedCluster {
-		isWorkerUpdating := isMcpUpdating("worker")
+		isWorkerUpdating := r.isMcpUpdating("worker")
 		r.Log.Info("MCP updating state", "MCP name", "worker", "is updating", isWorkerUpdating)
 		if isWorkerUpdating {
 			r.kataConfig.Status.InstallationStatus.IsInProgress = corev1.ConditionTrue
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -133,14 +133,12 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 		// indicated by the deletion timestamp being set.
 		if r.kataConfig.GetDeletionTimestamp() != nil {
 			res, err := r.processKataConfigDeleteRequest()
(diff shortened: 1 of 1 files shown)


pmores:
To explain the force-push, this was just properly rebasing this PR onto `peer-pods-tech-preview`.  It changed nothing in the PR itself.

controllers/openshift_controller.go
```diff
@@ -525,6 +556,18 @@ func (r *KataConfigOpenShiftReconciler) addFinalizer() error {
 	return nil
 }
 
+func (r *KataConfigOpenShiftReconciler) removeFinalizer() error {
+	r.Log.Info("Removing finalizer from the KataConfig")
+	controllerutil.RemoveFinalizer(r.kataConfig, kataConfigFinalizer)
+
+	err := r.Client.Update(context.TODO(), r.kataConfig)
+	if err != nil {
+		r.Log.Error(err, "Unable to update KataConfig")
```
bpradipt, line 565: Do you need to call stack here ?

controllers/openshift_controller.go
```diff
@@ -850,72 +850,74 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
+	}
+
 	r.clearUninstallStatus()
 	_, result, err2, done := r.updateStatus(machinePool)
 	if !done {
 		return result, err2
 	}
+	err = r.Client.Status().Update(context.TODO(), r.kataConfig)
+	if err != nil {
+		r.Log.Error(err, "Unable to update KataConfig status")
```
bpradipt: Do you need to handle the `Conflict` error separately and requeue after a timeout ?

bpradipt:
@pmores overall lgtm. Just few minor questions.

controllers/openshift_controller.go
pmores: I plan to clean up all logging as a separate task.

controllers/openshift_controller.go
pmores: That's right and this code is cleaned up in the penultimate commit of this series by removing it altogether.

(Even though it's presented as an addition by diff it's actually not new code, it's old code that wasn't even touched, it just appears moved to the diff algorithm.)

controllers/openshift_controller.go
```diff
@@ -850,72 +850,74 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
+			r.Log.Info("Deleting MachineConfigPool ", "machinePool", "kata-oc")
+			err = r.Client.Delete(context.TODO(), kataOcMcp)
+			if err != nil {
+				r.Log.Error(err, "Unable to delete kata-oc MachineConfigPool")
+				return ctrl.Result{}, err
+			}
+		} else if k8serrors.IsNotFound(err) {
+			r.Log.Info("MachineConfigPool not found", "machinePool", "kata-oc")
+		} else {
+			r.Log.Error(err, "Unable to get MachineConfigPool ", "machinePool", "kata-oc")
```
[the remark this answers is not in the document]
pmores: Most likely, yes, however this is a hard error so it _might_ make sense to log Errors in these cases.  I mean, I kind of doubt it but I'd like to give cases like this proper thought and testing as part of a separate logging clean-up (which is planned and necessary anyway since there's a lot dubious logging in parts of the controller code that wouldn't otherwise be hit by this ongoing clean-up).

controllers/openshift_controller.go
```diff
@@ -850,72 +850,74 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 			r.Log.Error(err, "Error found deleting machine config. If the machine config exists after installation it can be safely deleted manually.",
 				"mc", mc.Name)
 		}
-		// Sleep for MCP to reflect the changes
-		r.Log.Info("Pausing for a minute to make sure mcp has started syncing up")
-		time.Sleep(60 * time.Second)
+		r.Log.Info("Starting to wait for MCO to start")
```
jensfr, line 896: I assume "start" here means we are waiting for MCO to start reconciling the new config not "wait for the operator to start"?

pmores: Yes, exactly, the wording seems a bit clumsy.

pmores: Fixed.

controllers/openshift_controller.go
```diff
@@ -850,72 +850,74 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
+	targetPool := "worker"
+	if isConvergedCluster {
+		targetPool = "master"
 	}
-	r.Log.Info("Monitoring mcp", "mcp name", mcp.Name, "ready machines", mcp.Status.ReadyMachineCount,
-		"total machines", mcp.Status.MachineCount)
-	r.kataConfig.Status.UnInstallationStatus.InProgress.IsInProgress = corev1.ConditionTrue
+	isMcoUpdating := r.isMcpUpdating(targetPool)
+
+	if !isMcoUpdating && r.kataConfig.Status.WaitingForMcoToStart == true {
```
gkurz, line 917: Cosmetic: s/== true// for consistency with the `!isMcoUpdating` check maybe ?

pmores: Yeah, makes sense, will fix.

controllers/openshift_controller.go
```diff
@@ -304,6 +304,32 @@ func (r *KataConfigOpenShiftReconciler) processLogLevel(desiredLogLevel string)
...
+	err = r.Client.Delete(context.TODO(), ctrRuntimeCfg)
+	if err != nil {
+		r.Log.Info("error deleting ContainerRuntimeConfig", "err", err)
+		return err
+	}
+	r.Log.Info("logLevel ContainerRuntimeConfig deleted successfully")
+	return nil
+}
+
+
```
gkurz, line 337: Please drop the extra empty line (detected with `go fmt ./...`)

controllers/openshift_controller.go
```diff
@@ -846,72 +893,69 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 			r.Log.Error(err, "Error found deleting machine config. If the machine config exists after installation it can be safely deleted manually.",
 				"mc", mc.Name)
 		}
-		// Sleep for MCP to reflect the changes
-		r.Log.Info("Pausing for a minute to make sure mcp has started syncing up")
-		time.Sleep(60 * time.Second)
+		r.Log.Info("Starting to wait for MCO to start reconciliation")
+		r.kataConfig.Status.WaitingForMcoToStart = true
```
jensfr, line 897: I know this is not part of this PR, but when we added the new field `WaitingForMcoToStart` to the KataConfig CRD we changed the API. Did we increase the version of KataConfig and/or did we test upgrading the cluster?

pmores: No and no (well, that's pending but no so far afaik).  We handle it the same way as we did `KataConfig.spec.logLevel` addition or `KataConfig.spec.kataMonitorImage` removal.  From the docs I'm under the impression that once we start increasing version we have to go all the way through mutating webhooks to the heavyweight machinery necessary to do it right.  I'm further under the impression that we're still rather far from the point where this would start making sense and being worth the trouble but I might be wrong here, with first actual deployments underway.

gkurz:
> I reviewed the logic of the changes, less for style. It looks good to me. The concern about the API change that I stated in another commend shouldn't stop this one from being merged.

The current problem with this PR is that it is breaking `make test`. Looking at the logs it seems that kataconfig deletion times out likely because the operator is waiting for MCO actions that the test doesn't simulate. I've started modifying the test (basically updating the MCP statuses to please the expectations of the operator) and  I can get a bit further. Now I'm hitting some other errors I'm currently investigating.

controllers/openshift_controller.go
gkurz: We'll start testing upgrade cases soon and fix what needs to be fixed.

gkurz (approved):
Thanks for this huge effort @pmores !

pmores:
@gkurz Well thanks for the even more huge effort of reviewing it! :-)

jensfr (commented):
lgtm, thank you @pmores

PR #301, opened 2023-04-27
Revert "Include a direct dependency to controller-tools"

gkurz:
Reverts openshift/sandboxed-containers-operator#295 with same justification as #297 :

```
Reverting controller-tools implicit addition as there are more deps that are required to be directly included for successful downstream builds
```


pmores (approved):
It's kind of sad to see this go but lgtm. ;-)

littlejawa (approved):

PR #302, opened 2023-04-27
Merge main into peer-pods-tech-preview

gkurz:

The main goal is to bring the valuable changes of #291 to the feature branch for upcoming v1.4.0. Merging was chosen over rebasing because it is way simpler in terms of amount of conflicts to handle (especially in go.mod).

**- What I did**

```
git clone https://github.com/openshift/sandboxed-containers-operator.git
cd sandboxed-containers-operator/
git checkout -b merge-from-main origin/peer-pods-tech-preview 
git merge main
```

**- How to verify it**

Run all tests

Only two conflicts in `controllers/openshift_controller.go` :
- addition of the extension_mc_name const
- "IsInProgress = corev1.ConditionFalse" fix

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
 # Use OpenShift golang builder image
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11 AS builder
+# These images needs to be synced with the images in the Makefile.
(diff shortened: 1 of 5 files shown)


gkurz:
I could successfully run `make test` locallly.

pmores (approved):
Thanks @gkurz !

cpmeadors (approved):

PR #304, opened 2023-05-02
controller: fix incorrect logs in createAuthJsonSecret

esposem:
```   
 controller: fix incorrect logs in createAuthJsonSecret
    
    If the client is just updated with no error, the code was always falling
    back to r.Log.Info("Error creating auth-json-secret", "err", err)
    thus printing it also when it was not necessary.
    
    Signed-off-by: Emanuele Giuseppe Esposito <eesposit@redhat.com>
```


snir911 (approved):

bpradipt (approved):
/lgtm
Thanks @esposem

PR #305, opened 2023-05-08
Ensure KataConfigPoolSelector is used for PeerPodConfig CRD

bpradipt:
The peerpods config enablement is split into two parts - one part deals with creation of MachineConfigs before "kata-oc" MCP creation and another part deals with creation of PeerPodConfig CRD, runtimeclass and other misc configs after "kata-oc" MCP has updated.

The reason for this split is to ensure that MachineConfigs are created before "kata-oc" creation so as to optimise the number of reboots required when applying MachineConfigs. Further, the relevant node labels are only available when "kata-oc" MCP is updated, thereby making it available for the creation of PeerPodConfig CRD

The cloud-api-adaptor daemonset was not taking into account the KataConfigPoolSelector

Fixes: #[KATA-1873](https://issues.redhat.com//browse/KATA-1873)

**- What I did**
Ensured KataConfigPoolSelector is made available to the PeerPodConfig CRD which is then used by the PeerPodConfig controller to start the cloud-api-adaptor daemonset on the proper nodes

**- How to verify it**
Create a KataConfig with custom KataConfigPoolSelector and ensure `enablePeerPods: true`. Verify that the cloud-api-adaptor pods are created only on the nodes matched via the KataConfigPoolSelector

Ensure KataConfigPoolSelector is used for PeerPodConfig CRD

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -971,9 +971,9 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 
 	// peer pod enablement
 	if r.kataConfig.Spec.EnablePeerPods {
-		err := r.enablePeerPods()
+		err := r.enablePeerPodsMc()
 		if err != nil {
-			r.Log.Info("Enabling peerpods configuration failed", "err", err)
+			r.Log.Info("Enabling peerpods machineconfigs failed", "err", err)
 			return ctrl.Result{}, err
 		}
 	}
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -1120,6 +1120,17 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 			return reconcile.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
 		}
 
+		// create PeerPodConfig CRD and runtimeclass for peerpods
+		if r.kataConfig.Spec.EnablePeerPods {
+			err = r.enablePeerPodsMiscConfigs()
+			if err != nil {
+				r.Log.Info("Enabling peerpodconfig CR, runtimeclass etc", "err", err)
+				// Give sometime for the error to go away before reconciling again
+				return reconcile.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
```
pmores, line 1129: I would suggest that in a situation where we know nothing about the nature or cause of an error, a more appropriate handling would be to rely on the exponential back-off provided by `reconcile.Result{Requeue: true}` instead of a specific timeout which, considering the lack of information, has to be basically arbitrary.

But this will be a subject of a future PR to unify our reconciler return values so no need to change it now.

pmores (approved):
lgtm thanks @bpradipt !

jensfr (approved):
lgtm, @bpradipt

PR #307, opened 2023-05-15
Revert "Fix errors during envtest teardown"

gkurz:
`make test` is leaking the kube-apiserver and etcd processes. When unoticed, this causes `make test` to fail with weird errors like failing to open a file because of rlimit.

Investigation leads to commit 053f8664bf29d which was added to work around the very same problems at the time according to Pradipta. Reverting this commit fixes the issue.

It seems that updates to the controller-runtime and deps might have changed something and we no longer need the workaround.

Let's just revert 053f8664bf29d.

Fixes https://issues.redhat.com/browse/KATA-2168

**- What I did**

**- How to verify it**

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -20,11 +20,9 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"github.com/onsi/gomega/gexec"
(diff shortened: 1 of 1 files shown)


beraldoleal (approved):
LGTM.

pmores (approved):
I won't pretend to understand exactly what's happening here but if you tested the revert and it worked that's enough for me, and the state we're reverting to does look more natural to me, too. :-)

bpradipt (approved):
/lgtm

bpradipt (approved):
/lgtm

PR #310, opened 2023-05-15
Fix make test

gkurz:

`make test` is failing on the devel branch.

**- What I did**

- `git clone https://github.com/openshift/sandboxed-containers-operator.git`
- `git checkout devel`
- `make test`

**- How to verify it**

`make test` should succeed.

The real fixes are the second commit that adapts the tests to the new logic in the controller code introduced by #300,
and the third commit that adds the CRD of the ContainerRuntimeConfig to the test suite.

First commit is a cosmetic fix that make the timestamps human readable in the controller logs. It can only be tested in a situation where `make test` fails, e.g. before applying the other patches of this PR.

Fixes https://issues.redhat.com/browse/KATA-2167

### DIFF
--- a/config/extension-crds/containerruntimeconfig.crd.yaml
+++ b/config/extension-crds/containerruntimeconfig.crd.yaml
@@ -1,0 +1,170 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: containerruntimeconfigs.machineconfiguration.openshift.io
+  labels:
+    "openshift.io/operator-managed": ""
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  group: machineconfiguration.openshift.io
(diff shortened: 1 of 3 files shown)


gkurz:
Forced push because this needed rebase.

Added the `do-not-merge` label back because I'm now getting an error when tearing the test down, likely a regression caused by #307 that I wasn't getting before _of course_ ... :disappointed:

gkurz:
The test framework is basically creating a k8s cluster, starting the operator inside it and exercising some actions like a user could achieve with the `oc` CLI. This looks more like _integration_ testing than actual unit testing. It is very slow : up to 5 minutes, for just 6 high level trivial test cases and more than 50% of the code is there to mimick the behavior of the MCO as expected by the controller.

Adapting the tests to the changes introduced by #291 and #300 is only about adding a *lot* more of this MCO mimicking code. I've already added ~100 LoCs in my local repo and I'm still not there yet. It doesn't seem reasonable to invest so much effort for such a limited benefit.

@pmores suggested another direction : try to run a real MCO in the test cluster. This would allow to fully get rid of the mimicking code and to focus on actual test cases. This is future work, so I'm closing this PR.

PR #306, opened 2023-05-09
add image creation jobs for aws and azure

snir911:
apply scripts to create image in AWS/Azure

PR #311, opened 2023-05-16
Rebase devel on peer pods

gkurz:

The `peer-pods-tech-preview` and `devel` branches aren't currently in sync. They haven't diverged too much but we'd like them to stay as much as sync as possible until OSC 1.4.0 is released.

**- What I did**

`git rebase --ff peer-pods-tech-preview`

**- How to verify it**

PR branch should be just one patch ahead from `peer-pods-tech-preview`, i.e. `HEAD^` of the PR branch == `HEAD` of `peer-pods-tech-preview`

Pull all changes from `peer-pods-tech-preview` to `devel`. Next step will be to merge `devel` into `peer-pods-tech-preview` so that they are identical.

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -20,11 +20,9 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"github.com/onsi/gomega/gexec"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


littlejawa:
I think it needs another rebase now that https://github.com/openshift/sandboxed-containers-operator/pull/306 is merged :-(

pmores:
Yes, apparently.  Otherwise it looks good.

littlejawa (approved):
lgtm
Thanks @gkurz

gkurz:
Forced push because of #306 just being merged into `peer-pods-tech-preview`.

pmores (approved):
lgtm, thanks @gkurz !

PR #312, opened 2023-05-16
Rebase peer pods on devel

gkurz:

The peer-pods-tech-preview and devel branches aren't currently in sync. They haven't diverged too much but we'd like them to stay as much as sync as possible until OSC 1.4.0 is released.

**- What I did**

`git rebase --ff devel`

**- How to verify it**

PR branch should be identical to `devel`.

This makes `peer-pods-tech-preview` and `devel` identical.


littlejawa (approved):
lgtm
thanks @gkurz

gkurz:
GH is bugging again : the PR page is mentioning 4 commits but  only a604fed and 8e5e157 do exist in the [devel](https://github.com/openshift/sandboxed-containers-operator/commits/devel) branch.

pmores (approved):
thanks @gkurz !

PR #314, opened 2023-05-17
Fix kataconfig deletion when no kata nodes are on cluster

pmores:
This a follow-up fix for PR #300 and [KATA-2139](https://issues.redhat.com//browse/KATA-2139), please see the commit message for details.

PR #315, opened 2023-05-17
Fast-forward devel to peer-pods-tech-preview

pmores:
Overhead change to include PR #314 in devel and keep both branches in sync.


gkurz (approved):
LGTM. Thanks @pmores !

PR #316, opened 2023-05-17
Fix envtest tearing down

gkurz:
PR #307 was too optimistic. The original issue worked around by 053f8664bf29d still exists actually ; `make test` is now failing with :

  [FAILED] Unexpected error:
      <errors.aggregate | len:1, cap:1>: [
          <*errors.errorString | 0xc0005ebf80>{
              s: "timeout waiting for process kube-apiserver to stop",
          },
      ]
      timeout waiting for process kube-apiserver to stop
  occurred

This is caused by a missing cancellation of the certwatcher in the controller-runtime. The fix, as documented in [1], is to create a cancellable context, pass it to envtest and cancel it just before tearing down the test.

[1] https://github.com/kubernetes-sigs/kubebuilder/pull/2379

Fixes https://issues.redhat.com/browse/KATA-2168

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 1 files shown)


pmores:
Is there an explanation why #307 appeared to work?

gkurz:
> Is there an explanation why #307 appeared to work?

I tried to reproduce but I couldn't find my way in the multiple cross PRs that were merged after #307 obviously and I couldn't get it to work again. :disappointed:

PR #317, opened 2023-05-22
(draft) do not merge: Kata 2172

jensfr:

**- What I did**

**- How to verify it**


bpradipt:
This is no longer needed. @jensfr can we close this ?

PR #319, opened 2023-05-24
Install peerpods mutating webhook

bpradipt:
Fixes: #[KATA-2172](https://issues.redhat.com//browse/KATA-2172)

peer pods resource management relies on a mutating webhook to add the extended resources to the pod spec.
**- What I did**
Added mutating webhook installation to the controller
**- How to verify it**
1. Create a pod using `kata-remote-cc` runtimeclass
2. Verify the pod spec to check if the following is added
```
resources:
      limits:
        kata.peerpods.io/vm: "1"
      requests:
        kata.peerpods.io/vm: "1"
```

Install mutating webhook for peerpods

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -82,6 +82,8 @@ spec:
           value: kata-containers
         - name: RELATED_IMAGE_CAA
           value: quay.io/confidential-containers/cloud-api-adaptor
+        - name: RELATED_IMAGE_PEERPODS_WEBHOOK
+          value: "quay.io/confidential-containers/peer-pods-webhook:latest" 
         imagePullPolicy: Always
         resources:
           limits:
--- a/controllers/ppwebhook.go
+++ b/controllers/ppwebhook.go
@@ -1,0 +1,469 @@
+package controllers
+
+import (
+	"context"
+	"os"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	// import apps/v1 for Deployment
+	appsv1 "k8s.io/api/apps/v1"
+	// import metav1 for ObjectMeta
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	// import admissionregistration/v1 for MutatingWebhookConfiguration
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+)
+
+const (
+
+	// Define webhook name
+	webhookName = "mwebhook.peerpods.io"
+
+	// Define webhook deployment name
+	webhookDeploymentName = "peer-pods-webhook"
+
+	// Define webhook service name
+	webhookSvcName       = "peer-pods-webhook-svc"
+	webhookSvcSecretName = "peer-pods-webhook-svc-secret"
+	webhookDefaultImage  = "quay.io/confidential-containers/peer-pods-webhook:latest"
+
+	// Define webhook mutating config name
+	webhookConfigName = "mutating-webhook-configuration"
+)
+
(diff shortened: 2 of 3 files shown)


config/manager/manager.yaml
```diff
@@ -79,7 +79,9 @@ spec:
         - name: KATA_MONITOR_IMAGE
           value: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-monitor:latest
         - name: SANDBOXED_CONTAINERS_EXTENSION
-          value: kata-containers
+          value: sandboxed-containers
```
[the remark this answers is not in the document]
jensfr: true, we replace the value downstream. kata-containers was added for OKD I think.

jensfr (approved):
with Pavel's comment addressed and assuming you do a rebase on the peer-pods-tech-preview (required because I merged #313 first) it looks good to me. Thank you @bpradipt

config/manager/manager.yaml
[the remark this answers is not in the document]
bpradipt: yeah I'll change it

bpradipt:
@pmores addressed your comment. PTAL

controllers/ppwebhook.go
```diff
@@ -0,0 +1,470 @@
...
+			ClusterIP: "",
+			Type:      webhookServiceType,
+		},
+	}
+
+	// Create webhook service
+	if err := r.Client.Create(context.Background(), webhookService); err != nil {
+		// Check if the webhook service already exists
+		if !k8serrors.IsAlreadyExists(err) {
+			r.Log.Error(err, "peerpods mutating webhook service already exists")
```
gkurz, line 89: The error message is backwards with respect to the check. Since the error is already logged by the caller, what about just dropping this line ?

controllers/ppwebhook.go
```diff
@@ -0,0 +1,470 @@
...
+			Strategy: webhookDeploymentStrategy,
+			Template: webhookDeploymentPodTemplateSpec,
+		},
+	}
+
+	// Create webhook deployment
+	if err := r.Client.Create(context.Background(), webhookDeployment); err != nil {
+		// Check if the webhook deployment already exists
+		if !k8serrors.IsAlreadyExists(err) {
+			r.Log.Error(err, "peerpods mutating webhook deployment already exists")
```
gkurz, line 251: Ditto

controllers/ppwebhook.go
```diff
@@ -0,0 +1,470 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	// Create MutatingWebhookConfiguration object
+	if err := r.Client.Create(context.Background(), mutatingWebhookConfig); err != nil {
+		if !k8serrors.IsAlreadyExists(err) {
+			r.Log.Error(err, "Failed to create MutatingWebhookConfiguration object")
```
gkurz, line 402: The text is okay but the caller is also logging the error. Do we need both ? This is a general question for the rest of the patch actually.

gkurz: Also is it worth a call to `Error()` or `Info()` should be enough ?

config/manager/manager.yaml
```diff
@@ -82,6 +82,8 @@ spec:
           value: kata-containers
         - name: RELATED_IMAGE_CAA
           value: quay.io/confidential-containers/cloud-api-adaptor
+        - name: RELATED_IMAGE_PEERPODS_WEBHOOK
+          value: "quay.io/confidential-containers/peer-pods-webhook:latest"
```
gkurz, line 86: It looks a bit inconsistent to have one value with quotes when the one just above hasn't.

According to https://symfony.com/doc/current/reference/formats/yaml.html#strings , strings should be quoted in YAML when containing _special characters_ such as `:` or `-`. This change is thus ok but other values should be fixed in a subsequent PR.

bpradipt:
@gkurz ptal

PR #320, opened 2023-05-25
Allow to skip `go test`

gkurz:
`make test` has been broken since #300 was merged and fixing it is going to take time (see #310 for details). It is a problem for local development since the `test` target is a prequisite for `docker-build` in the makefile.

Add a variable to control whether `go test` should run when the `test` target is invoked.

 $ make test SKIP_TESTS=   # Run `go test` when unset

 $ make test SKIP_TESTS=1  # Don't run `go test` when set

Default is unset. [edit]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,12 @@ space := $() $()
 comma := ,
 GOFLAGS := -tags=$(subst $(space),$(comma),$(strip $(BUILTIN_CLOUD_PROVIDERS)))
 
+#
+# If SKIP_TESTS is set, the test target will *not* run `go test`.
+# This is to be able to temporarily work around test failures when doing
+# local development.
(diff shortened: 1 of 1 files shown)


pmores (approved):
I believe this is a good if not necessary change for the time being. Our approach to `make test` implementation needs serious rethinking. Until then, the current implementation seems fragile with a lot of false failures and very tedious to maintain while not bringing in much benefit (very few if any bugs were pointed out by `make test` to my knowledge).

gkurz:
Forced push with `default is unset` so that everyone keeps in mind that `make test` is still broken.

cpmeadors (approved):
Premerge tested per PR description.  All cases work.  Code looks good.

PR #323, opened 2023-06-08
image-job: fix payload url

snir911:


gkurz:
Hmm... isn't the `peer-pods-tech-preview` branch officialy dead now ?

cpmeadors:
Yes, peer-pods-tech-preview is dead.  Everything should go to devel now.

PR #324, opened 2023-06-08
image-job: fix payload url

snir911:
devel branch


littlejawa (approved):
lgtm
Thanks @snir911

gkurz:
> devel branch

The change is good but the `devel` branch isn't up to date AFAICT. It seems we should merge `main` into `devel` before going ahead with this fix.

gkurz (approved):
/lgtm

Thanks @snir911 !

PR #325, opened 2023-06-09
Merge main into devel after v1.4.0 release

gkurz:
Now that v1.4.0 is GA, let's merge all the changes back into the development branch.


pmores (approved):
So as it turns out, actually verifying the correctness of this merge is in fact very hard but since this should be a simple operation I'm assuming it will turn out right...

gkurz:
> So as it turns out, actually verifying the correctness of this merge is in fact very hard but since this should be a simple operation I'm assuming it will turn out right...

Yeah... we did some merge/rebase dance involving `devel`, `main` and `peer-pods-tech-preview` when we switched to the new workflow (see #311 #312 and #315).

Now, we want `devel` and `main` to point to the same commit, i.e. d255dac238a8aed813c904427b29f8b827037893 (tag v1.4.0). This PR does that.

cpmeadors (approved):
I don't seen any surprises.  All the changes look familiar.  Looks good.

PR #327, opened 2023-06-23
Fix kataconfig status handling to support installation updates

pmores:
This PR aims to fix KataConfig.status updating to handle installation modifications (adding and removing kata-enabled Nodes) to the extent I believe possible without changing the KataConfig.status structure itself.

Some not so intuitive aspects of the result, apart from names dictated by the current form of KataConfig.status, include:
- *all* Done Nodes will appear on one of the Completed lists, not just the ones affected by the ongoing or most recent operation.  This is because I'm not aware of any way of telling apart Nodes that had arrived in their state earlier as opposed to as a result of the most recent operation.
- Nodes scheduled to change state (=install/uninstall kata) as part of an ongoing operation but still waiting to do so are not listed anywhere since there doesn't seem to be a good place for them in the current KataConfig.status.  They will properly appear though on an InProgress list once they actually start updating and on a Completed list once they finish updating.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -651,23 +651,6 @@ func (r *KataConfigOpenShiftReconciler) checkNodeEligibility() error {
 	return nil
 }
 
-func (r *KataConfigOpenShiftReconciler) getMcpNameIfMcpExists() (string, error) {
-	r.Log.Info("Getting MachineConfigPool Name")
-
-	kataOC, err := r.kataOcExists()
-	if kataOC && err == nil {
-		r.Log.Info("kata-oc MachineConfigPool exists")
-		return "kata-oc", nil
-	}
-	isConvergedCluster, err := r.checkConvergedCluster()
-	if err == nil && isConvergedCluster {
-		r.Log.Info("Converged Cluster. Not creating kata-oc MCP")
-		return "master", nil
-	}
-	r.Log.Info("No valid MCP found")
-	return "", err
-}
-
 func (r *KataConfigOpenShiftReconciler) getMcpName() (string, error) {
 	isConvergedCluster, err := r.checkConvergedCluster()
 	if err != nil {
@@ -938,7 +921,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 		r.kataConfig.Status.WaitingForMcoToStart = false
 	}
 
-	err = r.updateStatusNew()
+	err = r.updateStatus()
 	if err != nil {
 		r.Log.Info("Error updating KataConfig.status", "err", err)
 	}
@@ -1143,7 +1126,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		r.kataConfig.Status.WaitingForMcoToStart = false
 	}
 
-	err = r.updateStatusNew()
+	err = r.updateStatus()
 	if err != nil {
 		r.Log.Info("Error updating KataConfig.status", "err", err)
 	}
@@ -1538,22 +1521,6 @@ func (r *KataConfigOpenShiftReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Complete(r)
 }
 
-func (r *KataConfigOpenShiftReconciler) getMcp() (*mcfgv1.MachineConfigPool, error) {
-	machinePool, err := r.getMcpNameIfMcpExists()
-	if err != nil {
-		return nil, err
-	}
-
-	foundMcp := &mcfgv1.MachineConfigPool{}
-	err = r.Client.Get(context.TODO(), types.NamespacedName{Name: machinePool}, foundMcp)
-	if err != nil {
-		r.Log.Error(err, "Getting MachineConfigPool failed ", "machinePool", machinePool)
-		return nil, err
-	}
-
-	return foundMcp, nil
-}
-
 func (r *KataConfigOpenShiftReconciler) getNodes() (error, *corev1.NodeList) {
 	nodes := &corev1.NodeList{}
 	labelSelector := labels.SelectorFromSet(map[string]string{"node-role.kubernetes.io/worker": ""})
@@ -1814,7 +1781,7 @@ func (r *KataConfigOpenShiftReconciler) processDoneNode(node *corev1.Node) error
 
 // If multiple errors occur during execution of this function the last one
 // will be returned.
-func (r *KataConfigOpenShiftReconciler) updateStatusNew() error {
+func (r *KataConfigOpenShiftReconciler) updateStatus() error {
 
 	if r.kataConfig.Status.UnInstallationStatus.InProgress.IsInProgress != corev1.ConditionTrue &&
 		r.kataConfig.Status.InstallationStatus.IsInProgress != corev1.ConditionTrue {
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1815,6 +1815,55 @@ func (r *KataConfigOpenShiftReconciler) processDoneNode(node *corev1.Node) error
 	return nil
 }
 
+// If multiple errors occur during execution of this function the last one
+// will be returned.
+func (r *KataConfigOpenShiftReconciler) updateStatusNew() error {
+
+	if r.kataConfig.Status.UnInstallationStatus.InProgress.IsInProgress != corev1.ConditionTrue &&
+		r.kataConfig.Status.InstallationStatus.IsInProgress != corev1.ConditionTrue {
+		return nil
+	}
+
+	err, nodeList := r.getNodes()
+	if err != nil {
+		return err
+	}
+
+	r.clearInstallStatus()
+	r.clearUninstallStatus()
+
+	r.kataConfig.Status.TotalNodesCount = func() int {
+		err, nodes := r.getNodesWithLabels(r.getNodeSelectorAsMap())
+		if err != nil {
+			r.Log.Info("Error retrieving kata-oc labelled Nodes to count them", "err", err)
+			return 0
+		}
+		return len(nodes.Items)
+	}()
+
+	for _, node := range nodeList.Items {
+		if annotation, ok := node.Annotations["machineconfiguration.openshift.io/state"]; ok {
+			switch annotation {
+			case NodeDone:
+				e := r.processDoneNode(&node)
+				if e != nil {
+					err = e
+				}
+			case NodeDegraded:
+			case NodeWorking:
+				e := r.processTransitioningNode(&node)
+				if e != nil {
+					err = e
+				}
+			default:
+				err = fmt.Errorf("Unexpected machineconfiguration.openshift.io/state: %v ", annotation)
+				r.Log.Info("Unexpected machineconfiguration.openshift.io/state", "node", node.GetName(), "state", annotation)
+			}
+		}
+	}
+	return err
+}
+
 func (r *KataConfigOpenShiftReconciler) updateStatus(machinePool string) (*mcfgv1.MachineConfigPool, ctrl.Result, error, bool) {
 	/* update KataConfig according to occurred error
 	 * We need to pull the status information from the machine config pool object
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -1679,6 +1679,69 @@ func (r *KataConfigOpenShiftReconciler) getMcpByName(mcpName string) (*mcfgv1.Ma
...
+				log.Info("putting done node on uninstallation completed list")
+				*uninstallCompletedList = append(*uninstallCompletedList, node.GetName())
+				r.kataConfig.Status.UnInstallationStatus.Completed.CompletedNodesCount = int(targetMcp.Status.UpdatedMachineCount)
+			} else {
+				log.Info("putting done node on installation completed list")
+				*installCompletedList = append(*installCompletedList, node.GetName())
+				r.kataConfig.Status.InstallationStatus.Completed.CompletedNodesCount = int(targetMcp.Status.UpdatedMachineCount)
+			}
+		} else if r.kataConfig.Status.UnInstallationStatus.InProgress.IsInProgress == corev1.ConditionTrue {
+			log.Info("putting done node on uninstallation completed list")
```
gkurz, line 1763: It could be valuable to distinguish this log from the other one above right away. Maybe make the first one "putting done worker node" ?

pmores: Yes, I understand where you're coming from, I was trying to find a good way to do that. I gave up since the philosophy of the logger used here is to use the key/value pairs to convey additional information rather than embed it in the message itself printf()-style, and I didn't feel comfortable going against it.

Note that although the two messages look the same in the source code they aren't the same actually since we're logging here via a custom logger with added keys/values - so the information that this is a member of "worker" is there already .

But I'm not 100% happy with the result either so if the difference doesn't seem enough I'm open to change this.

gkurz: The logger is :

```
r.Log.WithName("updateStatus").WithValues("node name", node.GetName(), "target pool", targetMcpName, "current config", currentNodeConfig, "target pool config", targetMcp.Spec.Configuration.Name)
```

In a non-converged scenario, `targetMcpName` will be `worker` in both cases. Only differentiator that comes to mind is the source MCP, either `kata-oc` (case 2) or `worker` (case 1).

In a converged scenario, the case 1 cannot be reached since "worker" will never be the target MCP and case 2 is `master`.

It isn't needed to determine the source MCP. Mentioning "worker" (as originating MCP) in case 1 seems to be simple enough.

pmores: I'm not sure if I'm following the argument correctly.  Let me recap the main aspects of the situation as I understand them to see if there are any problems with it:
- this code deals with Nodes that are Done and whose MachineConfig matches the MachineConfig of the MCP that they belong to. In other words, these Nodes are currently settled and not moving anywhere. So I'm not sure what you mean by "source MCP".
- because of the previous point, these Nodes go to a Completed list.  Both of the places you point out deal with Nodes that go to the uninstallation's Completed list.  A settled Node can end up on that list for one of two different reasons.
- one is because we're uninstalling kata from a cluster.  A settled Node in that situation can't have kata installed on it, kata has either been uninstalled already or it was never installed on the Node in the first place.  This the case 2.
- the other is because we're updating kata installation on a cluster (which is handled as installation internally) and part of that update was removing kata from a Node.  This is the case 1.  Such Node is identified by belonging to the "worker" MCP.
Note that this code also handles Nodes that belong to "worker" either because they never had kata installed on them or because they did but it was uninstalled previously already, since all these scenarios are indistinguishable I believe.

We simply see a Node that should and does belong to the "worker" MCP and we have no idea how it got there.

(It works the same on converged clusters but it's a bit simpler - since no Node can belong to "worker" the logic simplifies to "if we're putting kata on cluster any settled Node must have completed its kata installation, if we're removing kata from cluster any settled Node must have completed its kata uninstallation".)

gkurz: My point is that the two instances of `Log.Info()` produce exactly the same output (I've verified it) => it is impossible to guess which branch the logs comes from with a simple grep.

pmores: Okay, that's true, but if there's a difference between the two cases, it's not that the Node is in "worker" (which is the case in both log messages) but that we're "installing" in case 1 (in the confusingly wide meaning of the word) but uninstalling in case 2.  So do you want me to include _that_ information in the log messages?

pmores: Hm, but what does the proposed code do when `InstallationStatus.IsInProgress == corev1.ConditionTrue` and `targetMcpName == "worker"`?  Correct me if I'm wrong but it seems to me that it will (incorrectly) do nothing, considering that `InstallationStatus.IsInProgress == corev1.ConditionTrue` implies `UnInstallationStatus.InProgress.IsInProgress == corev1.ConditionFalse`.

I guess this could be fixed, perhaps by changing the `else if` condition to `UnInstallationStatus.InProgress.IsInProgress == corev1.ConditionTrue || targetMcpName == "worker"`, at which point however I'd strongly doubt that the new formulation is any better, in fact it seems to me quite a lot more tricky to understand and verify.

gkurz: Alright, maybe disentangle the checking and the processing then ?

```
		installingNode  := false
		uninstallingNode := false
		if r.kataConfig.Status.InstallationStatus.IsInProgress == corev1.ConditionTrue {
			if targetMcpName == "worker" {
				uninstallingNode = true
			} else {
				installingNode = true
			}
		} else if r.kataConfig.Status.UnInstallationStatus.InProgress.IsInProgress == corev1.ConditionTrue {
			uninstallingNode = true
		}

		if installingNode {
			log.Info("putting done node on installation completed list")
			*installCompletedList = append(*installCompletedList, node.GetName())
			r.kataConfig.Status.InstallationStatus.Completed.CompletedNodesCount = int(targetMcp.Status.UpdatedMachineCount)
		} else if uninstallingNode {
			log.Info("putting done node on uninstallation completed list")
			*uninstallCompletedList = append(*uninstallCompletedList, node.GetName())
			r.kataConfig.Status.UnInstallationStatus.Completed.CompletedNodesCount = int(targetMcp.Status.UpdatedMachineCount)
		}
```

gkurz: Yes you're right. Getting rid of the duplication isn't interesting here.

controllers/openshift_controller.go
```diff
@@ -1679,6 +1679,69 @@ func (r *KataConfigOpenShiftReconciler) getMcpByName(mcpName string) (*mcfgv1.Ma
...
+		putOnInstallList = func(node *corev1.Node) {
+			installFailedList := &r.kataConfig.Status.InstallationStatus.Failed.FailedNodesList
+			*installFailedList = append(*installFailedList, failedNodeStatus)
+		}
+		putOnUninstallList = func(node *corev1.Node) {
+			uninstallFailedList := &r.kataConfig.Status.UnInstallationStatus.Failed.FailedNodesList
+			*uninstallFailedList = append(*uninstallFailedList, failedNodeStatus)
+		}
+	}
+
```
gkurz, line 1678: This `if {} else {}`  tends to indicate the code up to here rather belongs to some caller IMHO since it is presumably already branching based on the "machineconfiguration.openshift.io/state" annotation.

Maybe move these lines to intermediate `processTransitioningWorkingNode()` and `processTransitioningDegradedNode()` functions and keep the rest in `processTransitioningNode()` called by the other two ?

controllers/openshift_controller.go
```diff
@@ -1805,6 +1805,55 @@ func (r *KataConfigOpenShiftReconciler) processDoneNode(node *corev1.Node) error
...
+	r.clearInstallStatus()
+	r.clearUninstallStatus()
+
+	r.kataConfig.Status.TotalNodesCount = func() int {
+		err, nodes := r.getNodesWithLabels(r.getNodeSelectorAsMap())
+		if err != nil {
+			r.Log.Info("Error retrieving kata-oc labelled Nodes to count them", "err", err)
+			return 0
```
gkurz, line 1793: Is this right ? Shouldn't `updateStatusNew()` bail out if it cannot retrieve the list of nodes ?

controllers/openshift_controller.go
```diff
@@ -1805,6 +1805,55 @@ func (r *KataConfigOpenShiftReconciler) processDoneNode(node *corev1.Node) error
...
+		}
+		return len(nodes.Items)
+	}()
+
+	for _, node := range nodeList.Items {
+		if annotation, ok := node.Annotations["machineconfiguration.openshift.io/state"]; ok {
+			switch annotation {
+			case "Done":
```
gkurz, line 1801: I see there are constants for the state annotation in the MCO source tree. I'm wondering if they should be used instead of hardcoded strings ?

gkurz (approved):
My overall impression is that this PR really improves things.
Most of my remarks, except maybe the one on the ignored error, can still be addressed in a followup, so I'm approving :smiley: 

Thanks @pmores !

controllers/openshift_controller.go
pmores: That's a good idea, I can't find them though, can you point me to them?  Are they public?

I only used literals because the legacy implementation used them and frankly it didn't occur to me that they might be defined somewhere in the MCO. :-)

controllers/openshift_controller.go
pmores: It's a judgment call I guess.  This failure would only prevent setting the `TotalNodesCount` correctly but the function would still be able to do the bulk of its job (it has already retrieved `nodeList` above successfully otherwise we wouldn't be here in the first place).

Given this, and also the fact that this problem will likely go away next time around `updateStatus()` is called, I thought it'd be a shame to leave all of the status not updated and stale just because we couldn't count nodes, probably temporarily.

controllers/openshift_controller.go
gkurz: These constants are defined at the machine config daemon level. They are accessible but I don't know if they are actually public.

States are defined here :

https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/constants/constants.go#L36-L44

Note there's also an `"Unreconcilable"` state as well. Not sure if it needs to be handled as well.

The annotation key is also defined BTW :

https://github.com/openshift/machine-config-operator/blob/master/pkg/daemon/constants/constants.go#L16

controllers/openshift_controller.go
```diff
@@ -1679,6 +1679,69 @@ func (r *KataConfigOpenShiftReconciler) getMcpByName(mcpName string) (*mcfgv1.Ma
 	return mcp, nil
 }
 
+func (r *KataConfigOpenShiftReconciler) processTransitioningNode(node *corev1.Node) error {
+	// Don't bother checking the second return value, the annotation is there
+	// otherwise we wouldn't be here (our caller checked already).
+	mcoNodeState, _ := node.Annotations["machineconfiguration.openshift.io/state"]
+
```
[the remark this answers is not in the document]
pmores: I actually don't agree with treating every single function like it was part of a public API, callable by anyone.  This is a private function, a helper to `updateStatus()` specifically.  It's only reason to be is that the code is too bulky to fit into a `case` arm of the `switch` statement in its only caller so I factored it out - it's not a function that anybody else will ever call.

So, because `updateStatus()` and `processTransitioningNode()` (and `processDoneNode()` too) form a tight subsystem, I believe it's appropriate for the helper to rely on work done by the caller.

controllers/openshift_controller.go
gkurz: Fair enough.

controllers/openshift_controller.go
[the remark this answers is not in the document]
gkurz: We don't agree then. Since we don't know who will have to maintain this code in the long term, we should do best efforts to have clear APIs, even internally. Putting all the code in the `case` statement of `updateStatusNew()` is certainly not what I'm suggesting.

I've approved the PR anyway :wink:

controllers/openshift_controller.go
pmores: Thanks, I'm working to fix this.

pmores:
Force-pushed a fix to the inappropriate string literals usage.

controllers/openshift_controller.go
gkurz: Fair enough if you've tried it out before then ;-)

gkurz:
> Force-pushed a fix to the inappropriate string literals usage.

Works much better :smiley:

PR #328, opened 2023-06-26
Allow monitor pods to run on tainted nodes

gkurz:

No metrics for sandboxed containers running on a tainted node.

**- What I did**

Allow monitor pods to tolerate all taints.

**- How to verify it**

1. Install OSC.
2. Taint worker nodes.
3. Create KataConfig.
4. Monitor pods should be running on all nodes in the kata-oc MCP.

The DaemonSet launched by the controller doesn't put tolerations on the monitor pods. This prevents the monitor pods to run on tainted nodes and break metrics for sandboxed containers.

The monitor pods must run on every node where kata is deployed, no matter any taint the user might have set. Enforce a toleration that matches all possible taints.

Fixes https://issues.redhat.com/browse/KATA-2121

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -384,6 +384,11 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.Dae
 				Spec: corev1.PodSpec{
 					ServiceAccountName: "monitor",
 					NodeSelector:       nodeSelector,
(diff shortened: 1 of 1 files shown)


tbuskey:
pre-merge launches kata-monitor on all nodes, included tainted one(s) - QE
/lgtm

pmores (approved):
Thanks @gkurz !

PR #329, opened 2023-06-30
New way of showing kata nodes in kataconfig status

pmores:
This PR adds a new section to `KataConfig.status`  called `kataNodes` which lists cluster nodes grouped by the status of kata installation on them which can be one of installed, installing, waiting to install, failed to install, not installed, uninstalling, waiting to uninstall and failed to uninstall.

By way of example, when applying a KataConfig that selects both nodes of a two-worker cluster, the newly added section might show

```
  kataNodes:
    waitingToInstall:
    - worker-0
    - worker-1
```
for a brief period of time, then something like
```
  kataNodes:
    installing:
    - worker-1
    waitingToInstall:
    - worker-0
```
Once `worker-1` is installed the other worker starts installing
```
  kataNodes:
    installed:
    - worker-1
    installing:
    - worker-0
```
and finally
```
  kataNodes:
    installed:
    - worker-0
    - worker-1
```
The new addition runs alongside the existing status reporting for now.

PR #330, opened 2023-07-04
Switch to condition for progress status reporting

pmores:
This PR adds a conventional Kubernetes Condition type and a Condition array to KataConfig.status.  The only Condition currently supported is InProgress with three possible Reasons - Installing, Updating and Uninstalling.

The final commit doesn't really fit here since it fixes an omission in #329 but I included it anyway since it's a tiny change that probably wouldn't be worth a separate PR IMHO.

With this PR I believe all is set to remove the legacy parts of KataConfig.status.  Anybody depending on them is encouraged to migrate to the new KataConfig.status structure.

As a migration guide, note that
- `installationStatus.IsInProgress == True` should correspond to `InProgress.status == True` and `InProgress.reason` either `Installing` or `Updating`
- `unInstallationStatus.inProgress.status == True` should correspond to `InProgress.status == True` and `InProgress.reason == Uninstalling`
- failure during installation or uninstallation is represented by  `InProgress.status == True` and `InProgress.reason == Failed` with an appropriate `InProgress.message`
- `totalNodesCount` should correspond to `kataNodes.nodeCount`
- installation's completed, inprogress and failed node lists should correspond to `kataNodes.installed`, `kataNodes.installing` and `kataNodes.failedToInstall` lists, respectively
- uninstallation's inprogress and failed node lists should correspond to `kataNodes.uninstalling` and `kataNodes.failedToUninstall` lists, respectively.  There's no equivalent of the completed list since it seems semantically cleaner to focus on kata-enabled nodes and there are no known users of that list to my knowledge.  Any node that is on no list in `kataNodes` is implicitly on a `notInstalled` list.

The planned removal of legacy `KataConfig.status` parts only concerns `installationStatus`, `unInstallationStatus`, `totalNodesCount` and `upgradeStatus` (which was never implemented anyway).  All other parts of `KataConfig.status` are meant to remain in place.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1828,6 +1828,8 @@ func (r *KataConfigOpenShiftReconciler) updateStatus() error {
 		return len(nodes.Items)
 	}()
 
+	r.kataConfig.Status.KataNodes.NodeCount = r.kataConfig.Status.TotalNodesCount
+
 	for _, node := range nodeList.Items {
 		if annotation, ok := node.Annotations["machineconfiguration.openshift.io/state"]; ok {
 
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -2025,12 +2025,83 @@ func (r *KataConfigOpenShiftReconciler) findInProgressCondition() *kataconfigura
 	return nil
 }
 
+func (r *KataConfigOpenShiftReconciler) getInProgressConditionValue() corev1.ConditionStatus {
+	cond := r.findInProgressCondition()
+	if cond == nil {
+		return corev1.ConditionUnknown
+	}
+	return cond.Status
+}
+
 func (r *KataConfigOpenShiftReconciler) addInProgressCondition() *kataconfigurationv1.KataConfigCondition {
 	r.kataConfig.Status.Conditions = append(r.kataConfig.Status.Conditions, kataconfigurationv1.KataConfigCondition{Type: kataconfigurationv1.KataConfigInProgress})
 
 	r.Log.Info("InProgress Condition added")
 
 	return &r.kataConfig.Status.Conditions[len(r.kataConfig.Status.Conditions)-1]
+}
+
+// This is just a technical helper to all InProgress Condition mutators,
+// factoring their common preamble out into an own function.
+func (r *KataConfigOpenShiftReconciler) retrieveInProgressConditionForChange() *kataconfigurationv1.KataConfigCondition {
+	cond := r.findInProgressCondition()
+	if cond == nil {
+		cond = r.addInProgressCondition()
+	}
+
+	cond.LastTransitionTime = metav1.Now()
+
+	return cond
+}
+
+func (r *KataConfigOpenShiftReconciler) setInProgressConditionToInstalling() {
+	cond := r.retrieveInProgressConditionForChange()
+	cond.Status = corev1.ConditionTrue
+	cond.Reason = "Installing"
+	cond.Message = "Performing initial installation of kata on cluster"
+
+	r.Log.Info("InProgress Condition set to Installing")
+}
+
+func (r *KataConfigOpenShiftReconciler) setInProgressConditionToUninstalling() {
+	cond := r.retrieveInProgressConditionForChange()
+	cond.Status = corev1.ConditionTrue
+	cond.Reason = "Uninstalling"
+	cond.Message = "Removing kata from cluster"
+
+	r.Log.Info("InProgress Condition set to Uninstalling")
+}
+
+func (r *KataConfigOpenShiftReconciler) setInProgressConditionToUpdating() {
+	cond := r.retrieveInProgressConditionForChange()
+	cond.Status = corev1.ConditionTrue
+	cond.Reason = "Updating"
+	cond.Message = "Adding and/or removing kata-enabled nodes"
+
+	r.Log.Info("InProgress Condition set to Updating")
+}
+
+func (r *KataConfigOpenShiftReconciler) setInProgressConditionToFailed(failingNode *corev1.Node) {
+	reasonForDegraded, ok := failingNode.Annotations["machineconfiguration.openshift.io/reason"]
+	if !ok {
+		r.Log.Info("Missing machineconfiguration.openshift.io/reason on Degraded node", "node", failingNode.GetName())
+	}
+
+	cond := r.retrieveInProgressConditionForChange()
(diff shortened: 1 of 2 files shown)


controllers/openshift_controller.go
```diff
@@ -2033,6 +2041,69 @@ func (r *KataConfigOpenShiftReconciler) addInProgressCondition() *kataconfigurat
...
+func (r *KataConfigOpenShiftReconciler) setInProgressConditionToFailed(failingNode *corev1.Node) {
+	reasonForDegraded, ok := failingNode.Annotations["machineconfiguration.openshift.io/reason"]
+	if !ok {
+		r.Log.Info("Missing machineconfiguration.openshift.io/reason on Degraded node", "node", failingNode.GetName())
+	}
+
+	cond := r.retrieveInProgressConditionForChange()
+	cond.Status = corev1.ConditionTrue
+	cond.Reason = "Failed"
+	cond.Message = "Node " + failingNode.GetName() + " Degraded: " + reasonForDegraded
```
gkurz, line 2109: Mention kata as well here ?

gkurz (approved):
LGTM. Thanks @pmores !

controllers/openshift_controller.go
```diff
@@ -2033,6 +2041,69 @@ func (r *KataConfigOpenShiftReconciler) addInProgressCondition() *kataconfigurat
...
+	cond.LastTransitionTime = metav1.Now()
+
+	return cond
+}
+
+func (r *KataConfigOpenShiftReconciler) setInProgressConditionToInstalling() {
+	cond := r.retrieveInProgressConditionForChange()
+	cond.Status = corev1.ConditionTrue
+	cond.Reason = "Installing"
+	cond.Message = "Performing initial installation on cluster"
```
gkurz, line 2077: To help grepability in some log blob and for consistency with the other mutators :

```suggestion
	cond.Message = "Performing initial installation of kata on cluster"
```

controllers/openshift_controller.go
```diff
@@ -2016,6 +2032,94 @@ func (r *KataConfigOpenShiftReconciler) clearUninstallStatus() {
...
+	reasonForDegraded, ok := failingNode.Annotations["machineconfiguration.openshift.io/reason"]
+	if !ok {
+		r.Log.Info("Missing machineconfiguration.openshift.io/reason on Degraded node", "node", failingNode.GetName())
+	}
+
+	cond := r.retrieveInProgressConditionForChange()
+	cond.Status = corev1.ConditionTrue
+	cond.Reason = "Failed"
+	cond.Message = "Node " + failingNode.GetName() + " Degraded: " + reasonForDegraded
+
```
gkurz, line 2110: Mention kata as well here ?

controllers/openshift_controller.go
pmores: Makes sense, thanks - it's fixed now.

controllers/openshift_controller.go
pmores: Not quite sure how though.

gkurz: > Not quite sure how though.

Callers of `setInProgressConditionToFailed()` know if install or uninstall is in progress. This could be passed as a bool and the message would be "Node XXX got degraded with reason YYY while  (un)installing kata".

pmores: Ah okay, that's true.  Although the information whether the affected Node was installing or uninstalling is readily available by looking at `kataNodes.failedToInstall` and `kataNodes.failedToUninstall` lists.

Generally, a Condition.message is a simple string and, I believe, not designed or meant to carry high-resolution information.  Consider an actually frequently happening case of multiple Nodes becoming Degraded at the same time.  Condition.message can never do justice to accurately describing this.  Sometimes Nodes can even become Degraded while installing kata as others while uninstalling it, possibly with different `machineconfiguration.openshift.io/reason`s.

The idea of InProgress Condition.message seems to me to be more like "something went wrong, please check out `kataNodes.failedToInstall` and `kataNodes.failedToUninstall` for more detail, or `Node.metadata.annotations.machineconfiguration.openshift.io/reason` for affected Nodes if even that is not enough".

Maybe the current form is misleading, trying to save some work to the user by including the MCO `reason` message for the simple and probably most common cases of a single Node becoming Degraded or a group of Nodes becoming Degraded for the same reason, and thus raising false expectations that does more than it actually can?

pmores: Come to think about it, how about we include each failed Node's `machineconfigmachineconfiguration.openshift.io/reason` in parens after Node name directly the `failedToInstall` and `failedToUninstall` lists and restrict the `Condition.message` to just something along the lines of "Node(s) failed, please see the failedToInstall and failedToUninstall lists for more information"?

snir911:
I'm not familiar with this part of the code but LGTM overall, one question, wouldn't it make sense to remove the legacy parts along with this PR? who depends on in? (sorry if it was discussed already)

littlejawa (approved):
Thanks @pmores

pmores:
> I'm not familiar with this part of the code but LGTM overall, one question, wouldn't it make sense to remove the legacy parts along with this PR? who depends on in? (sorry if it was discussed already)

Automated tests (QE/CI) can (and do afaik) parse status to find out if result of a test matches expectation, there might be others.

As much as I'm itching to remove the legacy parts, too, let's have a grace period where both approaches are available and migration between them can be done without pressure.

tbuskey:
Thanks for a transition.

I like the `{.status.conditions}` section.  A number of other things in k8s use that form.  The current legacy statuses always felt wrong; I had to watch 2 statuses to know if things were updating.

Should the status be a boolean instead of a string?  `{.status.waitingForMcoToStart}` is a bool

pmores:
> Should the status be a boolean instead of a string? `{.status.waitingForMcoToStart}` is a bool

Assuming you're talking about `Condition.status`, then no, most likely.  The `Condition` structure tries to follow the convention given in [k8s docs](https://github.com/kubernetes/community/blob/4c9ef2d/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) where `status` defined as `metav1.ConditionStatus` instead of boolean, mainly to allow it to be tristate IIUC.

(`{.status.waitingForMcoToStart}` is allowed to be whatever it wants since it's actually an undocumented private piece of data that the controller needs to persist. You can only see it because there's no notion of private with respect to k8s resources.)

tbuskey:
All tests passed with kata runtime

PR #333, opened 2023-07-11
Remove legacy kata config status parts

pmores:
This PR removes parts of `KataConfig.status` obsoleted by #330 (and #329 before it), `totalNodesCount`, `installationStatus`, `unInstallationStatus` (and the stub of `upgradeStatus`).

### DIFF
--- a/api/v1/kataconfig_types.go
+++ b/api/v1/kataconfig_types.go
@@ -85,9 +85,9 @@ type KataConfigStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=kataconfigs,scope=Cluster
-// +kubebuilder:printcolumn:name="InProgress",type=string,JSONPath=".status.installationStatus.IsInProgress",description="Status of Kata runtime installation"
-// +kubebuilder:printcolumn:name="Completed",type=integer,JSONPath=".status.installationStatus.completed.completedNodesCount",description="Number of nodes with Kata runtime installed"
-// +kubebuilder:printcolumn:name="Total",type=integer,JSONPath=".status.totalNodesCount",description="Total number of nodes"
+// +kubebuilder:printcolumn:name="InProgress",type=string,JSONPath=".status.conditions[?(@.type=='InProgress')].status",description="Status of Kata runtime installation"
+// +kubebuilder:printcolumn:name="Completed",type=integer,JSONPath=".status.kataNodes.readyNodeCount",description="Number of nodes with Kata runtime installed"
+// +kubebuilder:printcolumn:name="Total",type=integer,JSONPath=".status.kataNodes.nodeCount",description="Total number of nodes"
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=".metadata.creationTimestamp",description="Age of the KataConfig Custom Resource"
 type KataConfig struct {
 	metav1.TypeMeta   `json:",inline"`
--- a/api/v1/kataconfig_types.go
+++ b/api/v1/kataconfig_types.go
@@ -205,6 +205,11 @@ type KataNodesStatus struct {
 	// +optional
 	NodeCount int `json:"nodeCount"`
 
+	// Number of cluster nodes that have kata installed on them and are
+	// currently ready to run kata workloads.
+	// +optional
+	ReadyNodeCount int `json:"readyNodeCount"`
+
 	// +optional
 	Installed []string `json:"installed,omitempty"`
 	// +optional
--- a/api/v1/kataconfig_types.go
+++ b/api/v1/kataconfig_types.go
@@ -51,21 +51,6 @@ type KataConfigStatus struct {
 	// RuntimeClass is the name of the runtime class used in CRIO configuration
 	RuntimeClass string `json:"runtimeClass"`
 
-	// TotalNodesCounts is the total number of worker nodes targeted by this CR
-	TotalNodesCount int `json:"totalNodesCount"`
-
-	// InstallationStatus reflects the status of the ongoing kata installation
-	// +optional
-	InstallationStatus KataInstallationStatus `json:"installationStatus,omitempty"`
-
-	// UnInstallationStatus reflects the status of the ongoing kata uninstallation
-	// +optional
-	UnInstallationStatus KataUnInstallationStatus `json:"unInstallationStatus,omitempty"`
-
-	// Upgradestatus reflects the status of the ongoing kata upgrade
-	// +optional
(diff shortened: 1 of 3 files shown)


tbuskey:
Looks good with kata/peer-pods workloads

snir911 (approved):
LGTM thanks!

pmores:
Force-push to add a commit switching envtest to the new KataConfig.status (thanks @tbuskey for pointing this out)

controllers/openshift_controller.go
```diff
@@ -1828,31 +1682,15 @@ func (r *KataConfigOpenShiftReconciler) updateStatus() error {
...
-				if e != nil {
-					err = e
-				}
-			default:
-				err = fmt.Errorf("Unexpected machineconfiguration.openshift.io/state: %v ", annotation)
-				r.Log.Info("Unexpected machineconfiguration.openshift.io/state", "node", node.GetName(), "state", annotation)
-			}
+		e := r.putNodeOnStatusList(&node)
+		if e != nil {
+			err = e
```
littlejawa, line 1688: In case of multiple errors, we're keeping only the last one right?
Can we at least log the error here before we potentially lose it?
Or as an alternative: if we meet an error, is it useful to continue the loop?

api/v1/kataconfig_types.go
```diff
@@ -85,9 +70,9 @@ type KataConfigStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=kataconfigs,scope=Cluster
-// +kubebuilder:printcolumn:name="InProgress",type=string,JSONPath=".status.installationStatus.IsInProgress",description="Status of Kata runtime installation"
```
jensfr, line 88: Why not mark these fields as deprecated and leave them in? We don't know what tools people have written that rely on these fields, i.e. the .status section to me is part of the API. 

We could even add logging for the case that any of the fields are non-empty (unlikely that another software has the rights to update it, but possible). After a while we can bump the version and remove the fields. What do you think?

controllers/openshift_controller.go
pmores: Yes, a comment at the top of the functions make that point.

As for error handling, yes, my idea was that it's still useful to continue. The way I see it, if there's an error our status will always end up imprecise for the moment but it would be even more imprecise if we broke off on the first error.  That said, I'm not sure at the moment how rigorous we should be here, I'm guessing some practical experience will help.  IIRC the legacy status update behaved the same way and I don't remember any practical problems with that.

Errors that can come up here will likely tend to be transient.

littlejawa: Sorry, I missed the comment at the top.
Thanks for the answer. Makes sense

api/v1/kataconfig_types.go
tbuskey: I think I can change our automation to check both.
`-o=jsonpath='{.status.conditions[?(@.type=="InProgress")].status}'` is the new one
`-o=jsonpath={.status.installationStatus.IsInProgress}{.status.unInstallationStatus.inProgress.status}` is the current one

As long as `ToLower()` != false, I know things are not in progress

pmores: @tbuskey I think that checking the new one should be enough by now.

jensfr: sorry for the delay. I think we don't know who's using it and in what way (we were surprised a few times already. We may break users that we are not even aware of. To be honest I think we should take the slow path. What do others think? @bpradipt @snir911 ?

bpradipt: I'm considering these aspects when providing my view on this PR.
We are adding/improving new features - peer-pods, CoCo which will increase the adoption and thereby make it relativey more harder to make this sort of a change.
Further, this change will only be available in newer versions, and is only affecting the install/uninstall of the operator with no impact to the workloads. Aso it'll be documented so that users relying on earlier status field can make suitable changes to use the newer fields. 

Given the above aspects I'm in favour of getting this merged.

bpradipt: @pmores I don't have any comments on the code.  
lgtm on the code front.

tbuskey:
QE automation has already been changed to look for .status.conditions and will use  it instead if it exists

gkurz (approved):
/lgtm

Thanks @pmores !

tbuskey:
I was able to unsubscribe & resubscribe while running a kata workload
`oc get kataconfig example-kataconfig -o yaml` transitioned to:
```
apiVersion: kataconfiguration.openshift.io/v1
kind: KataConfig
metadata:
  creationTimestamp: "2023-08-16T13:50:04Z"
  finalizers:
  - kataconfiguration.openshift.io/finalizer
  generation: 2
  name: example-kataconfig
  resourceVersion: "146822"
  uid: c22a6445-8c02-47e7-80b6-eae10d2b93a3
spec:
  checkNodeEligibility: false
  enablePeerPods: false
  kataConfigPoolSelector: null
  logLevel: info
status:
  conditions:
  - lastTransitionTime: "2023-08-16T18:21:44Z"
    message: ""
    reason: ""
    status: "False"
    type: InProgress
  kataNodes:
    nodeCount: 0
    readyNodeCount: 0
  runtimeClass: kata
  waitingForMcoToStart: false
```
The kataNodes statuses should be 3:
```
tbuskey-230816-1-bjff5-worker-eastus1-2v54t   Ready    kata-oc,worker   5h35m   v1.27.3+4aaeaec
tbuskey-230816-1-bjff5-worker-eastus2-hnkcr   Ready    kata-oc,worker   5h36m   v1.27.3+4aaeaec
tbuskey-230816-1-bjff5-worker-eastus3-mc77d   Ready    kata-oc,worker   5h35m   v1.27.3+4aaeaec
```

I will test peer-pods & comment if it is different

tbuskey:
When running with `peerpodsEnabled: true`, I was also able to unsub/sub while a peer-pors workload was running.
The kataconfig yaml transitioned to this:
```
apiVersion: kataconfiguration.openshift.io/v1
kind: KataConfig
metadata:
  creationTimestamp: "2023-08-16T18:57:12Z"
  finalizers:
  - kataconfiguration.openshift.io/finalizer
  generation: 2
  name: example-kataconfig
  resourceVersion: "170584"
  uid: 8d3ee88b-ff2f-48ff-befb-16487f2a2acf
spec:
  checkNodeEligibility: false
  enablePeerPods: true
  kataConfigPoolSelector: null
  logLevel: info
status:
  runtimeClass: kata
  waitingForMcoToStart: false
```
Deleting kataconfig puts the `deletionTimestamp` in the metatdata but the deletion doesn't proceed, even after 15 minutes.

PR #339, opened 2023-08-08
Fix kataconfig status handling to support installation updates (port to main)

gkurz:

When you try to delete kataconfig while a workload is running, you will get an error that kataconfig cannot be deleted while it is running a workload.

You should be able to delete the workload and kataconfig will remove the status and will proceed to delete kataconfig.  This is the previous 1.3.3 behavior.

Fixes: https://issues.redhat.com/browse/KATA-2240

**- What I did**

Cherry-picked all patches from #327 

**- How to verify it**

1. build OSC from this PR
2. deploy it on OCP 4.13.x
3. create kataconfig with  enablePeerPods: false
4. create a sandboxed pod workload
6. delete kataconfig
7. see the expected `Existing pods using Kata Runtime found. Please delete the pods manually for KataConfig deletion to proceed` error in the kataconfig's yaml status
8. delete the sandboxed pod workload
9. kataconfig should be deleted

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1684,6 +1684,69 @@ func (r *KataConfigOpenShiftReconciler) getMcpByName(mcpName string) (*mcfgv1.Ma
 	return mcp, nil
 }
 
+func (r *KataConfigOpenShiftReconciler) processDoneNode(node *corev1.Node) error {
+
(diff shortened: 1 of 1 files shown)


pmores (approved):
Thanks @gkurz !

PR #341, opened 2023-09-05
Use raw Azure base image for peer-pods

bpradipt:
The LVM based image results in limited free space for the root filesystem. This creates issues when using peer-pods for container builds. Switching to raw images provides flexibility w.r.to root filesystem storage

Fixes: [KATA-2362](https://issues.redhat.com//browse/KATA-2362)


bpradipt:
This PR doesn't need any downstream builds. 
Merging this will ensure the new yaml file gets used for folks following the OSC documentation.

snir911 (approved):

PR #343, opened 2023-09-20
podvm: run image jobs from the controller automatically

snir911:
EDITED:

This PR contains several features:
1. **AWS**: if image is not specified it will generate AMI image and update the CM with it when KataConfig is created (and it will be deleted when KataConfig is deleted)
2.  **Azure**: if image is not specified it will generate Azure image and update the CM with it when KataConfig is created (and it will be deleted when KataConfig is deleted)
3. **Azure&AWS**: if cluster is installed with **fips=true** all created images will boot with kernel cmdline parameter fips=1

To test use:
```
quay.io/snir/osc-operator:v1.4.8
quay.io/snir/osc-operator-bundle:v1.4.8
quay.io/snir/osc-operator-catalog:v1.4.8
```

### DIFF
--- a/controllers/image_generator.go
+++ b/controllers/image_generator.go
@@ -1,0 +1,407 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/go-logr/logr"
+	configv1 "github.com/openshift/api/config/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+/*
+The image generator builds and deletes pod VM images for a cloud provider. It uses Kubernetes jobs to do this.
+Make sure
+Ensure that all job YAML files meet the following guidelines:
+1. Job files should be located at config/peerpods/
+2. all file names must follow this format: <cloud-provider>-[CVM|VM]-image-[create|delete]-job.yaml
+3. parallelism: 1
+4. completions: 1 (or none)
+5. backoffLimit: 1
+6. restartPolicy: Never
+7. job and pod names describes cloud-provider and operation
+8. Get configuration values from peer-pods-secret and peer-pods-cm volumes and predefined enviroment variables
+9. make sure create/delete jobs are based on the same configuration sources
+10. utilize precreated podvm binaries conatiner if possible
+11. all jobs must abort on image creation failure
+
+Every image creation job must obey to the following guidelines:
+1. create the podvm image and all other required cloud resouces needed for it
+2. Create a container named result that outputs the image ID (and nothing but the image ID) once the image is ready.
+3. image ID can be shared between containers using emptyDir volume
+
+Every image deletion job should preform the following:
+1. based on the exect same configurtion creation job is using
+2. deletes the image and all created resouces
+*/
+
+const (
+	peerpodsCMName          = "peer-pods-cm"
+	peerpodsCMAWSImageKey   = "PODVM_AMI_ID"
+	peerpodsCMAzureImageKey = "AZURE_IMAGE_ID"
+	fipsCMKey               = "BOOT_FIPS"
+	defaultVMType           = "VM"
+)
+
+type ImageGenerator struct {
+	client    client.Client        // controller-runtime client
+	clientset kubernetes.Interface // k8s clientset
+
+	provider     string
+	CMimageIDKey string
+	fips         bool
+}
+
+var ig *ImageGenerator = nil
+var igLogger logr.Logger = ctrl.Log.WithName("image-generator")
+
+// ImageCreate creates a podvm image for a cloud provider if not present
+func ImageCreate(c client.Client) (bool, ctrl.Result) {
+	if ig == nil { // initialize ImageGenerator if not exsits
+		var err error
+		ig, err = newImageGenerator(c)
+		if err != nil || ig == nil {
+			igLogger.Info("error initializing ImageGenerator instance", "err", err)
+			return false, ctrl.Result{Requeue: true}
+		}
+	}
+	return ig.imageJobRunner("create")
+}
+
+// ImageDelete deletes a podvm image for a cloud provider if present
+func ImageDelete(c client.Client) (bool, ctrl.Result) {
+	if ig == nil { // initialize ImageGenerator if not exsits
+		var err error
+		ig, err = newImageGenerator(c)
+		if err != nil || ig == nil {
+			igLogger.Info("error initializing ImageGenerator instance", "err", err)
+			return false, ctrl.Result{Requeue: true}
+		}
+	}
+	return ig.imageJobRunner("delete")
+}
+
+func newImageGenerator(client client.Client) (*ImageGenerator, error) {
+	var ig ImageGenerator
+	var procFIPS = "/proc/sys/crypto/fips_enabled"
+
+	ig.client = client
+	ig.provider = ""     // set by setupCloudProvider
+	ig.CMimageIDKey = "" // set by setupCloudProvider
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get in-cluster config: %v", err)
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get k8s clientset: %v", err)
+	}
+	ig.clientset = clientset
+
+	content, err := os.ReadFile(procFIPS)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read FIPS file: %v", err)
+	}
+
+	fips, err := strconv.Atoi(strings.Trim(string(content), "\n\t "))
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert FIPS file content to int: %v", err)
+	}
+	ig.fips = fips == 1
+
+	igLogger.Info("ImageGenerator instance has been initialized successfully", "fips", ig.fips)
+	return &ig, nil
+}
+
+func (r *ImageGenerator) getCloudProviderFromInfra() string {
+	// TODO: first check if it's indeed openshift
+	infrastructure := &configv1.Infrastructure{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infrastructure)
+	if err != nil {
+		igLogger.Info("getCloudProviderInfra: Error getting Infrastructure object", "err", err)
+		return ""
+	}
+
+	if infrastructure.Status.PlatformStatus == nil {
+		igLogger.Info("getCloudProviderInfra: Infrastructure.status.platformStatus is empty")
+		return ""
+	}
+
+	igLogger.Info("Got cloud provider from infrastructure object")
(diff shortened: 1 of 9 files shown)


snir911:
cc also @jensfr  @pmores

snir911:
@jensfr @bpradipt , fips changes add to this pr, based on https://github.com/confidential-containers/cloud-api-adaptor/pull/1474

config/peerpods/aws-VM-image-create-job.yaml
```diff
@@ -0,0 +1,103 @@
...
+      volumes:
+      - name: shared-data
+        emptyDir: {}
+        volumes:
+      - name: image-id
+        emptyDir: {}
+
+      securityContext:
+        runAsUser: 0 # needed for container mode dnf access
+
```
gkurz, line 23: It seems that only the "create" containers actually needs that. The other ones ("payload" and "result") can run unprivileged, right ? If so, please move the security context to the "create" container.

config/peerpods/aws-VM-image-create-job.yaml
```diff
@@ -0,0 +1,103 @@
...
+            name: peer-pods-secret
+        - configMapRef:
+            name: peer-pods-cm
+            optional: true
+        command:
+        - /bin/sh
+        - -c
+        - |
+          set -e
+          yum install -y yum-utils make git unzip
```
gkurz, line 61: `yum` ? Why not `dnf` as mentioned in the comment above ?

config/peerpods/aws-VM-image-delete-job.yaml
```diff
@@ -0,0 +1,50 @@
...
+          - name: PODVM_DISTRO
+            value: rhel
+          - name: IMAGE_NAME
+            value: "peer-pod-ami"
+        command:
+        - /bin/sh
+        - -c
+        - |
+          set -e
+          yum install -y unzip
```
gkurz, line 38: `dnf` ?

controllers/openshift_controller.go
```diff
@@ -1009,6 +1025,18 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 
 	// peer pod enablement
 	if r.kataConfig.Spec.EnablePeerPods {
+		if r.imageGenerator == nil {
+			if r.imageGenerator = NewImageGenerator(r.Client); r.imageGenerator == nil {
+				r.Log.Info("Failed to create image generator")
+				return ctrl.Result{Requeue: true}, nil
+			}
+		}
```
gkurz, line 1033: What about moving this to an helper function to avoid code duplication ?

main.go
```diff
@@ -135,6 +138,7 @@ func main() {
 			Client: mgr.GetClient(),
 			Log:    ctrl.Log.WithName("controllers").WithName("KataConfig"),
 			Scheme: mgr.GetScheme(),
+			// imageGenerator: &controllers.NewImageGenerator(mgr.GetClient()), TODO: legit? add later
```
gkurz, line 141: If this is only useful for peer pods, I'd say no.

gkurz:
Same remarks on `dnf` and the `securityContext` apply to the other yaml files as well.

controllers/image_generator.go
```diff
@@ -0,0 +1,359 @@
...
+	clientset kubernetes.Interface // k8s clientset
+	Log       logr.Logger
+
+	provider     string
+	CMimageIDKey string
+}
+
+func NewImageGenerator(client client.Client) *ImageGenerator {
+	config, err := rest.InClusterConfig()
+	if err != nil {
```
gkurz, line 97: Any reason for not propagating the error ?

controllers/image_generator.go
```diff
@@ -0,0 +1,359 @@
...
+		provider:     "", // set by getCloudProvider
+		CMimageIDKey: "", // set by getCloudProvider
+		Log:          ctrl.Log.WithName("image-generator"),
+	}
+}
+
+func (r *ImageGenerator) getCloudProviderFromInfra() string {
+	// TODO: first check if it's indeed openshift
+	infrastructure := &configv1.Infrastructure{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infrastructure)
```
gkurz, line 129: Where does this "cluster" namespace comes from ?

controllers/image_generator.go
```diff
@@ -78,26 +81,46 @@ type ImageGenerator struct {
...
+	var procFIPS = "/proc/sys/crypto/fips_enabled"
+
+	ig.client = client
+	ig.Log = ctrl.Log.WithName("image-generator")
+	ig.provider = ""     // set by setupCloudProvider
+	ig.CMimageIDKey = "" // set by setupCloudProvider
+
 	config, err := rest.InClusterConfig()
 	if err != nil {
+		ig.Log.Info("Failed to get in-cluster config", "err", err)
```
gkurz, line 98: It seems that logging should have been introduced in patch 2.

main.go
snir911: It is, i suggested as it is much cleaner to initialize it there, Delete/Create will be still called only if peerpods is enable

controllers/image_generator.go
snir911: it's the namespace of the cluster's "infrastructure" CR, it gets the cluster's region from there

snir911:
Hi @gkurz thanks for reviewing, some of the commits i addressed in HEAD^, the rest i believe are addressed in the the HEAD commit which hides the ImageGenerator initialization behind Create/Delete
(later I'll try to squash it all)

config/peerpods/aws-VM-image-create-job.yaml
```diff
@@ -0,0 +1,105 @@
...
+          dnf install -y make git unzip
+          curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+          unzip awscliv2.zip
+          ./aws/install
+          curl https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip -o packer_1.9.4_linux_amd64.zip
+          echo "6cd5269c4245aa8c99e551d1b862460d63fe711c58bec618fade25f8492e80d9  packer_1.9.4_linux_amd64.zip" | sha256sum -c
+          unzip packer_1.9.4_linux_amd64.zip -d /usr/bin/
+          PATH="/usr/bin:${PATH}"
+          packer plugins install github.com/hashicorp/amazon
+          git clone https://github.com/confidential-containers/cloud-api-adaptor.git -b v0.8.0
```
bpradipt: This should be from 0.7.0+cherry-picks right ? 0.8.0 is not yet there and also will not match with the CAA and the binaries

config/peerpods/aws-VM-image-create-job.yaml
```diff
@@ -0,0 +1,105 @@
...
+        - -c
+        - |
+          set -e
+          dnf install -y make git unzip
+          curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+          unzip awscliv2.zip
+          ./aws/install
+          curl https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip -o packer_1.9.4_linux_amd64.zip
+          echo "6cd5269c4245aa8c99e551d1b862460d63fe711c58bec618fade25f8492e80d9  packer_1.9.4_linux_amd64.zip" | sha256sum -c
+          unzip packer_1.9.4_linux_amd64.zip -d /usr/bin/
```
bpradipt, line 67: Good to pin this to 1.9.4 which is still using MPL

config/peerpods/azure-VM-image-create-job.yaml
```diff
@@ -0,0 +1,124 @@
...
+        - |
+          set -e
+          rpm --import https://packages.microsoft.com/keys/microsoft.asc
+          dnf install -y https://packages.microsoft.com/config/rhel/9.0/packages-microsoft-prod.rpm
+          dnf install -y make git azure-cli unzip
+          curl https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip -o packer_1.9.4_linux_amd64.zip
+          echo "6cd5269c4245aa8c99e551d1b862460d63fe711c58bec618fade25f8492e80d9  packer_1.9.4_linux_amd64.zip" | sha256sum -c
+          unzip packer_1.9.4_linux_amd64.zip -d /usr/bin/
+          PATH="/usr/bin:${PATH}"
+          git clone https://github.com/confidential-containers/cloud-api-adaptor.git -b v0.8.0
```
bpradipt: same comment as the AWS job on the code to be used for creating the image

config/peerpods/aws-VM-image-create-job.yaml
snir911: ah right, how should i tag it

config/peerpods/aws-VM-image-create-job.yaml
gkurz: Not sure about the meaning of your :+1: since `securityContext`  is still set at the POD level

config/peerpods/azure-VM-image-create-job.yaml
```diff
@@ -0,0 +1,127 @@
...
+    spec:
+      volumes:
+      - name: shared-data
+        emptyDir: {}
+      - name: image-id
+        emptyDir: {}
+
+
+      securityContext:
+        runAsUser: 0 # needed for container mode dnf access
```
gkurz, line 22: Same remark as with the AWS job.

controllers/image_generator.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+var ig *ImageGenerator = nil
+var igLogger logr.Logger = ctrl.Log.WithName("image-generator")
+
+// ImageCreate creates a podvm image for a cloud provider if not present
+func ImageCreate(c client.Client) (bool, ctrl.Result) {
+	if ig == nil { // initialize ImageGenerator if not exsits
+		var err error
+		ig, err = newImageGenerator(c)
```
gkurz, line 93: Hmm... my previous question on `newImageGenerator()` not returning an error also applies here, and more generally to any error path returning to the operator... why not propagating ?

controllers/image_generator.go
```diff
@@ -0,0 +1,407 @@
...
+	}, peerPodsCM)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: check in peer-pods-secret as well?
+
+	return peerPodsCM, nil
+}
+
+func (r *ImageGenerator) imageJobRunner(op string) (bool, ctrl.Result) {
```
gkurz, line 271: What are the exact semantics of the `bool` return value ? It seems to be basically `false` when some error occured and `true` otherwise... why not returning the error instead ?

config/peerpods/aws-VM-image-create-job.yaml
snir911: oops, correct, i got you wrong, I'll fix

controllers/image_generator.go
snir911: The variable name used in the controller.go implies it but maybe i should have add a comment..
The bool says whether the image creation job is completed/needed and can continue **or** it still running and should return the ctrl.Result to break the loop so it will be called again in future iteration

controllers/image_generator.go
snir911: Since the logic has to control the reconciliation loop actions (result and continuation) i let ImageCreate to return it, i avoided adding also the error because of two reasons:
a. it will require to handle 3 return values in multiple functions which i find a bit tedious
b. AFAIU from @pmores the common practice for this operator is not to return error along with ctrl.Result so is there a point in propagating it?

gkurz: We're not doing much with errors except logging them actually, which your PR already takes care for. Agreed that we don't return errors from Reconcile(). Let's go ahead. We still can revisit later if needed.

controllers/openshift_controller.go
```diff
@@ -71,6 +71,7 @@ const (
 	DEFAULT_PEER_PODS                   = "10"
 	peerpodConfigCrdName                = "peerpodconfig-openshift"
 	peerpodsMachineConfigPathLocation   = "/config/peerpods"
+	peerpodsImageJobsPathLocation       = "/config/peerpods"
```
gkurz, line 74: This will end up mixing machine config and job YAML files under the same directory, whose path suggests is was originally introduced for machine configs (see 6577693479bc0f181442923e93e5e2c2df0fa271). Maybe use a different path for the jobs ?

Why putting the image jobs at the same location as the machine config ?

controllers/utils.go
```diff
@@ -37,6 +38,24 @@ func IsOpenShift() (bool, error) {
...
+
+func readJobYAML(jobFileName string) ([]byte, error) {
+	jobFilePath := filepath.Join(peerpodsImageJobsPathLocation, jobFileName)
+	yamlData, err := os.ReadFile(jobFilePath)
+	if err != nil {
+		return nil, err
+	}
+	return yamlData, nil
+}
+
```
gkurz, line 58: It looks like a `readYAML(path string, fileName string)` generic helper can be factored out.

snir911: indeed, however, it's likely to be dropped in the future so i think we should leave it as is for now

gkurz (approved):
/lgtm . Thanks @snir911 !

bpradipt (approved):
/lgtm
Thanks @snir911

PR #349, opened 2023-10-17
don't let KataConfig deletion start while installation is still in progress

pmores:
When KataConfig deletion processing is allowed to start while a previous installation or update is still running both processes will interfere with each other, resulting in a variety of possible cluster states, most of them undesirable (most often, uninstallation blocks indefinitely).

This PR alleviates the situation by not allowing uninstallation to start while installation or an update is still underway.  Once the current operation finishes uninstallation should start immediately.

Fixes [KATA-1851](https://issues.redhat.com//browse/KATA-1851)

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -134,8 +134,10 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 		}
 
 		// Check if the KataConfig instance is marked to be deleted, which is
-		// indicated by the deletion timestamp being set.
-		if r.kataConfig.GetDeletionTimestamp() != nil {
+		// indicated by the deletion timestamp being set.  However, don't let
+		// uninstallation commence if another operation (installation, update)
+		// is underway.
+		if r.kataConfig.GetDeletionTimestamp() != nil && !r.isInstalling() && !r.isUpdating() {
 			res, err := r.processKataConfigDeleteRequest()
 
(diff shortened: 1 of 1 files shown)


tbuskey:
Premerge /lgtm

controllers/openshift_controller.go
```diff
@@ -1946,6 +1948,22 @@ func (r *KataConfigOpenShiftReconciler) resetInProgressCondition() {
...
+}
+
+func (r *KataConfigOpenShiftReconciler) isUpdating() bool {
+	cond := r.findInProgressCondition()
+	if cond == nil {
+		return false;
+	}
+	return cond.Status == corev1.ConditionTrue && cond.Reason == "Updating";
+}
+
```
gkurz, line 1966: I'm wondering if go is smart enough to detect that the code for these is identical.

gkurz (approved):
Code does as described. Thanks @pmores !

controllers/openshift_controller.go
pmores: I'm not quite sure if I understand you correctly but `isInstalling()` checks for `cond.Reason == "Installing"` whereas `isUpdating()` checks for `cond.Reason == "Updating"`, doesn't it?

gkurz: Yes, same code with _just_ a different literal string. Wondering if go can detect that and factor out the code internally.

pmores: I see. I've no idea but honestly, I doubt it. :-)

PR #351, opened 2023-10-25
coco-dev-preview: add CoCo defaulter job

snir911:


bpradipt (approved):
/lgtm
Thanks @snir911

bpradipt:
@jensfr @littlejawa any comments on this one before we merge?

littlejawa:
I could only review the file, not test it - I don't have the environment.
But I don't see any issue.

hack/coco-cm-defaulter.yaml
```diff
@@ -0,0 +1,55 @@
...
+              VXLAN_PORT=9000
+          EOF
+          else
+            echo "Uknown provider: \"${CLOUD_PROVIDER}\""
+          fi
+
+           cat /tmp/cm.env
+           oc create cm peer-pods-cm --from-env-file=/tmp/cm.env
+
+        image: 'registry.redhat.io/openshift4/ose-cli'
```
bpradipt, line 54: Is there an alternate public image available? I think pulling from registry.redhat.io needs auth, which is fine. But if there is a public equivalent it's better.

snir911: quay.io/openshift/origin-cli ?
It's uses oc 4.2 though

bpradipt: Or you'll need to download the cli from a public url like - https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.13.19/
Would it help to use a regular fedora image and use the oc from the above link ? or it's simply over engineering ;)

bpradipt: I think for generic usage we anyway need to adapt few things. And since this is specific for OCP, we can go with the the existing image as-is

PR #353, opened 2023-11-07
peer-pods: validate CM and Secret are set

snir911:


gkurz (approved):
LGTM. Thanks @snir911 !

cpmeadors (approved):

controllers/image_generator.go
```diff
@@ -142,6 +142,11 @@ func newImageGenerator(client client.Client) (*ImageGenerator, error) {
 	}
 	ig.fips = fips == 1
 
+	err = ig.setupCloudProvider()
+	if err != nil {
+		return nil, fmt.Errorf("failed to setup cloud provider: %v", err)
+	}
+
```
gkurz, line 156: Please provide a change log in this commit to explain why this needs to be done earlier.

controllers/image_generator.go
```diff
@@ -164,17 +169,8 @@ func (r *ImageGenerator) getCloudProviderFromInfra() string {
...
-		return nil
-	}
-
-	if cm.Data["CLOUD_PROVIDER"] != "" {
-		provider = cm.Data["CLOUD_PROVIDER"]
-	} else {
-		provider = r.getCloudProviderFromInfra()
-	}
+func (r *ImageGenerator) setupCloudProvider() error {
+	provider := r.getCloudProviderFromInfra()
```
gkurz, line 180: It isn't obvious why all this code is being dropped from `setupCloudProvider()`.

controllers/image_generator.go
```diff
@@ -96,6 +97,12 @@ func ImageCreate(c client.Client) (bool, ctrl.Result) {
 			return false, ctrl.Result{Requeue: true}
 		}
 	}
+
+	if err := ig.validatePeerPodsConfigs(); err != nil {
+		igLogger.Info("error validating peer-pods configs", "err", err)
+		return false, ctrl.Result{Requeue: true, RequeueAfter: 30 * time.Second}
+	}
+
```
gkurz, line 105: IIUC it is assumed that if the peer pods config is valid at image creation time, it cannot be invalidated in any way and doesn't need to be validated again at image delete time. If yes, maybe make it clear in a comment or in the change log. Otherwise maybe move the check to `imageJobRunner()`. My motivation is that `imageJobRunner()` re-gets the CM from the server and we have no formal guarantee it wasn't changed by some external action.

gkurz (changes requested):

controllers/image_generator.go
snir911: This is indeed assumed.
I added the validation also to deletion (originally i didn't add it as deletion requires less parameters in some cases but i guess it won't hurt to over-validate if it isn't expected to be changed and shouldn't harm functioning.. )

gkurz: Yes it isn't expected to change but since it is still technically possible, it seems better to make the code proof to that. Thanks !

controllers/image_generator.go
```diff
@@ -396,3 +418,35 @@ func (r *ImageGenerator) getImageIDFromJobLogs(job *batchv1.Job) (string, error)
...
+	case "azure":
+		azureEssentials := []string{"AZURE_TENANT_ID", "AZURE_CLIENT_ID", "AZURE_CLIENT_SECRET", "AZURE_RESOURCE_GROUP", "AZURE_SUBSCRIPTION_ID", "AZURE_REGION"}
+		essentialKeys = append(essentialKeys, azureEssentials...)
+	default:
+		return fmt.Errorf("validatePeerPodsConfigs: Unsupported cloud provider: %s", r.provider)
+	}
+
+	// check if all essential keys are present in either ConfigMap or Secret
+	for _, k := range essentialKeys {
+		if len(peerPodsCM.Data[k]) < 1 && len(peerPodsSecret.Data[k]) < 1 {
```
gkurz, line 447: This assumes that `k` cannot overflow the `Data[]` arrays and the go runtime will panic if it does. Maybe also check that both arrays have the length of essentialKeys or greater ?

controllers/image_generator.go
snir911: I'm not sure it's possible to change it, it shouldn't harm i think, however, I'd  be worried about messing the whole cluster

controllers/image_generator.go
gkurz: My actual concern is that the CM is retrieved from the server in `imageJobRunner()` and it looks like this is the place where we should ensure it is valid.

controllers/image_generator.go
```diff
@@ -259,11 +271,21 @@ func (r *ImageGenerator) getPeerPodsCM() (*corev1.ConfigMap, error) {
...
+	err := r.client.Get(context.TODO(), types.NamespacedName{
+		Name:      peerPodsSecretName,
+		Namespace: "openshift-sandboxed-containers-operator",
+	}, peerPodsSecret)
+	if err != nil {
+		return nil, err
+	}
+	return peerPodsSecret, nil
+}
+
```
gkurz, line 288: This is only called by the validate function. How are these secrets actually used ?

controllers/image_generator.go
gkurz: Well... I tend to assume that anything can be changed in k8s by default and it is up to the operator to cope with that :wink: . Now if changing the infra object in our back has a greater impact than just breaking the operator, I guess it is acceptable.

controllers/image_generator.go
snir911: This secret is being mounted along with the CAA pods, the image creation Jobs and fetched by the peerpod-ctrl, here this function was created only for consistency with the getPeerPodCM function.

controllers/image_generator.go
snir911: Data is a Map so it should be ok

controllers/image_generator.go
snir911: Technically the validation is for the values used for the job itself (which mounts the secret&cm), and practically it's also validates the values used by the CAA which starts much later (this is why at first i did it in the main reconcile loop( and switched to avoid fetchCloudProvider code duplication))

gkurz (approved):
Thanks for the work and patience @snir911 !

PR #354, opened 2023-11-08
coco-dev-preview: hack: fix azure login

snir911:


hack/azure-CVM-image-create-job.yaml
```diff
@@ -89,7 +89,7 @@ spec:
           [[ ! "${AZURE_SUBSCRIPTION_ID}" ]] && echo "AZURE_SUBSCRIPTION_ID is missing" && exit 1
           [[ ! "${AZURE_CLIENT_SECRET}" ]] && echo "AZURE_CLIENT_SECRET is missing" && exit 1
           [[ ! "${AZURE_CLIENT_ID}" ]] && echo "AZURE_CLIENT_ID is missing" && exit 1
-          az login --service-principal -u ${AZURE_CLIENT_ID} -p ${AZURE_CLIENT_SECRET} --tenant ${AZURE_TENANT_ID}
+          az login --service-principal -u ${AZURE_CLIENT_ID} -p=${AZURE_CLIENT_SECRET} --tenant ${AZURE_TENANT_ID}
```
gkurz, line 92: This works but it isn't _pretty_. Legacy command line handling for single character arguments is to be followed by a space and to consume the next parameter if it needs to. For some reason, az is not sticking to that.
Equal sign usually comes with richer syntax arguments like `--password=${AZURE_CLIENT_SECRET}` so my suggestion would be to fully switch to the richer syntax. This would guarantee the command line to be proof against such issues in the future.

```suggestion
          az login --service-principal --user=${AZURE_CLIENT_ID} --password=${AZURE_CLIENT_SECRET} --tenant=${AZURE_TENANT_ID}
```

gkurz (approved):
LGTM. Thanks @snir911 !

PR #355, opened 2023-11-10
Bump OSC to 1.5.0

gkurz:
As a preparatory step for the upcoming release, this bumps the version to 1.5.0 in Makefile, the CSV and images.

Note that the updated images aren't available at this stage. They will be added later. Until then the image links aren't valid.


cpmeadors (approved):
looks good

PR #358, opened 2023-11-14
Merge main into devel after v1.5.0 release

gkurz:

Now that v1.5.0 was tagged on main, let's merge it back to devel.

**- What I did**

Create this PR.

**- How to verify it**

HEAD of this PR should have the same SHA1 as v1.5.0.


pmores (approved):
lgtm thanks @gkurz !

PR #361, opened 2023-11-19
makefile: match operator, bundle and catalog versioning

snir911:
It seems to me the right approach, cc @gkurz @pmores @bpradipt for confirmation this is legit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 endif
 
 # Image URL to use all building/pushing image targets
-IMG ?= $(IMAGE_TAG_BASE):$(VERSION)
+IMG ?= $(IMAGE_TAG_BASE):v$(VERSION)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.24
 
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1962,17 +1962,17 @@ func (r *KataConfigOpenShiftReconciler) resetInProgressCondition() {
 func (r *KataConfigOpenShiftReconciler) isInstalling() bool {
 	cond := r.findInProgressCondition()
 	if cond == nil {
-		return false;
+		return false
 	}
-	return cond.Status == corev1.ConditionTrue && cond.Reason == "Installing";
+	return cond.Status == corev1.ConditionTrue && cond.Reason == "Installing"
 }
 
 func (r *KataConfigOpenShiftReconciler) isUpdating() bool {
 	cond := r.findInProgressCondition()
 	if cond == nil {
-		return false;
+		return false
 	}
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -49,7 +49,7 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 endif
 
 # Image URL to use all building/pushing image targets
-IMG ?= $(IMAGE_TAG_BASE):$(VERSION)
```
bpradipt, line 52: Do you know if the skeleton generated by the latest operator-sdk makes this change ?
If not, will it break bundle and catalog generation if there is an implicit assumption to use version without `v` prefix ?

controllers/openshift_controller.go
```diff
@@ -1962,17 +1962,17 @@ func (r *KataConfigOpenShiftReconciler) resetInProgressCondition() {
...
 }
 
 func (r *KataConfigOpenShiftReconciler) isUpdating() bool {
 	cond := r.findInProgressCondition()
 	if cond == nil {
-		return false;
+		return false
 	}
-	return cond.Status == corev1.ConditionTrue && cond.Reason == "Updating";
+	return cond.Status == corev1.ConditionTrue && cond.Reason == "Updating"
```
bpradipt, line 1975: /lgtm for this change.

Makefile
snir911: it seems the generated skeleton suggests to push this image to "latest", hence, i assumed this was changed locally and it's fine to change it again

gkurz: Operator-sdk introduced the `v` in the bundle and catalog versioning in v1.6.0 with https://github.com/operator-framework/operator-sdk/pull/4406. Specifically this intermediate change in the PR https://github.com/operator-framework/operator-sdk/compare/9ffbb583609404fa3ab27d59c1e1c624255ad8aa..f8a08b782a90e7a1af8493bdaa9693bc61692dd9#diff-97e56898a92a31cc551cc061b2eb64809ad7fc45c2bacd13c660ae207632538fL139

I've checked the review comments that preceded this change but I don't find any suggestion about the extra `v`... Maybe worth asking operator-sdk folks to sort this out.

gkurz (commented):
> This will also allow convenient using of IMG_TAG_BUILD for custom builds

This variable doesn't exist in the tree. Assuming you are actually referring to `IMAGE_TAG_BASE`, I'm not sure to see how this change makes things more convenient. Could you elaborate @snir911 ?

Makefile
gkurz: I could chat with an operator-sdk developer on [slack](https://kubernetes.slack.com/archives/C017UU45SHL/p1700562561803609). The outcome is that the `v` was introduced for _convenience_, for whatever that means, and we can do pretty much what we want.

gkurz (approved):
/lgtm
Thanks @snir911 !

Makefile
gkurz: @bpradipt can we merge this change ?

bpradipt: sure..

bpradipt (approved):
/lgtm

PR #362, opened 2023-11-22
image-generator: skip image creation for unsupported providers

snir911:
this allows installtion completion without being blocked waiting
for image creation when image creation isn't implemented for
cloud-provider

### DIFF
--- a/controllers/image_generator.go
+++ b/controllers/image_generator.go
@@ -67,12 +67,13 @@ Every image deletion job should preform the following:
 */
 
 const (
-	peerpodsCMName          = "peer-pods-cm"
-	peerPodsSecretName      = "peer-pods-secret"
-	peerpodsCMAWSImageKey   = "PODVM_AMI_ID"
-	peerpodsCMAzureImageKey = "AZURE_IMAGE_ID"
-	fipsCMKey               = "BOOT_FIPS"
-	defaultVMType           = "VM"
+	unsupportedCloudProvider = "unsupported"
+	peerpodsCMName           = "peer-pods-cm"
+	peerPodsSecretName       = "peer-pods-secret"
+	peerpodsCMAWSImageKey    = "PODVM_AMI_ID"
+	peerpodsCMAzureImageKey  = "AZURE_IMAGE_ID"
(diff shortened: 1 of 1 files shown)


controllers/image_generator.go
```diff
@@ -98,6 +99,11 @@ func ImageCreate(c client.Client) (bool, ctrl.Result) {
 		}
 	}
 
+	if ig.provider == unsupportedCloudProvider {
+		igLogger.Info("unsupported cloud provider, skipping image creation")
+		return true, ctrl.Result{Requeue: true}
```
gkurz, line 104: It seems that we shouldn't requeue in this case.

controllers/image_generator.go
```diff
@@ -117,6 +123,11 @@ func ImageDelete(c client.Client) (bool, ctrl.Result) {
 		}
 	}
 
+	if ig.provider == unsupportedCloudProvider {
+		igLogger.Info("unsupported cloud provider, skipping image deletion")
+		return true, ctrl.Result{Requeue: true}
```
gkurz, line 128: Ditto.

gkurz (changes requested):
Thanks @snir911 ! The change seems to be good but I still have some comments :

1. please add some context in the commit log : motivation, link to jira...
2. this moves `setupCloudProvider()` into `newImageGenerator()` : not sure to understand why and this makes the change harder to review. Could the refactoring happen in a preliminary commit ?

controllers/image_generator.go
snir911: it won't, returning true means it will ignore the returning value, I can't return nil here, how about `ctrl.Result{}`?

gkurz:
Now waiting for @tbuskey to pre-merge test this PR for regressions with supported providers.

snir911:
@tbuskey , no, it fetches the cloud-provider name from the the infrastructure object of the cluster

gkurz:
Fixes https://issues.redhat.com/browse/KATA-2572

tbuskey:
Azure works
AWS works
/lgtm

PR #366, opened 2023-12-07
Fix upgrade from OSC 1.4.1

gkurz:

This fixes [KATA-2593 (Kata Operator upgrade failed 1.4.1 to 1.5.0)](https://issues.redhat.com/browse/KATA-2593) .

**- What I did**

Renamed the offending field in KataConfigStatus and improved idempotency of the controller.

**- How to verify it**

1. Install OSC 1.4.1 with automatic upgrade enabled [use catalog image at *]
2. Create a KataConfig
3. Setup a catalog source with this PR [use catalog image at **]
4. Patch the subscription of the operator to select the new source
```
oc patch subscription -n openshift-sandboxed-containers-operator sandboxed-containers-operator -p '{"spec":{"source":"my-operator-catalog", "startingCSV":"sandboxed-containers-operator.v1.5.1", "channel":"candidate"}}' --type merge
```
5. Upgrade should start and complete successfully

The following scenarios should be checked :

- [ ] User has 1.4.1 installed (described above)
- [ ] User has 1.4.1/1.5.0 in pending state ([KATA-2593](https://issues.redhat.com/browse/KATA-2593)).
       I believe this requires to first uninstall OSC completely, aka `oc delete ns openshift-sandboxed-containers-operator`
- [ ] User has 1.5.0 installed

[*] quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator-catalog:1.4.1-10
[**] quay.io/rhgkurz/openshift-sandboxed-containers-operator-catalog:v1.5.1

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -561,7 +561,7 @@ func (r *KataConfigOpenShiftReconciler) listKataPods() error {
 	}
 	for _, pod := range podList.Items {
 		if pod.Spec.RuntimeClassName != nil {
-			if contains(r.kataConfig.Status.RuntimeClass, *pod.Spec.RuntimeClassName) {
+			if contains(r.kataConfig.Status.RuntimeClasses, *pod.Spec.RuntimeClassName) {
 				return fmt.Errorf("Existing pods using \"%v\" RuntimeClass found. Please delete the pods manually for KataConfig deletion to proceed", *pod.Spec.RuntimeClassName)
 			}
 		}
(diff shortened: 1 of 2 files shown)


controllers/openshift_controller.go
```diff
@@ -1667,10 +1667,6 @@ const (
 // will be returned.
 func (r *KataConfigOpenShiftReconciler) updateStatus() error {
 
-	if r.getInProgressConditionValue() != corev1.ConditionTrue {
-		return nil
-	}
-
```
pmores, line 1678: Regarding the commit message, why does just editing KataConfig externally trigger this problem?  I'd say predating PR #329 should be the only scenario.

tbuskey:
premerge testing: see [KATA-2593](https://issues.redhat.com/browse/KATA-2593)

controllers/openshift_controller.go
tbuskey: Is  un-rebuilding nodes a blocker?

gkurz: > Is un-rebuilding nodes a blocker?

Not really as it doesn't prevent the operator to be functional.

pmores: Oh I suspect there would be quite a lot of things that user could do to sabotage the controller which wouldn't recover. ;-) But that's a fact indepedent of the idea of this PR - in fact, this has always been true and continues to be true even after this PR I believe.

My idea was not to mix independent facts in the message and not put them in the same context as if they were related since that could confuse a future reader. I'm not insisting though.

pmores: Ah this is merged already... never mind.

PR #369, opened 2023-12-13
image-generator: fix image creation edge cases

snir911:
* at creation: name images with unique naming to aviod name reuse failures
* at deletion: ignore if image is already deleted

Fixes: #[KATA-2618](https://issues.redhat.com//browse/KATA-2618)

to test:
```
quay.io/snir/osc-operator:v9.9.2618
quay.io/snir/osc-operator-catalog:v9.9.2618
quay.io/snir/osc-operator-bundle:v9.9.2618
```
in OCP make sure to set also:
`oc set env deployment.apps/controller-manager SANDBOXED_CONTAINERS_EXTENSION=sandboxed-containers -n openshift-sandboxed-containers-operator`

### DIFF
--- a/config/peerpods/podvm/aws-VM-image-create-job.yaml
+++ b/config/peerpods/podvm/aws-VM-image-create-job.yaml
@@ -44,8 +44,6 @@ spec:
             value: aws
           - name: PODVM_DISTRO
             value: rhel
-          - name: IMAGE_NAME
-            value: "peer-pod-ami"
 #          - name: INSTANCE_TYPE
 #            value: "t3.small" # default is t3.small, uncomment and modify if not available in your region
         envFrom:
@@ -59,6 +57,7 @@ spec:
         - -c
         - |
           set -e
+          [[ ! "${IMAGE_NAME}" ]] && UUID=$(uuidgen) && export IMAGE_NAME="peer-pod-ami-${UUID::6}" && echo "IMAGE_NAME:${IMAGE_NAME}"
           dnf install -y make git unzip
           curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
           unzip awscliv2.zip
--- a/config/peerpods/podvm/aws-VM-image-create-job.yaml
+++ b/config/peerpods/podvm/aws-VM-image-create-job.yaml
@@ -46,6 +46,8 @@ spec:
             value: rhel
 #          - name: INSTANCE_TYPE
 #            value: "t3.small" # default is t3.small, uncomment and modify if not available in your region
+#          - name: IMAGE_NAME
+#            value: "aws-podvm-image-name" # set custom image name for custom image if you wish to avoid its deletion
         envFrom:
         - secretRef:
             name: peer-pods-secret
--- a/config/peerpods/podvm/aws-VM-image-delete-job.yaml
+++ b/config/peerpods/podvm/aws-VM-image-delete-job.yaml
@@ -26,13 +26,13 @@ spec:
         env:
           - name: PODVM_DISTRO
             value: rhel
-          - name: IMAGE_NAME
-            value: "peer-pod-ami"
         command:
         - /bin/sh
         - -c
         - |
           set -e
+          [[ ! "${PODVM_AMI_ID}" ]] && echo "PODVM_AMI_ID is missing, it's unknown which image to delete" && exit 1
+          [[ "${IMAGE_NAME}" ]] && echo "IMAGE_NAME:${IMAGE_NAME} is set, it implies image was not automatically created, delete it manually" && exit 0
           dnf install -y unzip
           curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
           unzip awscliv2.zip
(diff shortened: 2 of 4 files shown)


abhbaner:
**Successfully Pre-Merge tested it**

_Edge case 1 :_  If there is an existing podvm Image with the name 'peer-pod-XXXX' (on the cloud provider) the kataconfig install does not begin as the IMAGE ID creation job fails (due to the same name of the existing podvm image) and that causes the kataconfig install to be stuck in and there is no progress

_Edge case 2:_ If by mistake the user deletes the podvm IMAGE  (from the console of the cloud provider ) the kataconfig deletion will not progress as the process waits for the image deletion job to complete which it wont in this case as the user has deleted the podvm image already and the job fails , unable to find the podvm image id. 

**Pre-Merge test Steps followed** 

- Applied the catalog source (with the catalog  [quay.io/snir/osc-operator-catalog:v9.9.2618](http://quay.io/snir/osc-operator-catalog:v9.9.2618) in it)
- Installed the OSC Operator
- Post install of the operator I issued the cmd oc set env deployment.apps/controller-manager SANDBOXED_CONTAINERS_EXTENSION=sandboxed-containers -n openshift-sandboxed-containers-operator
- Created the the peer pod cm and secret
- Created the kataconfig with enablepeerpods set to 'true'
 
 **Edge case 1 tested** 
- The AWS console had an existing AMI with name (and ID) as  peer-pod-ami (ami-0cc7239b859af5641). Despite that the image id creation job was able to create a new AMI ID and named it peer-pod-ami-3c37 (ami-0cd66ff9c8c21755c)
- The kataconfig creation went through fine as well

**Edge case 2 tested**
- Deregistered all AMIs from the AWS console with peer-pod-ami in their name
- Successfully deleted kataconfig

cpmeadors:
@abhbaner can you approve since you did the pre-merge testing?

config/peerpods/podvm/aws-VM-image-create-job.yaml
```diff
@@ -59,6 +57,7 @@ spec:
         - -c
         - |
           set -e
+          [[ ! "${IMAGE_NAME}" ]] && UUID=$(uuidgen) && export IMAGE_NAME="peer-pod-ami-${UUID::6}" && echo "IMAGE_NAME:${IMAGE_NAME}"
```
cpmeadors, line 60: If we are not requiring IMAGE_NAME to be explicitly unset we probably need to add that to the docs.

config/peerpods/podvm/aws-VM-image-delete-job.yaml
```diff
@@ -26,13 +26,13 @@ spec:
...
             value: rhel
-          - name: IMAGE_NAME
-            value: "peer-pod-ami"
         command:
         - /bin/sh
         - -c
         - |
           set -e
+          [[ ! "${PODVM_AMI_ID}" ]] && echo "PODVM_AMI_ID is missing, it's unknown which image to delete" && exit 1
+          [[ "${IMAGE_NAME}" ]] && echo "IMAGE_NAME:${IMAGE_NAME} is set, it implies image was not automatically created, delete it manually" && exit 0
```
cpmeadors, line 35: I think we need to make sure it is clear in docs as to what happens with manually created images and that user is responsible for deleting them as the message states.

config/peerpods/podvm/aws-VM-image-delete-job.yaml
```diff
@@ -41,8 +41,8 @@ spec:
...
           [[ ! "${AWS_REGION}" ]] && echo "AWS_REGION is missing" && exit 1
-          export IMAGE_NAME=${IMAGE_NAME:-peer-pod-ami}
-          PODVM_AMI_ID=$(aws ec2 describe-images --query "Images[*].[ImageId]" --filters "Name=name,Values=${IMAGE_NAME}" --region ${AWS_REGION} --output text) && \
           echo "Deleting AMI: ${PODVM_AMI_ID}"
-          aws ec2 deregister-image --image-id ${PODVM_AMI_ID} --region ${AWS_REGION}
-          echo "Deleting AMI: ${PODVM_AMI_ID} - DONE"
+          RES=$(aws ec2 deregister-image --image-id "${PODVM_AMI_ID}" --region "${AWS_REGION}" 2>&1) || ERR=true
+          echo ${RES}
+          [[ ${ERR} ]] && [[ "$RES" =~ InvalidAMIID\.(Unavailable|NotFound) ]] # if deregister returned error and image is already deleted, continue
+          echo "Deleted AMI: ${PODVM_AMI_ID} - DONE"
```
cpmeadors, line 48: Are users supposed to see the error message in the output of the job?  Is that captured in a log anywhere?  Also a little weird that we say the AMI is deleted when it didn't exist.  I feel like there is an "if" statement missing.

cpmeadors (commented):
Just some questions for docs (who should probably be reviewing as well) and a question about the output and user visibility.

config/peerpods/podvm/aws-VM-image-delete-job.yaml
snir911: It's possible to see the error in the job logs, however, in this case the error means image is already deleted and there's nothing else to do which is fine, would it make more sense to change the wording to something like: "image deletion job completed"?

config/peerpods/podvm/aws-VM-image-delete-job.yaml
snir911: Is custom images supported ATM? AFAIK the reason it's supported is for tech/dev-preview features only

config/peerpods/podvm/aws-VM-image-create-job.yaml
snir911: User shouldn't even know it's possible to set it, this variable is not mentioned in the docs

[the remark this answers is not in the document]
gkurz: Ok, so IMAGE_NAME is a kind of hidden API that we don't want to document for users but we still want to be usable, correct ? It would be very nice to document this in some document for developers at least, or even comments in the yaml files, because the questions in this PR clearly show that it isn't obvious.

config/peerpods/podvm/aws-VM-image-delete-job.yaml
gkurz: > Are users supposed to see the error message in the output of the job?

Users can see these logs with `oc logs` but they aren't supposed to look at them. This is an internal job.

> Is that captured in a log anywhere? 

This should probably be captured by the OSC must-gather if it isn't the case already.

> It's possible to see the error in the job logs, however, in this case the error means image is already deleted and there's nothing else to do which is fine, would it make more sense to change the wording to something like: "image deletion job completed"?

IIUC the only case where we get the `InvalidAMIID\.(Unavailable|NotFound)` error is when someone deleted the image automatically created by us in our back, right ? Assuming this is `fine`, then the job just needs to be idempotent and report something like `"PODVM_AMI_ID: \"$PODVM_AMI_ID\" is deleted"`.

gkurz (approved):
Thanks @snir911 !

PR #371, opened 2024-01-26
RFC: feature-gates: adding initial structure

beraldoleal:
This commit introduces the `FeatureGates` struct in the operator's configuration, enabling a flexible and controlled approach to feature rollout. The `FeatureGates` struct allows for the enabling or disabling of specific features, with a default behaviour based on each feature's maturity level.

Uses Kubebuilder default value annotations for CRD fields to automatically set default behaviours in the Kubernetes API.

I'm trying to use a simple approach here where we could change promote/depromote a feature by just changing the comments and the default values.

Please let know what I'm missing (probably I'm missing a lot) and what you think about this approach.


snir911 (commented):
Added some comments, however, looks like the right direction to me

controllers/openshift_controller.go
```diff
@@ -135,6 +135,13 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 			return ctrl.Result{}, nil
 		}
 
+		if r.kataConfig.Spec.Features.EnableTimeTravel {
```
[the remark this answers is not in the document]
bpradipt: Instead of having featuregate in the CRD, can we have feature gates in the configmap and read it in the reconcile loop?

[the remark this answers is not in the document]
pmores: That was my impression from the initial discussions, too.

Who controls the gating?  Unless we want to enable users to do it we shouldn't put it in KataConfig.

api/v1/kataconfig_types.go
```diff
@@ -44,6 +44,30 @@ type KataConfigSpec struct {
...
 	EnablePeerPods bool `json:"enablePeerPods"`
+
+	// FeatureGates is used to enable/disable specific features. If not
+	// specified, features will be enabled only if enabled by default. Each
+	// feature has a default behaviour, depending on the maturity level.
+	// *optional
+	Features FeatureGates `json:"features,omitempty"`
+}
+
+type FeatureGates struct {
```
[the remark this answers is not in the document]
beraldoleal: Agree. I just would like to confirm if we are going to migrate peer-pods to feature gates or not.

controllers/openshift_controller.go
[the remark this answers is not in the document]
beraldoleal: Hi @snir911 and @bpradipt, thanks for the feedback. My thoughts:

There isn't a "standard" method for this, but since we have a CRD in place, it avoids extra complexity and additional objects.

I was concerned about managing default values and documentation with a ConfigMap (CM). However, using kubebuilder:default= simplifies default values, and we can leverage tools like swagger for documentation, keeping everything in one place.

In my testing, yes: a kataconfig is necessary, but spec.Features doesn't need to be explicitly defined; also the reconcile loop was catching the changes normally. I chose CRD to make it easier for users.

I'm open to using CM if there are benefits we are missing. My initial choice was just to avoid additional complexities.

api/v1/kataconfig_types.go
```diff
@@ -44,6 +44,30 @@ type KataConfigSpec struct {
 	// +optional
 	// +kubebuilder:default:=false
 	EnablePeerPods bool `json:"enablePeerPods"`
+
+	// FeatureGates is used to enable/disable specific features. If not
+	// specified, features will be enabled only if enabled by default. Each
+	// feature has a default behaviour, depending on the maturity level.
+	// *optional
```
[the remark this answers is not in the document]
beraldoleal: good catch. thanks. fixed.

controllers/openshift_controller.go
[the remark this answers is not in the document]
beraldoleal: @pmores initiall requirements is to have a "_a user-controlled feature gating mechanism_".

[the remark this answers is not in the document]
pmores: We need to make sure that whatever we put in KataConfig needs to be OK to be controlled by the user, simply because the user _owns_ their KataConfig instance.  We need to avoid repeating the kata monitor image problem.  If we put control of the features to users' hands we need to understand and accept that they e.g. might leave the features off even after we'd like them on.

Technically both CM and KataConfig allow users to set the gate values, however a CM could be undocumented and thus out of reach for users (unless they're fine with losing support). KataConfig is explicitly documented as controlled by users.

Also, changing the KataConfig CRD has been a major PITA, both because RH downstream build tools don't handle that, and because of the CRD versioning & deprecation complexity.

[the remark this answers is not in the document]
jensfr: I would also say that we should try to avoid using the CRD, it's very complex with respect to upgrades, versioning of the CRD, conversion webhooks etc.

internal/featuregates/featuregates.go
```diff
@@ -0,0 +1,41 @@
...
+	Client    client.Client
+	Namespace string
+	MapName   string
+}
+
+var DefaultFeatureGates = map[string]bool{
+	"timeTravel":              false,
+	"quantumEntanglementSync": false,
+	"autoHealingWithAI":       true,
+}
```
snir911, line 20: Can we have this map as part of the FeatureGates struct?

controllers/openshift_controller.go
```diff
@@ -1527,8 +1533,8 @@ func (r *KataConfigOpenShiftReconciler) SetupWithManager(mgr ctrl.Manager) error
 			&McpEventHandler{r}).
 		Watches(
 			&source.Kind{Type: &corev1.Node{}},
-			&NodeEventHandler{r}).
-		Complete(r)
+			&NodeEventHandler{r},
+		).Complete(r)
```
snir911, line 1537: What if user switches a feature at some point? should we support it? if we do we should watch ConfigMaps here

beraldoleal: Right. Makes sense and I just sent.

api/v1/kataconfig_types.go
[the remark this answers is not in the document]
beraldoleal: @snir911, chatting with @jensfr and @bpradipt, we're holding off on moving peer pods for now. It requires CRD changes and a bit of work. Also, I would like to start by implementing the mechanism first, and future PRs can use it. Sound good to you?

internal/featuregates/featuregates.go
snir911: I thought it will be only single instance of `FeatureGates` struct, isn't it?

snir911: This is not critical for me, you can keep it as you prefer, however, AFAIU the FeatureGates will be used only once by the controller and it shouldn't be used elsewhere and the map is used in the same context so make sense to me to include it. Also AFAIU map cannot be static, so it's simply defined as global variable ATM...

beraldoleal: I will keep it for now, and latter we can reevaluate if needed.

Thank you again, and let me know if you ack/nack the PR.

snir911:
@beraldoleal thanks! LGTM overall, couple of questions:
1. are we merging it with the dummy features? i.e. timeTrabel, AIhealing etc...?
2. consider filtering the events using Predicate rather than handler, see this [reference](https://github.com/openshift/sandboxed-containers-operator/blob/721edad2c77f4c8ad7153acb15cdbfc7f0cdd2db/controllers/credentials_controller.go#L116-L140) (IIUC it would filter the events before actually emitted and handled)
3. Should we deploy also the  osc-feature-gates CM with the default values so it can be used as a template for the user? IDK if it's even possible...

beraldoleal:
Hi @snir911 , my comments bellow:
 
>     1. are we merging it with the dummy features? i.e. timeTrabel, AIhealing etc...?

Good question. I was planning to keep the default map empty (since is a code block) but keep the `config/samples/featuregates.yaml` there with some notes, saying those are fake... (since is just an example and until we have our first feature on boarded). Wdyt?

>     2. consider filtering the events using Predicate rather than handler, see this [reference](https://github.com/openshift/sandboxed-containers-operator/blob/721edad2c77f4c8ad7153acb15cdbfc7f0cdd2db/controllers/credentials_controller.go#L116-L140) (IIUC it would filter the events before actually emitted and handled)

When considering the implementation, my initial thought was to use Predicates. However, since the other watches were utilizing EventHandlers, I opted to maintain consistency and avoid a mix of approaches. To be honest, I also prefer using Predicates. If you see no issue with incorporating both EventHandlers and Predicates within the same setup, we could explore this option.
 
>     3. Should we deploy also the  osc-feature-gates CM with the default values so it can be used as a template for the user? IDK if it's even possible...

I believe it's possible, yes, but it might be unnecessary. The default values are intended for use as-is, without the need for user intervention. Introducing another ConfigMap into users' deployments for those adhering to the default values might not be ideal. The `config/samples/` directory serves as a source of "templates" for users. So, unless there is something I'm missing, I would say to keep it just in `config/samples/`.

controllers/openshift_controller.go
```diff
@@ -120,6 +122,10 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
+	if r.FeatureGates.IsEnabled(ctx, "timeTravel") {
```
littlejawa, line 125: Maybe it is inferred, but we need to take disabling into account too.

If a feature is enabled by default, we need a way to disable it from the configmap.
And when the reconciler is called on a "Delete" operation, I would assume it should cause the selected options to be disabled.

Also it seems the reconciler is not called on an "Update" event, so if we modify the configmap to disable something we initially enabled, it will not happen. We would need to delete/recreate it to change the config.

This seems contradictory with the documentation:
```Feature gates can be enabled or disabled by editing the `osc-feature-gates` ConfigMap
resource.```

Comment (snir911) [reply controllers/openshift_controller.go:125]: Oh, Right,  however, I guess if we make sure after every **if IsEnabled==true {** to enable the feature **} else {** to disable the feature} it should keep everything idempotent, no?
In any case I'd try to look for a reference as i have no idea what is the best practice 
(also cc @pmores just in case he knows something about it as he's operators Pro)

**Comment by snir911** [comment]
> Hi @snir911 , my comments bellow:
> 
> > 1. are we merging it with the dummy features? i.e. timeTrabel, AIhealing etc...?
> 
> Good question. I was planning to keep the default map empty (since is a code block) but keep the `config/samples/featuregates.yaml` there with some notes, saying those are fake... (since is just an example and until we have our first feature on boarded). Wdyt?

IDK, we can either start with a real feature (maybe peer-pods? but just print a msg instead of actually enabling it ATM?) or maybe to leave only single dummy feature?
> 
> > 2. consider filtering the events using Predicate rather than handler, see this [reference](https://github.com/openshift/sandboxed-containers-operator/blob/721edad2c77f4c8ad7153acb15cdbfc7f0cdd2db/controllers/credentials_controller.go#L116-L140) (IIUC it would filter the events before actually emitted and handled)
> 
> When considering the implementation, my initial thought was to use Predicates. However, since the other watches were utilizing EventHandlers, I opted to maintain consistency and avoid a mix of approaches. To be honest, I also prefer using Predicates. If you see no issue with incorporating both EventHandlers and Predicates within the same setup, we could explore this option.
AFAIK there's no problem to mix (we do it in the credentials controller), the handlers are usually used when there are some ops needed IIUC.. 
> 
> > 3. Should we deploy also the  osc-feature-gates CM with the default values so it can be used as a template for the user? IDK if it's even possible...
> 
> I believe it's possible, yes, but it might be unnecessary. The default values are intended for use as-is, without the need for user intervention. Introducing another ConfigMap into users' deployments for those adhering to the default values might not be ideal. The `config/samples/` directory serves as a source of "templates" for users. So, unless there is something I'm missing, I would say to keep it just in `config/samples/`.

Sure, anyway we can do it at any point..

File: controllers/openshift_controller.go (hunk 6 above)
Comment (littlejawa) [reply controllers/openshift_controller.go:125]: Yes, I think the "else" part would be required. No big deal for this example code, but we really need to keep it in mind on every feature we add.
Also we need to be careful when doing the "to disable the feature" part : this code will be called for every feature that is disabled by default, even when no configmap is involved. It needs to work when there is nothing to disable.

File: controllers/openshift_controller.go [hunk 7]
```diff
@@ -1572,7 +1617,10 @@ func (r *KataConfigOpenShiftReconciler) SetupWithManager(mgr ctrl.Manager) error
 		Watches(
 			&corev1.Node{},
 			&NodeEventHandler{r}).
-		Complete(r)
+		Watches(
+			&corev1.ConfigMap{},
+			&ConfigMapEventHandler{r},
```
bpradipt, line 1622: It needs to filter out events and only act on specific configmap. Otherwise it'll be too noisy.
An example is in the credentials_controller.go

config/samples/featuregates.yaml
```diff
@@ -0,0 +1,22 @@
...
+  # using principles of quantum mechanics. This advanced feature ensures
+  # data is synchronized across different locations without traditional
+  # network latency. Default is "false".
+  # quantumEntanglementSync: "false"
+
+  # autoHealingWithAI employs artificial intelligence to automatically
+  # detect and resolve cluster issues. It leverages machine learning algorithms
+  # to predict potential problems before they occur, ensuring high availability.
+  # Default is "true".
+  # autoHealingWithAI: "true"
```
bpradipt, line 22: Should we use a real featuregate in this PR to make it complete ?
@littlejawa @snir911 wdyt ?

beraldoleal:
> IDK, we can either start with a real feature (maybe peer-pods? but just print a msg instead of actually enabling it ATM?) or maybe to leave only single dummy feature?

Currently peerpods is enabled by another way... adding peer-pods to the feature feature gates without disabling the other way it will be confuse. Also, if we add a new feature with just a print message (lets say "CoCo") it will give the users the wrong impression. IMO it will be confusing. So yes, I will leave just a single dummy feature for now.

> AFAIK there's no problem to mix (we do it in the credentials controller), the handlers are usually used when there are some ops needed IIUC..

Ok, I will send a new version with the Predicates.

Thanks again, @snir911 .

controllers/openshift_controller.go
beraldoleal: Yes, I'm going to submit a Predicate version soon.

controllers/openshift_controller.go
beraldoleal: > If a feature is enabled by default, we need a way to disable it from the configmap.
>  And when the reconciler is called on a "Delete" operation, I would assume it should cause the selected options to be disabled.

Currently you can disable from the configmap, you just need to set the value to `false` and use the else logic  as mentioned by @snir911 

> Also it seems the reconciler is not called on an "Update" event, so if we modify the configmap to disable something we initially enabled, it will not happen. We would need to delete/recreate it to change the config.

Nice catch, actually there is a missing queue.add . I will fix when implementing the Predicates.

beraldoleal: And yes, I'm open to ways to improve this. If someone has a better approach, is more than welcome.

beraldoleal:
@snir911 @bpradipt just pushed with the Predicate implementation, let me know if I'm missing something.

controllers/openshift_controller.go
littlejawa: I don't have a better approach, I think you're on the right track.
My comments are probably not for this PR really, and more for our future selves that will have to be careful when implementing it.

config/samples/featuregates.yaml
littlejawa: I think this PR is fine with fake examples - it is setting a framework that we can build upon, but doesn't really change the current workflow.
Implementing an actual feature gate would raise feature-specific questions that we don't really want to address here I think.

bpradipt (approved):
/lgtm
Thanks @beraldoleal

controllers/openshift_controller.go
```diff
@@ -1563,6 +1572,38 @@ func (eh *NodeEventHandler) Delete(ctx context.Context, event event.DeleteEvent,
...
+			return false
+		}
+
+		configMap, ok := obj.(*corev1.ConfigMap)
+		if !ok {
+			return false
+		}
+
+		relevantNamespace := configMap.Namespace == reconciler.FeatureGates.Namespace
+		relevantName := configMap.Name == reconciler.FeatureGates.ConfigMapName
```
bpradipt, line 1587: One thing which is unclear to me is who is setting the FeatureGates struct that is being compared here ?

beraldoleal: Hi @bpradipt, we are doing this at main:146 when creating the KataConfigOpenShiftReconciler.

bpradipt (approved):
/lgtm
Thanks @beraldoleal

beraldoleal:
@snir911 ?

snir911 (approved):
LGTM, thanks! (assuming it was tested)

PR #380, opened 2024-03-21
Bump golang version to 1.21

spotlesstofu:

**- What I did**

**- How to verify it**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Use OpenShift golang builder image
 # These images needs to be synced with the images in the Makefile.
-ARG BUILDER_IMAGE=${BUILDER_IMAGE:-registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.13}
-ARG TARGET_IMAGE=${TARGET_IMAGE:-registry.ci.openshift.org/ocp/4.13:base}
+ARG BUILDER_IMAGE=${BUILDER_IMAGE:-registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16}
+ARG TARGET_IMAGE=${TARGET_IMAGE:-registry.ci.openshift.org/ocp/4.16:base}
 FROM ${BUILDER_IMAGE} AS builder
 
 WORKDIR /workspace
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,8 @@ GOBIN=$(shell go env GOBIN)
 endif
 
 # These images needs to be synced with the default values in the Dockerfile.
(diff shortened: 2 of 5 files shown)


gkurz (approved):
/lgtm

Thanks @spotlesstofu !

bpradipt:
@spotlesstofu shouldn't this be targeted for devel branch ?

gkurz:
> @spotlesstofu shouldn't this be targeted for devel branch ?

Indeed I overlooked it was targetting a dummy branch... @spotlesstofu is the PR ok to go to `devel` ?

spotlesstofu:
@gkurz it is OK now. The bot tells it needs rebase, but it shouldn't. I changed the target branch and rebased already.

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Use OpenShift golang builder image
 # These images needs to be synced with the images in the Makefile.
-ARG BUILDER_IMAGE=${BUILDER_IMAGE:-registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.13}
+ARG BUILDER_IMAGE=${BUILDER_IMAGE:-registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.21-openshift-4.13}
```
bpradipt: does it makes sense to use the following images ?
BUILDER_IMAGE=registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
TARGET_IMAGE=registry.ci.openshift.org/ocp/4.16:base-rhel9

Ref: https://github.com/openshift/release/blob/077ae802a5461f63a34ba5157cc2bc2783db7219/ci-operator/config/openshift/oc/openshift-oc-release-4.16.yaml

gkurz: These are  likely the images @spotlesstofu is looking for. Thanks @bpradipt !

bpradipt (approved):
/lgtm

gkurz:
Thanks @spotlesstofu !

We don't formally ask  for it but it is generally good practice to add some extra info in the commit message and a S-o-b tag.

e.g.

```
Bump golang version to 1.21

Downstream OSC was already bumped to use golang 1.21. We should at least use
that version for any custom builds from now on.

Next version of OCP we'll need to support is 4.16. Their CI operator relies on golang
1.21 builder image at [1]. Let's use this same image.

[1] https://github.com/openshift/release/blob/077ae802a5461f63a34ba5157cc2bc2783db7219/ci-operator/config/openshift/oc/openshift-oc-release-4.16.yaml#L25

Signed-off-by: Camilla Conte <cconte@redhat.com>
```

spotlesstofu:
Thanks @gkurz, just updated with your commit message.

PR #382, opened 2024-03-27
tests: cloud credentials controller 

snir911:
This series disables the kataConfig controller testing and adds basic testing for the credentials operator against KataConfig events.
Since KataConfig Controller tests are broken ATM we need to improve mimicking standard cluster and then add more test cases

### DIFF
--- a/controllers/credentials_controller_test.go
+++ b/controllers/credentials_controller_test.go
@@ -1,0 +1,256 @@
+package controllers
+
+import (
+	"context"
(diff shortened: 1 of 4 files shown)


bpradipt (approved):
nice work @snir911 
This lays the foundation to add tests for other pieces. 
We really need to figure out how to test the reconcile loop for KataConfig which depends on MCP. But that's for other time :-)

PR #376, opened 2024-02-21
credentials-controller: fetch credentials using cloud-credentials-operator

snir911:
flow is:

credentials-controller creates credentialsRequests for all supported providers upon
KataConfig+PeerPods creation ->
CCO creates cco-secret in response to provider's credentialsRequests ->
credentials-controller triggers by cco-secret creation and translates cco-secrets
to peer-pods-secret owned by the cco-secret

at deletion of KataConfig:
credentials-controller delets credentialsRequests ->
CCO deletes the cco-secret ->
GC deletes owned peer-pods-secret

to test:
```
quay.io/snir/osc-operator-catalog:v9.9.114
quay.io/snir/osc-operator-bundle:v9.9.114
quay.io/snir/osc-operator:v9.9.114

```

**avoiding OOM patch introduced by #375 , remove it**

PR #384, opened 2024-03-27
Bump OSC 1.6.0

bpradipt:


gkurz (changes requested):
Hi @bpradipt ! The usual workflow for a major or minor release is to merge devel into main. We try to avoid cherry-picks as much as we can because they don't coexist well with merges.

Any reason for not merging ?

I have raised the request changes flag to prevent accidental merge since the button isn't grayed out.

bpradipt:
> Hi @bpradipt ! The usual workflow for a major or minor release is to merge devel into main. We try to avoid cherry-picks as much as we can because they don't coexist well with merges.
> 
> Any reason for not merging ?
> 
> I have raised the request changes flag to prevent accidental merge since the button isn't grayed out.

We have the cloud-credentials request feature that's not part of 1.6.0 and hence I didn't cherry-pick it. What do you suggest ?

gkurz:
> > Hi @bpradipt ! The usual workflow for a major or minor release is to merge devel into main. We try to avoid cherry-picks as much as we can because they don't coexist well with merges.
> > Any reason for not merging ?
> > I have raised the request changes flag to prevent accidental merge since the button isn't grayed out.
> 
> We have the cloud-credentials request feature that's not part of 1.6.0 and hence I didn't cherry-pick it. What do you suggest ?

Are you referring to #376 ? If yes, what would be the impact of having this in 1.6.0 @bpradipt ?

bpradipt:
> > > Hi @bpradipt ! The usual workflow for a major or minor release is to merge devel into main. We try to avoid cherry-picks as much as we can because they don't coexist well with merges.
> > > Any reason for not merging ?
> > > I have raised the request changes flag to prevent accidental merge since the button isn't grayed out.
> > 
> > 
> > We have the cloud-credentials request feature that's not part of 1.6.0 and hence I didn't cherry-pick it. What do you suggest ?
> 
> Are you referring to #376 ? If yes, what would be the impact of having this in 1.6.0 @bpradipt ?

Additional code will be included which we don't expect users to consume yet. Although the code will not be triggered if the  required peerpods secrets are manually created (which is what we expect the users to do for 1.6).

How about I reorder the commits when merging to main from devel so that we can select a specific commit downstream which doesn't include cloud-creds and the featuregate code. @gkurz would this be fine ?

bpradipt:
Closing this PR as cherry-pick is to be avoided when merging from devel to main

PR #386, opened 2024-04-02
Bump OSC to 1.5.3

gkurz:
Prepare the release of OSC 1.5.3 on the main branch. A subsequent PR will be created to merge this change back into the devel branch.

Fixes: [KATA-2876](https://issues.redhat.com//browse/KATA-2876)

PR #387, opened 2024-04-02
Merge 1.5.3 to devel

gkurz:
This is to bring to the devel branch #386 that was done on the main branch.

EDIT: The PR's branch is indeed 45 commits away from devel but all these commits already exist in devel and git merge will do FF.


bpradipt (approved):
/lgtm
Thanks @gkurz

PR #389, opened 2024-04-02
KATA-2881: add new annotations to CSV

cpmeadors:
CVP downstream requires new annotations to exist in the CSV. https://docs.openshift.com/container-platform/4.15/operators/operator_sdk/osdk-generating-csvs.html#osdk-csv-annotations-infra_osdk-generating-csvs

Closes: [KATA-2881](https://issues.redhat.com//browse/KATA-2881)

**- What I did**
Added the annotations to CSV.  I set disconnected to true and all the rest false.  We can update them if that is wrong, but this should satisfy CVP.

**- How to verify it**
Make bundle.  Check bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml to see if new annotations are in rendered CSV

### DIFF
--- a/config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -13,9 +13,15 @@ metadata:
         }
       ]
     capabilities: Seamless Upgrades
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "false"
+    features.operators.openshift.io/proxy-aware: "false"
+    features.operators.openshift.io/tls-profiles: "false"
+    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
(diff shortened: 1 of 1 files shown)


beraldoleal (approved):
LGTM.

bpradipt (approved):
/lgtm

config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -13,6 +13,13 @@ metadata:
         }
       ]
     capabilities: Seamless Upgrades
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "false"
```
bpradipt, line 17: nit: this should be true given
`operators.openshift.io/infrastructure-features: '["disconnected", "fips"]'`

cpmeadors: ProdSec invalidated our FIPS compliance by invalidating anything based On UBI builds.  We must build on rhel-els base images for FIPS compliance now.  I wasn't sure what to do here.  This pr is about resolving the CVP issues and it doesn't care what value is set, just that it is set.  I think it would be clearer if we create additional stories to resolve the FIPS compliance issue.

PR #390, opened 2024-04-03
Merge 1.5.3 to devel (part 2)

gkurz:
This is to bring #389 to devel.


Dockerfile
```diff
@@ -15,7 +15,6 @@ COPY main.go main.go
 COPY api api/
 COPY config config/
 COPY controllers controllers/
-COPY internal internal/
```
bpradipt, line 18: Shouldn't this revert commit be dropped as this is deleting new code in devel that we want?

gkurz: Right, merging main as is reverts all the 1.6.0 code from devel. We need to think harder :wink:

gkurz:
New approach in #391.

PR #388, opened 2024-04-02
Revert "Merge pull request #385 from openshift/1.6-bump"

gkurz:
Code for 1.6.0 landed early in the main branch by mistake. We need proper builds from devel to be tested by QE first and we still have 1 sprint to go until the release.

This reverts commit 108248312ff31abcf47cbad5758a65e53e90926a, reversing changes made to 2a9ed5eaea02442f98e9e610cbd9c9bb20147a43.

PR #391, opened 2024-04-04
Merge csv change to devel

gkurz:
This is to bring #389 and #392 to devel without #388 that would remove 1.6.0 code from devel :wink:

Solution is to revert #388 inside this PR.


cpmeadors (approved):
Changes look good.  Thanks for removing the infrastructure-features line.  That resolves the conflict for FIPS.

bpradipt (approved):
/lgtm

gkurz:
We're taking the different approach with a release branch.

PR #392, opened 2024-04-04
Introduce `replaces` field in CSV

gkurz:
This is now mandated by CVP if the CSV uses the `skipRange` field.


cpmeadors (approved):
Matches downstream change which is proven to work.

PR #393, opened 2024-04-05
Fix the podvm-payload image references

bpradipt:
Ensure release tag is used


gkurz (approved):
/lgtm. Thanks @bpradipt !

PR #394, opened 2024-04-06
Add two feature gates : Image based deployment and additional runtimeClasses

bpradipt:
This is a feature enhancement adding two feature gates which typically goes together.
- Image based deployment and additional runtimeClasses
An example configmap enabling the feature
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: osc-feature-gates
  namespace: openshift-sandboxed-containers-operator
data:
  LayeredImageDeployment: "false"
  AdditionalRuntimeClasses: "false"
```

Sample ConfigMap with configs for individual features
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: layeredimagedeployment-config
  namespace: openshift-sandboxed-containers-operator
data:
  osImageURL: "quay.io/bpradipt/coco:ocp415"
  kernelArguments: "kvm_intel.tdx=1"
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: additionalruntimeclasses-config
  namespace: openshift-sandboxed-containers-operator
data:
  runtimeClassConfig: "kata-cc-tdx, kata-cc-sim"
```

This code can be tested on an OCP cluster by following the instructions from this repo:
https://github.com/bpradipt/coco-install/tree/main/osc-image-deploy
SNO should be fine as well for testing.

### DIFF
--- a/controllers/cm_event_handler.go
+++ b/controllers/cm_event_handler.go
@@ -1,0 +1,59 @@
+package controllers
+
+import (
+	"context"
+
+	"github.com/openshift/sandboxed-containers-operator/internal/featuregates"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+type ConfigMapEventHandler struct {
+	reconciler *KataConfigOpenShiftReconciler
+}
+
+func (ch *ConfigMapEventHandler) Create(ctx context.Context, event event.CreateEvent, queue workqueue.RateLimitingInterface) {
+
+	if ch.reconciler.kataConfig == nil {
+		return
+	}
+
+	cm := event.Object
+
+	// Check if the configmap name is one of the feature gates configmap or
+	// feature config
+	if cm.GetNamespace() != OperatorNamespace || !featuregates.IsFeatureGateConfigMap(cm.GetName()) {
+		return
+	}
+	log := ch.reconciler.Log.WithName("CMCreate").WithValues("cm name", cm.GetName())
+	log.Info("FeatureGates configmap created")
+
+	queue.Add(ch.reconciler.makeReconcileRequest())
+}
+
+func (ch *ConfigMapEventHandler) Update(ctx context.Context, event event.UpdateEvent, queue workqueue.RateLimitingInterface) {
+
+	if ch.reconciler.kataConfig == nil {
+		return
+	}
+
+	cm := event.ObjectNew
+
+	// Check if the configmap name is one of the feature gates configmap or
+	// feature config
+	if cm.GetNamespace() != OperatorNamespace || !featuregates.IsFeatureGateConfigMap(cm.GetName()) {
+		return
+	}
+
+	log := ch.reconciler.Log.WithName("CMUpdate").WithValues("cm name", cm.GetName())
+	log.Info("FeatureGates configmap updated")
+
+	queue.Add(ch.reconciler.makeReconcileRequest())
+
+}
+
+func (ch *ConfigMapEventHandler) Delete(ctx context.Context, event event.DeleteEvent, queue workqueue.RateLimitingInterface) {
+}
+
+func (ch *ConfigMapEventHandler) Generic(ctx context.Context, event event.GenericEvent, queue workqueue.RateLimitingInterface) {
+}
--- a/controllers/cm_event_handler.go
+++ b/controllers/cm_event_handler.go
@@ -53,6 +53,21 @@ func (ch *ConfigMapEventHandler) Update(ctx context.Context, event event.UpdateE
 }
 
 func (ch *ConfigMapEventHandler) Delete(ctx context.Context, event event.DeleteEvent, queue workqueue.RateLimitingInterface) {
+	if ch.reconciler.kataConfig == nil {
+		return
+	}
+
+	cm := event.Object
+
+	// Check if the configmap name is one of the feature gates configmap or
+	// feature config
+	if cm.GetNamespace() != OperatorNamespace || !featuregates.IsFeatureGateConfigMap(cm.GetName()) {
+		return
+	}
+	log := ch.reconciler.Log.WithName("CMDelete").WithValues("cm name", cm.GetName())
+	log.Info("FeatureGates configmap deleted")
+
+	queue.Add(ch.reconciler.makeReconcileRequest())
 }
 
 func (ch *ConfigMapEventHandler) Generic(ctx context.Context, event event.GenericEvent, queue workqueue.RateLimitingInterface) {
--- a/controllers/image_generator.go
+++ b/controllers/image_generator.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
@@ -242,7 +243,9 @@ func newImageGenerator(client client.Client) (*ImageGenerator, error) {
 func (r *ImageGenerator) createJobFromFile(jobFileName string) (*batchv1.Job, error) {
 	igLogger.Info("Create Job out of YAML file", "jobFileName", jobFileName)
 
-	yamlData, err := readJobYAML(jobFileName)
+	jobYamlFile := filepath.Join(peerpodsImageJobsPathLocation, jobFileName)
+
+	yamlData, err := readYamlFile(jobYamlFile)
 	if err != nil {
 		return nil, err
 	}
@@ -630,7 +633,8 @@ func (r *ImageGenerator) createImageConfigMapFromFile() error {
 	}
 
 	filename := r.provider + "-podvm-image-cm.yaml"
-	yamlData, err := readConfigMapYAML(filename)
+	configMapYamlFile := filepath.Join(peerpodsImageJobsPathLocation, filename)
+	yamlData, err := readYamlFile(configMapYamlFile)
 	if err != nil {
 		return err
 	}
--- a/controllers/mc_handler.go
+++ b/controllers/mc_handler.go
@@ -1,0 +1,195 @@
+package controllers
+
+/*
+This code handles the installation of Kata artifacts using an RHCOS extension or image and deploying it
+via MachineConfig
+*/
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	ignTypes "github.com/coreos/ignition/v2/config/v3_2/types"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	"github.com/openshift/sandboxed-containers-operator/internal/featuregates"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	extension_mc_name = "50-enable-sandboxed-containers-extension"
+	image_mc_name     = "50-enable-sandboxed-containers-image"
+)
+
+type DeploymentState struct {
+	PreviousMethod string
+	PreviousMcName string
+}
+
+// If the first return value is 'true' it means that the MC was just created
+// by this call, 'false' means that it's already existed.  As usual, the first
+// return value is only valid if the second one is nil.
+func (r *KataConfigOpenShiftReconciler) createMc(machinePool string) (bool, error) {
+
+	// In case we're returning an error we want to make it explicit that
+	// the first return value is "not care".  Unfortunately golang seems
+	// to lack syntax for creating an expression with default bool value
+	// hence this work-around.
+	var dummy bool
+
+	/* Create Machine Config object to deploy sandboxed containers artifacts */
+	mcName := extension_mc_name
+	if r.FeatureGatesStatus[featuregates.LayeredImageDeployment] {
+		mcName = image_mc_name
+	}
+
+	r.Log.Info("Checking if MachineConfig exists", "mc.Name", mcName)
+	mc := &mcfgv1.MachineConfig{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: mcName}, mc)
+	if err != nil && (k8serrors.IsNotFound(err) || k8serrors.IsGone(err)) {
+		r.Log.Info("creating new MachineConfig")
+		mc, err = r.newMCForCR(machinePool)
+		if err != nil {
+			return dummy, err
+		}
+
+		err = r.Client.Create(context.TODO(), mc)
+		if err != nil {
+			r.Log.Error(err, "Failed to create a new MachineConfig ", "mc.Name", mc.Name)
+			return dummy, err
+		}
+		r.Log.Info("MachineConfig successfully created", "mc.Name", mc.Name)
+		return true, nil
+	} else if err != nil {
+		r.Log.Info("failed to retrieve MachineConfig", "err", err)
+		return dummy, err
+	} else {
+		r.Log.Info("MachineConfig already exists")
+		return false, nil
+	}
+}
+
+// Create a new MachineConfig object for the Custom Resource
+func (r *KataConfigOpenShiftReconciler) newMCForCR(machinePool string) (*mcfgv1.MachineConfig, error) {
+	r.Log.Info("Creating MachineConfig for Custom Resource")
+
+	ic := ignTypes.Config{
+		Ignition: ignTypes.Ignition{
+			Version: "3.2.0",
+		},
+	}
+
+	icb, err := json.Marshal(ic)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.FeatureGatesStatus[featuregates.LayeredImageDeployment] {
+		r.Log.Info("Creating MachineConfig for Custom Resource with image")
+		return r.newImageMCForCR(machinePool, icb)
+	} else {
+		r.Log.Info("Creating MachineConfig for Custom Resource with extension")
+		return r.newExtensionMCForCR(machinePool, icb)
+	}
+
+}
+
+// Create a new MachineConfig object for the Custom Resource with the extension
+func (r *KataConfigOpenShiftReconciler) newExtensionMCForCR(machinePool string, icb []byte) (*mcfgv1.MachineConfig, error) {
+	extension := getExtensionName()
+
+	mc := mcfgv1.MachineConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "machineconfiguration.openshift.io/v1",
+			Kind:       "MachineConfig",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: extension_mc_name,
+			Labels: map[string]string{
+				"machineconfiguration.openshift.io/role": machinePool,
+				"app":                                    r.kataConfig.Name,
+			},
+			Namespace: OperatorNamespace,
+		},
+		Spec: mcfgv1.MachineConfigSpec{
+			Extensions: []string{extension},
+			Config: runtime.RawExtension{
+				Raw: icb,
+			},
+		},
+	}
+
+	return &mc, nil
+}
+
+// Create a new MachineConfig object for the Custom Resource with the image
+func (r *KataConfigOpenShiftReconciler) newImageMCForCR(machinePool string, icb []byte) (*mcfgv1.MachineConfig, error) {
+
+	// Get the LayeredImageDeployment feature gate parameters
+	imageParams := r.FeatureGates.GetFeatureGateParams(context.TODO(), featuregates.LayeredImageDeployment)
+	// Ensure at least osImageURL is set
+	if _, ok := imageParams["osImageURL"]; !ok {
+		return nil, fmt.Errorf("osImageURL not set in feature gate %s", featuregates.LayeredImageDeployment)
+	}
+
+	mc := mcfgv1.MachineConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "machineconfiguration.openshift.io/v1",
+			Kind:       "MachineConfig",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: image_mc_name,
+			Labels: map[string]string{
+				"machineconfiguration.openshift.io/role": machinePool,
+				"app":                                    r.kataConfig.Name,
+			},
+			Namespace: OperatorNamespace,
+		},
+		Spec: mcfgv1.MachineConfigSpec{
+			OSImageURL: imageParams["osImageURL"],
+			Config: runtime.RawExtension{
+				Raw: icb,
+			},
+		},
+	}
+
+	if kernelArguments, ok := imageParams["kernelArguments"]; ok {
+		// Parse the kernel arguments and set them in the MachineConfig
+		// Note that in the configmap the kernel arguments are stored as a single string
+		// eg. "a=b c=d ..." and we need to split them into individual arguments
+		// eg ["a=b", "c=d", ...]
+		// Split the kernel arguments string into individual arguments
+		mc.Spec.KernelArguments = strings.Fields(kernelArguments)
+	}
+
+	return &mc, nil
+}
+
+// Delete the MachineConfig object
+func (r *KataConfigOpenShiftReconciler) deleteMc(mcName string) error {
+	r.Log.Info("Deleting MachineConfig", "mc.Name", mcName)
+	mc := &mcfgv1.MachineConfig{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: mcName}, mc)
+	if err != nil {
+		if k8serrors.IsNotFound(err) || k8serrors.IsGone(err) {
+			r.Log.Info("MachineConfig not found. Most likely it was already deleted.")
+			return nil
+		}
+		r.Log.Info("failed to retrieve MachineConfig", "err", err)
+		return err
+	}
+
+	r.Log.Info("deleting MachineConfig")
+	err = r.Client.Delete(context.TODO(), mc)
+	if err != nil {
+		r.Log.Error(err, "Failed to delete MachineConfig")
+		return err
+	}
+
+	r.Log.Info("MachineConfig successfully deleted")
+	return nil
+}
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -17,7 +17,6 @@ package controllers
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
 	"reflect"
@@ -29,7 +28,6 @@ import (
 
 	"k8s.io/apimachinery/pkg/labels"
 
-	ignTypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/go-logr/logr"
 	secv1 "github.com/openshift/api/security/v1"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
@@ -66,6 +64,7 @@ type KataConfigOpenShiftReconciler struct {
 	kataConfig         *kataconfigurationv1.KataConfig
 	FeatureGates       *featuregates.FeatureGates
 	FeatureGatesStatus featuregates.FeatureGateStatus
+	deploymentState    DeploymentState
 }
 
 const (
@@ -73,7 +72,6 @@ const (
 	dashboard_configmap_name            = "grafana-dashboard-sandboxed-containers"
 	dashboard_configmap_namespace       = "openshift-config-managed"
 	container_runtime_config_name       = "kata-crio-config"
-	extension_mc_name                   = "50-enable-sandboxed-containers-extension"
 	DEFAULT_PEER_PODS                   = "10"
 	peerpodConfigCrdName                = "peerpodconfig-openshift"
 	peerpodsMachineConfigPathLocation   = "/config/peerpods"
@@ -492,46 +490,6 @@ func getExtensionName() string {
 		extension = "kata-containers"
 	}
 	return extension
-}
-
-func (r *KataConfigOpenShiftReconciler) newMCForCR(machinePool string) (*mcfgv1.MachineConfig, error) {
-	r.Log.Info("Creating MachineConfig for Custom Resource")
-
-	ic := ignTypes.Config{
-		Ignition: ignTypes.Ignition{
-			Version: "3.2.0",
-		},
-	}
-
-	icb, err := json.Marshal(ic)
-	if err != nil {
-		return nil, err
-	}
-
-	extension := getExtensionName()
-
-	mc := mcfgv1.MachineConfig{
-		TypeMeta: metav1.TypeMeta{
-			APIVersion: "machineconfiguration.openshift.io/v1",
-			Kind:       "MachineConfig",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name: extension_mc_name,
-			Labels: map[string]string{
-				"machineconfiguration.openshift.io/role": machinePool,
-				"app":                                    r.kataConfig.Name,
-			},
-			Namespace: OperatorNamespace,
-		},
-		Spec: mcfgv1.MachineConfigSpec{
-			Extensions: []string{extension},
-			Config: runtime.RawExtension{
-				Raw: icb,
-			},
-		},
-	}
-
-	return &mc, nil
 }
 
 func (r *KataConfigOpenShiftReconciler) addFinalizer() error {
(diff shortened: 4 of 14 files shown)


bpradipt:
@spotlesstofu ptal

spotlesstofu:
What's the case for exposing the RuntimeClasses definition to the user? I understand that's the current custom in the CoCo operator, but I couldn't find out why.

bpradipt:
> What's the case for exposing the RuntimeClasses definition to the user? I understand that's the current custom in the CoCo operator, but I couldn't find out why.

Good point @spotlesstofu. When using image based deployment, it may require additional runtimeclass to be created depending on the configs included in the image. 
Likewise when enabling CoCo, additional runtimeclass will be required. One idea was to include runtimeClass as a feature specific config param like this
eg
```
imageBasedDeployment.params: |
    osImageURL=quay.io/....
    kernelArguments=a=b c=d ...
    runtimeClassConfig=name1:cpuOverHead1:memOverHead1, name2:cpuOverHead2:memOverHead2
```

Likewise for any feature that might need additional runtimeclasses they can have a feature.params.

However, I was unsure whether to use the above approach or make runtimeClass config as a separate feature itself.

What is your suggestion?

spotlesstofu:
So I understand that the node configuration (image, kernel, etc.) is tied to the RuntimeClass. Then, what we should expose to the user is a CustomResource (**CR**) to represent this. A `KataRuntimeClass`, to instantiate multiple times for each runtime that's needed.

This CR would include the node configuration that's desired for that runtime. The operator should take care of composing the multiple `KataRuntimeClass` and produce a working configuration. Also the default runtime configuration would be moved to this CR at some point.

Does this make sense, would it address our requirements?

bpradipt:
> So I understand that the node configuration (image, kernel, etc.) is tied to the RuntimeClass. Then, what we should expose to the user is a CustomResource (**CR**) to represent this. A `KataRuntimeClass`, to instantiate multiple times for each runtime that's needed.
> 

Node configuration may or may not be tied to a runtimeclass. But if the image do provide configuration for additional runtimeclasses (crio configs), then we need a way to expose these to end-user via creating runtimeclass objects
Using CR seems too heavy weight for this imho

> This CR would include the node configuration that's desired for that runtime. The operator should take care of composing the multiple `KataRuntimeClass` and produce a working configuration. Also the default runtime configuration would be moved to this CR at some point.
> 
> Does this make sense, would it address our requirements?

Like with everything in K8s using CR is definitely a way to implement. We might as well decide to include these in KataConfig CR with additional attributes.  Question is do we want to use CR or let it be part of the feature configmap ..

internal/featuregates/features.go
```diff
@@ -0,0 +1,59 @@
+package featuregates
```
beraldoleal, line 1: This commit message says is adding a new feature and it would be behind the feature gate mechanism. Here you are introducing featuregates module and changing the approach a "little bit"...

Do you mind to split this commit into two so we can review them independently, with the reason for that on its own commit message, please?

internal/featuregates/featuregates.go
```diff
@@ -42,3 +48,59 @@ func (fg *FeatureGates) IsEnabled(ctx context.Context, feature string) bool {
...
 	return false
 }
+
+// Method to read the feature specific config parameters from the configmap
+// The feature specific config params are stored in their own configmap like this
+// data: |
+//  key1=value1
+//  key2=value2
+
+func (fg *FeatureGates) GetFeatureGateParams(ctx context.Context, feature string) map[string]string {
```
snir911, line 37: Should we bind between featureGates to parameter CMs? you think simply making it generic (e.g. GetParamsCM) would be useful also for existing features or things that won't be necessarily under featuregate

internal/featuregates/features.go
```diff
@@ -0,0 +1,59 @@
...
+- feature gate is only for experimental features
+- Keep the feature gate as simple boolean.
+- Any config specific for a feature gate should be in it’s own configMap. This aligns with our current implementation of peer-pods and image generator feature
+- When we decide to make the feature as a stable feature, it should move to kataConfig.spec
+*/
+
+const (
+	FeatureGatesConfigMapName     = "osc-feature-gates"
+	ImageBasedDeployment          = "imageBasedDeployment"
+	ImageBasedDeploymentConfigMap = "osc-feature-gate-image-deploy-config"
```
snir911: I'd remove "feature-gate" from the name, i don't see why it needs to be mentioned
also i wonder if we can find a name for this feature which will be less confusing with podvm image creation stuff, maybe `layered-image-deployment` or something WDYT?

controllers/cm_event_handler.go
```diff
@@ -0,0 +1,59 @@
+package controllers
```
snir911, line 1: Earlier reconcile loop was called with CM?

internal/featuregates/features.go
bpradipt: @beraldoleal sorry I didn't understand the message "changing the approach a "little bit"...". Can you please clarify ?

beraldoleal: Sorry @bpradipt, I just took a second look and it seems you are just setting the config per feature, so please ignore my comment here.

internal/featuregates/features.go
```diff
@@ -0,0 +1,74 @@
...
+- When we decide to make the feature as a stable feature, it should move to kataConfig.spec
+*/
+
+const (
+	FeatureGatesConfigMapName         = "osc-feature-gates"
+	ImageBasedDeployment              = "imageBasedDeployment"
+	ImageBasedDeploymentConfigMap     = "osc-feature-gate-image-deploy-config"
+	AdditionalRuntimeClasses          = "additionalRuntimeClasses"
+	AdditionalRuntimeClassesConfigMap = "osc-feature-gate-additional-rc-config"
+)
```
beraldoleal, line 18: Any reason for using "osc-feature-gate-" prefix ? I'm wondering if we couldn't just use the convention: FeatureName + '-config' for all features. That way we would avoid the need of most of this code here.

internal/featuregates/features.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Sample ConfigMap with Features
```
beraldoleal, line 20: Maybe moving this comment into `config/samples` folder?

internal/featuregates/features.go
bpradipt: With some assumptions let me try to answer it..
1. The reason for using package is for better code organisation. Since earlier what we had were just dummy features there was not enough thinking about how the feature gates will be used, how state management will happen etc. Now that real feature is being implemented, I took the liberty to organise the code in a modular fashion. I can rework to split this commit into two, but not sure if it's really needed. But if you insist I'll do it. 
2. There is a different logic to figure out the status of featuregates in the reconcile loop. The earlier logic only handled single feature and reading configmap  repeatedly for each individual feature is sub-optimal in the reconcile loop. Hence I use a map for the same. Let me know if this helps. I'll add this in the commit message as well

internal/featuregates/features.go
bpradipt: It helps to know which configmaps are for what by just looking at the names. May be I can just use `fg` instead of `feature-gate`. 
I'll rename the feature gate to layered image deployment. It makes sense

controllers/cm_event_handler.go
bpradipt: Yes. So had to use an explicit event handler. Similar to how node events are handled

internal/featuregates/features.go
bpradipt: See the response here - https://github.com/openshift/sandboxed-containers-operator/pull/394#discussion_r1574991356
The only reason was to make it easier for admins to know by just looking at the name when listing the configmaps.
May be I'll use `fg` instead of the full name.

internal/featuregates/features.go
bpradipt: This is for someone later on looking at the code to know how it's being used.  I can add a dummy config file to the config/samples as well. But typically config/samples is for CRD. May be it's ok to add additional examples there

internal/featuregates/featuregates.go
bpradipt: Since this is featuregate specific it's better to be this way. Also gives us the flexibility to modify the implementation with changes to feature gate implementations. If a generic function makes sense then it should go in generic util imho.

internal/featuregates/features.go
bpradipt: @beraldoleal @snir911 let me know if you would prefer to use FeatureName + '-config' and I'll change the code accordingly.

beraldoleal: I agree that we should easily identify what the ConfigMap is for. Either the "fg" prefix/suffix or a common namespace would help with that. But I would also try (if possible) to avoid having those hard coded for each feature and eventually remove the need of most of the code here.

So if possible, I would give a try with something like FeatureName + "-fg-cfg/-fg-config" suffix as a convention without the need of hardcode it.

internal/featuregates/features.go
beraldoleal: Hi @bpradipt, improving the existing code is definitely the way to go here.

I recommend splitting the changes into separate commits. Merging them into another feature might cause some reviewers to miss them. I won't "insist", but it would be really helpful, if you don't mind.

bpradipt:
@beraldoleal @snir911 I have addressed all your comments. PTAL

internal/featuregates/featuregates.go
```diff
@@ -42,3 +50,35 @@ func (fg *FeatureGates) IsEnabled(ctx context.Context, feature string) bool {
...
+	featureGateStatus := make(FeatureGateStatus)
+
+	// Read the configmap to get the status of all the feature gates and populate the status struct
+	// Ignore key with feature.params as these are the feature gate specific config params
+	cfgMap := &corev1.ConfigMap{}
+	err := fg.Client.Get(ctx,
+		client.ObjectKey{Name: fg.ConfigMapName, Namespace: fg.Namespace},
+		cfgMap)
+	if err != nil {
+		fgLogger.Info("Error fetching feature gates", "err", err)
```
beraldoleal, line 70: fgLogger.Error ?

internal/featuregates/featuregates.go
```diff
@@ -42,3 +50,35 @@ func (fg *FeatureGates) IsEnabled(ctx context.Context, feature string) bool {
...
+		if _, exists := featureGateStatus[key]; !exists {
+			featureGateStatus[key] = value
+		}
+	}
+
+	fgLogger.Info("Feature gate status", "status", featureGateStatus)
+
+	return featureGateStatus
+
+}
```
beraldoleal, line 89: This method is indeed handy, thanks for the addition. Unless I'm missing something, it seems to have a lot of duplicated code from `IsEnabled`. Any reason for not re-using it and concentrate the logic there? Something like this:

```
    for feature := range DefaultFeatureGates {
        featureGateStatus[feature] = fg.IsEnabled(ctx, feature)
      }
```

internal/featuregates/featuregates.go
bpradipt: As a convention we avoid using Error as it throws a stack trace which is not that useful.

internal/featuregates/featuregates.go
bpradipt: Yes.. IsEnabled will do a Get (to K8s API server) for every featuregate and is suboptimal. In this method there is a single Get and then reading the configmap to populate the status map.

beraldoleal: Hum I got it. And that was what I was missing. thanks.

It concerns me that we have two methods with the "same logic", and we should avoid one. Since IsEnabled is not used anymore and we are trying to avoid it, what about just drop it?

beraldoleal (commented):
Hi @bpradipt thanks for the patience and splitting the commits. From the Feature Gates PoV, lgtm.

internal/featuregates/featuregates.go
bpradipt: Yeah you are right. I have removed it.

controllers/openshift_controller.go
```diff
@@ -63,8 +63,9 @@ type KataConfigOpenShiftReconciler struct {
 	Log    logr.Logger
 	Scheme *runtime.Scheme
 
-	kataConfig   *kataconfigurationv1.KataConfig
-	FeatureGates *featuregates.FeatureGates
+	kataConfig         *kataconfigurationv1.KataConfig
+	FeatureGates       *featuregates.FeatureGates
```
pmores, line 61: Do we need this structure? Seeing as it contains two constant values and Client which is the same value as `KataConfigOpenShiftReconciler::Client` it looks kind of superfluous to me.

I'm asking because I'm a bit worried about bloating `KataConfigOpenShiftReconciler`.

internal/featuregates/featuregates.go
```diff
@@ -13,12 +15,17 @@ type FeatureGates struct {
 	ConfigMapName string
 }
 
+// FeatureGate Status Struct map of string and bool
+type FeatureGateStatus map[string]bool
+
 var DefaultFeatureGates = map[string]bool{
```
pmores: I think it would make sense if `DefaultFeatureGates` was of the `FeatureGateStatus` type, defined just above, wouldn't it?

controllers/openshift_controller.go
bpradipt: Right now it's two features. But the number of features also going to increase, although I don't expect it to be too many features. Also currently I'm not sure if there won't be any new members to the struct.
I'm open for suggestions. Let me know if you prefer an alternative way.

internal/featuregates/featuregates.go
```diff
@@ -42,3 +50,35 @@ func (fg *FeatureGates) IsEnabled(ctx context.Context, feature string) bool {
...
+			featureGateStatus[key] = value == "true"
+		}
+	}
+
+	// Populate the status struct with the default values for the feature gates which are not present in the configmap
+	for key, value := range DefaultFeatureGates {
+		if _, exists := featureGateStatus[key]; !exists {
+			featureGateStatus[key] = value
+		}
+	}
```
pmores, line 83: Frankly it would seems more natural/idiomatic to me if we initialised `featureGateStatus` with `DefaultFeatureGates`  values in the first place, rather than another loop to fill in defaults when necessary after the fact.

internal/featuregates/featuregates.go
bpradipt: Right.

controllers/mc_handler.go
```diff
@@ -0,0 +1,153 @@
...
+
+const (
+	extension_mc_name = "50-enable-sandboxed-containers-extension"
+	image_mc_name     = "50-enable-sandboxed-containers-image"
+)
+
+// If the first return value is 'true' it means that the MC was just created
+// by this call, 'false' means that it's already existed.  As usual, the first
+// return value is only valid if the second one is nil.
+func (r *KataConfigOpenShiftReconciler) createImageMc(machinePool string) (bool, error) {
```
pmores: I can see nothing image-specific in this function - I think it could be called just `createMc()`.   Whether the extension or image variant, seems to be only decided in  `newMCForCR()` below.

controllers/openshift_controller.go
```diff
@@ -1035,9 +992,23 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
-	if err != nil {
-		return ctrl.Result{Requeue: true}, nil
+	// Based on LayeredImageDeployment feature, we'll either create an ExtensionMachineConfig or
+	// just apply the image
+	wasMcJustCreated := false
+
+	// If featuregate LayeredImageDeployment is enabled then create the image based MC
+	// TBD: Handling switching from image based deployment to extension and vice versa
+
+	if r.FeatureGatesStatus[featuregates.LayeredImageDeployment] {
```
pmores, line 952: This feature gate testing seems to duplicate the one in `newMCForCR()`.  Furthermore, `createImageMc()` and `createExtensionMc()` seem to duplicate quite a bit of code.  I'd suggest that we call just a `createMc()` function here and let its implementation sort out the variant.

controllers/cm_event_handler.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	log := ch.reconciler.Log.WithName("CMUpdate").WithValues("cm name", cm.GetName())
+	log.Info("FeatureGates configmap updated")
+
+	queue.Add(ch.reconciler.makeReconcileRequest())
+
+}
+
+func (ch *ConfigMapEventHandler) Delete(ctx context.Context, event event.DeleteEvent, queue workqueue.RateLimitingInterface) {
```
pmores, line 55: If we reconcile on Create I guess it would make sense to also reconcile on Delete, wouldn't it?  I understand handling Create since we want to handle creation of both the feature gate master configmap and individual gated features' configmaps.  With the same reasoning though, wouldn't we want handle their deletion as well?

internal/featuregates/features.go
```diff
@@ -45,3 +46,16 @@ func GetFeatureGateConfigMapName(feature string) string {
...
 
 }
+
+// Check if the configmap is a feature gate configmap
+// We use explicit check for the feature gate configmap
+// and avoid reconstructing the feature name from configmap name
+// and do the match
+func IsFeatureGateConfigMap(configMapName string) bool {
+	switch configMapName {
+	case FeatureGatesConfigMapName, LayeredImageDeploymentConfig:
```
pmores: It could be worth the effort to find a more dynamic approach here.  Otherwise we're risking bugs due to forgetting to update this function, I'm afraid.

controllers/openshift_controller.go
```diff
@@ -631,60 +631,64 @@ func (r *KataConfigOpenShiftReconciler) createScc() error {
...
-					corev1.ResourceCPU:    resource.MustParse(cpuOverhead),
-					corev1.ResourceMemory: resource.MustParse(memoryOverhead),
+type RuntimeClassConfig struct {
+	ClassName      string
+	CPUOverhead    string
+	MemoryOverhead string
+}
+
+func (r *KataConfigOpenShiftReconciler) createRuntimeClasses(runtimeClassConfigs []RuntimeClassConfig) error {
+	for _, config := range runtimeClassConfigs {
```
pmores: Frankly, I'd prefer if this loop just called the original `createRuntimeClass()` function rather than inline it here...

controllers/openshift_controller.go
bpradipt: Yeah makes sense

controllers/cm_event_handler.go
bpradipt: How do you think the deletion of configmaps be handled?  Everything roll backed ? Or the configmap being recreated based on known state?
My thinking was to only handle create/update and document that on deletion the behaviour is not deterministic. And handle these cases when feature gate graduates to stable and part of KataConfig.

internal/featuregates/features.go
bpradipt: I would prefer to keep it static here and document it in the code. It's possible to reconstruct by checking if configmap has the suffix `-config` and then string match the prefix with the featuregates. It's documented in the code as well so at least a minimal expectation from a new implementation ;)

internal/featuregates/featuregates.go.new
```diff
@@ -0,0 +1,106 @@
+package featuregates
```
snir911: Accidentally committed ?

controllers/openshift_controller.go
```diff
@@ -125,9 +116,8 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
```
snir911, line 117: nit, maybe worth mentioning in the comment above (under `if  k8serrors.IsNotFound(err) `) aso the new case it may not be found (feguregate CM modified before KataConfig creation)

controllers/cm_event_handler.go
pmores: Let me start by stating the wider context so that we can be sure we're on the same page.

I assume that we're only watching configmaps because we want to support modifications on-the-fly (ie. while a KataConfig exists).  If we didn't we'd just sample the configmaps during installation and then no further modification would be possible.

I believe it would then make sense to support all such modifications rather than a (pretty much arbitrary) subset.  I mean, we _could_ say "you can enable a gated feature subsequently if you didn't enable it initially but once you do that you're stuck with it, no disabling possible" but does it make much sense?

I'm not sure what you mean when you mention non-deterministic behaviour.  The feature gates are booleans so they fundamentally switch between two states, both of which necessarily need to be well defined.  If you delete a configmap that enabled both images and additional runtimeclasses I imagine that the controller should deterministically respond by switching back to extension and by deleting the extra runtimeclasses.

bpradipt: > if you didn't enable it initially but once you do that you're stuck with it, no disabling possible" but does it make much sense?

What we say is you disable it by turning it off in the configmap and not by deleting the configmap.

I'm not comfortable to do a complete rollback for all extensions to default values if the configmap is deleted. IOW switching to extension, removing all runtimeClasses and other features we add going forward. 
Will be a heavy operation for say an accidental delete. Although one can argue that it's an admin operation and need to be done carefully.
Do you think this is how we should handle or there is an alternative?

pmores: The question is, do we even have a choice?  Cluster admin can delete the configmap as they see fit.  Our only choice that I can see is, do we handle the deletion or do we allow the cluster to stay in an inconsistent state?

It's true that depending on the gated feature set a complete rollback might be a heavy operation and it could be done in error.  But the same is true of just about any KataConfig operation, including creation, modification and deletion.

bpradipt: One approach I was thinking was to use the in-memory state of the featuregate to decide. But then it deviates from the using etcd as the source of truth. I'll do some experiments and update.

pmores: In-memory state wouldn't be reliable enough, the controller could be restarted at any time.  And yes, it could be rather confusing for cluster admin, too, if the controller is guided by its hidden in-memory state rather the visible cluster resources...

snir911: Is there a k8s way to "lock" the CM once it's created? or to block its deletion (finalizers?)? would it help in this case?

pmores: Adding a finalizer would indeed delay deletion, I'm afraid however that it would be quite a misuse (or even abuse) of finalizer.

bpradipt: Yeah adding finalizer will not give us anything here imho..
Anyways I have added some changes to handle deletion.. PTAL

controllers/mc_handler.go
```diff
@@ -0,0 +1,190 @@
...
+			Name: image_mc_name,
+			Labels: map[string]string{
+				"machineconfiguration.openshift.io/role": machinePool,
+				"app":                                    r.kataConfig.Name,
+			},
+			Namespace: OperatorNamespace,
+		},
+		Spec: mcfgv1.MachineConfigSpec{
+			OSImageURL:      imageParams["osImageURL"],
+			KernelArguments: []string{imageParams["kernelArguments"]},
```
gkurz: According to https://github.com/openshift/api/blob/cb0966a7db72bce9aaacea4a6650c59914c4f326/machineconfiguration/v1/types.go#L324, `kernelArguments contains a list of kernel arguments to be added`.

IIUC the cm can optionally provide something like `"a=b c=d ..."`. It seems that such a string should be parsed and turned into `[]string{"a=b", "c=d", ...}` rather than `[]string{"a=b c=d ..."}`.

Also if the cm doesn't have `kernelArguments`, no need to set it in the `MachineConfigSpec` since the field is marked as `+optional`.

What about making `kernelArguments` an optional array of string in the cm as well ? And we'd just need to copy it to the `MachineConfigSpec` if it exists.

controllers/mc_handler.go
```diff
@@ -0,0 +1,190 @@
...
+	return &mc, nil
+}
+
+// Create a new MachineConfig object for the Custom Resource with the image
+func (r *KataConfigOpenShiftReconciler) newImageMCForCR(machinePool string, icb []byte) (*mcfgv1.MachineConfig, error) {
+
+	// Get the LayeredImageDeployment feature gate parameters
+	imageParams := r.FeatureGates.GetFeatureGateParams(context.TODO(), featuregates.LayeredImageDeployment)
+	// Ensure at least osImageURL is set
+	if _, ok := imageParams["osImageURL"]; !ok {
```
gkurz, line 135: https://github.com/openshift/sandboxed-containers-operator/blob/e73f421617caf243568920f0e1509d67bbe62ccc/internal/featuregates/featuregates.go#L58 can return `nil`, e.g. cm doesn't exist, causing this line to crash.

bpradipt: yes.. This is fixed in my next version.

controllers/rc_handler.go
```diff
@@ -0,0 +1,385 @@
...
+//
+// or
+// runtimeClassConfig="name1, name2"
+func (r *KataConfigOpenShiftReconciler) getRuntimeClassConfigs() []RuntimeClassConfig {
+	runtimeClassConfigs := []RuntimeClassConfig{}
+
+	// Get the AdditionalRuntimeClasses feature gate parameters
+	additionalRuntimeClassesParams := r.FeatureGates.GetFeatureGateParams(context.TODO(), featuregates.AdditionalRuntimeClasses)
+	// If the runtimeClassConfig key is not present, return empty runtimeClassConfig
+	if _, ok := additionalRuntimeClassesParams["runtimeClassConfig"]; !ok {
```
gkurz, line 87: Same issue as reported in https://github.com/openshift/sandboxed-containers-operator/pull/394/files#r1579489744.

controllers/mc_handler.go
gkurz: Ok a CM doesn't seem to allow that... only structured API I can think of is CRD. Would it be possible for each feature to be described with its own CRD ?

controllers/mc_handler.go
bpradipt: Oh btw this wont' crash in golang - https://go.dev/ref/spec#Index_expressions.
However I have changed the method in the next iteration to return empty map.

controllers/mc_handler.go
bpradipt: Right. It's not possible to have this with CM. So for now only the single string containing the entire kernel argument string needs to be provided and it gets used as-is.
CRDs are best imho for structured info.  Once we stablise the features and move it to KataConfig CRD (with versioning of course) we can have a full-fledged struct to handle these better.

controllers/rc_handler.go
bpradipt: Addressed in previous comment.

controllers/mc_handler.go
bpradipt: May be another approach (for now) could be to use a machineconfig yaml and embed it in the configmap. So for the admin the requirement will be to create a machineconfig yaml and then create the configmap out of the machineconfig yaml. We can't allow the admin user to apply the machineconfig directly as the operator needs to control the application and rollback.

controllers/openshift_controller.go
bpradipt: Changed the flow a bit. The status CM is created by operator to start with.

controllers/mc_handler.go
gkurz: > Right. It's not possible to have this with CM. So for now only the single string containing the entire kernel argument string needs to be provided and it gets used as-is.

Does this work ? If so, I'm wondering why `KernelArguments` is `string[]` and not just a `string`. In case this works by chance, maybe worth having some `// FIXME` or `// HACK` comment ?

> CRDs are best imho for structured info. Once we stablise the features and move it to KataConfig CRD (with versioning of course) we can have a full-fledged struct to handle these better.

Yeah in the end it will be proper CRD. I was just wondering if it would be help to introduce them right away.

gkurz: > May be another approach (for now) could be to use a machineconfig yaml and embed it in the configmap. So for the admin the requirement will be to create a machineconfig yaml and then create the configmap out of the machineconfig yaml. We can't allow the admin user to apply the machineconfig directly as the operator needs to control the application and rollback.

Yes this would allow the user to set the kernel arguments correctly.

bpradipt: > Does this work ? If so, I'm wondering why KernelArguments is string[] and not just a string. In case this works by chance, maybe worth having some // FIXME or // HACK comment ?

Yeah it works.  The configmap value is a string. And currently I take the value and put it as the only entry for string[]. I'll add a //FIXME and indicate the current limitation. 
For example
```
apiVersion: v1
data:
  kernelArguments: "kvm_intel.tdx=1 rhgb quiet"
  osImageURL: quay.io/bpradipt/coco:ocp415
kind: ConfigMap
```
And "kvm_intel.tdx=1 rhgb quiet" is the only element when creating the machineConfig struct

bpradipt: Fixed. Now the elements are converted to an array and passed.

controllers/openshift_controller.go
```diff
@@ -66,14 +64,14 @@ type KataConfigOpenShiftReconciler struct {
 	kataConfig         *kataconfigurationv1.KataConfig
 	FeatureGates       *featuregates.FeatureGates
 	FeatureGatesStatus featuregates.FeatureGateStatus
+	deploymentState    DeploymentState
```
pmores, line 63: I'm quite afraid this breaks the controller's idempotency.  The controller can be migrated to another worker at any time and if I'm not mistaken this will wipe the deployment state information.  This is not a theoretical concern either.  The controller normally runs on workers so any worker reboot is an opportunity for the controller to be killed and restarted elsewhere.  I've seen many times during kata deployment on all workers of a two-worker cluster the controller being migrated twice (ie. the number of workers).  This happens when the controller happens to be running on the first worker to get kata deployed.  It's then migrated to the second worker, only to be migrated back to the first worker when the second worker has kata deployed on it subsequently.

Instead of storing the deployment state here, we should ideally figure out the deployment state by examining the cluster.  In case that's impossible or impractical (I don't actually know if that's possible but I suspect it might be) the state should still be stored somewhere in the cluster, perhaps in `KataConfig.status`.

controllers/openshift_controller.go
pmores: My point is that `FeatureGates` is just the Client which is already stored here in `KataConfigOpenShiftReconciler` and some constant values which could be just constants.  So I don't see the utility of this new `KataConfigOpenShiftReconciler` member.

internal/featuregates/featuregates.go
```diff
@@ -19,7 +19,7 @@ type FeatureGates struct {
 // FeatureGate Status Struct map of string and bool
 type FeatureGateStatus map[string]bool
 
-var DefaultFeatureGates = map[string]bool{
+var DefaultFeatureGatesStatus = map[string]bool{
```
pmores, line 24: Cool, was just about to propose this change! :-)

Though this should be squashed to a previous commit, really.

controllers/openshift_controller.go
```diff
@@ -119,6 +119,14 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 	// Check for enabled FeatureGates by retrieving the FeatureGateStatus
 	r.FeatureGatesStatus = r.FeatureGates.GetFeatureGateStatus(ctx)
 
+	// If FeatureGateStatus is nil, we cannot proceed. Log the error and reconcile
+	// This ensures we don't take any action if we cannot determine the FeatureGateStatus
+	// Which could be due to an error in fetching the ConfigMap or the ConfigMap not being present
+	if r.FeatureGatesStatus == nil {
+		r.Log.Info("FeatureGateStatus is nil, cannot proceed. Check if the feature gate configmap is present")
+		return ctrl.Result{}, nil
```
pmores, line 128: This was discussed elsewhere so here just briefly for public record: I believe that just continuing here as if a default-valued feature gate configmap was present would lead to both simpler implementation and better user experience.

controllers/image_generator.go
```diff
@@ -242,7 +243,9 @@ func newImageGenerator(client client.Client) (*ImageGenerator, error) {
 func (r *ImageGenerator) createJobFromFile(jobFileName string) (*batchv1.Job, error) {
 	igLogger.Info("Create Job out of YAML file", "jobFileName", jobFileName)
 
-	yamlData, err := readJobYAML(jobFileName)
+	jobYamlFile := filepath.Join(peerpodsImageJobsPathLocation, jobFileName)
+
+	yamlData, err := readYamlFile(jobYamlFile)
```
pmores, line 248: Perhaps I'm confused but I don't see how this is related to feature gating. (?)

internal/featuregates/features.go
```diff
@@ -10,11 +10,11 @@ import "strings"
 */
 
 const (
-	FeatureGatesConfigMapName      = "osc-feature-gates"
-	LayeredImageDeployment         = "LayeredImageDeployment"
-	LayeredImageDeploymentConfig   = "layeredimagedeployment-config"
-	AdditionalRuntimeClasses       = "AdditionalRuntimeClasses"
-	AdditionalRuntimeClassesConfig = "additionalruntimeclasses-config"
+	FeatureGatesStatusConfigMapName = "osc-feature-gates"
```
pmores, line 13: Again, a good change but should be squashed with the commit where the original name was first introduced.

controllers/openshift_controller.go
```diff
@@ -888,6 +888,13 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	// Revert the feature gate status configmap to default
+	err = r.FeatureGates.RevertFeatureGateStatusConfigMapToDefault(context.TODO())
```
pmores, line 892: I believe it would be cleaner, and simpler too, to just remove the cm here.  As I said elsewhere, I'm far from convinced that we should insist on this configmap's existence in the first place, but it definitely serves no purpose when there's even no KataConfig instance, right?

controllers/openshift_controller.go
```diff
@@ -1035,9 +942,75 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
+			err := fmt.Errorf("LayeredImageDeployment feature is enabled but no configuration provided")
+			return ctrl.Result{Requeue: true}, err
+		}
+
+		// Check if the previous deployment method was different
+		// When we first time enter this condition the PreviousMethod will be empty
+		if r.deploymentState.PreviousMethod != "image" {
+			// Rollback the previous deployment
+			if r.deploymentState.PreviousMcName != "" {
+				// Delete the previous MachineConfig
```
gkurz, line 968: It seems that we should just check for the existence of the `extension_mc_name`  MC and delete it if it exists.

controllers/openshift_controller.go
```diff
@@ -1035,9 +942,75 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
+
+	} else {
+		r.Log.Info("PreviousMcName and PreviousMethod, ImageBasedDeployment=false",
+			"PreviousMcName", r.deploymentState.PreviousMcName, "PreviousMethod", r.deploymentState.PreviousMethod)
+		// Check if the previous deployment method was different
+		// When we first time enter this condition the PreviousMethod will be empty
+		if r.deploymentState.PreviousMethod != "extension" {
+			// Rollback the previous deployment
+			if r.deploymentState.PreviousMcName != "" {
+				// Delete the previous MachineConfig
```
gkurz, line 996: Same with `image_mc_name` MC.

controllers/openshift_controller.go
```diff
@@ -1035,9 +993,75 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
+				err := r.deleteMc(r.deploymentState.PreviousMcName)
+				if err != nil {
+					r.Log.Info("Error deleting previous MachineConfig", "MachineConfig",
+						r.deploymentState.PreviousMcName, "err", err)
+					return ctrl.Result{Requeue: true}, err
+				}
+			}
+			wasMcJustCreated, err = r.createMc(machinePool)
```
pmores, line 976: I'm not quite sure admittedly but is the case handled where the user stays with the image-based deployment mechanism _but_ has changed the image URL?  Because I suspect it's not.

controllers/openshift_controller.go
```diff
@@ -1035,9 +993,75 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
+			if err != nil {
+				r.Log.Info("Error creating image MachineConfig for Image based deployment", "err", err)
+				return ctrl.Result{Requeue: true}, err
+			}
```
pmores, line 980: Also, is it guaranteed that if we delete the previous MC and create the new one right away, the MCO will handle this correctly and sequence the operations properly, is the result well-defined?  If not or if we can't tell we'd better return and reschedule after the deletion and only create the new MC once we can verify that the operation triggered by the deletion of the previous one has finished.

controllers/openshift_controller.go
pmores: > the state should still be stored somewhere in the cluster, perhaps in `KataConfig.status`

Actually, no, `status` is most likely no good for this either since as per k8s principles the pertinent resource controller is expected to recover even if its resource's `status` subobject is lost.

controllers/openshift_controller.go
beraldoleal: Yes, I'm also not sure why do we need to update the user's CM? This might look strange from the user's PoV having some settings changed. We should not mix default values and user's values. Whoever create it shouldn't be the one responsible for deleting?

Not a k8s expert, and I was not following all the discussions here closely, so I'm probably missing something.

pmores: That is a very good point, we should avoid muddying the notion of resource ownership to avoid problems like we had a while ago with `KataConfig.spec.kataMonitorImage` which was basically an OSC-owned field within a user-owner `KataConfig` instance.

However, my takeaway from the latest round of discussions is that we're now leaning towards the controller _not_ requiring the existence of any of the feature-gating configmaps.  This solves the ownership issue quite neatly by putting the configmap resources wholly under user control.

bpradipt:
Based on internal discussions I'll close this PR and split this into separate PRs.

PR #396, opened 2024-04-08
Mass revert 1.5.3 main

gkurz:

This does a couple of reverts to bring main back to the same state as the v1.5.2 tag. A release branch will be forked from there to push all the 1.5.3 specific fixes.

Fixes: [KATA-2876](https://issues.redhat.com//browse/KATA-2876)

**- What I did**

Reverted all merges commit in main since v1.5.2.

**- How to verify it**

- Checkout this PR
- No output expected with `git diff v1.5.2`


littlejawa (approved):
lgtm
Thanks @gkurz

PR #403, opened 2024-04-10
Fix stale doc links for 1.5.3

gkurz:

OSC documentation has moved out of OCP. Fix the links in the repo.

Fixes: [KATA-2523](https://issues.redhat.com//browse/KATA-2523)

**- What I did**

Fixed the links everywhere

**- How to verify it**

- Checkout the PR
- `git grep docs.openshift.com` is empty
- Verify links in `git grep /access.redhat.com` are valid


bpradipt (approved):
/lgtm
Tested the links and see it points to 1.5

PR #404, opened 2024-05-11
Pod VM image handling fix

bpradipt:
Fixes: #[KATA-2958](https://issues.redhat.com//browse/KATA-2958) #[KATA-2962](https://issues.redhat.com//browse/KATA-2962)

### DIFF
--- a/config/peerpods/podvm/azure-podvm-image-handler.sh
+++ b/config/peerpods/podvm/azure-podvm-image-handler.sh
@@ -637,22 +637,6 @@ function delete_image_gallery() {
     echo "Azure image gallery deleted successfully"
 }
 
-# Function to delete the image from Azure given the image name
-# Resource group is must
-# Input is of the form /subscriptions/<subscription-id>/resourceGroups/<resource-group>/providers/Microsoft.Compute/images/<image-name>
-
-function delete_image_using_name() {
-    echo "Deleting Azure image"
-    # If any error occurs, exit the script with an error message
-
-    # Delete the image
-    az image delete --resource-group "${AZURE_RESOURCE_GROUP}" \
-        --name "${IMAGE_NAME}" ||
-        error_exit "Failed to delete the image"
-
-    echo "Azure image deleted successfully"
-}
-
 # Function to delete the image from Azure given the image id
 # Input is of the form
 # /subscriptions/<subscription-id>/resourceGroups/<resource-group>/providers/Microsoft.Compute/galleries/<gallery-name>/images/<image-name>/versions/<image-version>
--- a/config/peerpods/podvm/azure-podvm-image-handler.sh
+++ b/config/peerpods/podvm/azure-podvm-image-handler.sh
@@ -654,8 +654,8 @@ function delete_image_using_name() {
 }
 
 # Function to delete the image from Azure given the image id
-# Input is of the form /subscriptions/<subscription-id>/resourceGroups/<resource-group>/providers/Microsoft.Compute/images/<image-name>
-# or /subscriptions/<subscription-id>/resourceGroups/<resource-group>/providers/Microsoft.Compute/galleries/<gallery-name>/images/<image-name>/versions/<image-version>
+# Input is of the form
+# /subscriptions/<subscription-id>/resourceGroups/<resource-group>/providers/Microsoft.Compute/galleries/<gallery-name>/images/<image-name>/versions/<image-version>
 
 function delete_image_using_id() {
     echo "Deleting Azure image"
@@ -665,7 +665,7 @@ function delete_image_using_id() {
     [[ -z "${IMAGE_ID}" ]] && error_exit "IMAGE_ID is empty"
 
     # Delete the image
-    az image delete --ids "${IMAGE_ID}" ||
+    az sig image-version delete --ids "${IMAGE_ID}" ||
         error_exit "Failed to delete the image"
 
     # Remove the image id annotation from peer-pods-cm configmap
--- a/config/peerpods/podvm/azure-podvm-image-handler.sh
+++ b/config/peerpods/podvm/azure-podvm-image-handler.sh
@@ -47,7 +47,9 @@ function verify_vars() {
     # Ensure that the image defintion variables are set
     [[ -z "${IMAGE_DEFINITION_PUBLISHER}" ]] && error_exit "IMAGE_DEFINITION_PUBLISHER is empty"
     [[ -z "${IMAGE_DEFINITION_OFFER}" ]] && error_exit "IMAGE_DEFINITION_OFFER is empty"
+
     [[ -z "${IMAGE_GALLERY_NAME}" ]] && error_exit "IMAGE_GALLERY_NAME is empty"
+
     [[ -z "${IMAGE_DEFINITION_SKU}" ]] && error_exit "IMAGE_DEFINITION_SKU is empty"
     [[ -z "${IMAGE_DEFINITION_OS_TYPE}" ]] && error_exit "IMAGE_DEFINITION_OS_TYPE is empty"
     [[ -z "${IMAGE_DEFINITION_OS_STATE}" ]] && error_exit "IMAGE_DEFINITION_OS_STATE is empty"
@@ -155,9 +157,14 @@ function create_image_gallery() {
     # If any error occurs, exit the script with an error message
 
     # Create the image gallery
+    echo "Creating image gallery ${IMAGE_GALLERY_NAME}"
+
     az sig create --resource-group "${AZURE_RESOURCE_GROUP}" \
         --gallery-name "${IMAGE_GALLERY_NAME}" ||
         error_exit "Failed to create Azure image gallery"
+
+    # Update peer-pods-cm configmap with the gallery name
+    add_image_gallery_annotation_to_peer_pods_cm
 
     echo "Azure image gallery created successfully"
 
@@ -361,15 +368,6 @@ function create_or_update_image_configmap() {
         IMAGE_ID_LIST="${IMAGE_ID}"
     fi
 
-    # Create or update the value of the azure key in podvm-images configmap with all the images
-    # If any error occurs, exit the script with an error message
-    kubectl create configmap podvm-images \
-        -n openshift-sandboxed-containers-operator \
-        --from-literal=azure="${IMAGE_ID_LIST}" \
-        --dry-run=client -o yaml |
-        kubectl apply -f - ||
-        error_exit "Failed to create or update podvm-images configmap"
-
     echo "podvm-images configmap created or updated successfully"
 }
 
@@ -397,7 +395,7 @@ function recreate_image_configmap() {
 # Function to add the image id as annotation in the peer-pods-cm configmap
 
 function add_image_id_annotation_to_peer_pods_cm() {
-    echo "Adding image id to peer-pods-cm configmap"
+    echo "Adding image id annotation to peer-pods-cm configmap"
 
     # Check if the peer-pods-cm configmap exists
     if ! kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator >/dev/null 2>&1; then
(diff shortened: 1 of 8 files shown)


controllers/image_generator.go
```diff
@@ -640,6 +647,26 @@ func (r *ImageGenerator) createImageConfigMapFromFile() error {
...
+	// The IMAGE_GALLERY_NAME is set to IMAGE_GALLERY_NAME_PREFIX + "_" + clusterID
+	// Also add tags to IMAGE_GALLERY_TAGS
+	if r.provider == AzureProvider {
+		if cm.Data["IMAGE_GALLERY_NAME"] == "" {
+			image_gallery_name := cm.Data["IMAGE_GALLERY_NAME_PREFIX"] + "_" + r.clusterId
+			cm.Data["IMAGE_GALLERY_NAME"] = image_gallery_name
+			igLogger.Info("Setting IMAGE_GALLERY_NAME", "image_gallery_name", image_gallery_name)
+		}
+
+		// Add tags to IMAGE_GALLERY_TAGS
```
snir911: what is the use case of having multiple tags?

controllers/openshift_controller.go
```diff
@@ -90,7 +90,7 @@ const (
 // +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets;replicasets;statefulsets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apps,resources=daemonsets/finalizers,resourceNames=manager-role,verbs=update
 // +kubebuilder:rbac:groups=node.k8s.io,resources=runtimeclasses,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusterversions,verbs=get;list;watch
```
snir911, line 89: note that it doesn't seems like manifests were regenerated in this PR (maybe these rules are already included)

controllers/utils.go
```diff
@@ -202,3 +202,15 @@ func getCloudProviderFromInfra(c client.Client) (string, error) {
...
+func getClusterID(c client.Client) (string, error) {
+	clusterVersion := &configv1.ClusterVersion{}
+	err := c.Get(context.TODO(), types.NamespacedName{Name: "version"}, clusterVersion)
+	if err != nil {
+		return "", err
+	}
+
+	// Return first 8 characters of the cluster id
+	return string(clusterVersion.Spec.ClusterID[:8]), nil
+}
```
snir911, line 194: Why not doing this directly in the scripts IIRC we do have kubectl/oc there available there

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -410,7 +410,64 @@ function add_image_id_annotation_to_peer_pods_cm() {
...
+    echo "Adding IMAGE_GALLERY_NAME annotation to peer-pods-cm configmap"
+
+    # Check if the peer-pods-cm configmap exists
+    if ! kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator >/dev/null 2>&1; then
+        echo "peer-pods-cm configmap does not exist. Skipping adding the IMAGE_GALLERY_NAME annotation"
+        return
+    fi
+
+    # Add IMAGE_GALLERY_NAME annotation to peer-pods-cm configmap
+    kubectl annotate configmap peer-pods-cm -n openshift-sandboxed-containers-operator \
```
snir911, line 445: can you explain the rational set the annotations, I think i may have missed something

controllers/image_generator.go
bpradipt: For OSC, we only need one tag (ie, clusterId). But then realised that end users might want to have specific tags like with all other resources for billing etc. And since I'm adding support for tags, felt it's better to have this flexibility

controllers/openshift_controller.go
bpradipt: Yes, the manifests file are now ignored. Take a look at .gitignore

controllers/utils.go
bpradipt: You are right, this can be done in the scripts as well. The reason for doing it here is to feed the input config for gallery (azure-podvm-image-cm) via the operator. The azure-podvm-image-cm is input only config and I consciously avoided updating it in the script.

config/peerpods/podvm/azure-podvm-image-handler.sh
bpradipt: The peer-pods-cm is the single source of truth for dynamically updated info. Hence it's used to store the image as well as the gallery name. The podvm-handling.md explains it as well.

controllers/image_generator.go
```diff
@@ -640,6 +647,26 @@ func (r *ImageGenerator) createImageConfigMapFromFile() error {
 		return err
 	}
 
+	// For Azure provider set the IMAGE_GALLERY_NAME if its empty
```
snir911, line 662: AFAIU it will reach this point only when the default CM is used (and created from file) and not when the use is providing the a CM which misses the IMAGE_GALLERY_NAME or set it to ""

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -180,12 +206,27 @@ function delete_podvm_image_gallery() {
...
+
+  # Get the IMAGE_GALLERY_NAME from the IMAGE_GALLERY_NAME annotation key in peer-pods-cm configmap
+  IMAGE_GALLERY_NAME=$(kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator -o jsonpath='{.metadata.annotations.IMAGE_GALLERY_NAME}')
+
+  # If IMAGE_GALLERY_NAME is not set, then exit
+  if [ -z "${IMAGE_GALLERY_NAME}" ]; then
+    echo "IMAGE_GALLERY_NAME is not set in peer-pods-cm. Skipping image gallery deletion"
+    exit 0
+  fi
+
```
snir911, line 218: Am i getting this right? Image gallery will get deleted only if was set by OSC, based on the annotation we set?
And the image itself will get deleted based on IMAGE_ID value (if not changed) right? however, we do have the image_id as annotation as well, would it make sense to use the same mechanism?

controllers/image_generator.go
bpradipt: right..

config/peerpods/podvm/podvm-builder.sh
bpradipt: The way to check if the image gallery was created by OSC is the tag. The annotation just mentions the image gallery being used. The value in the annotation could be a pre-existing gallery that is entered by the admin user in the azure-podvm-image-cm

snir911: Sorry, i don't get why the difference in implementation is needed, ie. why galleries are deleted based on the annotations values and images are deleted based on the CM key value (data)?

bpradipt: The reason being a gallery can hold multiple images which is not the case for an image. But you do have a point. I think having a tag for the image is also required to ensure we only delete the image created by OSC and not any external image

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -410,7 +410,64 @@ function add_image_id_annotation_to_peer_pods_cm() {
         "LATEST_IMAGE_ID=${IMAGE_ID}" ||
         error_exit "Failed to add the image id as annotation to peer-pods-cm configmap"
 
-    echo "Image id added as annotation to peer-pods-cm configmap successfully"
+    echo "Image id annotation added as annotation to peer-pods-cm configmap successfully"
```
gkurz: This line doesn't seem to require change since it already mentions the annotation or am I missing something ?

config/peerpods/podvm/podvm-builder.sh
bpradipt: Oh forgot to mention that the image is created by packer and the only way to add tag is after creation of the image. But I think as you suggested I can use annotation as the source of truth for OSC created resources and use the same logic for gallery and image.

config/peerpods/podvm/azure-podvm-image-handler.sh
bpradipt: Yeah..

config/peerpods/podvm/podvm-builder.sh
bpradipt: I have changed it to use the gallery value from annotation and avoid using tags completely.

snir911: > Oh forgot to mention that the image is created by packer and the only way to add tag is after creation of the image. But I think as you suggested I can use annotation as the source of truth for OSC created resources and use the same logic for gallery and image. 

oh i missed that, now it make sense, thanks

controllers/image_generator.go
snir911: So if the image CM exist already (637) shouldn't we make sure gallery image name is set? IIUC it will fail on gallery creation? is it ok to assume if user create he's own CM he's know he should set it

bpradipt: If user creates the CM, then user should set the gallery name. We'll update in the README or instructions which provides steps on manual creation of the image creation job (outside of the operator)

snir911 (approved):
LGTM 
Thanks!

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -87,9 +86,28 @@ function create_podvm_image() {
...
 # These are checked in individual cloud provider scripts and if not set, the script will exit
+# Accepts two optional arguments
+# -f : force delete the image
+# -g : delete the image gallery
 
 function delete_podvm_image() {
 
+  local max_args=3
+  if (("${#@}" > max_args)); then
+    error_exit "Too many arguments passed to delete_podvm_image (max ${max_args} allowed)"
```
gkurz: Why checking the number of arguments ?

bpradipt: This is to ensure the function is used in the intended way with the required number of arguments. Ideally I wanted to have a similar check for all the functions to ensure only the accepted number of arguments are passed but that's best suited for another PR.
If this doesn't make sense I'm happy to remove it.

bpradipt: @gkurz can we move ahead with this or you want this check to be removed ?

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -47,7 +47,10 @@ function verify_vars() {
     # Ensure that the image defintion variables are set
     [[ -z "${IMAGE_DEFINITION_PUBLISHER}" ]] && error_exit "IMAGE_DEFINITION_PUBLISHER is empty"
     [[ -z "${IMAGE_DEFINITION_OFFER}" ]] && error_exit "IMAGE_DEFINITION_OFFER is empty"
+
+    [[ -z "${IMAGE_GALLERY_NAME_PREFIX}" ]] && error_exit "IMAGE_GALLERY_NAME_PREFIX is empty"
```
gkurz: What's the motivation behind this check ? The variable isn't used in the script AFAICT (like most of the variables verified in this function).

bpradipt: Afaicr I added this to ensure when the script is executed via the operator (as part of job execution) it's always set

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -413,6 +412,63 @@ function add_image_id_annotation_to_peer_pods_cm() {
...
+        echo "peer-pods-cm configmap does not exist. Skipping deleting the image id"
+        return
+    fi
+
+    # Delete the image id annotation from peer-pods-cm configmap
+    kubectl annotate configmap peer-pods-cm -n openshift-sandboxed-containers-operator \
+        "LATEST_IMAGE_ID-" ||
+        error_exit "Failed to delete the image id annotation from peer-pods-cm configmap"
+
+    echo "Image id added as annotation to peer-pods-cm configmap successfully"
```
gkurz:
```suggestion
    echo "Image id annotation deleted from peer-pods-cm configmap successfully"
```

config/peerpods/podvm/azure-podvm-image-handler.sh
bpradipt: I went through this again and some of the older iterations of this code. I don't see a logical reason to keep this check anymore. Instead this variable is not needed and I can set the prefix directly in the controller code.

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -555,6 +612,7 @@ function delete_image_gallery() {
     # Check if the gallery has any image versions
     get_all_image_ids
 
+    # This will set the IMAGE_ID_LIST variable
```
gkurz: `IMAGE_ID_LIST` is set by `get_all_image_ids`. Maybe move the comment before the call to the function.

controllers/image_generator.go
```diff
@@ -644,6 +651,17 @@ func (r *ImageGenerator) createImageConfigMapFromFile() error {
...
+	// The IMAGE_GALLERY_NAME is set to IMAGE_GALLERY_NAME_PREFIX + "_" + clusterID
+	if r.provider == AzureProvider {
+		if cm.Data["IMAGE_GALLERY_NAME"] == "" {
+			image_gallery_name := cm.Data["IMAGE_GALLERY_NAME_PREFIX"] + "_" + r.clusterId
+			cm.Data["IMAGE_GALLERY_NAME"] = image_gallery_name
```
gkurz, line 667: Should we check that the generated name is valid ?

```
  # Note that the gallery name must be unique across the subscription and not exceed 80 characters.
  # Also, the allowed characters are English alphanumeric characters,
  # with underscores and periods allowed in the middle
```

config/peerpods/podvm/podvm-builder.sh
gkurz: FWIW usual scheme is to check that the function is passed arguments it understands rather than the number of arguments.

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -180,12 +206,27 @@ function delete_podvm_image_gallery() {
...
+    echo "IMAGE_GALLERY_NAME is not set in peer-pods-cm. Skipping image gallery deletion"
+    exit 0
+  fi
+
   if [ "$1" == "-f" ]; then
     /scripts/azure-podvm-image-handler.sh -G force
   else
     /scripts/azure-podvm-image-handler.sh -G
   fi
+
```
gkurz: nit: empty line

controllers/image_generator.go
bpradipt: Based on this comment https://github.com/openshift/sandboxed-containers-operator/pull/404#discussion_r1618405423
if I change the code to 
```
const osc_gallery_name_prefix = PodVMGallery
image_gallery_name :=  osc_gallery_name_prefix + "_" + r.clusterId
```
This will always be valid as clusterId is only the first 8 chars. So imho no need to recheck if the generated name is valid. Wdyt ?

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -87,9 +86,28 @@ function create_podvm_image() {
 # Function to delete podvm image
 # IMAGE_ID or AMI_ID is the input and expected to be set
 # These are checked in individual cloud provider scripts and if not set, the script will exit
+# Accepts two optional arguments
+# -f : force delete the image
+# -g : delete the image gallery
```
gkurz, line 91: Please add `[-g]` to the usage at https://github.com/openshift/sandboxed-containers-operator/pull/404/commits/8caf648d06893619d282ae13b4e085cbe97f82a1#diff-6544cd9ec466fe5e11123ce3f835bf520a39dd0e245eb676ccff7bebda9c4757R233.

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -666,7 +666,7 @@ function delete_image_using_id() {
     [[ -z "${IMAGE_ID}" ]] && error_exit "IMAGE_ID is empty"
 
     # Delete the image
-    az image delete --ids "${IMAGE_ID}" ||
+    az image-version delete --ids "${IMAGE_ID}" ||
```
gkurz: Providing some reference URLs to clarify this change would be appreciated (but not strongly required).

bpradipt: Looking at your comment made me realise a typo. The command should be `az sig image-version delete ..`. I have fixed it.

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -189,7 +224,7 @@ function delete_podvm_image_gallery() {
 }
 
 function display_usage() {
-  echo "Usage: $0 {create|delete [-f]|delete-gallery [-f]}"
+  echo "Usage: $0 {create|delete [-f]|delete-gallery [-g]}"
```
gkurz:
```suggestion
  echo "Usage: $0 {create|delete [-f] [-g]|delete-gallery [-f]}"
```

gkurz (approved):
/lgtm @bpradipt ! Just one last nit to fix in the usage string and we're good to go.

config/peerpods/podvm/podvm-builder.sh
bpradipt: Thanks.. Not sure what I was doing when making this change :-(

PR #406, opened 2024-05-14
Remove duplicate code for reading yamls

bpradipt:
We had different methods to read yaml files depending on the kubernetes object. However this is really not needed as all the methods simply were reading files with the only difference being the file location.
This commit simplifies the code and uses a single readYamlFile method. The caller should set the full path of the yaml file to be read. Further any additional requirements to read different yaml files will no longer need a new method

### DIFF
--- a/controllers/utils.go
+++ b/controllers/utils.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


littlejawa (approved):
lgtm
Thanks @bpradipt !

controllers/utils.go
```diff
@@ -73,9 +72,11 @@ func parseJobYAML(yamlData []byte) (*batchv1.Job, error) {
 	return job, nil
 }
 
-func readJobYAML(jobFileName string) ([]byte, error) {
-	jobFilePath := filepath.Join(peerpodsImageJobsPathLocation, jobFileName)
-	yamlData, err := os.ReadFile(jobFilePath)
+// Method to read yaml file.
+// The full path of the job yaml file is passed as an argument
```
gkurz:
```suggestion
// The full path of the yaml file is passed as an argument
```
since this is a generic method.

gkurz (approved):
/lgtm

Thanks @bpradipt !

PR #407, opened 2024-05-14
Add separate event handler for configmap updates

bpradipt:
With the previous code of predicate func, the reconcile loop was not receiving the KataConfig struct. Instead the reconcile loop was called with the OSC featuregate configmap.

The new method ensures that for featuregate configMap changes, the reconcile loop is called with the KataConfig struct.

### DIFF
--- a/controllers/cm_event_handler.go
+++ b/controllers/cm_event_handler.go
@@ -1,0 +1,82 @@
+package controllers
+
+import (
+	"context"
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+type ConfigMapEventHandler struct {
+	reconciler *KataConfigOpenShiftReconciler
+}
+
+func (ch *ConfigMapEventHandler) Create(ctx context.Context, event event.CreateEvent, queue workqueue.RateLimitingInterface) {
+
+	if ch.reconciler.kataConfig == nil {
+		return
+	}
+
+	cm := event.Object
+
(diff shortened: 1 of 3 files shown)


controllers/cm_event_handler.go
```diff
@@ -0,0 +1,70 @@
...
+	if ch.reconciler.kataConfig == nil {
+		return
+	}
+
+	cm := event.ObjectNew
+
+	// Check if the configMap name is relevant to the operator
+	if cm.GetNamespace() != OperatorNamespace || !isConfigMapRelevant(cm.GetName()) {
+		return
+	}
```
pmores, line 46: We might want to check here whether the `spec` subobjects of the old and new configmap versions actually differ, otherwise we end up reconciliating on spurious metadata or status changes.

bpradipt: I see..

pmores (approved):
Thanks @bpradipt , lgtm!  I'm leaving the `spec`s check on configmap update to your discretion.  While it would certainly be the correct thing to do it could also turn out in practice that changes to configmap status or metadata are so infrequent that the point is effectively moot.

bpradipt:
> Thanks @bpradipt , lgtm! I'm leaving the `spec`s check on configmap update to your discretion. While it would certainly be the correct thing to do it could also turn out in practice that changes to configmap status or metadata are so infrequent that the point is effectively moot.

@pmores I added the check as you suggested to avoid unnecessary reconciles. PTAL

beraldoleal (approved):
LGTM, thanks

gkurz (approved):
/lgtm

Thanks @bpradipt !

PR #408, opened 2024-05-16
Handle processing of a sample featuregate

bpradipt:
Few key things to note
- Use a single sample feature gate.
- Move the feature gate as a string constant.
- The feature gates are handled within the KataConfig reconcile loop.
- If there are any errors with feature gate processing, the reconciliation process continues and doesn't re-queue
- Feature gate processing, in case of any errors in determining whether it is enabled or not (for example if the configMap is deleted or some other API server errors), is same as the default compiled in state of the respective feature gate

Related to #[KATA-2947](https://issues.redhat.com//browse/KATA-2947)

### DIFF
--- a/controllers/fg_handler.go
+++ b/controllers/fg_handler.go
@@ -1,0 +1,52 @@
+package controllers
+
+import (
+	"github.com/openshift/sandboxed-containers-operator/internal/featuregates"
+)
+
+// Create enum to represent the state of the feature gates
+type FeatureGateState int
+
+const (
+	Enabled FeatureGateState = iota
+	Disabled
+)
+
+// Function to handle the feature gates
+func (r *KataConfigOpenShiftReconciler) processFeatureGates() error {
+
+	fgStatus, err := featuregates.NewFeatureGateStatus(r.Client)
+	if err != nil {
+		r.Log.Info("There were errors in getting feature gate status.", "err", err)
+		return err
+	}
+
+	// Check which feature gates are enabled in the FG ConfigMap and
+	// perform the necessary actions
+	// The feature gates are defined in internal/featuregates/featuregates.go
+	// and are fetched from the ConfigMap in the namespace
+	// Eg. TimeTravelFeatureGate
+
+	if featuregates.IsEnabled(fgStatus, featuregates.TimeTravelFeatureGate) {
+		r.Log.Info("Feature gate is enabled", "featuregate", featuregates.TimeTravelFeatureGate)
+		// Perform the necessary actions
+		r.handleTimeTravelFeature(Enabled)
+	} else {
+		r.Log.Info("Feature gate is disabled", "featuregate", featuregates.TimeTravelFeatureGate)
+		// Perform the necessary actions
+		r.handleTimeTravelFeature(Disabled)
+	}
+
+	return err
+
+}
+
+// Function to handle the TimeTravel feature gate
+func (r *KataConfigOpenShiftReconciler) handleTimeTravelFeature(state FeatureGateState) {
+	// Perform the necessary actions for the TimeTravel feature gate
+	if state == Enabled {
+		r.Log.Info("Starting TimeTravel")
+	} else {
+		r.Log.Info("Stopping TimeTravel")
+	}
+}
--- a/internal/featuregates/featuregates.go
+++ b/internal/featuregates/featuregates.go
@@ -2,43 +2,62 @@ package featuregates
 
 import (
 	"context"
+
 	corev1 "k8s.io/api/core/v1"
-	"log"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type FeatureGates struct {
-	Client        client.Client
-	Namespace     string
-	ConfigMapName string
-}
+const (
+	TimeTravelFeatureGate = "timeTravel"
+	FgConfigMapName       = "osc-feature-gates"
+	OperatorNamespace     = "openshift-sandboxed-containers-operator"
+)
 
 var DefaultFeatureGates = map[string]bool{
-	"timeTravel":              false,
-	"quantumEntanglementSync": false,
-	"autoHealingWithAI":       true,
+	"timeTravel": false,
 }
 
-func (fg *FeatureGates) IsEnabled(ctx context.Context, feature string) bool {
-	if fg == nil {
-		return false
-	}
-	cfgMap := &corev1.ConfigMap{}
-	err := fg.Client.Get(ctx,
-		client.ObjectKey{Name: fg.ConfigMapName, Namespace: fg.Namespace},
-		cfgMap)
+type FeatureGateStatus struct {
+	FeatureGates map[string]bool
+}
 
-	if err != nil {
-		log.Printf("Error fetching feature gates: %v", err)
-	} else {
-		if value, exists := cfgMap.Data[feature]; exists {
-			return value == "true"
+// This method returns a new FeatureGateStatus object
+// that contains the status of the feature gates
+// defined in the ConfigMap in the namespace
+// Return default values if the ConfigMap is not found.
+// Return values from the ConfigMap if the ConfigMap is not found. Use default values for missing entries in the ConfigMap.
+// Return an error for any other reason, such as an API error.
+func NewFeatureGateStatus(client client.Client) (*FeatureGateStatus, error) {
+	fgStatus := &FeatureGateStatus{
+		FeatureGates: make(map[string]bool),
+	}
+
+	cfgMap := &corev1.ConfigMap{}
+	err := client.Get(context.TODO(), types.NamespacedName{Name: FgConfigMapName,
+		Namespace: OperatorNamespace}, cfgMap)
+	if err == nil {
+		for feature, value := range cfgMap.Data {
+			fgStatus.FeatureGates[feature] = value == "true"
 		}
 	}
 
-	defaultValue, exists := DefaultFeatureGates[feature]
-	if exists {
-		return defaultValue
+	// Add default values for missing feature gates
+	for feature, defaultValue := range DefaultFeatureGates {
+		if _, exists := fgStatus.FeatureGates[feature]; !exists {
+			fgStatus.FeatureGates[feature] = defaultValue
+		}
 	}
-	return false
+
+	if k8serrors.IsNotFound(err) {
+		return fgStatus, nil
+	} else {
+		return fgStatus, err
+	}
+}
+
+func IsEnabled(fgStatus *FeatureGateStatus, feature string) bool {
+
+	return fgStatus.FeatureGates[feature]
 }
--- a/config/samples/featuregates.yaml
+++ b/config/samples/featuregates.yaml
@@ -8,15 +8,3 @@ data:
   # It is useful for scenarios where you want to view historical data or
   # predict future states based on current trends. Default is "false".
   # timeTravel: "false"
-
-  # quantumEntanglementSync enables instant state consistency across clusters
-  # using principles of quantum mechanics. This advanced feature ensures
-  # data is synchronized across different locations without traditional
-  # network latency. Default is "false".
-  # quantumEntanglementSync: "false"
-
-  # autoHealingWithAI employs artificial intelligence to automatically
-  # detect and resolve cluster issues. It leverages machine learning algorithms
-  # to predict potential problems before they occur, ensuring high availability.
-  # Default is "true".
-  # autoHealingWithAI: "true"
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -36,7 +36,6 @@ import (
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	mcfgconsts "github.com/openshift/machine-config-operator/pkg/daemon/constants"
 	kataconfigurationv1 "github.com/openshift/sandboxed-containers-operator/api/v1"
-	"github.com/openshift/sandboxed-containers-operator/internal/featuregates"
 	corev1 "k8s.io/api/core/v1"
 	nodeapi "k8s.io/api/node/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -61,8 +60,7 @@ type KataConfigOpenShiftReconciler struct {
 	Log    logr.Logger
 	Scheme *runtime.Scheme
 
-	kataConfig   *kataconfigurationv1.KataConfig
-	FeatureGates *featuregates.FeatureGates
+	kataConfig *kataconfigurationv1.KataConfig
 }
 
 const (
@@ -123,8 +121,10 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
-	if r.FeatureGates.IsEnabled(ctx, "timeTravel") {
-		r.Log.Info("TimeTravel feature is enabled. Performing feature-specific logic...")
+	err = r.processFeatureGates()
+	if err != nil {
+		r.Log.Info("Unable to process feature gates", "err", err)
+		return ctrl.Result{}, err
 	}
 
 	return func() (ctrl.Result, error) {
--- a/docs/FEATURE-GATES.md
+++ b/docs/FEATURE-GATES.md
@@ -2,9 +2,26 @@
 
 Feature Gates in the Openshift Sandboxed Containers operator allow
 administrators to enable or disable specific features. These features are
-controlled through the `FeatureGates` struct in the operator's configuration,
+controlled through the feature gates configmap (`osc-feature-gates`),
 providing a flexible approach to testing new features and managing the rollout
 of stable functionalities.
+
+## Design choices
+
+The feature gates are handled within the KataConfig reconcile loop.  If there
+are any errors with feature gate processing, the reconciliation process
+continues and doesn't re-queue.  In case of any errors in determining whether a
+feature gate is enabled or not (for example if the configMap is deleted or some
+other API server errors), the processing of the  feature gate is same as the
+default compiled in state of the respective feature gate.
+
+The status of individual feature gates is stored in the `FeatureGateStatus`
+struct that is populated in the beginning of the reconcile loop with the status
+of the feature gates from the  configMap. This ensures that the entire feature
+gate configMap is only read once from the API server instead of making repeated
+calls to the API server for checking individual feature gates.
+Any errors in reading the configMap from the API server will requeue a reconciliation request,
+except for when configMap is not found.
 
 ## Maturity Levels
 
@@ -41,7 +58,10 @@ adjusting your cluster's behavior as needed.
 
 ### Enabling and disabling features
 
-To enable a feature, you modify your `ConfigMap` object like so:
+A feature is enabled by explicitly setting it's value to `"true"` (case sensitive)
+and disabled by setting it's value to `"false"` (case sensitive)
+
+To enable a feature, you modify the `ConfigMap` object like so:
 
 ```yaml
 apiVersion: v1
@@ -51,11 +71,8 @@ metadata:
   namespace: openshift-sandboxed-containers-operator
 data:
   timeTravel: "true"
-  quantumEntanglementSync: "false"
-  autoHealingWithAI: "true"
 ```
 
-In this example, `timeTravel` is explicitly enabled, while
-`quantumEntanglementSync` is disabled, and `autoHealingWithAI` is enabled,
+In this example, `timeTravel` is explicitly enabled,
 showcasing how to manage the state of each feature individually. Regardless the
 default values they have.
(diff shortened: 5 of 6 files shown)


internal/featuregates/featuregates.go
```diff
@@ -13,10 +14,12 @@ type FeatureGates struct {
...
 }
 
+const (
+	TimeTravelFeatureGate = "timeTravel"
+)
+
 var DefaultFeatureGates = map[string]bool{
-	"timeTravel":              false,
-	"quantumEntanglementSync": false,
-	"autoHealingWithAI":       true,
```
beraldoleal, line 19: If we are removing here, do you mind removing from `config/samples/featuregates.yaml` too?

beraldoleal: Also from docs/FEATURE-GATES.md

controllers/fg_handler.go
```diff
@@ -0,0 +1,52 @@
...
+		// Perform the necessary actions
+		r.handleTimeTravelFeature(Disabled)
+	}
+
+	return nil
+
+}
+
+// Function to handle the TimeTravel feature gate
+func (r *KataConfigOpenShiftReconciler) handleTimeTravelFeature(state FeatureGateState) {
```
beraldoleal, line 45: nit that is not a problem for now, but maybe for the future: maybe move this to a specific file, so we can isolate code from different features.

bpradipt: You mean feature specific file right ?

internal/featuregates/featuregates.go
bpradipt: makes sense

pmores:
@bpradipt This PR looks good to me as it stands and I believe I'll be able to approve shortly.

I'd like to ask though if you have any plans to deal with the `FeatureGates` structure (since this PR relies on it implicitly).  To summarise and expand what I said on PR #394: the structure looks superfluous to me frankly.  Looking at its members, `ConfigMapName` is a compile-time constant so it can live in the binary's static data, no need to store it dynamically.  `Namespace` can be treated either as a compile-time constant as well, or we could go slightly more dynamic and derive it from the `KataConfig` instance namespace - in either case I can see little to no need to store it in a separate structure.  Finally, `Client` is already stored in `KataConfigOpenShiftReconciler`.

Also, its `IsEnabled()` member fetches the corresponding configmap on each query anew.  Is that necessary?  I think it would be preferable to fetch the configmap once per reconciliation somewhere near the top of `Reconcile()` then pass it to feature gate processing.  It could be stored in `KataConfigOpenShiftReconciler` for convenience if need be (and provided it's re-fetched on each reconcile).  Then `IsEnabled()` could just fetch a value from the cm if present or replace it with a default it not.

controllers/fg_handler.go
beraldoleal: yes.

bpradipt:
> @bpradipt This PR looks good to me as it stands and I believe I'll be able to approve shortly.
> 
> I'd like to ask though if you have any plans to deal with the `FeatureGates` structure (since this PR relies on it implicitly). To summarise and expand what I said on PR #394: the structure looks superfluous to me frankly. Looking at its members, `ConfigMapName` is a compile-time constant so it can live in the binary's static data, no need to store it dynamically. `Namespace` can be treated either as a compile-time constant as well, or we could go slightly more dynamic and derive it from the `KataConfig` instance namespace - in either case I can see little to no need to store it in a separate structure. Finally, `Client` is already stored in `KataConfigOpenShiftReconciler`.
> 

Valid point. I'll rethink about the FG struct. And even if we need this struct, we might not need it to be part of KataConfigOpenShiftReconciler struct. 


> Also, its `IsEnabled()` member fetches the corresponding configmap on each query anew. Is that necessary? I think it would be preferable to fetch the configmap once per reconciliation somewhere near the top of `Reconcile()` then pass it to feature gate processing. It could be stored in `KataConfigOpenShiftReconciler` for convenience if need be (and provided it's re-fetched on each reconcile). Then `IsEnabled()` could just fetch a value from the cm if present or replace it with a default it not.

While reading more about controller-runtime I found that it already uses [caching](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/client.go#L351). This was unclear to me earlier and I preferred retrieving the configMap once and using it for processing. With this new info I'm not sure and hence left the existing code as-is. What do you recommend ?

bpradipt:
> > @bpradipt This PR looks good to me as it stands and I believe I'll be able to approve shortly.
> > I'd like to ask though if you have any plans to deal with the `FeatureGates` structure (since this PR relies on it implicitly). To summarise and expand what I said on PR #394: the structure looks superfluous to me frankly. Looking at its members, `ConfigMapName` is a compile-time constant so it can live in the binary's static data, no need to store it dynamically. `Namespace` can be treated either as a compile-time constant as well, or we could go slightly more dynamic and derive it from the `KataConfig` instance namespace - in either case I can see little to no need to store it in a separate structure. Finally, `Client` is already stored in `KataConfigOpenShiftReconciler`.
> 
> Valid point. I'll rethink about the FG struct. And even if we need this struct, we might not need it to be part of KataConfigOpenShiftReconciler struct.
> 
> > Also, its `IsEnabled()` member fetches the corresponding configmap on each query anew. Is that necessary? I think it would be preferable to fetch the configmap once per reconciliation somewhere near the top of `Reconcile()` then pass it to feature gate processing. It could be stored in `KataConfigOpenShiftReconciler` for convenience if need be (and provided it's re-fetched on each reconcile). Then `IsEnabled()` could just fetch a value from the cm if present or replace it with a default it not.
> 
> While reading more about controller-runtime I found that it already uses [caching](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/client.go#L351). This was unclear to me earlier and I preferred retrieving the configMap once and using it for processing. With this new info I'm not sure and hence left the existing code as-is. What do you recommend ?

@pmores I have added new commits to remove the FG struct and read the configmap during reconciliation. 
Please check and let me know if this is what you were looking at ?

bpradipt:
> @pmores I have added new commits to remove the FG struct and read the configmap during reconciliation. Please check and let me know if this is what you were looking at ?

@pmores would it be better to have the FG status struct as part of the KataConfigOpenShiftReconciler struct itself in case there is a need for it's use in later part of the reconcile method ?

controllers/fg_handler.go
```diff
@@ -17,17 +17,13 @@ const (
 // Function to handle the feature gates
 func (r *KataConfigOpenShiftReconciler) processFeatureGates() error {
 
-	if r.FeatureGates == nil {
-		return nil
-	}
-
```
beraldoleal: Hi @bpradipt this seems to be part of the previous commit. Do you mind squashing them?

controllers/fg_handler.go
```diff
@@ -17,13 +15,18 @@ const (
...
+	}
+
 	// Check which feature gates are enabled in the FG ConfigMap and
 	// perform the necessary actions
 	// The feature gates are defined in internal/featuregates/featuregates.go
 	// and are fetched from the ConfigMap in the namespace
 	// Eg. TimeTravelFeatureGate
 
-	if featuregates.IsEnabled(context.TODO(), featuregates.TimeTravelFeatureGate) {
+	if featuregates.IsEnabled(fgStatus, featuregates.TimeTravelFeatureGate) {
```
beraldoleal, line 30: same here.

controllers/fg_handler.go
bpradipt: Yeah I'll do it. Are you ok with the logic ?

controllers/fg_handler.go
```diff
@@ -0,0 +1,51 @@
+package controllers
+
+import (
+	"github.com/openshift/sandboxed-containers-operator/internal/featuregates"
+)
+
+// Create two enums to represent the state of the feature gates
```
beraldoleal: two or one enum with two possible values?

internal/featuregates/featuregates.go
```diff
@@ -19,23 +18,39 @@ var DefaultFeatureGates = map[string]bool{
...
-	if err != nil {
-		log.Printf("Error fetching feature gates: %v", err)
-	} else {
-		if value, exists := cfgMap.Data[feature]; exists {
-			return value == "true"
+	err := client.Get(context.TODO(), types.NamespacedName{Name: FgConfigMapName,
+		Namespace: OperatorNamespace}, cfgMap)
+	if err == nil {
+		for feature, value := range cfgMap.Data {
+			fgStatus.FeatureGates[feature] = value == "true"
```
pmores, line 42: Do we want to allow for "True" or "TRUE" as well?  This comparison is case-sensitive so it will only match "true".

pmores:
> > @pmores I have added new commits to remove the FG struct and read the configmap during reconciliation. Please check and let me know if this is what you were looking at ?
> 
> @pmores would it be better to have the FG status struct as part of the KataConfigOpenShiftReconciler struct itself in case there is a need for it's use in later part of the reconcile method ?

Yeah, that occured to me as well. :-)  As long as we can keep gated feature processing limited to `processFeatureGates()` we don't have to put the FG status into `KataConfigOpenShiftReconciler`.  However my hunch is that sooner or later (actually rather sooner than later :-)) there will be a feature that needs to affect the control flow of the main controller algorithms (installation, uninstallation), and then having the FG status in the reconciler struct will come handy.  So I think it's fine to put it there - as long as we take care to load it at the beginning of each reconciliation - and we can do it now or we can wait until it's actually needed (I'd probably slightly prefer the latter).

internal/featuregates/featuregates.go
```diff
@@ -2,43 +2,55 @@ package featuregates
...
-	if exists {
-		return defaultValue
+	// Add default values for missing feature gates
+	for feature, defaultValue := range DefaultFeatureGates {
+		if _, exists := fgStatus.FeatureGates[feature]; !exists {
+			fgStatus.FeatureGates[feature] = defaultValue
+		}
 	}
-	return false
+	return fgStatus, err
```
beraldoleal: TBH, it feels strange to me to return an error to the upper layers, assuming they will ignore the error (or just log it as an info) and continue the flow.

IMO, err should be used only during failures that could potentially block the execution, or at least give us error messages. For this case, since we are assuming default values, I would move the log on inside this method, and be straightforward here and do not return an error. 

I"m missing the use cases where we need to differentiate one case from another, and if we need it.

internal/featuregates/featuregates.go
```diff
@@ -2,43 +2,55 @@ package featuregates
...
+}
+
+func IsEnabled(fgStatus *FeatureGateStatus, feature string) bool {
+
+	return fgStatus.FeatureGates[feature]
```
beraldoleal, line 62: Since feature is a string, and anything could be used, I'm missing here a KeyError handler.

beraldoleal (commented):
Hi @bpradipt, I left a few comments for you.

Also, it would be much appreciated if the commit messages was explaining why you are proposing the changes instead of listing what is doing. This would help reviewers to think what you have in mind with those changes. Do you mind expanding the reasoning on the commit message, please?

internal/featuregates/featuregates.go
bpradipt: @beraldoleal can you please share an example..

internal/featuregates/featuregates.go
bpradipt: @pmores what do you recommend ?

controllers/fg_handler.go
bpradipt: You are right. one enum type with two values.

internal/featuregates/featuregates.go
bpradipt: Currently we are only allowing "true" (case sensitive).

internal/featuregates/featuregates.go
pmores: Off-hand, I think we might want to differentiate between an err due to the configmap not existing as opposed to other causes (failures to access the cluster, generally).  If former we should just fill in the FG status as if we succeeded retrieving the configmap and it was full of default values.  If latter we should return the err to the caller and the caller shouldn't ignore it - yes, log it for sure, but also break off and reschedule the reconciliation.

Does that make sense?

bpradipt:
> Hi @bpradipt, I left a few comments for you.
> 
> Also, it would be much appreciated if the commit messages was explaining why you are proposing the changes instead of listing what is doing. This would help reviewers to think what you have in mind with those changes. Do you mind expanding the reasoning on the commit message, please?

I'll add it. But specifically for this PR there is a long history of discussions. I would also suggest to take a look at them for the context.

internal/featuregates/featuregates.go
pmores: Okay, fair enough.  We need to keep in mind documenting it though.

bpradipt:
> > Hi @bpradipt, I left a few comments for you.
> > Also, it would be much appreciated if the commit messages was explaining why you are proposing the changes instead of listing what is doing. This would help reviewers to think what you have in mind with those changes. Do you mind expanding the reasoning on the commit message, please?
> 
> I'll add it. But specifically for this PR there is a long history of discussions. I would also suggest to take a look at them for the context.

@beraldoleal for some context on the new changes w.r.to removal of the FG struct and reading the configMap once during entering the reconcile loop - https://github.com/openshift/sandboxed-containers-operator/pull/408#issuecomment-2115295426

bpradipt:
> > > @pmores I have added new commits to remove the FG struct and read the configmap during reconciliation. Please check and let me know if this is what you were looking at ?
> > 
> > 
> > @pmores would it be better to have the FG status struct as part of the KataConfigOpenShiftReconciler struct itself in case there is a need for it's use in later part of the reconcile method ?
> 
> Yeah, that occured to me as well. :-) As long as we can keep gated feature processing limited to `processFeatureGates()` we don't have to put the FG status into `KataConfigOpenShiftReconciler`. However my hunch is that sooner or later (actually rather sooner than later :-)) there will be a feature that needs to affect the control flow of the main controller algorithms (installation, uninstallation), and then having the FG status in the reconciler struct will come handy. So I think it's fine to put it there - as long as we take care to load it at the beginning of each reconciliation - and we can do it now or we can wait until it's actually needed (I'd probably slightly prefer the latter).

Ok, let's do it later when it's actually needed :-)

internal/featuregates/featuregates.go
pmores: Well, as far as I can see this function will only ever be called with one of a limited set of compile-time string constants as a parameter - it's not like we read a string externally (e.g. from the user) and pass it to `IsEnabled()`.  If that's the case, key existence checking doesn't seem strictly necessary, and adding it could arguably confuse a reader by suggesting a non-existent key _can_ be passed in.

So I'd say, double check my premise above is correct and if so, make this an explicit feature of code design by not including checking (maybe add a comment to that effect).  However if there _is_ a chance that the premise doesn't hold, even a slight one, let's stay on the safe side and do the checking.

pmores:
> > Hi @bpradipt, I left a few comments for you.
> > Also, it would be much appreciated if the commit messages was explaining why you are proposing the changes instead of listing what is doing. This would help reviewers to think what you have in mind with those changes. Do you mind expanding the reasoning on the commit message, please?
> 
> I'll add it. But specifically for this PR there is a long history of discussions. I would also suggest to take a look at them for the context.

I know it's tedious but some of the preliminary discussion was not public, and I agree with @beraldoleal that at least main design decisions should ideally be publicly documented.

internal/featuregates/featuregates.go
bpradipt: Sorry, I'm unable to understand this and hence my question. The FG status is populated during every reconcile and checked by the operator code so the code has to ensure right feature is checked.
Also if a missing feature is checked it'll return the default bool value, ie false. 
Check this - thttps://play.golang.com/p/X2N0sc5fNi6

pmores: Right, maybe I misinterpreted Beraldo's concern.  @beraldoleal could you elaborate what you're worried about?

bpradipt:
@pmores @beraldoleal I have squashed the commits and addressed the comments.
One thing pending is the following - https://github.com/openshift/sandboxed-containers-operator/pull/408#discussion_r1608357767

Please tell me what you prefer and I'll make the changes accordingly. I think it's a matter of taste and doesn't impact the logic.

pmores:
> @pmores @beraldoleal I have squashed the commits and addressed the comments. One thing pending is the following - [#408 (comment)](https://github.com/openshift/sandboxed-containers-operator/pull/408#discussion_r1608357767)
> 
> Please tell me what you prefer and I'll make the changes accordingly. I think it's a matter of taste and doesn't impact the logic.

Correct me if I'm wrong but I'm under the impression that it does make a difference.  If the function never returns an error then we're bound to revert gated features to defaults on any intermittent failure to communicate with the control plane, right?  Wouldn't we rather want to reschedule and try again?  There might be gated features whose status change can trigger heavyweight operations in the cluster.

internal/featuregates/featuregates.go
bpradipt: We can check if the configMap is not found, then fill it default values. For any other error we can reschedule the reconciliation. Would this be ok ?

bpradipt:
>  There might be gated features whose status change can trigger heavyweight operations in the cluster.

This was precisely my argument w.r.to differentiating between deleting a configMap and disabling the FGs vs explicitly disabling an FG in the configMap :-(

pmores:
> > There might be gated features whose status change can trigger heavyweight operations in the cluster.
> 
> This was precisely my argument w.r.to differentiating between deleting a configMap and disabling the FGs vs explicitly disabling an FG in the configMap :-(

I know, but I see a significant difference between triggering a heavyweight operation due to an explicit user action (i.e. deleting the configmap) vs due to a intermittent failure to talk to the control plane.  I believe we should work hard to avoid the latter as its result would be a terrible user experience - the controller all of a sudden changing the state of the cluster to an undesired one, seemingly for no reason.

internal/featuregates/featuregates.go
beraldoleal: I completely missed that accessing a map with a key that does not exist in Go returns the zero value for the type of the map's values, which is false for a bool. So, we are good here.

However, since we are refactoring this, I'm wondering if showing that a feature is disabled is acceptable even if it is nonexistent. This could lead to users misinterpreting logs. While this behaviour is valid currently because we are using constants and only existing features, future changes could break this assumption. Maybe we should differentiate this via logs (not necessarily  at `IsEnabled`)

beraldoleal:
> > > Hi @bpradipt, I left a few comments for you.
> > > Also, it would be much appreciated if the commit messages was explaining why you are proposing the changes instead of listing what is doing. This would help reviewers to think what you have in mind with those changes. Do you mind expanding the reasoning on the commit message, please?
> > 
> > 
> > I'll add it. But specifically for this PR there is a long history of discussions. I would also suggest to take a look at them for the context.
> 
> @beraldoleal for some context on the new changes w.r.to removal of the FG struct and reading the configMap once during entering the reconcile loop - [#408 (comment)](https://github.com/openshift/sandboxed-containers-operator/pull/408#issuecomment-2115295426)

I understand your approach and I got where you are coming from, but it's important to consider the reviewer's perspective. Reading through all comments in every thread (eventually from multiple reviewers) to understand the changes can be quite unproductive.

Commits should provide immediate context and reasoning, allowing reviewers to quickly grasp the proposed changes without needing to dig through lengthy discussions.

Also, a PR is a proposal to update the git tree, with commits reflecting the desired final state. Reviewing everything twice (code and later git log) isn't very efficient. Let's aim to keep our commit messages clear and comprehensive for smoother reviews.

internal/featuregates/featuregates.go
bpradipt: Let's handle this case in future please..

pmores: Frankly, it'd probably leave it as is.  First, gated feature identifiers are very likely to stay the way they are - compile-time constants controlled by this project.  In other words, I don't see much scope for them to become somewhat arbitrary strings.  Also, it could be argued that if a feature doesn't exist then it's necessarily disabled :-) so even the worst case semantics doesn't really disturb me.

bpradipt: Note that the caller (ie the operator code) needs to explicitly query to check a feature gate. So why would we willingly write a code for non-existent feature? If there are cases about using featuregate module elsewhere then it might make sense. Let's handle it when there is a requirement.

pmores: Yes, I think it's robust.

pmores (approved):
lgtm, thanks @bpradipt !

beraldoleal (approved):
Thanks, LGTM.

bpradipt:
Thanks guys.. Merging this without delay ;)

internal/featuregates/featuregates.go
beraldoleal: > So why would we willingly write a code for non-existent feature?

> First, gated feature identifiers are very likely to stay the way they are - compile-time constants controlled by this project. In other words, I don't see much scope for them to become somewhat arbitrary strings. 

We can't guarantee future code, refactoring, or features won't break this assumption. Writing code based on an assumption without enforcement is risky. However, I agree we can't account for every possibility and need to find a balance. Over-engineering is bad, but lacking some checks is also dangerous.

So, for the sake of balance, let's move forward. Approved.

PR #409, opened 2024-05-23
Bump OSC to 1.6.0

gkurz:
The rationale is that the version should be bumped on the devel branch at the beginning of the release development cycle according to [1]. The version is still 1.5.2. Let's catch up and bump now. We'll then create the release-1.6 branch from here and be able to start the 1.7.0 development cycle on devel.

The release activities process described in [1] basically means that we should only have X.Y.0 releases on the devel branch from now on. Any patch version that we may release before the next minor or major version will only exist in the associated release branch. BUT we should still remember to bump the version in the `replaces:` field of the CSV on the devel branch (e.g. this commit is bumping from 1.5.2 to 1.6.0 but it replaces 1.5.3 which is the actual latest release).

[1] https://docs.google.com/document/d/1Sp0z7hKlHg8C_KIGIMY-42xppWhjRbMtvee0FAJSREQ/edit#heading=h.rd5u14lkzviz


hack/aws-image-job.yaml
```diff
@@ -21,7 +21,7 @@ spec:
 
       initContainers:
       - name: payload
-        image: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9:1.5.2
+        image: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9:1.6.0
```
bpradipt: These are no longer needed for 1.6.0. The aws-image-job.yaml and azure-image-job.yaml can be removed

bpradipt: Additionally the tag in these files needs to be changed to 1.6.0
https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/osc-podvm-create-job.yaml#L18
https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/osc-podvm-create-job.yaml#L32
https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/osc-podvm-delete-job.yaml#L22
https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/osc-podvm-gallery-delete-job.yaml#L15

bpradipt (approved):
/lgtm
Thanks @gkurz

littlejawa (approved):
/lgtm
Thanks @gkurz

PR #416, opened 2024-06-04
Kata installation using RHCOS image layer

bpradipt:


config/samples/featuregates.yaml
```diff
@@ -7,4 +7,8 @@ data:
   # timeTravel allows navigating through cluster states across time.
   # It is useful for scenarios where you want to view historical data or
   # predict future states based on current trends. Default is "false".
-  # timeTravel: "false"
+  timeTravel: "false"
```
pmores: Is enabling the dummy gated feature intended here?

config/samples/layered-image-deploy-cm.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+data:
+  osImageURL: "quay.io/openshift/tdx:ocp415"
```
pmores: Is this a suitable upstream default value?

controllers/openshift_controller.go
```diff
@@ -1239,30 +1272,25 @@ func (r *KataConfigOpenShiftReconciler) createExtensionMc(machinePool string) (b
...
-	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: extension_mc_name}, mc)
-	if err != nil && (k8serrors.IsNotFound(err) || k8serrors.IsGone(err)) {
 
-		r.Log.Info("creating RHCOS extension MachineConfig")
-		mc, err = r.newMCForCR(machinePool)
-		if err != nil {
-			return dummy, err
-		}
+	if r.ExistingMc != nil {
+		r.Log.Info("Existing MachineConfig already exists")
```
pmores: This log message sounds kind of tautological, was it meant to say "there is an existing MC already" or "using an existing MC" or something like that?

config/samples/layered-image-deploy-cm.yaml
bpradipt: Not at all ;)

controllers/openshift_controller.go
bpradipt: I meant to say "there is an existing MC already".

config/samples/featuregates.yaml
bpradipt: Do you prefer to keep all the featuregates commented out in the sample file ?
The reason I uncommented it were to prevent formatting errors in the IDE
```
Incorrect type. Expected "object".yaml-schema: kubernetes://schema/v1%40configmap
```

pmores: In that case, since we're adding actual gated features already, maybe it would be best to get rid of the dummy feature as the first commit of this series?

As for the contents of the sample file, admittedly I'm not sure how it's used.  If it's an actual sample for users to use then I assume its contents should ideally match the defaults the controller uses.

bpradipt: > In that case, since we're adding actual gated features already, maybe it would be best to get rid of the dummy feature as the first commit of this series?

I agree..

spotlesstofu:
Avoid growing the file `openshift_controller.go`. The code that you added to `openshift_controller.go` can be moved to `layered_image_handler.go` as one or more functions. Then you call those functions from `openshift_controller.go`.

With this approach, you could also avoid the new variables `ImgMc` and `ExistingMc`.

For example, let's suppose you have a `getImageMc` function that returns the image MachineConfig when the feature is enabled. You could call `getImageMc` from `newMCForCR`. If that returns nil, you continue the usual flow, extension-based. If `getImageMc` returns a MachineConfig, you apply that.

Can this work?

bpradipt:
> Avoid growing the file `openshift_controller.go`. The code that you added to `openshift_controller.go` can be moved to `layered_image_handler.go` as one or more functions. Then you call those functions from `openshift_controller.go`.
> 
> With this approach, you could also avoid the new variables `ImgMc` and `ExistingMc`.
> 
> For example, let's suppose you have a `getImageMc` function that returns the image MachineConfig when the feature is enabled. You could call `getImageMc` from `newMCForCR`. If that returns nil, you continue the usual flow, extension-based. If `getImageMc` returns a MachineConfig, you apply that.
> 
> Can this work?

Thanks for your suggestion @spotlesstofu . I'll spend some time to figure out how to incorporate the changes and get back.

bpradipt:
@spotlesstofu I have incorporated your suggestion and added it as a separate commit. 

If the approach looks good, I'll squash the new commit with the first one.

controllers/openshift_controller.go
```diff
@@ -69,6 +69,7 @@ const (
 	dashboard_configmap_namespace       = "openshift-config-managed"
 	container_runtime_config_name       = "kata-crio-config"
 	extension_mc_name                   = "50-enable-sandboxed-containers-extension"
+	image_mc_name                       = "50-enable-sandboxed-containers-image"
```
spotlesstofu: Nit: move this to the feature handler file.

controllers/openshift_controller.go
```diff
@@ -494,6 +495,24 @@ func getExtensionName() string {
...
 
+	// Return ImgMc if present
+	mc := r.getImageMc()
+	if mc != nil {
+		r.Log.Info("Using Image based MachineConfig")
+		// Add the required MachineConfig labels
+		mc.Labels = map[string]string{
+			"machineconfiguration.openshift.io/role": machinePool,
+			"app":                                    r.kataConfig.Name,
+		}
```
spotlesstofu: Can you move setting the labels to `layered_image_handler.go`?

controllers/layered_image_handler.go
```diff
@@ -0,0 +1,149 @@
...
+		// Split the kernel arguments string into individual arguments
+		mc.Spec.KernelArguments = strings.Fields(kernelArguments)
+	}
+
+	return mc, nil
+}
+
+// Method to retrieve Image MachineConfig from the cluster
+func (r *KataConfigOpenShiftReconciler) getImageMc() *mcfgv1.MachineConfig {
+	return ImgMc
```
spotlesstofu: Is it possible to generate the `MachineConfig` when `getImageMc` is called?

controllers/layered_image_handler.go
```diff
@@ -0,0 +1,149 @@
...
+	LayeredImageDeployCm = "layered-image-deploy-cm"
+)
+
+// Global struct to store the Image based MachineConfig struct
+var ImgMc *mcfgv1.MachineConfig
+
+func (r *KataConfigOpenShiftReconciler) handleLayeredImageDeploymentFeature(state FeatureGateState) error {
+
+	// Check if MachineConfig exists and return the same without changing anything
+	mc, err := r.getExistingMachineConfig()
```
spotlesstofu, line 36: Is this needed? The possibility of a MachineConfig already existing is managed in the function `createMc`.

controllers/layered_image_handler.go
bpradipt: Yeah, it is possible. 
I can check if the featuregate is enabled and generate the MachineConfig in this method.
However, since the approach I took is to do all these in the handler function, I avoided generating it again in this function.

controllers/layered_image_handler.go
bpradipt: Let's consider this sequence 
getImgMc updated to generate MachineConfig if FG is enabled.

Now let's assume FG is disabled and KataConfig is created
Reconcile -> createMc -> newMCForCR -> getImgMc (returns imageMc or nil based on FG status - returns nil) 
createMc checks for mc.Name (extension in this case) and creates it if not present

Now let's assume FG is enabled after KataConfig creation

Reconcile -> createMc -> newMCForCR -> getImgMc (returns imageMc or nil based on FG status - returns imageMc) 
createMc checks for mc.Name (imageMc in this case) and creates it if not present. 
So we end up re-installing.

Now I think this can be avoided by checking for either of the extension in createMc.
But I took the current approach of checking for either of the MCs and creating the image MC in the handler function 

I might have missed something as well.  Please let me know your suggestion on the best way forward.

spotlesstofu: Right. It makes sense that you check for the "image" MachineConfig.

You could avoid checking for the "extension" MachineConfig, since that already happens in `createMc`.

controllers/layered_image_handler.go
spotlesstofu: I find it unusual, and probably confusing, to define `ImgMc` before it's needed, and to keep it in memory until it's applied.

controllers/fg_handler.go
```diff
@@ -95,6 +97,15 @@ func (r *KataConfigOpenShiftReconciler) processFeatureGates() error {
 		}
 	}
 
-	return err
+	// Check layered Image deployment FG
+	if IsEnabled(fgStatus, LayeredImageDeployment) {
+		r.Log.Info("Feature gate is enabled", "featuregate", LayeredImageDeployment)
+		// Perform the necessary actions
```
snir911, line 103: remove?

controllers/layered_image_handler.go
```diff
@@ -0,0 +1,138 @@
...
+	mcList := &mcfgv1.MachineConfigList{}
+	err := r.Client.List(context.Background(), mcList)
+	if err != nil {
+		r.Log.Info("Error in listing MachineConfigs", "err", err)
+		return nil, err
+	}
+
+	for _, mc := range mcList.Items {
+		if mc.Labels["app"] == r.kataConfig.Name &&
+			(mc.Name == extension_mc_name || mc.Name == image_mc_name) {
```
snir911, line 96: It seems like it's assumed the extension_mc and the image mc are mutually exclusive and used for the same purpose, in that case, i'd expect them to be similarly set from the user's POV, why the user is required to set the layered-image-deploy-cm? (extension is automatically set IIRC)

bpradipt: The reason for having `layered-image-deploy-cm` to specify details like image and kernel params is because we want to enable the admins to change it as needed. So while  extension_mc and the image mc are mutually exclusive, the key difference is that image mc is configurable by the admin.

controllers/layered_image_handler.go
bpradipt: Let's take a step back and see if we can have a consensus on the basic pattern that needs to be used for setting the FG configs. 

1. Pattern-1: Check if FG is enabled at the reconciliation loop's beginning and set FG-specific variables (e.g., using ImgMc as part of the KataConfigOpenShiftReconciler struct or otherwise ). If needed, use the required FG-specific variables later in the reconciliation loop.
2. Pattern-2: Where ever needed in the reconcile loop, check if FG is enabled and set and use the required FG-specific variables

I find pattern-1 easier to understand and maintain going forward for different FGs. 

@spotlesstofu @pmores @snir911 do you see any downsides in using pattern-1 ? Do you have a strong preference for pattern-2 ?

controllers/layered_image_handler.go
snir911: In this case shouldn't we set default layered-image-deploy-cm with default values when this feature is enabled? instead of expecting user to set it

bpradipt: We may set it going forward. We'll also provide examples in blogs when we describe using this feature for some scenarios. Right now we don't have any default image to set. 
Also note that these FGs are for enabling experiments. Eventually if we decide to enable the FG by default, then it goes in the CRD.

controllers/layered_image_handler.go
snir911: Nothing obvious jump in my mind, i'm find with pattern-1

bpradipt: @pmores @spotlesstofu can I proceed with pattern-1 and complete the pending changes?

spotlesstofu: Yes. Sorry for not answering before.

snir911 (approved):
LGTM overall, thanks.
( I'd still prefer to have fixed image value also if it's FG, however, no a deal breaker according to what you mentioned :) )

spotlesstofu (approved):

controllers/layered_image_handler.go
```diff
@@ -0,0 +1,164 @@
...
+		r.Log.Info("LayeredImageDeployment feature is enabled")
+
+		cm := &corev1.ConfigMap{}
+		err := r.Client.Get(context.Background(), types.NamespacedName{
+			Name:      LayeredImageDeployCm,
+			Namespace: OperatorNamespace,
+		}, cm)
+		if err != nil {
+			r.Log.Info("Error in retrieving LayeredImageDeployment ConfigMap", "err", err)
+			return err
```
pmores, line 61: Should we here again handle "not found" separately from other errors?

If the error is other than "not found" then it likely means just an (probably) intermittent failure to talk to the control plane.

If it's "not found" (and if I understand the logic of this PR correctly) then it's a user error which we should try to communicate to the user ("you enabled image-based deployment but gave no image URL and we can't work with that") _unless_ we also want to provide a default image URL.  In that case, I think we should proceed as if we found a configmap with the default image URL and empty kernel args.

PR #417, opened 2024-06-04
Add feature confidential

spotlesstofu:
https://issues.redhat.com/browse/KATA-2870


controllers/fg_handler.go
```diff
@@ -12,41 +30,66 @@ const (
 	Disabled
 )
 
+// This method returns a new FeatureGateStatus object
+// that contains the status of the feature gates
+// defined in the ConfigMap in the namespace
+// Return default values if the ConfigMap is not found.
+// Return values from the ConfigMap if the ConfigMap is not found. Use default values for missing entries in the ConfigMap.
+// Return an error for any other reason, such as an API error.
+func NewFeatureGateStatus(client client.Client) (*FeatureGateStatus, error) {
```
bpradipt: since this method is now part of controllers package can it be made part of the KataConfigOpenShiftReconciler struct ?

controllers/confidential_handler.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return client.Update(context.TODO(), configMap)
+}
+
+func toggleFeatureConfidential(client client.Client, enabled bool) error {
+
+	var podvmData map[string]string
+	var peerpodsData map[string]string
+
+	if enabled {
```
bpradipt: Few questions:
- Should the code check if `enablePeerPods == true` before (re)setting the peer-pods cm ?
- Is there a need to check and reset the `AZURE_IMAGE_ID`, since if this is set then image won't be generated.
- Should the code be structured in a way to handle other providers when needed instead of hard-coding it to azure specifics?

controllers/fg_handler.go
spotlesstofu: Sure thing. Can you help me understand why this is desirable?

bpradipt: We can use the client and other variables that are part of the KataConfigOpenShiftReconciler struct if needed.

spotlesstofu: Addressed with [aaeaea6](https://github.com/openshift/sandboxed-containers-operator/pull/417/commits/aaeaea6dd3e8638b9a15775b6f2cb65025214f06).

controllers/confidential_handler.go
```diff
@@ -0,0 +1,53 @@
...
+		podvmData = map[string]string{"CONFIDENTIAL_COMPUTE_ENABLED": "yes"}
+		peerpodsData = map[string]string{"DISABLECVM": "false"}
+	} else {
+		podvmData = map[string]string{"CONFIDENTIAL_COMPUTE_ENABLED": "no"}
+		peerpodsData = map[string]string{"DISABLECVM": "true"}
+	}
+
+	// Patch PodVM image configMap.
+	podvmName := "azure-podvm-image-cm"
```
snir911: nit, maybe podvmCMName? also for peerpodsName

controllers/confidential_handler.go
snir911: > Should the code be structured in a way to handle other providers when needed instead of hard-coding it to azure specifics?

Wouldn't it make sense to implement this logic be in the image creation scripts?

And some more questions:
Will we support switching confidential/non-confidential  at runtime?
what about verifying the instance type user sets is confidential? should we enforce it? (maybe in the scripts?) what happens if it's non confidential ? (IIRC DISABLECVM=true) will cause it to fail

spotlesstofu:
@snir911 @bpradipt I tried to address your feedback and force-pushed new changes.

What I did **not** address:
  - Support switching confidential/non-confidential at runtime.
  - Reset the `AZURE_IMAGE_ID`, because of the previous point.
  - Verify the instance type. It's a big cost to keep the instance types up to date. May reconsider for future releases.

bpradipt:
> * Verify the instance type. It's a big cost to keep the instance types up to date. May reconsider for future releases.

I don't think this is needed since an error will be thrown during VM creation if instance type is incorrect. Also it's hard to keep track of different instance types.  
In case we want to verify if the instance type is right or not, the same can be implemented in cloud-api-adaptor itself by enhancing the ConfigVerifier method - https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-providers/aws/provider.go#L321

bpradipt:
> Support switching confidential/non-confidential at runtime.

To start with, I think we can tie this featuregate to KataConfig creation/deletion (similar to layered image fg). So any changes to this feature gate config post KataConfig creation will be a no-op.  Likewise for disabling CoCo, the KataConfig needs to be deleted and recreated. 
@spotlesstofu does this make sense ?
cc @pmores

spotlesstofu:
Yes @bpradipt, I agree with both your comments.

controllers/image_generator.go
```diff
@@ -631,6 +631,10 @@ func checkKeysPresentAndNotEmpty(data interface{}, keys []string) bool {
 	return true
 }
 
+func (r *ImageGenerator) getImageConfigMapName() string {
+	return r.provider + "-podvm-image-cm"
+}
+
```
bpradipt, line 637: nice..

controllers/confidential_handler.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func patchConfigMap(client client.Client, name string, namespace string, data map[string]string) error {
```
bpradipt: nit: s/name/cmName

controllers/confidential_handler.go
```diff
@@ -0,0 +1,71 @@
...
+	for key, value := range data {
+		configMap.Data[key] = value
+	}
+
+	return client.Update(context.TODO(), configMap)
+}
+
+func (r *KataConfigOpenShiftReconciler) toggleFeatureConfidential(enabled bool) error {
+
+	var ImageConfigMapData map[string]string
```
bpradipt: nit: s/ImageConfigMapData/imageConfigMapData
since this is not an exported var

controllers/confidential_handler.go
```diff
@@ -0,0 +1,71 @@
...
+	// Changes to make before pods start:
+	//   - Edit peer pods config to enable confidential.
+
+	// Patch peer pods configMap, if it exists.
+	if err := patchConfigMap(r.Client, peerpodsCMName, OperatorNamespace, peerpodsCMData); err != nil && !k8serrors.IsNotFound(err) {
+		return err
+	}
+
+	return nil
+}
```
bpradipt, line 78: Should the configmap values be switched to default when this fg is disabled (as part of kataconfig deletion)?

controllers/confidential_handler.go
```diff
@@ -62,9 +62,11 @@ func (r *KataConfigOpenShiftReconciler) toggleFeatureConfidential(enabled bool)
...
 	//   - Edit peer pods config to enable confidential.
 
-	// Patch peer pods configMap, if it exists.
-	if err := patchConfigMap(r.Client, peerpodsCMName, OperatorNamespace, peerpodsCMData); err != nil && !k8serrors.IsNotFound(err) {
-		return err
+	if r.kataConfig.Spec.EnablePeerPods {
+		// Patch peer pods configMap, if it exists.
+		if err := patchConfigMap(r.Client, peerpodsCMName, OperatorNamespace, peerpodsCMData); err != nil && !k8serrors.IsNotFound(err) {
+			return err
+		}
```
bpradipt, line 74: nit: this commit should be squashed with the previous one adding the confidential_handler

controllers/confidential_handler.go
bpradipt: Also pls change `enabled bool`  to `state FeatureGateState` to be consistent. 
While today we just have two states for the featuregate (fg), so bool is ok, but having the fg state as enum will retain the flexibility in case we want to introduce some additional states.

controllers/confidential_handler.go
bpradipt: Also, I think since any changes to the image configmap values are only applicable before ImageCreate, other than resetting the configmap values on disabling of this fg, there is nothing else to be done.

bpradipt:
@spotlesstofu overall this is pretty close and I added some minor comments. 
Would love to hear comments from others
cc @snir911 @pmores @gkurz

gkurz:
> Merge featuregates.go in fg_handler.go
> Signed-off-by: Camilla Conte <cconte@redhat.com>

Hi @spotlesstofu ! Sorry for reviewing this late :-\

Maybe it was discussed elsewhere but I'd like to know the motivation for this initial commit. Is it just to consolidate all the feature gates code into one place and get rid of the `./internal` sub directory or is there more to it ?

Please explain it in the commit description.

controllers/confidential_handler.go
gkurz: Frankly, I don't understand this function. There are subtle differences in the handing of each config map and it is hard to guess the outcome depending when it is called.

@spotlesstofu please explain the flow and expected result in the commit description, including any special situation (e.g. why is it fine to lack a config map or not).

controllers/fg_handler.go
```diff
@@ -78,27 +78,18 @@ func (r *KataConfigOpenShiftReconciler) processFeatureGates() error {
...
-		r.handleTimeTravelFeature(Enabled)
-	} else {
-		r.Log.Info("Feature gate is disabled", "featuregate", TimeTravelFeatureGate)
-		// Perform the necessary actions
-		r.handleTimeTravelFeature(Disabled)
+	if r.kataConfig.Spec.EnablePeerPods {
+		if IsEnabled(fgStatus, ConfidentialFeatureGate) {
+			r.Log.Info("Feature gate is enabled", "featuregate", ConfidentialFeatureGate)
+			// Perform the necessary actions
+			r.toggleFeatureConfidential(true)
```
gkurz: `toggleFeatureConfidential()` can fail. Ignoring the error seems wrong.

controllers/fg_handler.go
```diff
@@ -78,27 +78,18 @@ func (r *KataConfigOpenShiftReconciler) processFeatureGates() error {
...
+		} else {
+			r.Log.Info("Feature gate is disabled", "featuregate", ConfidentialFeatureGate)
+			// Perform the necessary actions
+			r.toggleFeatureConfidential(false)
```
gkurz: ditto

controllers/confidential_handler.go
```diff
@@ -0,0 +1,71 @@
...
+	configMap := &corev1.ConfigMap{}
+	if err := client.Get(context.TODO(), types.NamespacedName{
+		Name:      name,
+		Namespace: namespace,
+	}, configMap); err != nil {
+		return err
+	}
+
+	for key, value := range data {
+		configMap.Data[key] = value
```
snir911: Since we are patching configuration that was set by the user it worth having some log mentioning this, WDYT?
Maybe i think it's more true when it comes to the peer-pods-cm rather than the image one..

controllers/confidential_handler.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	// Changes to make before image creation job starts:
+	//   - Edit ImageConfigMap to create a confidential image.
+
+	ig := GetImageGenerator()
+
+	if enabled {
+		// Create ImageConfigMap, if it doesn't exist already.
+		if err := ig.createImageConfigMapFromFile(); err != nil {
```
snir911: Should we think about cases where Confidential is supported but image generation isn't (libvirt?), i'm not sure it will go through successfully here
maybe we should consider confidential featureGating to supported providers only? (i.e. have a function to check provider is supported)

snir911 (commented):
overall LGTM, thanks! added some comments

spotlesstofu:
Thanks @snir911 @gkurz @bpradipt. I should've addressed all your comments now.

To make the code clearer, I did:
- Split function into enable and disable.
- Add some comments to explain the flow.
- Expand the code.

spotlesstofu:
I also noticed that `GetImageGenerator` can return `nil`, so I added a call to `InitializeImageGenerator` first.

I wanted to update `GetImageGenerator` to initialize automatically as needed, but that wasn't immediate. Indeed, the initialize function needs the client as an additional argument.

controllers/confidential_handler.go
```diff
@@ -0,0 +1,132 @@
...
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// disableFeatureConfidential resets config maps to non-confidential values.
+func (r *KataConfigOpenShiftReconciler) disableFeatureConfidential() error {
```
snir911: would it be more logically sense to enable/disable the image CM from within ig.updateImageConfigMap based on if confidential is enabled/disabled?
And similarly for the peer-pods-cm at the processKataConfigInstallRequest maybe
Sorry for the late suggestion, feel free to ignore it.. :)

spotlesstofu: Yes, but that would mean integrating the feature in the bulk of the code and not keeping it in a separate file. I understand that the whole point of this feature gates thing is to keep things separate, or isn't it?

snir911: Right, well, the question is if editing the same objects in multiple locations/phases of the reconcile loop for the sake of code separation wroth it, IDK..
We are at the point we can still track it, so i guess it's fine, however, we do expect more features that touches the same CMs (automatic setting of peer-pods-cm, GPU enablement etc..), we'll have to make it more generic some point..

controllers/confidential_handler.go
```diff
@@ -0,0 +1,132 @@
...
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (r *KataConfigOpenShiftReconciler) updateConfigMap(cmName string, namespace string, newData map[string]string) error {
```
bpradipt: @spotlesstofu would it make sense to move this method to utils.go as commonly used methods are kept in that file and this method could be useful for other featuregates as well ?

spotlesstofu:
Force-pushed new changes:
- Add description to main commit message.
- Move `updateConfigMap` to `utils.go`.
- Remove `if EnablePeerPods` from `confidential_handler.go`. It was redundant, it's already in `fg_handler.go`.

bpradipt (approved):
/lgtm
Thanks @spotlesstofu

snir911 (approved):
Thanks @spotlesstofu! (Assuming the CI failures unrelated :) )

bpradipt:
@spotlesstofu the `/prow/images` CI failure is genuine one. Please update the Dockerfile to remove this line as it's deleted due to code reorg- https://github.com/openshift/sandboxed-containers-operator/blob/devel/Dockerfile#L18

bundle-custom.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Use OpenShift golang builder image
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
```
gkurz, line 2: @cpmeadors Can you remind where `bundle-custom.Dockerfile` is used ?

Dockerfile
```diff
@@ -15,7 +15,6 @@ COPY main.go main.go
 COPY api api/
 COPY config config/
 COPY controllers controllers/
-COPY internal internal/
```
gkurz, line 18: This change should go to https://github.com/openshift/sandboxed-containers-operator/pull/417/commits/e01db87b68081a42f4484c360ebbef9fb515822c which is actually removing the directory. It is important to preserve `git bisect` when we can.

bundle-custom.Dockerfile
spotlesstofu: I saw [this test](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_sandboxed-containers-operator/417/pull-ci-openshift-sandboxed-containers-operator-devel-ci-index-openshift-sandboxed-containers-operator-bundle/1803708952588849152) was failing because of it

Dockerfile
spotlesstofu: Oh, I didn't realize that

bundle-custom.Dockerfile
gkurz: The test you mention is failing because of `COPY internal internal/` step, not because of the golang builder.
Ideally this should go to some separate PR but it is acceptable to make it a preparatory commit here.

controllers/confidential_handler.go
```diff
@@ -0,0 +1,93 @@
...
+func (r *KataConfigOpenShiftReconciler) disableFeatureConfidential() error {
+
+	if err := InitializeImageGenerator(r.Client); err != nil {
+		return err
+	}
+	ig := GetImageGenerator()
+
+	if ig.provider == unsupportedCloudProvider {
+		r.Log.Info("unsupported cloud provider, skipping confidential image configuration")
+	} else {
```
gkurz, line 65: `enableFeatureConfidential()` has the very same leading lines...

controllers/confidential_handler.go
```diff
@@ -0,0 +1,93 @@
...
+	peerpodsCMData := map[string]string{"DISABLECVM": "true"}
+	if err := updateConfigMap(r.Client, r.Log, peerpodsCMName, OperatorNamespace, peerpodsCMData); err != nil {
+		if k8serrors.IsNotFound(err) {
+			// Nothing to do, feature is disabled and configMap doesn't exist.
+		} else {
+			return err
+		}
+	}
+
+	return nil
```
gkurz, line 77: ... and, the comment let aside, this is again the same code as in `enableFeatureConfidential()`.

Either factor out the duplicate code to separate functions or come up with a single `toggleFeatureConfidential()` feature with bool argument.

bundle-custom.Dockerfile
gkurz: > That was left out by previous golang bump (https://github.com/openshift/sandboxed-containers-operator/commit/5accd071a837dc74c789dae0fd0de8f3cb6046a7).
>
> This change is not necessary for this pull request. I'm doing it because I'm touching
> the same file in the next commit.

Excellent commit message ! :+1:

controllers/confidential_handler.go
spotlesstofu: That's how I did it in the beginning, then I split to two functions (enabled/disabled) because I understood the code wasn't clear enough. Happy to know that my first approach was appreciated. I just force-pushed with the refactoring back to a single function.

gkurz: Assuming you're referring to https://github.com/openshift/sandboxed-containers-operator/pull/417#discussion_r1636330904, I had indeed difficulties to review the single function when I looked at it at the time. Splitting the function like you did clearly helped me to understand the flow and spot the _code duplication_ :wink:

From a reviewing perspective, this means that this commit should have been split in two :

1. introduce enabled/disabled as separate functions so that paths are easier to follow
2. merge the functions which should just be a mechanical change that preserves the paths reviewed in the previous commit

Not asking to change the PR again since I could review 1. and https://github.com/openshift/sandboxed-containers-operator/compare/1b5e206665b282941dc55130ea9940f0b48c39a6..8a92c657a108e6edbd1ad094af6c5e65da722291 which is what 2. would have been.

gkurz (approved):
/lgtm

Thanks @spotlesstofu !

bpradipt:
Merging this as all reviews completed and the required CI is green

PR #418, opened 2024-06-06
Fixes for annotation handling related to pod vm image creation and deletion

bpradipt:


littlejawa (approved):
LGTM
Thanks @bpradipt

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -259,14 +259,35 @@ function add_ami_id_annotation_to_peer_pods_cm() {
         return
     fi
 
-    # Add the image id as annotation to peer-pods-cm configmap
-    kubectl annotate configmap peer-pods-cm -n openshift-sandboxed-containers-operator \
+    # Add the ami id as annotation to peer-pods-cm configmap
+    # Overwrite any existing values
+    kubectl annotate --overwrite configmap peer-pods-cm -n openshift-sandboxed-containers-operator \
```
gkurz, line 264: IIUC the issue was caused by the lack of code to delete the annotation (while we have it for Azure BTW :wink:). This commit adds the missing code. Do we still need to `--overwrite` ?

gkurz: After some private discussion with @bpradipt on slack, it turns out that this isn't strictly needed for operator created/deleted images. It doesn't hurt though and ensures that `add_ami_id_annotation_to_peer_pods_cm()` will succeed even in the case the peer pods CM was tampered externally.

gkurz (approved):
/lgtm.

Thanks @bpradipt !

PR #421, opened 2024-06-11
Add libvirt podvm image support

supritai:
Add libvirt image creation job and libvirt image type for image generation.

**- What I did**

**- How to verify it**

### DIFF
--- a/config/peerpods/peerpodssecret.yaml
+++ b/config/peerpods/peerpodssecret.yaml
@@ -9,3 +9,5 @@ stringData:
   #LIBVIRT_URI: "qemu+ssh://root@192.168.122.1/system?no_verify=1"
   #LIBVIRT_NET: "default
   #LIBVIRT_POOL: "default"
+  #REDHAT_OFFLINE_TOKEN: "" #Required to download rhel base image : Download token from https://access.redhat.com/management/api
+  
--- a/config/peerpods/podvm/Dockerfile.podvm-builder
+++ b/config/peerpods/podvm/Dockerfile.podvm-builder
@@ -12,7 +12,7 @@ LABEL kata_src_commit=stable-3.6
 
 RUN mkdir -p /scripts
 
-ADD lib.sh aws-podvm-image-handler.sh azure-podvm-image-handler.sh  /scripts/
+ADD lib.sh libvirt-podvm-image-handler.sh aws-podvm-image-handler.sh azure-podvm-image-handler.sh /scripts/
 
 RUN /scripts/azure-podvm-image-handler.sh -- install_rpms
 
--- a/config/peerpods/podvm/lib.sh
+++ b/config/peerpods/podvm/lib.sh
@@ -29,6 +29,7 @@ function install_rpm_packages() {
         "make"
         "unzip"
         "skopeo"
+        "jq"
     )
 
     # Create a new array to store rpm packages that are not installed
@@ -65,20 +66,28 @@ function install_rpm_packages() {
 # are available in the variable REQUIRED_BINARY_PACKAGES
 # the function will download the packages, extract them and install them in /usr/local/bin
 # Following are the packages that are installed:
-# TBD: add multi-arch support for these binaries
 #"packer=https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip"
 #"kubectl=https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.14.9/openshift-client-linux.tar.gz"
 #"yq=https://github.com/mikefarah/yq/releases/download/v4.35.2/yq_linux_amd64.tar.gz"
 #"umoci=https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64"
 
 install_binary_packages() {
-    # Define the required binary packages
-    REQUIRED_BINARY_PACKAGES=(
-        "packer=https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip"
-        "kubectl=https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.14.9/openshift-client-linux.tar.gz"
-        "yq=https://github.com/mikefarah/yq/releases/download/v4.35.2/yq_linux_amd64.tar.gz"
-        "umoci=https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64"
-    )
+    ARCH=$(uname -m)
+    if [ "${ARCH}" == "s390x" ]; then
+        # Define the required binary packages
+        REQUIRED_BINARY_PACKAGES=(
+            "kubectl=https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp/4.14.9/openshift-client-linux.tar.gz"
+            "yq=https://github.com/mikefarah/yq/releases/download/v4.35.1/yq_linux_s390x.tar.gz"
+        )
+    else
+        # Define the required binary packages
+        REQUIRED_BINARY_PACKAGES=(
+            "packer=https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip"
+            "kubectl=https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.14.9/openshift-client-linux.tar.gz"
+            "yq=https://github.com/mikefarah/yq/releases/download/v4.35.2/yq_linux_amd64.tar.gz"
+            "umoci=https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64"
+        )
+    fi
 
     # Specify the installation directory
     local install_dir="/usr/local/bin"
@@ -110,9 +119,9 @@ install_binary_packages() {
             fi
 
             echo "Marking  ${install_dir}/${package_name} executable"
-            # yq extracted file name is yq_linux_amd64. Rename it
+            # yq extracted file name is yq_linux_${ARCH}. Rename it
             [[ "${package_name}" == "yq" ]] &&
-                mv "${install_dir}/yq_linux_amd64" "${install_dir}/yq"
+                mv "${install_dir}"/yq_linux_"${ARCH/x86_64/amd64}" "${install_dir}"/yq
 
             chmod +x "${install_dir}/${package_name}" ||
                 error_exit "Failed to mark ${package_name} executable"
--- a/config/peerpods/podvm/libvirt-podvm-image-cm.yaml
+++ b/config/peerpods/podvm/libvirt-podvm-image-cm.yaml
@@ -1,0 +1,24 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: libvirt-podvm-image-cm
+  namespace: openshift-sandboxed-containers-operator
+data:
+  # PodVM image distro
+  PODVM_DISTRO: "rhel"
+
+  # Pod VM sources
+  # If changing the source, then ensure the respective payload binaries are available
+  # for the new source
+  CAA_SRC: "https://github.com/confidential-containers/cloud-api-adaptor"
+  CAA_REF: "v0.8.2"
+
+  # Booleans
+  DOWNLOAD_SOURCES: "no"
+  CONFIDENTIAL_COMPUTE_ENABLED: "no"
+  UPDATE_PEERPODS_CM: "yes"
+
+  # Libvirt specific
+  ORG_ID: ""
+  ACTIVATION_KEY: ""
+  BASE_OS_VERSION: "9.2"
--- a/config/peerpods/podvm/libvirt-podvm-image-handler.sh
+++ b/config/peerpods/podvm/libvirt-podvm-image-handler.sh
@@ -1,0 +1,280 @@
+#!/bin/bash
+# FILEPATH: libvirt-podvm-image-handler.sh
+
+# This script is used to create or delete libvirt qcow2 image for podvm
+# The basic assumption is that the required variables are set as environment variables in the pod
+# Typically the variables are read from configmaps and set as environment variables in the pod
+# The script will be called with one of the following options:
+# Create image (-c)
+# Delete image (-C)
+
+set -x
+# include common functions from lib.sh
+# shellcheck source=/dev/null
+# The directory is where libvirt-podvm-image-handler.sh is located
+source /scripts/lib.sh
+
+# Function to verify that the required variables are set
+
+function verify_vars() {
+    # Ensure CLOUD_PROVIDER is set to libvirt
+    [[ -z "${CLOUD_PROVIDER}" || "${CLOUD_PROVIDER}" != "libvirt" ]] && error_exit "CLOUD_PROVIDER is empty or not set to libvirt"
+
+    [[ -z "${ORG_ID}" ]] && error_exit "ORG_ID is not set"
+    [[ -z "${ACTIVATION_KEY}" ]] && error_exit "ACTIVATION_KEY is not set"
+    [[ -z "${BASE_OS_VERSION}" ]] && error_exit "BASE_OS_VERSION is not set"
+
+    [[ -z "${PODVM_DISTRO}" ]] && error_exit "PODVM_DISTRO is not set"
+
+    [[ -z "${CAA_SRC}" ]] && error_exit "CAA_SRC is empty"
+    [[ -z "${CAA_REF}" ]] && error_exit "CAA_REF is empty"
+
+    [[ -z "${REDHAT_OFFLINE_TOKEN}" ]] && error_exit "Redhat token is not set"
+
+    # Ensure booleans are set
+    [[ -z "${DOWNLOAD_SOURCES}" ]] && error_exit "DOWNLOAD_SOURCES is empty"
+}
+
+
+
+function create_libvirt_image() {
+    echo "Creating libvirt qcow2 image"
+
+    # If any error occurs, exit the script with an error message
+
+    if [[ "${DOWNLOAD_SOURCES}" == "yes" ]]; then
+        # Download source code from GitHub
+        download_source_code
+    fi
+
+    # Prepare the source code for building the libvirt
+    prepare_source_code
+
+    # Dowload the base rhel image for packer
+    download_rhel_kvm_guest_qcow2
+
+    # Prepare the pause image for embedding into the libvirt image
+    download_and_extract_pause_image "${PAUSE_IMAGE_REPO}" "${PAUSE_IMAGE_VERSION}" "${PAUSE_IMAGE_REPO_AUTH_FILE}" 
+
+    cd "${CAA_SRC_DIR}"/podvm || \
+	    error_exit "Failed to change directory to "${CAA_SRC_DIR}"/podvm"
+    LIBC=gnu make BINARIES= PAUSE_BUNDLE= image
+
+    export PODVM_IMAGE_PATH=/payload/podvm-libvirt.qcow2
+    cp -pr "${CAA_SRC_DIR}"/podvm/output/*.qcow2 "${PODVM_IMAGE_PATH}"
+
+    # Upload the created qcow2 to the volume
+    upload_libvirt_image
+
+    # Add the libvirt_volume_name to peer-pods-cm configmap
+    add_libvirt_vol_to_peer_pods_cm
+}
+
+# Function to dowload the rhel base image
+
+function download_rhel_kvm_guest_qcow2() {
+    ARCH=$(uname -m)
+    export ARCH
+
+    # Define the API endpoints
+    TOKEN_GENERATOR_URI=https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token
+    IMAGES_URI=https://api.access.redhat.com/management/v1/images/rhel/"${BASE_OS_VERSION}"/"${ARCH}"
+
+    filename="rhel-"${BASE_OS_VERSION}"-"${ARCH}"-kvm.qcow2"
+
+    token=$(curl "${TOKEN_GENERATOR_URI}" \
+        -d grant_type=refresh_token -d client_id=rhsm-api -d refresh_token="${REDHAT_OFFLINE_TOKEN}" | jq --raw-output .access_token)
+    images=$(curl -X 'GET' "${IMAGES_URI}" \
+        -H 'accept: application/json' -H "Authorization: Bearer "${token}"" | jq )
+
+    download_href=$(echo "${images}" | jq -r --arg fn "${filename}" '.body[] | select(.filename == $fn) | .downloadHref')
+
+    download_url=$(curl -X 'GET' "${download_href}" \
+        -H "Authorization: Bearer "${token}"" -H 'accept: application/json' | jq -r .body.href )
+
+    curl -X GET "${download_url}" -H "Authorization: Bearer "${token}"" --output rhel-"${BASE_OS_VERSION}"-"${ARCH}"-kvm.qcow2
+
+    cp -pr rhel-"${BASE_OS_VERSION}"-"${ARCH}"-kvm.qcow2 "${CAA_SRC_DIR}"/podvm/rhel-"${BASE_OS_VERSION}"-"${ARCH}"-kvm.qcow2
+
+    export IMAGE_URL="${CAA_SRC_DIR}"/podvm/rhel-"${BASE_OS_VERSION}"-"${ARCH}"-kvm.qcow2
+    export IMAGE_CHECKSUM=$(sha256sum "${IMAGE_URL}" | awk '{ print $1 }')
+
+}
+
+# Function to upload the qcow2 image to volume
+
+function upload_libvirt_image() {
+    echo "LIBVIRT_VOL_NAME: "${LIBVIRT_VOL_NAME}"" && echo "LIBVIRT_POOL: "${LIBVIRT_POOL}"" && \
+	    echo "LIBVIRT_URI: "${LIBVIRT_URI}"" && echo "PODVM_IMAGE_PATH: "${PODVM_IMAGE_PATH}"" 
+    echo "Starting to upload the image."
+    virsh -d 0 -c "${LIBVIRT_URI}" vol-upload --vol "${LIBVIRT_VOL_NAME}" "${PODVM_IMAGE_PATH}" --pool "${LIBVIRT_POOL}" --sparse
+    if [ $? -eq 0 ]; then
+        echo "Uploaded the image successfully"
+    fi
+}
+
+# Function to add the libvirt_volume_name in the peer-pods-cm configmap
+
+function add_libvirt_vol_to_peer_pods_cm(){
+    if [ "${UPDATE_PEERPODS_CM}" == "yes" ]; then
+
+        # Check if the peer-pods-cm configmap exists
+        if ! kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator >/dev/null 2>&1; then
+            echo "peer-pods-cm configmap does not exist. Skipping adding the libvirt volume name"
+            return
+        fi
+
+        # Add the libvirt image id to peer-pods-cm configmap
+        echo "Updating peer-pods-cm configmap with LIBVIRT_IMAGE_ID=${LIBVIRT_VOL_NAME}"
+        kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator \
+        --type merge -p "{\"data\":{\"LIBVIRT_IMAGE_ID\":\"${LIBVIRT_VOL_NAME}\"}}" ||
+            error_exit "Failed to add the libvirt image id to peer-pods-cm configmap"
+    fi
+}
+
+# function to delete the libvirt
+# LIBVIRT_POOL name must be set as an environment variable
+
+function delete_libvirt_image() {
+    echo "Deleting Libvirt image"
+
+    # Delete the Libvirt pool 
+    # If any error occurs, exit the script with an error message
+
+    # LIBVIRT_POOL shouldn't be empty
+    [[ -z "${LIBVIRT_POOL}" ]] && error_exit "LIBVIRT_POOL is empty"
+
+    # Delete the libvirt pool
+    echo "Deleting libvirt pool."
+    virsh -d 0 -c "${LIBVIRT_URI}" pool-destroy "${LIBVIRT_POOL}" ||
+        error_exit "Failed to destroy the libvirt pool"
+    
+    virsh -d 0 -c "${LIBVIRT_URI}" pool-undefine "${LIBVIRT_POOL}" ||
+        error_exit "Failed to undefine the libvirt pool"
+
+    echo "Deleted libvirt image successfully"
+
+    # Remove the libvirt_volume_name from peer-pods-cm configmap
+    delete_libvirt_vol_from_peer_pods_cm
+
+}
+
+# Function to delete the libvirt image id  from the peer-pods-cm configmap
+
+function delete_libvirt_vol_from_peer_pods_cm() {
+    echo "Deleting libvirt volume from peer-pods-cm configmap"
+
+    # Check if the peer-pods-cm configmap exists
+    if ! kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator >/dev/null 2>&1; then
+        echo "peer-pods-cm configmap does not exist. Skipping deleting the ami id"
+        return
+    fi
+
+    # Delete the libvirt image id from peer-pods-cm configmap
+    kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator \
+    --type merge -p "{\"data\":{\"LIBVIRT_IMAGE_ID\":\"\"}}" ||
+        error_exit "Failed to delete the libvirt image id from peer-pods-cm configmap"
+    echo "libvirt image id deleted from peer-pods-cm configmap successfully"
+}
+
+# display help message
+
+function display_help() {
+    echo "This script is used to create libvirt qcow2 for podvm"
+    echo "Usage: $0 [-c|-C] [-- install_binaries]"
+    echo "Options:"
+    echo "-c  Create image"
+    echo "-C  Delete image"
+}
+
+function install_packages(){
+
+    install_binary_packages
+
+    #Install other libvirt specific binaries packages
+    ARCH=$(uname -m)
+    export ARCH
+    if [[ -n "${ACTIVATION_KEY}" && -n "${ORG_ID}" ]]; then
+        subscription-manager register --org="${ORG_ID}" --activationkey="${ACTIVATION_KEY}"
+    fi
+    subscription-manager repos --enable codeready-builder-for-rhel-9-"${ARCH}"-rpms
+    dnf install -y gcc genisoimage qemu-kvm libvirt-client
+
+    GO_VERSION="1.21.9"
+    curl https://dl.google.com/go/go"${GO_VERSION}".linux-"${ARCH/x86_64/amd64}".tar.gz -o go"${GO_VERSION}".linux-"${ARCH/x86_64/amd64}".tar.gz && \
+    rm -rf /usr/local/go && tar -C /usr/local -xzf go"${GO_VERSION}".linux-"${ARCH/x86_64/amd64}".tar.gz && \
+    rm -f go"${GO_VERSION}".linux-"${ARCH/x86_64/amd64}".tar.gz
+    export PATH="/usr/local/go/bin:"${PATH}""
+    export GOPATH="/src"
+
+    if [ "${ARCH}" == "s390x" ]; then
+        # Build packer from source for s390x as there are no prebuilt binaries for the required packer version
+        PACKER_VERSION="v1.9.4"
+        git clone --depth 1 --single-branch https://github.com/hashicorp/packer.git -b "${PACKER_VERSION}"
+        cd packer
+        sed -i -- "s/ALL_XC_ARCH=.*/ALL_XC_ARCH=\"${ARCH}\"/g" scripts/build.sh
+	    sed -i -- "s/ALL_XC_OS=.*/ALL_XC_OS=\"Linux\"/g" scripts/build.sh
+        make bin && cp bin/packer /usr/local/bin/
+
+        # Build umoci from source for s390x as there are no prebuilt binaries
+        mkdir -p umoci
+        git clone https://github.com/opencontainers/umoci.git
+        cd umoci
+        make
+        cp -pr umoci /usr/local/bin/
+    fi
+
+    # set a correspond qemu-system-* named link to qemu-kvm
+    ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-"${ARCH}"
+
+    # Build cloud-utils package from source as prebuilt binary is not available
+    git clone https://github.com/canonical/cloud-utils
+    cd cloud-utils && make install
+
+}
+
+# main function
+
+if [ "$#" -eq 0 ]; then
+    display_help
+    exit 1
+fi
+
+if [ "$1" = "--" ]; then
+    shift
+    # Handle positional parameters
+    case "$1" in
+
+    install_binaries)
+        install_packages
+        ;;
+    *)
+        echo "Unknown argument: $1"
+        exit 1
+        ;;
+    esac
+else
+    while getopts "cCh" opt; do
+        verify_vars
+        case ${opt} in
+        c)
+            # Create the libvirt image
+            create_libvirt_image
+            ;;
+        C)
+            # Delete the libvirt image
+            delete_libvirt_image
+            ;;
+        h)
+            # Display help
+            display_help
+            exit 0
+            ;;
+        *)
+            # Invalid option
+            display_help
+            exit 1
+            ;;
+        esac
+    done
+fi
--- a/config/peerpods/podvm/osc-podvm-create-job.yaml
+++ b/config/peerpods/podvm/osc-podvm-create-job.yaml
@@ -49,16 +49,30 @@ spec:
             - configMapRef:
                 name: aws-podvm-image-cm
                 optional: true
+            - configMapRef:
+                name: libvirt-podvm-image-cm
+                optional: true
           command: ["/podvm-builder.sh", "create"]
           volumeMounts:
             - name: payload
               mountPath: /payload
             - name: regauth
-              mountPath: /tmp/regauth
+              mountPath: /tmp/regauth  
+            - name: ssh-key-secret
+              mountPath: "/root/.ssh/"
+              readOnly: true
+              optional: true             
       volumes:
         - name: payload
           emptyDir: {}
         - name: regauth
           secret:
             secretName: auth-json-secret
+        - name: ssh-key-secret
+          secret:
+            secretName: ssh-key-secret
+            items:
+            - key: id_rsa
+              path: "id_rsa"
+            defaultMode: 0400
       restartPolicy: Never
--- a/config/peerpods/podvm/osc-podvm-delete-job.yaml
+++ b/config/peerpods/podvm/osc-podvm-delete-job.yaml
@@ -27,6 +27,8 @@ spec:
               value: "" # Set this to the aws ami id to delete
             - name: IMAGE_ID
               value: "" # Set this to the azure image id to delete
+            - name: LIBVIRT_IMAGE_ID
+              value: "" # Set this to the libvirt image id to delete
           envFrom:
             - secretRef:
                 name: peer-pods-secret
@@ -39,6 +41,22 @@ spec:
             - configMapRef:
                 name: aws-podvm-image-cm
                 optional: true
+            - configMapRef:
+                name: libvirt-podvm-image-cm
+                optional: true
           command: ["/podvm-builder.sh", "delete", "-f"]
+          volumeMounts:
+            - name: ssh-key-secret
+              mountPath: "/root/.ssh/"
+              readOnly: true
+              optional: true
+      volumes:
+        - name: ssh-key-secret
+          secret:
+            secretName: ssh-key-secret
+            items:
+            - key: id_rsa
+              path: "id_rsa"
+            defaultMode: 0400
 
       restartPolicy: Never
--- a/config/peerpods/podvm/podvm-builder.sh
+++ b/config/peerpods/podvm/podvm-builder.sh
@@ -17,6 +17,13 @@ function install_aws_deps() {
   /scripts/aws-podvm-image-handler.sh -- install_binaries
 }
 
+# Function to install libvirt deps
+function install_libvirt_deps() {
+  echo "Installing libvirt deps"
+  # Install the required packages
+  /scripts/libvirt-podvm-image-handler.sh -- install_binaries
+}
+
 # Function to check if peer-pods-cm configmap exists
 function check_peer_pods_cm_exists() {
   if kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator >/dev/null 2>&1; then
@@ -76,8 +83,19 @@ function create_podvm_image() {
       kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator --type merge -p "{\"data\":{\"PODVM_AMI_ID\":\"${AMI_ID}\"}}"
     fi
     ;;
+  libvirt)
+    echo "Creating Libvirt qcow2"
+    /scripts/libvirt-podvm-image-handler.sh -c
+    if [ "${UPDATE_PEERPODS_CM}" == "yes" ]; then
+      # Check if peer-pods-cm configmap exists
+      if ! check_peer_pods_cm_exists; then
+        echo "peer-pods-cm configmap does not exist. Skipping the update of peer-pods-cm"
+        exit 1
+      fi
+    fi
+    ;;
   *)
-    echo "CLOUD_PROVIDER is not set to azure or aws"
+    echo "CLOUD_PROVIDER is not set to azure or aws or libvirt"
     exit 1
     ;;
   esac
@@ -182,8 +200,41 @@ function delete_podvm_image() {
     fi
 
     ;;
+    libvirt)
+    # If LIBVIRT_IMAGE_ID is not set, then exit
+    if [ -z "${LIBVIRT_IMAGE_ID}" ]; then
+      echo "LIBVIRT_IMAGE_ID is not set. Skipping the deletion of libvirt image"
+      exit 1
+    fi
+
+    LIBVIRT_IMAGE=$(kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator -o jsonpath='{.data.LIBVIRT_IMAGE_ID}')
+
+    # If LIBVIRT_IMAGE is not set, then exit
+    if [ -z "${LIBVIRT_IMAGE}" ]; then
+      echo "LIBVIRT_IMAGE_ID is not set in peer-pods-cm. Skipping the deletion of Libvirt image"
+      exit 1
+    fi
+
+    # check if the LIBVIRT_IMAGE value in peer-pods-cm is same as the input LIBVIRT_IMAGE_ID
+    # If yes, then don't delete the image unless force option is provided
+    if [ "${LIBVIRT_IMAGE_ID}" == "${LIBVIRT_IMAGE}" ]; then
+      if [ "$1" != "-f" ]; then
+        echo "LIBVIRT_IMAGE_ID in peer-pods-cm is same as the input image to be deleted. Skipping the deletion of Libvirt image"
+        exit 0
+      fi
+    fi
+
+    echo "Deleting Libvirt image id"
+    /scripts/libvirt-podvm-image-handler.sh -C
+
+    # Update the peer-pods-cm configmap and remove the LIBVIRT_IMAGE_ID value
+    if [ "${UPDATE_PEERPODS_CM}" == "yes" ]; then
+      kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator --type merge -p "{\"data\":{\"LIBVIRT_IMAGE_ID\":\"\"}}"
+    fi
+
+    ;;
   *)
-    echo "CLOUD_PROVIDER is not set to azure or aws"
+    echo "CLOUD_PROVIDER is not set to azure or aws or libvirt"
     exit 1
     ;;
   esac
@@ -227,7 +278,7 @@ function display_usage() {
   echo "Usage: $0 {create|delete [-f] [-g]|delete-gallery [-f]}"
 }
 
-# Check if CLOUD_PROVIDER is set to azure or aws
+# Check if CLOUD_PROVIDER is set to azure or aws or libvirt
 # Install the required dependencies
 case "${CLOUD_PROVIDER}" in
 azure)
@@ -236,8 +287,11 @@ azure)
 aws)
   install_aws_deps
   ;;
+libvirt)
+  install_libvirt_deps
+  ;;
 *)
-  echo "CLOUD_PROVIDER is not set to azure or aws"
+  echo "CLOUD_PROVIDER is not set to azure or aws or libvirt"
   display_usage
   exit 1
   ;;
--- a/controllers/image_generator.go
+++ b/controllers/image_generator.go
@@ -51,10 +51,12 @@ const (
 	peerpodsCMName                = "peer-pods-cm"
 	peerpodsCMAWSImageKey         = "PODVM_AMI_ID"
 	peerpodsCMAzureImageKey       = "AZURE_IMAGE_ID"
+	peerpodsLibvirtImageKey       = "LIBVIRT_IMAGE_ID"
 	fipsCMKey                     = "BOOT_FIPS"
 	procFIPS                      = "/proc/sys/crypto/fips_enabled"
 	AWSProvider                   = "aws"
 	AzureProvider                 = "azure"
+	LibvirtProvider               = "libvirt"
 	peerpodsImageJobsPathLocation = "/config/peerpods/podvm"
 	azureImageGalleryPrefix       = "PodVMGallery"
 )
@@ -237,6 +239,10 @@ func newImageGenerator(client client.Client) (*ImageGenerator, error) {
 	case AzureProvider:
 		ig.CMimageIDKey = peerpodsCMAzureImageKey
 		ig.provider = provider
+	case LibvirtProvider:
+		igLogger.Info("libvirt is our provider", "provider", provider)
+		ig.CMimageIDKey = peerpodsLibvirtImageKey
+		ig.provider = provider
 	default:
 		igLogger.Info("unsupported cloud provider, image creation/deletion will be disabled", "provider", ig.provider)
 		ig.provider = unsupportedCloudProvider
@@ -293,6 +299,11 @@ func (r *ImageGenerator) createJobFromFile(jobFileName string) (*batchv1.Job, er
 		} else if r.provider == AWSProvider {
 			job.Spec.Template.Spec.Containers[0].Env = append(job.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
 				Name:  "AMI_ID",
+				Value: imageId,
+			})
+		} else if r.provider == LibvirtProvider {
+			job.Spec.Template.Spec.Containers[0].Env = append(job.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
+				Name:  "LIBVIRT_IMAGE_ID",
 				Value: imageId,
 			})
 		}
@@ -382,6 +393,7 @@ func (r *ImageGenerator) getPeerPodsCM() (*corev1.ConfigMap, error) {
 // cloud provider specific podvm image config is present
 // azure-podvm-image-cm.yaml for Azure
 // aws-podvm-image-cm.yaml for AWS
+// libvirt-podvm-image-cm.yaml for Libvirt
 
 func (r *ImageGenerator) imageCreateJobRunner() (int, error) {
 	igLogger.Info("imageCreateJobRunner: Start")
@@ -462,6 +474,7 @@ func (r *ImageGenerator) imageCreateJobRunner() (int, error) {
 // cloud provider specific podvm image config is present
 // azure-podvm-image-cm.yaml for Azure
 // aws-podvm-image-cm.yaml for AWS
+// libvirt-podvm-image-cm.yaml for Libvirt
 // Successful deletion removes the image ID from peer-pods-cm
 
 func (r *ImageGenerator) imageDeleteJobRunner() (int, error) {
@@ -575,6 +588,10 @@ func (r *ImageGenerator) validatePeerPodsConfigs() error {
 	azureSecretKeys := []string{"AZURE_TENANT_ID", "AZURE_CLIENT_ID", "AZURE_CLIENT_SECRET", "AZURE_SUBSCRIPTION_ID"}
 	// azure ConfigMap Keys
 	azureConfigMapKeys := []string{"AZURE_RESOURCE_GROUP", "AZURE_REGION", "CLOUD_PROVIDER"}
+	// libvirt Secret Keys
+	libvirtSecretKeys := []string{"CLOUD_PROVIDER", "LIBVIRT_URI", "LIBVIRT_POOL", "LIBVIRT_VOL_NAME"}
+	// libvirt ConfigMap Keys
+	libvirtConfigMapKeys := []string{"CLOUD_PROVIDER"}
 
 	// Check for each cloud provider if respective ConfigMap keys are present in the peerPodsConfigMap
 	switch r.provider {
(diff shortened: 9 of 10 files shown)


config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -29,6 +27,9 @@ RUN if [[ -n "$CERT_RPM" ]] ; then \
 
 RUN git clone ${CAA_SRC} -b ${CAA_REF} /src/cloud-api-adaptor
 
+ADD rhel-guest-image-9.2-20230414.17.s390x.qcow2 /src/cloud-api-adaptor/src/cloud-api-adaptor/podvm/rhel-guest-image-9.2-20230414.17.s390x.qcow2
```
supritai: Need to change : download the image while running the job than bundling it along the builder image

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,34 @@
...
+  CONFIDENTIAL_COMPUTE_ENABLED: "no"
+
+  # Livirt specific
+  ORG_ID: ""
+  ACTIVATION_KEY: ""
+  YQ_VERSION : "v4.35.1"
+  YQ_CHECKSUM: "sha256:4e6324d08630e7df733894a11830412a43703682d65a76f1fc925aac08268a45"
+  ARCH: "s390x"
+  GO_VERSION: "1.21.9"
+  IMAGE_URL: "rhel-guest-image-9.2-20230414.17.s390x.qcow2"
```
supritai: Need to change : download the image while running the job than bundling it along the builder image

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+    echo "Starting to upload the image."
+    virsh -d 0 -c $LIBVIRT_URI vol-upload --vol $LIBVIRT_VOL_NAME $PODVM_IMAGE_PATH --pool $LIBVIRT_POOL --sparse
+    if [ $? -eq 0 ]; then
+    echo "Uploaded the image successfully" 
+    LIBVIRT_POOL_UUID=$(virsh -d 0 -c $LIBVIRT_URI pool-list --name --uuid | grep $LIBVIRT_POOL | awk '{print $1}')
+    echo "Updating peer-pods-cm configmap with IMAGE_ID=${LIBVIRT_POOL_UUID}" 
+    else 
+    echo "Couldn't upload the image." 
+    fi
+    sleep 120s
```
supritai: Remove it

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+    export PODVM_DISTRO=${PODVM_DISTRO}
+    export CLOUD_PROVIDER=${CLOUD_PROVIDER}
+    export ARCH=${ARCH}
+
+    cd "${CAA_SRC_DIR}"/src/cloud-api-adaptor/podvm || \
+	    error_exit "Failed to change directory to ${CAA_SRC_DIR}/src/cloud-api-adaptor/podvm"
+    make image
+    sleep 120s
+    cp -pr /src/cloud-api-adaptor/src/cloud-api-adaptor/podvm/output/*.qcow2 /src/cloud-api-adaptor/src/cloud-api-adaptor/podvm/output/libvirt-podvm.qcow2
+    sleep 120s
```
supritai: Remove it

config/peerpods/podvm/osc-podvm-create-job.yaml
```diff
@@ -13,9 +13,11 @@ spec:
     spec:
       # Add initContainers to pull the image from the registry and copy
       # /podvm-binaries.tar.gz /payload/podvm-binaries.tar.gz
+      imagePullSecrets:
+        - name: regcred
       initContainers:
         - name: copy
-          image: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9:1.7.0
+          image: quay.io/suprit_iti/openshift-sandboxed-containers/osc-podvm-payload-rhel9:latest
```
supritai: Change the image source to registry.redhat.io and use it once the changes related to Dockerfile is approved

config/peerpods/podvm/osc-podvm-create-job.yaml
```diff
@@ -29,12 +31,11 @@ spec:
         - name: create
           # Binaries like kubectl, packer and yq are expected to be under /usr/local/bin
           # podvm binaries are expected to be under /payload/podvm-binaries.tar.gz
-          image: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-builder-rhel9:1.7.0
+          image: quay.io/suprit_iti/openshift-sandboxed-containers/osc-podvm-builder-rhel9:testv29
```
supritai: Change the image source to registry.redhat.io and use it once the changes related to Dockerfile is approved

config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -29,6 +27,9 @@ RUN if [[ -n "$CERT_RPM" ]] ; then \
...
+
+
 ADD podvm-builder.sh /podvm-builder.sh
 
-ENTRYPOINT ["/podvm-builder.sh"]
+ENTRYPOINT ["/podvm-builder.sh"]
```
savitrilh: No change here, it is showing as changed. Can you correct this

controllers/image_generator.go
```diff
@@ -200,6 +202,7 @@ func ImageDelete(c client.Client) (int, error) {
 }
 
 func newImageGenerator(client client.Client) (*ImageGenerator, error) {
+	fmt.Println("triggering NewImageGenerator")
```
savitrilh: Remove the printf and add a log

controllers/utils.go
```diff
@@ -168,11 +168,12 @@ func getCloudProviderFromInfra(c client.Client) (string, error) {
...
 	}
-
 	if infrastructure.Status.PlatformStatus == nil {
-		return "", fmt.Errorf("Infrastructure.status.platformStatus is empty")
+		return "", fmt.Errorf("Infrastructure.status.platformStatus is empty and is not libvirt")
+	} else if string(infrastructure.Status.PlatformStatus.Type) == "None" {
+		// For libvirt provider, PlatformStatus.Type is empty string, so returning libvirt provider
+		return "libvirt", nil
 	}
-
```
savitrilh, line 176: don't remove newline

controllers/utils.go
```diff
@@ -168,11 +168,12 @@ func getCloudProviderFromInfra(c client.Client) (string, error) {
 	if err != nil {
 		return "", err
 	}
-
```
savitrilh, line 172: don't remove newline

config/peerpods/podvm/osc-podvm-create-job.yaml
```diff
@@ -49,11 +50,45 @@ spec:
...
+                    key: LIBVIRT_POOL
+                    name: LIBVIRT_POOL
+                optional: true
           command: ["/podvm-builder.sh", "create"]
           volumeMounts:
             - name: payload
               mountPath: /payload
+            - mountPath: "/root/.ssh/"
+              name: ssh-key-secret
+              readOnly: true
```
savitrilh, line 63: Cann't we do -mountpath as optional as it is not used by other cloud providers

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai55: done

controllers/utils.go
supritai55: ok

controllers/utils.go
supritai55: ok

config/peerpods/podvm/Dockerfile.podvm-builder
supritai55: ok

config/peerpods/podvm/Dockerfile.podvm-builder
supritai55: I will update the required changes in the next patch

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai55: I will update the required changes in the next patch

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai55: done

config/peerpods/podvm/osc-podvm-create-job.yaml
supritai55: done

config/peerpods/podvm/osc-podvm-create-job.yaml
supritai55: I will update the required changes in the next patch

config/peerpods/podvm/osc-podvm-create-job.yaml
supritai55: ok

controllers/image_generator.go
supritai55: ok

supritai55:
This PR is ready for review. Since I do not have write access, unable to remove the 'draft' tag. Can someone remove this tag please?

bpradipt:
> This PR is ready for review. Since I do not have write access, unable to remove the 'draft' tag. Can someone remove this tag please?

Since you are the PR creator, you should be able to change this "Ready".  I believe you raised this PR as a draft pull request.

bpradipt:
@supritai @savitrilh can you please rebase the PR..

supritai55:
> @supritai @savitrilh can you please rebase the PR..

Sure @bpradipt

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,34 @@
...
+  # for the new source
+  CAA_SRC: "https://github.com/confidential-containers/cloud-api-adaptor"
+  CAA_REF: "v0.8.2"
+
+  # Booleans
+  INSTALL_PACKAGES: "no"
+  DOWNLOAD_SOURCES: "no"
+  CONFIDENTIAL_COMPUTE_ENABLED: "no"
+
+  # Livirt specific
```
bpradipt: s/Livirt/Libvirt

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,34 @@
...
+  ORG_ID: ""
+  ACTIVATION_KEY: ""
+  YQ_VERSION : "v4.35.1"
+  YQ_CHECKSUM: "sha256:4e6324d08630e7df733894a11830412a43703682d65a76f1fc925aac08268a45"
+  ARCH: "s390x"
+  GO_VERSION: "1.21.9"
```
bpradipt: Is GO_VERSION needed here ?

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+    prepare_source_code
+
+    export PODVM_DISTRO=${PODVM_DISTRO}
+    export CLOUD_PROVIDER=${CLOUD_PROVIDER}
+    export ARCH=${ARCH}
+
+    cd "${CAA_SRC_DIR}"/src/cloud-api-adaptor/podvm || \
+	    error_exit "Failed to change directory to ${CAA_SRC_DIR}/src/cloud-api-adaptor/podvm"
+    make image
+    sleep 120s
```
bpradipt: What's the reason for sleep here ?

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+    cp -pr /src/cloud-api-adaptor/src/cloud-api-adaptor/podvm/output/*.qcow2 /src/cloud-api-adaptor/src/cloud-api-adaptor/podvm/output/libvirt-podvm.qcow2
+    sleep 120s
+    #mkdir -p /src/cloud-api-adaptor/src/cloud-api-adaptor/podvm/output
```
bpradipt: Leftover ?

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,34 @@
...
+  INSTALL_PACKAGES: "no"
+  DOWNLOAD_SOURCES: "no"
+  CONFIDENTIAL_COMPUTE_ENABLED: "no"
+
+  # Livirt specific
+  ORG_ID: ""
+  ACTIVATION_KEY: ""
+  YQ_VERSION : "v4.35.1"
+  YQ_CHECKSUM: "sha256:4e6324d08630e7df733894a11830412a43703682d65a76f1fc925aac08268a45"
+  ARCH: "s390x"
```
snir911: Have you tested with x86 also by any chance? i don't see a reason why it wouldn't work, it will allow easy testing.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+
+    # Install packages if INSTALL_PACKAGES is set to yes
+
+    if [[ "${INSTALL_PACKAGES}" == "yes" ]]; then
+
+        # Install required rpm packages
+        install_rpm_packages
+
+        # Install required binary packages
+        install_libvirt_binary_packages
```
snir911: This function is not defined

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+        # Download source code from GitHub
+        download_source_code
+    fi
+
+    # Prepare the source code for building the ami
+    prepare_source_code
+
+    export PODVM_DISTRO=${PODVM_DISTRO}
+    export CLOUD_PROVIDER=${CLOUD_PROVIDER}
+    export ARCH=${ARCH}
```
snir911: If exporting for this 3 vars is required I'd guess it's required also for the others no? (alternatively maybe exporting is not required?)

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+source /scripts/lib.sh
+
+# Function to verify that the required variables are set
+
+function verify_vars() {
+    # Ensure CLOUD_PROVIDER is set to libvirt
+    [[ -z "${CLOUD_PROVIDER}" || "${CLOUD_PROVIDER}" != "libvirt" ]] && error_exit "CLOUD_PROVIDER is empty or not set to libvirt"
+
+    [[ -z "${ORG_ID}" ]] && error_exit "ORG_ID is not set"
+    [[ -z "${ACTIVATION_KEY}" ]] && error_exit "ACTIVATION_KEY is not set"
```
snir911, line 24: are these always mandatory ?  i mean. is it possible that image is already registered in certain cases? or that the installing the pkgs doesn't really requires registering?

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+# libvirt_id must be set as an environment variable
+
+function delete_libvirt_using_id() {
+    echo "Deleting Libvirt image"
+
+    # Delete the AMI
+    # If any error occurs, exit the script with an error message
+
+    # LIBVIRT_ID shouldn't be empty
+    [[ -z "${LIBVIRT_ID}" ]] && error_exit "LIBVIRT_ID is empty"
```
snir911: where LIBVIRT_ID is coming from? how it's being set?

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+function display_help() {
+    echo "This script is used to create libvirt qcow2 for podvm"
+    echo "Usage: $0 [-c|-C] [-- install_binaries|install_rpms|install_cli]"
+    echo "Options:"
+    echo "-c  Create image"
+    echo "-C  Delete image"
+    echo "-R Recreate podvm-images configMap"
+}
+
+function install_packages(){
```
snir911, line 190: all of these are actually needed? are you building binaries from source or being based on payload container? I'd expect mostly the libvirt pkgs & packer to be needed

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,34 @@
...
+  GO_VERSION: "1.21.9"
+  IMAGE_URL: "rhel-guest-image-9.2-20230414.17.s390x.qcow2"
+  IMAGE_CHECKSUM: "c57928279b376f9711c9e4aaa1d79d9225eaeedac679cadbc02bd01ba01a4d75"
+  PROTOC_VERSION: "26.1"
+  PODVM_TAG: "v1.0"
+  PACKER_VERSION: "v1.9.4"
+  RUST_VERSION: "1.72.0"
```
snir911: See my comment on the installed pkgs, i suspect all of the versions are not needed

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+    if [[ -n "${ACTIVATION_KEY}" && -n "${ORG_ID}" ]]; then \
+    subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY}; \
```
snir911: nit, I think the style is commonly to indent the lines within conditionals

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,220 @@
...
+    #mkdir -p /src/cloud-api-adaptor/src/cloud-api-adaptor/podvm/output
+    export PODVM_IMAGE_PATH=/src/cloud-api-adaptor/src/cloud-api-adaptor/podvm/output/libvirt-podvm.qcow2
+    echo "LIBVIRT_VOL_NAME: $LIBVIRT_VOL_NAME" && echo "LIBVIRT_POOL: $LIBVIRT_POOL" && \
+	    echo "LIBVIRT_URI: $LIBVIRT_URI" && echo "PODVM_IMAGE_PATH: $PODVM_IMAGE_PATH" 
+    echo "Starting to upload the image."
+    virsh -d 0 -c $LIBVIRT_URI vol-upload --vol $LIBVIRT_VOL_NAME $PODVM_IMAGE_PATH --pool $LIBVIRT_POOL --sparse
+    if [ $? -eq 0 ]; then
+    echo "Uploaded the image successfully" 
+    LIBVIRT_POOL_UUID=$(virsh -d 0 -c $LIBVIRT_URI pool-list --name --uuid | grep $LIBVIRT_POOL | awk '{print $1}')
+    echo "Updating peer-pods-cm configmap with IMAGE_ID=${LIBVIRT_POOL_UUID}"
```
snir911: where's the actual update of the cm?

config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -32,4 +32,4 @@ ADD rhel-guest-image-9.2-20230414.17.s390x.qcow2 /src/cloud-api-adaptor/src/clou
 
 ADD podvm-builder.sh /podvm-builder.sh
 
-ENTRYPOINT ["/podvm-builder.sh"]
\ No newline at end of file
+ENTRYPOINT ["/podvm-builder.sh"]
```
snir911: nit, if possible pls squash the addressed comments and resolve them in github

snir911:
Thanks! I like the idea, added few comments

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai: Okay, will change

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai: Sorry, I have not tested on x86.

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai: Yes, As using this variable installing GO as part of 'config/peerpods/podvm/libvirt-podvm-image-handler.sh:128'

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: It looks mandatory for the builder image which I have been using. If the builder image registered, then no need of it. https://github.com/openshift/sandboxed-containers-operator/blob/a77c2e93f2e46d8e0b568f92a3ff51be938eab50/config/peerpods/podvm/Dockerfile.podvm-builder#L1

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: This is not required, removed it.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: These are used as environment variables for https://github.com/confidential-containers/cloud-api-adaptor/blob/db56a2d3b30ff7e42cc100e6616ef049cdcc03b8/src/cloud-api-adaptor/podvm/Dockerfile.podvm.rhel#L23

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: yes, removed

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: I will add function to add the same.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: LIBVIRT_POOL_UUID should be supplied as env variable

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: Binaries are taken from the payload. Yes, mostly libvirt related packages are need to be installed.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: sure, will indent wherever required

config/peerpods/podvm/libvirt-podvm-image-handler.sh
snir911: right, but i don't think it's required to export them as they are already defined

config/peerpods/podvm/libvirt-podvm-image-handler.sh
snir911: please remove all of those which aren't required (go, rust etc..)

config/peerpods/podvm/Dockerfile.podvm-builder
supritai: I will do

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: go is required for building the some of the packages:
Ref: https://github.com/confidential-containers/cloud-api-adaptor/blob/0480804f8f7d06202a5a18fef3a966ec6e4e59c7/src/cloud-api-adaptor/podvm/Makefile.inc#L125

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: Right, will remove them

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: rust not needed, removed

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,273 @@
...
+    echo "Adding libvirt uuid to peer-pods-cm configmap"
+
+    # Check if the peer-pods-cm configmap exists
+    if ! kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator >/dev/null 2>&1; then
+        echo "peer-pods-cm configmap does not exist. Skipping adding the libvirt uuid"
+        return
+    fi
+
+    # Add the libvirt pool uuid as annotation to peer-pods-cm configmap
+    kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator \
```
supritai: @snir911 awa/azure are adding as the annotation : https://github.com/openshift/sandboxed-containers-operator/blob/b13fc7111e920795df3a6f7a9e6e734cfa2539a9/config/peerpods/podvm/aws-podvm-image-handler.sh#L263 
But getImageID method looks for Data in configmap. Not sure how it is works.
https://github.com/openshift/sandboxed-containers-operator/blob/b13fc7111e920795df3a6f7a9e6e734cfa2539a9/controllers/image_generator.go#L556
I have patched as Data field in configmap.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
snir911: > go is required for building the some of the packages:
Ref: https://github.com/confidential-containers/cloud-api-adaptor/blob/0480804f8f7d06202a5a18fef3a966ec6e4e59c7/src/cloud-api-adaptor/podvm/Makefile.inc#L125

The forwarder should be included in the payload image, hence, it should not be built as part of creating the image process, if it is, something might have been missed in the process of building the payload image

config/peerpods/podvm/libvirt-podvm-image-handler.sh
snir911: Right, in Azure and AWS the IMAGE_ID in the CM is used in order to identify the created podvm image, however, in libvirt, the equivalent is the volume name, which by default is "podvm-base.qcow2", unless LIBVIRT_VOL_NAME is set.
Actually i think you should drop LIBVIRT_POOL_UUID completely in favor of LIBVIRT_VOL_NAME, as the uuid is not consumed by CAA. 
cc @bpradipt for advise

config/peerpods/peerpodssecret.yaml
```diff
@@ -9,3 +9,4 @@ stringData:
   #LIBVIRT_URI: "qemu+ssh://root@192.168.122.1/system?no_verify=1"
   #LIBVIRT_NET: "default
   #LIBVIRT_POOL: "default"
+  #REDHAT_OFFLINE_TOKEN: ""
```
bpradipt: What is this token for? Please add a comment describe the use of this token.

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,31 @@
...
+  # Booleans
+  DOWNLOAD_SOURCES: "no"
+  CONFIDENTIAL_COMPUTE_ENABLED: "no"
+
+  # Libvirt specific
+  ORG_ID: ""
+  ACTIVATION_KEY: ""
+  YQ_VERSION : "v4.35.1"
+  YQ_CHECKSUM: "sha256:4e6324d08630e7df733894a11830412a43703682d65a76f1fc925aac08268a45"
+  ARCH: "s390x"
```
bpradipt: Why to set ARCH default to `s390x` here? Also is ARCH needed?  
Can't it be derived in the handler itself by running `uname -m` ?

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,31 @@
...
+  GO_VERSION: "1.21.9"
+  PROTOC_VERSION: "26.1"
+  PODVM_TAG: "v1.0"
+  PACKER_VERSION: "v1.9.4"
+  OS_VERSION: "9.2"
```
bpradipt: I don't think PROTOC and PACKER version is needed here. PROTOC will be used for binary builds and will be part of payload. PACKER version is not configurable as we need to stick to 1.9.4

bpradipt: Also is OS_VERSION the base OS version? if yes, then may be the var can be called as BASE_OS_VERSION or similar ?

config/peerpods/podvm/libvirt-podvm-image-handler.sh
bpradipt: I agree.. if LIBVIRT_POOL_UUID is not used, then there is no point in adding.
If this will be required later to support per pod VM image, it can be added later

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,301 @@
...
+    fi
+
+    subscription-manager repos --enable codeready-builder-for-rhel-9-${ARCH}-rpms; \
+    dnf install -y libvirt-client; \
+    dnf groupinstall -y 'Development Tools'; \
+    dnf install -y jq; \
+    dnf install -y yum-utils gnupg git --allowerasing curl pkg-config clang perl libseccomp-devel gpgme-devel \
+    device-mapper-devel qemu-kvm unzip wget libassuan-devel genisoimage cloud-utils-growpart cloud-init;
+
+    curl -L -o /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_${ARCH/x86_64/amd64} \
```
bpradipt: I would recommend updating the install_binary_packages() in lib.sh to support multi-arch.
For binaries which you need to build may be have a separate method in lib.sh and use it in the handler.

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
snir911: Is PROTOC_VERSION needed? AFAIU we never build binaries at runtime no?

bpradipt: Right

config/peerpods/podvm/libvirt-podvm-image-handler.sh
bpradipt: If some binaries can be built downstream and made part of the builder image it will be good. Otherwise the process will be so time consuming. Wdyt @snir911 ?

gkurz:
General remarks :
- I see multiple merge commits in this PR. Please reorganize this PR to be a list of regular commits on top of current devel.
- Please amend all commits to have proper description and Signed-off-by tags

config/peerpods/podvm/libvirt-podvm-image-handler.sh
snir911: I agree, however, @supritai said binaries are taken from payload image, hence, I'm not sure what is left to be built here, IIUC most of the preparations in this file can be dropped (these are taken from the upstream Dockerfile.podvm_builder*)

config/peerpods/peerpodssecret.yaml
supritai: This is to download base rhel image, I will add the comment

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai: okay, removed ARCH var

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai: Yes not needed, removed

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: My bad, GO is not required there. It is required for umoci ( as need to be installed from source due to unavailability of s390x package)

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: Included only needed packages.

gkurz:
@supritai please do proper rebases instead of merging upstream devel again and again in your repo. We don't want this complexity in our `git log`. A PR should be a list of consecutive regular non-merge commits that apply to the target branch.

Each commit must have a description that helps understanding the change and have proper DCO (Signed-off-by).

This is hard requirement for this PR to get merged in addition to the requests from reviewers.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
savitrilh: We need one Libvirt related ID( CMimageIDKey)  LIBVIRT_POOL_UUID or LIBVIRT_VOL_NAME. This check is  required whether to  skip or perform the image configuration operation. 

We thought to use LIBVIRT_POOL_UUID_<Vol NAME> as a CMimageIDKey to differentiate between different Libvirt images. We planned to do corresponding changes in CAA also. 
This decided this way because two different LibVirt Pool can have same Volume names. This differentiation will be useful to support multiple PodVM images.

Now we will set CMimageIDKey = LIBVIRT_VOL_NAME for libVirt provider. In future, we will change as we said above.

Is it okay?

supritai:
> @supritai please do proper rebases instead of merging upstream devel again and again in your repo. We don't want this complexity in our `git log`. A PR should be a list of consecutive regular non-merge commits that apply to the target branch.
> 
> Each commit must have a description that helps understanding the change and have proper DCO (Signed-off-by).
> 
> This is hard requirement for this PR to get merged in addition to the requests from reviewers.

@gkurz Sure. I will do that. I am waiting for review to reach final stage so that I can squash all the commits.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
bpradipt: yeah sounds good. Let's use LIBVIRT_VOL_NAME for now

controllers/image_generator.go
```diff
@@ -295,6 +301,11 @@ func (r *ImageGenerator) createJobFromFile(jobFileName string) (*batchv1.Job, er
 				Name:  "AMI_ID",
 				Value: imageId,
 			})
+		} else if r.provider == LibvirtProvider {
+			job.Spec.Template.Spec.Containers[0].Env = append(job.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
+				Name:  "LIBVIRT_VOL_NAME",
```
bpradipt: nit 's/LIBVIRT_VOL_NAME/peerpodsLibvirtImageKey`

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,289 @@
...
+        -H "Authorization: Bearer $token" -H 'accept: application/json' | jq -r .body.href )
+
+    curl -X GET $download_url -H "Authorization: Bearer $token" --output rhel-$BASE_OS_VERSION-s390x-kvm.qcow2
+
+    cp -pr rhel-$BASE_OS_VERSION-s390x-kvm.qcow2 $CAA_SRC_DIR/podvm/rhel-$BASE_OS_VERSION-s390x-kvm.qcow2
+
+    export IMAGE_URL=$CAA_SRC_DIR/podvm/rhel-$BASE_OS_VERSION-s390x-kvm.qcow2
+    export IMAGE_CHECKSUM=$(sha256sum ${IMAGE_URL} | awk '{ print $1 }')
+
+}
```
snir911, line 102: nice! maybe use $ARCH instead of s390x for future muliarch support

config/peerpods/peerpodssecret.yaml
```diff
@@ -9,3 +9,4 @@ stringData:
   #LIBVIRT_URI: "qemu+ssh://root@192.168.122.1/system?no_verify=1"
   #LIBVIRT_NET: "default
   #LIBVIRT_POOL: "default"
+  #REDHAT_OFFLINE_TOKEN: "" #Required to download rhel base image
```
snir911: maybe mention also how to get it (https://access.redhat.com/management/api IIUC)

snir911: Will it support getting the image from container payload? if not, consider validating REDHAT_OFFLINE_TOKEN is set (IIUC it will be mandatory)

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
snir911: IIUC yq and go are used only in order build the tools right? (i.e. not peerpods/coco components) in that  case, i think it would be ok to use fixed versions in script and don't allow the user to set these.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
snir911: I'd check again if this is still mandatory as IIUC some of the pkgs installation were dropped

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,289 @@
...
+    ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-$(uname -m)
+
+    # cloud-utils package is not available for rhel.
+    git clone https://github.com/canonical/cloud-utils
+    cd cloud-utils && make install 
+    
+    # Install openshift-client
+    mkdir -p /tmp/openshift-client
+    curl https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp/4.2.13-s390x/openshift-client-linux-4.2.13-s390x.tar.gz -o /tmp/openshift-client/openshift-client-linux-4.2.13-s390x.tar.gz
+    tar -xvf /tmp/openshift-client/openshift-client-linux-4.2.13-s390x.tar.gz -C /tmp/openshift-client
```
snir911: nit, no pkg for openshfit client?

snir911 (commented):
thanks!

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai: GO, 'Development Tools'(gcc) - used for building umoci binary
yq, genisoimage and cloud-utils - for packer 
Agree, these can be fixed version than user setting up.

config/peerpods/peerpodssecret.yaml
supritai: No, I will add condition to check the variable

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: If the host is a subscribed, no need for subscription of the container.
If I remove the subscription and test :
While running the job, neither worker node nor container would be registered and package installation fails.
```
_This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 17:37:29 ago on Tue Jul 16 12:23:50 2024.
Module or Group 'Development Tools' is not available.
Error: Nothing to do.
++ dnf install -y genisoimage qemu-kvm libvirt-client
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 17:37:30 ago on Tue Jul 16 12:23:50 2024.
No match for argument: genisoimage
No match for argument: qemu-kvm
No match for argument: libvirt-client
```

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: okay

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: It is there. I have included in install_binary_packages() in lib.sh. I will remove this redundant code.

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,25 @@
...
+  CAA_REF: "v0.8.2"
+
+  # Booleans
+  DOWNLOAD_SOURCES: "no"
+  CONFIDENTIAL_COMPUTE_ENABLED: "no"
+
+  # Libvirt specific
+  ORG_ID: ""
+  ACTIVATION_KEY: ""
+  PODVM_TAG: "v1.0"
```
snir911: redundant?

config/peerpods/podvm/lib.sh
```diff
@@ -72,13 +73,22 @@ function install_rpm_packages() {
...
-        "packer=https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip"
-        "kubectl=https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.14.9/openshift-client-linux.tar.gz"
-        "yq=https://github.com/mikefarah/yq/releases/download/v4.35.2/yq_linux_amd64.tar.gz"
-        "umoci=https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64"
-    )
+    ARCH=$(uname -m)
+    if [ "${ARCH}" == "s390x" ]; then
+        # Define the required binary packages
+        REQUIRED_BINARY_PACKAGES=(
+            "kubectl=https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp/4.2.13-s390x/openshift-client-linux-4.2.13-s390x.tar.gz"
```
bpradipt: You can use a similar url like x86_64 with just the arch being different - https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp/4.14.9/openshift-client-linux.tar.gz

config/peerpods/podvm/lib.sh
```diff
@@ -110,9 +120,9 @@ install_binary_packages() {
             fi
 
             echo "Marking  ${install_dir}/${package_name} executable"
-            # yq extracted file name is yq_linux_amd64. Rename it
+            # yq extracted file name is yq_linux_${ARCH}. Rename it
             [[ "${package_name}" == "yq" ]] &&
-                mv "${install_dir}/yq_linux_amd64" "${install_dir}/yq"
+                mv "${install_dir}/yq_linux_${ARCH}" "${install_dir}/yq"
```
bpradipt: This will fail for x86_64 since ARCH will be `x86_64` whereas the expected filename has `amd64`

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -17,6 +17,13 @@ function install_aws_deps() {
   /scripts/aws-podvm-image-handler.sh -- install_binaries
 }
 
+# Function to install libvirt deps
+function install_libvirt_deps() {
+  echo "Installing libvirt deps"
+  # Install the required packages
+  . /scripts/libvirt-podvm-image-handler.sh -- install_binaries
```
bpradipt:
```suggestion
/scripts/libvirt-podvm-image-handler.sh -- install_binaries
```

config/peerpods/podvm/lib.sh
bpradipt:
```suggestion
            "kubectl=https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp/4.14.9/openshift-client-linux.tar.gz"
```

config/peerpods/podvm/osc-podvm-delete-job.yaml
```diff
@@ -27,6 +27,8 @@ spec:
               value: "" # Set this to the aws ami id to delete
             - name: IMAGE_ID
               value: "" # Set this to the azure image id to delete
+            - name: LIBVIRT_ID
```
bpradipt: I think you can use IMAGE_ID itself. LIBVIRT_ID sounds incorrect.

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -181,9 +199,42 @@ function delete_podvm_image() {
       kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator --type merge -p "{\"data\":{\"PODVM_AMI_ID\":\"\"}}"
     fi
 
+    ;;
+    libvirt)
+    # If LIBVIRT_ID is not set, then exit
+    if [ -z "${LIBVIRT_ID}" ]; then
```
bpradipt: you can use IMAGE_ID, which will be more apt

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -181,9 +199,42 @@ function delete_podvm_image() {
...
+      echo "LIBVIRT_ID is not set. Skipping the deletion of libvirt image"
+      exit 1
+    fi
+
+    LIBVIRT_IMAGE_ID=$(kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator -o jsonpath='{.data.LIBVIRT_IMAGE_ID}')
+
+    # If LIBVIRT_IMAGE_ID is not set, then exit
+    if [ -z "${LIBVIRT_IMAGE_ID}" ]; then
+      echo "PODVM_AMI_ID is not set in peer-pods-cm. Skipping the deletion of AWS AMI"
```
bpradipt:
```suggestion
echo "LIBVIRT_IMAGE_ID is not set in peer-pods-cm. Skipping the deletion of the libvirt image"
```

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -181,9 +199,42 @@ function delete_podvm_image() {
...
+      if [ "$1" != "-f" ]; then
+        echo "LIBVIRT_IMAGE_ID in peer-pods-cm is same as the input image to be deleted. Skipping the deletion of Libvirt image"
+        exit 0
+      fi
+    fi
+
+    echo "Deleting Libvirt image id"
+    /scripts/libvirt-podvm-image-handler.sh -C
+
+    # Update the peer-pods-cm configmap and remove the PODVM_AMI_ID value
```
bpradipt: s/PODVM_AMI_ID/LIBVIRT_IMAGE_ID

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -181,9 +199,42 @@ function delete_podvm_image() {
...
+
+    LIBVIRT_IMAGE_ID=$(kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator -o jsonpath='{.data.LIBVIRT_IMAGE_ID}')
+
+    # If LIBVIRT_IMAGE_ID is not set, then exit
+    if [ -z "${LIBVIRT_IMAGE_ID}" ]; then
+      echo "PODVM_AMI_ID is not set in peer-pods-cm. Skipping the deletion of AWS AMI"
+      exit 1
+    fi
+
+    # check if the LIBVIRT_IMAGE_ID value in peer-pods-cm is same as the input LIBVIRT_IMAGE_ID
```
bpradipt:
```suggestion
     # check if the LIBVIRT_IMAGE_ID value in peer-pods-cm is same as the input IMAGE_ID
```

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: libvirt-podvm-image-cm
+  namespace: openshift-sandboxed-containers-operator
+data:
+  CLOUD_PROVIDER: "libvirt"
```
bpradipt: not needed

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+    [[ -z "${ORG_ID}" ]] && error_exit "ORG_ID is not set"
+    [[ -z "${ACTIVATION_KEY}" ]] && error_exit "ACTIVATION_KEY is not set"
+    [[ -z "${BASE_OS_VERSION}" ]] && error_exit "BASE_OS_VERSION is not set"
+
+    [[ -z "${PODVM_DISTRO}" ]] && error_exit "PODVM_DISTRO is not set"
+
+    [[ -z "${CAA_SRC}" ]] && error_exit "CAA_SRC is empty"
+    [[ -z "${CAA_REF}" ]] && error_exit "CAA_REF is empty"
+
+    [[ -z "${REDHAT_OFFLINE_TOKEN}" ]] && error_exit "Redhat token is not set"
```
bpradipt, line 32: REDHAT_OFFLINE_TOKEN is not there in `libvirt-podvm-image-cm`

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+    echo "-c  Create image"
+    echo "-C  Delete image"
+}
+
+function install_packages(){
+
+    install_binary_packages
+
+    #Install other libvirt specific binaries packages
+    export ARCH=$(uname -m)
```
bpradipt: please split command o/p assignment and export into separate statements
```
ARCH=$(uname -m)
export ARCH
```

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+    if [[ -n "${ACTIVATION_KEY}" && -n "${ORG_ID}" ]]; then
+        subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY}
+    fi
+    subscription-manager repos --enable codeready-builder-for-rhel-9-${ARCH}-rpms
+    dnf install -y gcc genisoimage qemu-kvm libvirt-client
+
+    GO_VERSION="1.21.9"
+    curl https://dl.google.com/go/go${GO_VERSION}.linux-${ARCH/x86_64/amd64}.tar.gz -o go${GO_VERSION}.linux-${ARCH/x86_64/amd64}.tar.gz && \
```
bpradipt: please quote all shell vars eg `"${ORG_ID}"`

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+    PACKER_VERSION="v1.9.4"
+    if [ "${ARCH}" == "s390x" ]; then
+        git clone --depth 1 --single-branch https://github.com/hashicorp/packer.git -b ${PACKER_VERSION}
+        cd packer
+        sed -i -- "s/ALL_XC_ARCH=.*/ALL_XC_ARCH=\"${ARCH}\"/g" scripts/build.sh
+	    sed -i -- "s/ALL_XC_OS=.*/ALL_XC_OS=\"Linux\"/g" scripts/build.sh
+        make bin && cp bin/packer /usr/local/bin/
+        cd $OLDPWD
+    else
+        yum-config-manager --add-repo https://rpm.releases.hashicorp.com/RHEL/hashicorp.repo && \
```
bpradipt: packer is already being installed for x86_64 as part of install_binary_packages. So this is not needed here

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+        dnf install -y packer
+    fi
+
+    # set a correspond qemu-system-* named link to qemu-kvm
+    ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-$(uname -m)
```
bpradipt: You'll need to quote `"$(uname -m)` .. Applies to the entire code.

controllers/image_generator.go
```diff
@@ -295,6 +301,11 @@ func (r *ImageGenerator) createJobFromFile(jobFileName string) (*batchv1.Job, er
 				Name:  "AMI_ID",
 				Value: imageId,
 			})
+		} else if r.provider == LibvirtProvider {
+			job.Spec.Template.Spec.Containers[0].Env = append(job.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{
+				Name:  "LIBVIRT_IMAGE_ID",
```
bpradipt, line 306: s/LIBVIRT_IMAGE_ID/IMAGE_ID" based on previous comments w.r.to the var for delete job

controllers/image_generator.go
```diff
@@ -575,6 +588,10 @@ func (r *ImageGenerator) validatePeerPodsConfigs() error {
 	azureSecretKeys := []string{"AZURE_TENANT_ID", "AZURE_CLIENT_ID", "AZURE_CLIENT_SECRET", "AZURE_SUBSCRIPTION_ID"}
 	// azure ConfigMap Keys
 	azureConfigMapKeys := []string{"AZURE_RESOURCE_GROUP", "AZURE_REGION", "CLOUD_PROVIDER"}
+	// libvirt Secret Keys
+	libvirtSecretKeys := []string{"CLOUD_PROVIDER", "LIBVIRT_URI", "LIBVIRT_POOL", "LIBVIRT_VOL_NAME"}
+	// libvirt ConfigMap Keys
+	libvirtConfigMapKeys := []string{"CLOUD_PROVIDER", "PROXY_TIMEOUT"}
```
bpradipt: Do you need PROXY_TIMEOUT as mandatory key ?

config/peerpods/podvm/lib.sh
supritai: okay

config/peerpods/podvm/lib.sh
supritai: changed

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai: okay

config/peerpods/podvm/libvirt-podvm-image-cm.yaml
supritai: yes, removed

controllers/utils.go
```diff
@@ -171,7 +171,10 @@ func getCloudProviderFromInfra(c client.Client) (string, error) {
 	}
 
 	if infrastructure.Status.PlatformStatus == nil {
-		return "", fmt.Errorf("Infrastructure.status.platformStatus is empty")
+		return "", fmt.Errorf("Infrastructure.status.platformStatus is empty and is not libvirt")
+	} else if string(infrastructure.Status.PlatformStatus.Type) == "None" {
+		// For libvirt provider, PlatformStatus.Type is empty string, so returning libvirt provider
+		return "libvirt", nil
```
bpradipt:
```suggestion
                return LibvirtProvider, nil
```

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: REDHAT_OFFLINE_TOKEN it is in peerpod secret.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: okay

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: okay

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: But we need packer for s390 systems. Am I missing something?

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: ya, replaced with "$ARCH"

config/peerpods/podvm/osc-podvm-delete-job.yaml
supritai: okay

config/peerpods/podvm/podvm-builder.sh
supritai: okay

config/peerpods/podvm/podvm-builder.sh
supritai: okay

config/peerpods/podvm/podvm-builder.sh
supritai: ok

config/peerpods/podvm/podvm-builder.sh
supritai: sorry, missed it

controllers/utils.go
supritai: ok

controllers/image_generator.go
supritai: ok

config/peerpods/podvm/podvm-builder.sh
supritai: corrected

controllers/image_generator.go
supritai: It is required for cloud-api-adaptor

config/peerpods/podvm/libvirt-podvm-image-handler.sh
snir911: I assume this is sensitive value, wouldn't be better to keep in the  Secret?

controllers/image_generator.go
bpradipt: There is a default already set. You can try without this parameter in your env.

bpradipt:
@supritai once you have resolved a specific comment, can you please click on the associated "Resolve conversation" button, so that it's evident which comments have been addressed and what is still pending.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: removed it

controllers/image_generator.go
supritai: okay, removing it

config/peerpods/podvm/libvirt-podvm-image-handler.sh
bpradipt: You are already building packer here for s390x. When this code runs on x86_64, then the original install_binary_packages (which you call in install_packages) should install the packer binary. 
I don't understand why the hashicorp repo needs to be added here  ? What am I missing ?

supritai: Got your point. No need to hashicorp repo. Added condition to install additional packages for s390x only

supritai:
> @supritai once you have resolved a specific comment, can you please click on the associated "Resolve conversation" button, so that it's evident which comments have been addressed and what is still pending.

Sure @bpradipt , most of them I have marked as resolved. Only where agreement of reviewer needed are not closed

config/peerpods/podvm/lib.sh
```diff
@@ -110,9 +119,14 @@ install_binary_packages() {
             fi
 
             echo "Marking  ${install_dir}/${package_name} executable"
-            # yq extracted file name is yq_linux_amd64. Rename it
-            [[ "${package_name}" == "yq" ]] &&
-                mv "${install_dir}/yq_linux_amd64" "${install_dir}/yq"
+            # yq extracted file name is yq_linux_${ARCH}. Rename it
+            if [[ "${package_name}" == "yq" ]]; then
+                if [ "${ARCH}" == "s390x" ]; then
```
bpradipt: Can you please try with this one liner. 
`mv "${install_dir}"/yq_linux_"${ARCH/x86_64/amd64}" "${install_dir}"/yq`

bpradipt:
Thanks @supritai for addressing the comments. 
I added a minor suggestion to incorporate.  In case you have tried the latest code on s390x, can you post the test results.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+
+    GO_VERSION="1.21.9"
+    curl https://dl.google.com/go/go"${GO_VERSION}".linux-"${ARCH/x86_64/amd64}".tar.gz -o go"${GO_VERSION}".linux-"${ARCH/x86_64/amd64}".tar.gz && \
+    rm -rf /usr/local/go && tar -C /usr/local -xzf go"${GO_VERSION}".linux-"${ARCH/x86_64/amd64}".tar.gz && \
+    rm -f go"${GO_VERSION}".linux-"${ARCH/x86_64/amd64}".tar.gz
+    export PATH="/usr/local/go/bin:"${PATH}""
+    export GOPATH="/src"
+
+    if [ "${ARCH}" == "s390x" ]; then
+        # Install packer. Packer doesn't does not have prebuilt s390x arch binaries above Packer version 0.1.5
```
bpradipt:
```suggestion
        # Build packer from source for s390x as there is no prebuilt binary for the required packer version
```

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+    if [ "${ARCH}" == "s390x" ]; then
+        # Install packer. Packer doesn't does not have prebuilt s390x arch binaries above Packer version 0.1.5 
+        PACKER_VERSION="v1.9.4"
+        git clone --depth 1 --single-branch https://github.com/hashicorp/packer.git -b "${PACKER_VERSION}"
+        cd packer
+        sed -i -- "s/ALL_XC_ARCH=.*/ALL_XC_ARCH=\"${ARCH}\"/g" scripts/build.sh
+	    sed -i -- "s/ALL_XC_OS=.*/ALL_XC_OS=\"Linux\"/g" scripts/build.sh
+        make bin && cp bin/packer /usr/local/bin/
+
+        # Install umoci source as binary for s390x is not available
```
bpradipt:
```suggestion
        # Build umoci from source for s390x as there is no prebuilt binary
```

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+        git clone https://github.com/opencontainers/umoci.git
+        cd umoci
+        make
+        cp -pr umoci /usr/local/bin/
+    fi
+
+    # set a correspond qemu-system-* named link to qemu-kvm
+    ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-"${ARCH}"
+
+    # cloud-utils package is not available for rhel.
```
bpradipt:
```suggestion
    # Build cloud-utils package from source as prebuilt binary is not available
```

config/peerpods/podvm/osc-podvm-delete-job.yaml
```diff
@@ -27,6 +27,8 @@ spec:
               value: "" # Set this to the aws ami id to delete
             - name: IMAGE_ID
               value: "" # Set this to the azure image id to delete
+            - name: IMAGE_ID
+              value: "" # Set this to the libvirt image id to delete
```
snir911, line 31: Maybe LIBVIRT_VOL_NAME? because this is what will actually be deleted and to differ from Azure? (or LIBVIRT_IMAGE_ID to support future usage as you mentioned)

config/peerpods/podvm/libvirt-podvm-image-handler.sh
```diff
@@ -0,0 +1,280 @@
...
+    # Check if the peer-pods-cm configmap exists
+    if ! kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator >/dev/null 2>&1; then
+        echo "peer-pods-cm configmap does not exist. Skipping adding the libvirt volume name"
+        return
+    fi
+
+    # Add the libvirt image id to peer-pods-cm configmap
+    kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator \
+    --type merge -p "{\"data\":{\"LIBVIRT_IMAGE_ID\":\"${LIBVIRT_VOL_NAME}\"}}" ||
+        error_exit "Failed to add the libvirt image id to peer-pods-cm configmap"
```
snir911: I think UPDATE_PEERPODS_CM is ignored at creation time, shouldn't it avoid updating CM if set to false?

snir911 (commented):
added minor comments, overall LGTM

config/peerpods/podvm/lib.sh
supritai: sure, thanks it reduced the code

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: I added, thanks for pointing it

config/peerpods/podvm/osc-podvm-delete-job.yaml
supritai: Better to have LIBVIRT_IMAGE_ID now to avoid confusion of image and volume name.

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: sure

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: okay

config/peerpods/podvm/libvirt-podvm-image-handler.sh
supritai: okay

supritai (commented):
Attached the logs.
[osc-podvm-image-creation.log](https://github.com/user-attachments/files/16330389/osc-podvm-image-creation.log)
[osc-podvm-image-deletion.log](https://github.com/user-attachments/files/16330390/osc-podvm-image-deletion.log)

bpradipt (approved):
/lgtm
Thanks @supritai

snir911 (approved):
LGTM, thank you

savitrilh (approved):

PR #422, opened 2024-06-14
Changes for file based catalog in Makefile and DEVELOPMENT.md

tbuskey:
[KATA-3122](https://issues.redhat.com/browse/KATA-3122) Changes for FBC catalog to replace sqlite

**- What I did**
- Makefile
  - add an token variable for docker-build so the public images are not used
  - change the (downloaded) opm version to 1.28.0
  - change catalog-build for file based catalogs, replacing sqlite
- docs/DEVELOPMENT.md
  - explain how to use the token instead of public images
  - sed to change SANDBOXED_CONTAINERS_EXTENSION for openshift
  - where to add SKIP_TESTS

**- How to verify it**
Build per DEVELOPMENT.md
Create a catsrc using your catalog & install in your cluster
Install the operator in your cluster

Changes for file based catalog instead of sqlite

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,28 +267,21 @@ OPM = $(shell which opm)
 endif
 endif
 
-# A comma-separated list of bundle images (e.g. make catalog-build BUNDLE_IMGS=example.com/operator-bundle:v0.1.0,example.com/operator-bundle:v0.2.0).
-# These images MUST exist in a registry and be pull-able.
-BUNDLE_IMGS ?= $(BUNDLE_IMG)
-
 # The image tag given to the resulting catalog image (e.g. make catalog-build CATALOG_IMG=example.com/operator-catalog:v0.2.0).
 CATALOG_IMG ?= $(IMAGE_TAG_BASE)-catalog:v$(VERSION)
 
-# Set CATALOG_BASE_IMG to an existing catalog image tag to add $BUNDLE_IMGS to that image.
-ifneq ($(origin CATALOG_BASE_IMG), undefined)
-FROM_INDEX_OPT := --from-index $(CATALOG_BASE_IMG)
-endif
+# Validate & build a file based catalog image using opm.
+# https://docs.openshift.com/container-platform/4.16/operators/admin/olm-managing-custom-catalogs.html
+# https://github.com/operator-framework/community-operators/blob/master/docs/packaging-operator.md
 
-# Build a catalog image by adding bundle images to an empty catalog using the operator package manager tool, 'opm'.
-# This recipe invokes 'opm' in 'semver' bundle add mode. For more information on add modes, see:
-# https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
+# if IMAGE_TAG_BASE was modified or set in shell, see docs/DEVELOPMENT.md for recreating catalog.Dockerfile and catalog/index.yaml
 .PHONY: catalog-build
-catalog-build: opm ## Build a catalog image.
-	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
+catalog-build: opm
+	$(OPM) validate catalog
+	docker build . -f catalog.Dockerfile -t $(CATALOG_IMG)
 
-# Push the catalog image.
 .PHONY: catalog-push
-catalog-push: ## Push a catalog image.
+catalog-push:
 	$(MAKE) docker-push IMG=$(CATALOG_IMG)
 
 ##@ Cleanup
--- a/Makefile
+++ b/Makefile
@@ -259,7 +259,7 @@ ifeq (,$(shell which opm 2>/dev/null))
 	set -e ;\
 	mkdir -p $(dir $(OPM)) ;\
 	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
-	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.23.0/$${OS}-$${ARCH}-opm ;\
+	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.28.0/$${OS}-$${ARCH}-opm ;\
 	chmod +x $(OPM) ;\
 	}
 else
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,8 @@ endif
 # These images needs to be synced with the default values in the Dockerfile.
 BUILDER_IMAGE ?= registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16
 TARGET_IMAGE  ?= registry.ci.openshift.org/ocp/4.16:base
+# Set if token was obtained
+AUTH_FILE ?= ""
 
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # This is a requirement for 'setup-envtest.sh' in the test target.
@@ -145,7 +147,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build \
-		-t ${IMG} \
+		-t ${IMG} --authfile $(AUTH_FILE) \
 		--build-arg BUILDER_IMAGE=$(BUILDER_IMAGE) \
 		--build-arg TARGET_IMAGE=$(TARGET_IMAGE) \
 		.
--- a/catalog/index.yaml
+++ b/catalog/index.yaml
@@ -1,0 +1,243 @@
+---
+defaultChannel: stable
+description: "<!-- TOC start -->\n- [Introduction to sandboxed containers](#introduction-to-sandboxed-containers)\n
+  \ * [Features & benefits of sandboxed containers](#features-benefits-of-sandboxed-containers)\n-
+  [OpenShift sandboxed containers Operator](#openshift-sandboxed-containers-operator)\n
+  \ * [Operator Architecture](#operator-architecture)\n  * [KataConfig Custom Resource
+  Definition](#kataconfig-custom-resource-definition)\n  * [Getting Started](#getting-started)\n-
+  [Operator Development](#operator-development)\n- [Demos](#demos)\n- [Further Reading](#further-reading)\n<!--
+  TOC end -->\n\n\n## Introduction to sandboxed containers\n\n[OpenShift sandboxed
+  containers](https://www.redhat.com/en/openshift-sandboxed-containers), based on
+  the [Kata Containers](https://katacontainers.io/) open source project, provides
+  an Open Container Initiative (OCI) compliant container runtime using lightweight
+  virtual machines, running your workloads in their own isolated kernel and therefore
+  contributing an additional layer of isolation back to OpenShift’s Defense-in-Depth
+  strategy. \n\n### Features & benefits of sandboxed containers\n\n- **Isolated Developer
+  Environments & Privileges Scoping** \nAs a developer working on debugging an application
+  using state-of-the-art tooling you might need elevated privileges such as `CAP_ADMIN`
+  or `CAP_BPF`. With OpenShift sandboxed containers, any impact will be limited to
+  a separate dedicated kernel.\n\n- **Legacy Containerized Workload Isolation** \nYou
+  are mid-way in converting a containerized monolith into cloud-native microservices.
+  However, the monolith still runs on your cluster unpatched and unmaintained. OpenShift
+  sandboxed containers helps isolate it in its own kernel to reduce risk.\n\n- **Safe
+  Multi-tenancy & Resource Sharing (CI/CD Jobs, CNFs, ..)** \nIf you are providing
+  a service to multiple tenants, it could mean that the service workloads are sharing
+  the same resources (e.g., worker node). By deploying in a dedicated kernel, the
+  impact of these workloads have on one another is greatly reduced.\n\n- **Additional
+  Isolation with Native Kubernetes User Experience**\nOpenShift sandboxed containers
+  is used as a compliant OCI runtime. Therefore, many operational patterns used with
+  normal containers are still preserved including but not limited to image scanning,
+  GitOps, Imagestreams, and so on.\n\nPlease refer to this [blog](https://cloud.redhat.com/blog/the-dawn-of-openshift-sandboxed-containers-overview)
+  for a detailed overview of sandboxed containers use cases and other related details.\n\n##
+  OpenShift sandboxed containers Operator\n\nThe operator manages the lifecycle (`install/configure/update`)
+  of sandboxed containers runtime (`Kata containers`) on OpenShift clusters.\n\n###
+  Operator Architecture\n\nThe following diagram shows how the operator components
+  are connected to the OpenShift overall architecture:\n\n![High Level Overview](./docs/arch.png)\n\n\nHere
+  is a brief summary of the components:\n\n- OpenShift clusters consist of controller
+  and worker nodes organized as  machine config pools. \n- The Machine Config Operator
+  (MCO) manages the operating system and keeps the cluster up to date and configured.\n-
+  The control-plane nodes run all the services that are required to control the cluster
+  such as the API server, etcd, controller-manager, and the scheduler. \n- The OpenShift
+  sandboxed containers operator runs on a control plane node.\n- The cluster worker
+  nodes run all the end-user workloads. \n- The container engine `CRI-O` uses either
+  the default container runtime `runc` or, in sandboxed containers case, the `Kata`
+  containers runtime.\n\n### KataConfig Custom Resource Definition\n\nThe operator
+  owns and control the `KataConfig` Custom Resource Definition (CRD).\nPlease refer
+  to the [code](https://github.com/openshift/sandboxed-containers-operator/blob/main/api/v1/kataconfig_types.go)
+  to find details of the `KataConfig` CRD.\n\n### Getting Started\n\nPlease refer
+  to the OpenShift release specific documentation for getting started with sandboxed
+  containers. \n- For OpenShift latest documentation please follow this [doc](https://docs.openshift.com/container-platform/latest/sandboxed_containers/deploying-sandboxed-container-workloads.html)\n\nFurther
+  note that starting with OpenShift 4.9, the branch naming is tied to the operator
+  version and not the OpenShift version.\nFor example `release-1.1` corresponds to
+  the Operator release verson `1.1.x`.\n\n## Operator Development\n\nPlease take a
+  look at the following [doc](./docs/DEVELOPMENT.md). \nContributions are most welcome!!\n\n##
+  Demos\n\nYou can find various demos in the following [youtube channel](https://www.youtube.com/channel/UC6PCt2zbug9cF4SpnMrE68A).\n\n##
+  Further Reading\n\n- [OpenShift sandboxed containers 101](https://cloud.redhat.com/blog/openshift-sandboxed-containers-101)\n-
+  [Operator overview](https://cloud.redhat.com/blog/openshift-sandboxed-containers-operator-from-zero-to-hero-the-hard-way)\n-
+  [Troubleshooting](https://cloud.redhat.com/blog/sandboxed-containers-operator-from-zero-to-hero-the-hard-way-part-2)\n-
+  [Hybrid Cloud Blog Series](https://www.redhat.com/en/openshift-sandboxed-containers)\n"
+name: sandboxed-containers-operator
+schema: olm.package
+---
+image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator-bundle:v1.6.0
+name: sandboxed-containers-operator.v1.6.0
+package: sandboxed-containers-operator
+properties:
+- type: olm.gvk
+  value:
+    group: confidentialcontainers.org
+    kind: PeerPod
+    version: v1alpha1
+- type: olm.gvk
+  value:
+    group: confidentialcontainers.org
+    kind: PeerPodConfig
+    version: v1alpha1
+- type: olm.gvk
+  value:
+    group: kataconfiguration.openshift.io
+    kind: KataConfig
+    version: v1
+- type: olm.package
+  value:
+    packageName: sandboxed-containers-operator
+    version: 1.6.0
+- type: olm.csv.metadata
+  value:
+    annotations:
+      alm-examples: |-
+        [
+          {
+            "apiVersion": "kataconfiguration.openshift.io/v1",
+            "kind": "KataConfig",
+            "metadata": {
+              "name": "example-kataconfig"
+            }
+          }
+        ]
+      capabilities: Seamless Upgrades
+      createdAt: "2024-05-08T15:43:54Z"
+      features.operators.openshift.io/disconnected: "true"
+      features.operators.openshift.io/fips-compliant: "false"
+      features.operators.openshift.io/proxy-aware: "false"
+      features.operators.openshift.io/tls-profiles: "false"
+      features.operators.openshift.io/token-auth-aws: "false"
+      features.operators.openshift.io/token-auth-azure: "false"
+      features.operators.openshift.io/token-auth-gcp: "false"
+      olm.skipRange: '>=1.1.0 <1.5.2'
+      operatorframework.io/suggested-namespace: openshift-sandboxed-containers-operator
+      operators.openshift.io/valid-subscription: '["OpenShift Container Platform",
+        "OpenShift Platform Plus"]'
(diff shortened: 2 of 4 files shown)


gkurz:
Hi @tbuskey !

This is conflating way to many changes into a single commit. Please break it down into smaller items.
First starting with preparatory patches and then adding the actual support for file based catalog.

- add authentication support
- automate download of operator-sdk
- bump the version of opm
- file based catalog

You can regroup all the documentation changes in a final commit, for simplicity.

Makefile
```diff
@@ -143,7 +145,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build \
-		-t ${IMG} \
+		-t ${IMG} $(IMAGE_AUTH) \
```
gkurz: The meaning of IMAGE_AUTH is is a bit unclear. It exposes a direct dependency to the docker command line for no good reason. I'd prefer IMAGE_AUTH to be called something like AUTH_FILE and do
```suggestion
		-t ${IMG} \
		--authfile $(AUTH_FILE) \
```

Makefile
```diff
@@ -246,6 +248,23 @@ bundle-build: bundle ## Build the bundle image.
 bundle-push: ## Push the bundle image.
 	$(MAKE) docker-push IMG=$(BUNDLE_IMG)
 
+.PHONY: operator-sdk
```
bpradipt: I personally prefer to keep the dev prerequisites (golang, operator-sdk etc) outside of the Makefile, since there could be different ways to install and configure it. For example operator-sdk might be on my system path for usage across different operators.
Instead, checking for the presence of specific version of operator-sdk binary (based on osc operator requirement) in the Makefile and throwing error is portable and useful imho.

docs/DEVELOPMENT.md
```diff
@@ -2,30 +2,35 @@
...
-- login to one of the clusters' console
-- use the console's shortcut to get the commandline login command
-- log in from the command line with the provided command
-- use "oc registry login" to save the token locally
-
-### Using public images
+#### Using the token
+Set the authfile used by docker-build if login was successful.  Ex:
+```
+export IMAGE_AUTH='--authfile=~/.docker/config.json'
```
bpradipt: Is this really needed? Doesn't docker honour the default auth file (`$HOME/.docker/config.json`) automatically if present ?

tbuskey: `docker-build` did not use the auth I had in `$HOME/.docker/config.json` until I added
 `--authfile=$HOME/.docker/config.json`

If the user has not set IMAGE_AUTH, the Makefile will leave it out

Makefile
tbuskey: Sounds reasonable.

Does this work?
1. check if operator-sdk already exists in the path
    - download into ./bin IFF it doesn't
 2.  check the version of operator-sdk against $OPERATOR-SDK-VERSION (allow overrides)
     - fail make if version isn't matched

tbuskey: I'm leaving the operator-sdk section unchanged

tbuskey:
@gkurz @bpradipt 
Changes are now in 5 commits
There are no changes for `operator-sdk`

Makefile
```diff
@@ -270,17 +272,25 @@ BUNDLE_IMGS ?= $(BUNDLE_IMG)
...
+# Build a file based catalog image using opm.
+# https://docs.openshift.com/container-platform/4.12/operators/admin/olm-managing-custom-catalogs.html
 # https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
 .PHONY: catalog-build
-catalog-build: opm ## Build a catalog image.
-	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
+catalog-build: opm
+	@[ ! -d catalog ] && mkdir -p catalog || true
+	@[ ! -f catalog.Dockerfile ] && $(OPM) generate dockerfile catalog || true
+	$(OPM) init  sandboxed-containers-operator --default-channel=$(DEFAULT_CHANNEL) --description ./README.md --output yaml > catalog/index.yaml
```
cpmeadors: This should be done once and then checked in to the repo.  We need static files that we can modify as needed to make thing clear.

Makefile
```diff
@@ -310,7 +320,13 @@ bundle-clean: ## Clean generated bundle files
 
 .PHONY: bin-clean
 bin-clean: ## Clean downloaded binaries
+	@[ -d bin ] && chmod -R u+rwx bin || true
 	$(RM) -r bin
 
+.PHONY: catalog-clean
+catalog-clean: ## Clean catalog files
+	$(RM) -r catalog
```
cpmeadors: This needs to be removed once the catalog dir has been checked in to the repo.

cpmeadors:
Probably need to add stuff to .gitignore as describe at https://olm.operatorframework.io/docs/reference/file-based-catalogs/#structure

tbuskey:
> Probably need to add stuff to .gitignore as describe at https://olm.operatorframework.io/docs/reference/file-based-catalogs/#structure

bundle/ is

tbuskey:
@gkurz @bpradipt could you approve?

gkurz:
> @gkurz @bpradipt could you approve?

Sorry for coming back late @tbuskey. Thanks for splitting the patches !

Some more general polishing is needed though :
- provide a description in every commit (e.g. first commit adds authentication but we don't know why)
- add you Signed-off-by:

I'll leave some comments in the commits.

Makefile
```diff
@@ -312,6 +312,7 @@ bundle-clean: ## Clean generated bundle files
 
 .PHONY: bin-clean
 bin-clean: ## Clean downloaded binaries
+	@[ -d bin ] && chmod -R u+rwx bin || true
```
gkurz: I don't quite understand how `chmod` would succeed in a scenario where `rm -f -r` would fail... what was the problem ?

Makefile
```diff
@@ -257,7 +257,7 @@ ifeq (,$(shell which opm 2>/dev/null))
 	set -e ;\
 	mkdir -p $(dir $(OPM)) ;\
 	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
-	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.23.0/$${OS}-$${ARCH}-opm ;\
+	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.28.0/$${OS}-$${ARCH}-opm ;\
```
gkurz, line 262: Please provide justification for this bump.

Makefile
```diff
@@ -272,17 +272,25 @@ BUNDLE_IMGS ?= $(BUNDLE_IMG)
...
 
-# Set CATALOG_BASE_IMG to an existing catalog image tag to add $BUNDLE_IMGS to that image.
-ifneq ($(origin CATALOG_BASE_IMG), undefined)
-FROM_INDEX_OPT := --from-index $(CATALOG_BASE_IMG)
-endif
-
-# Build a catalog image by adding bundle images to an empty catalog using the operator package manager tool, 'opm'.
-# This recipe invokes 'opm' in 'semver' bundle add mode. For more information on add modes, see:
+# Build a file based catalog image using opm.
+# https://docs.openshift.com/container-platform/4.12/operators/admin/olm-managing-custom-catalogs.html
```
gkurz: Please point to latest documentation at https://docs.openshift.com/container-platform/4.16/operators/admin/olm-managing-custom-catalogs.html.

Makefile
```diff
@@ -272,17 +272,25 @@ BUNDLE_IMGS ?= $(BUNDLE_IMG)
...
 # https://github.com/operator-framework/community-operators/blob/7f1438c/docs/packaging-operator.md#updating-your-existing-operator
 .PHONY: catalog-build
-catalog-build: opm ## Build a catalog image.
-	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
+catalog-build: opm
+	@[ ! -d catalog ] && mkdir -p catalog || true
```
gkurz: `mkdir -p` just exits 0 if the directory already exists.
```suggestion
	mkdir -p catalog
```

Makefile
```diff
@@ -272,17 +272,25 @@ BUNDLE_IMGS ?= $(BUNDLE_IMG)
...
+	@[ ! -f catalog.Dockerfile ] && $(OPM) generate dockerfile catalog || true
```
gkurz: It looks like `catalog.Dockerfile` should be a prereq for the `catalog-build` target. This will allow to get rid of the `|| true` that would hide a failure of `$(OPM)` AFAICT.

Makefile
tbuskey: Changed to AUTH_FILE 
Explanation is in DEVELOPMENT.md

Makefile
gkurz: If we are to check in generated files, then I'd rather not have them created as part of the main build flow.

Makefile
gkurz: Both `catalog` initial creation and removal should be dropped from the rules in this case.

Makefile
tbuskey: opm 1.28 matches the current version of operator-sdk

- go version bumped to 1.18 -> 1.19
- k8s bumped 1.24 -> 1.25 -> 1.26
- go libraries bumped
- FBC caching refactored
- fix channel sorting
- improve selinux labeling
- bug fixes

Makefile
tbuskey: It was exiting the make when I was testing

Makefile
tbuskey: Permissions in bin/k8s/1.24.2-linux-amd64 were readonly:
dr-xr-xr-x. 2 tbuskey tbuskey 4096 Jul  3 13:07 1.24.2-linux-amd64

Makefile
tbuskey: _opm generate dockerfile catalog_ gives an error if **catalog.Dockerfile** exists

Makefile
tbuskey: catalog-init is not in the main build flow.  Normal is make catalog-build; make catalog-push

If you are using a private quay.io, the catalog files need to be re-written.  make catalog-clean; make catalog-init will do that for you.

Makefile
tbuskey: removing catalog-init/catalog-clean from Makefile
Adding directions to generate catalog to DEVELOPMENT.md

Makefile
tbuskey: catalog/index.yaml and catalog.Dockerfile are now checked in
Regenerating them is now in DEVELOPMENT.md

Makefile
tbuskey: Dropped

gkurz:
This PR isn't ready to be merged as is. Some more polishing is needed.

- when addressing a reviewer's requests, it is preferrable to squash the fixes in the original commits and force-push as suggested at https://docs.google.com/document/d/1aZEuqUubGWassHyURs7t_4sX1fMAnxyaBAlKsS8KbvU/edit#heading=h.6nn0bdwndby1 and https://docs.google.com/document/d/1aZEuqUubGWassHyURs7t_4sX1fMAnxyaBAlKsS8KbvU/edit#heading=h.xocsdceerbc
e.g. Makefile changes from https://github.com/openshift/sandboxed-containers-operator/pull/422/commits/0a61d070f2bc2a44bede7700e21b9d6cda1e5d16 should be squashed into https://github.com/openshift/sandboxed-containers-operator/pull/422/commits/42536b539025279511170ebf70ddad8acb3aba3e
and documentation changes should be squashed into https://github.com/openshift/sandboxed-containers-operator/pull/422/commits/461059c48cb6b16cf6baf77539c08915743886ea
- all commits should have a descriptive title, a good commit message and a `Signed-off-by` tag as suggested by https://docs.google.com/document/d/1aZEuqUubGWassHyURs7t_4sX1fMAnxyaBAlKsS8KbvU/edit#heading=h.jwlp2948280n and https://docs.google.com/document/d/1aZEuqUubGWassHyURs7t_4sX1fMAnxyaBAlKsS8KbvU/edit#heading=h.iwk7eyleqzrc

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+  look at the following [doc](./docs/DEVELOPMENT.md). \nContributions are most welcome!!\n\n##
+  Demos\n\nYou can find various demos in the following [youtube channel](https://www.youtube.com/channel/UC6PCt2zbug9cF4SpnMrE68A).\n\n##
+  Further Reading\n\n- [OpenShift sandboxed containers 101](https://cloud.redhat.com/blog/openshift-sandboxed-containers-101)\n-
+  [Operator overview](https://cloud.redhat.com/blog/openshift-sandboxed-containers-operator-from-zero-to-hero-the-hard-way)\n-
+  [Troubleshooting](https://cloud.redhat.com/blog/sandboxed-containers-operator-from-zero-to-hero-the-hard-way-part-2)\n-
+  [Hybrid Cloud Blog Series](https://www.redhat.com/en/openshift-sandboxed-containers)\n"
+name: sandboxed-containers-operator
+schema: olm.package
+---
+image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator-bundle:v1.6.0
```
gkurz:
```suggestion
image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator-bundle:v1.7.0
```

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+name: sandboxed-containers-operator.v1.6.0
```
gkurz:
```suggestion
name: sandboxed-containers-operator.v1.7.0
```

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+    version: v1alpha1
+- type: olm.gvk
+  value:
+    group: kataconfiguration.openshift.io
+    kind: KataConfig
+    version: v1
+- type: olm.package
+  value:
+    packageName: sandboxed-containers-operator
+    version: 1.6.0
```
gkurz:
```suggestion
    version: 1.7.0
```

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+      capabilities: Seamless Upgrades
+      createdAt: "2024-05-08T15:43:54Z"
+      features.operators.openshift.io/disconnected: "true"
+      features.operators.openshift.io/fips-compliant: "false"
+      features.operators.openshift.io/proxy-aware: "false"
+      features.operators.openshift.io/tls-profiles: "false"
+      features.operators.openshift.io/token-auth-aws: "false"
+      features.operators.openshift.io/token-auth-azure: "false"
+      features.operators.openshift.io/token-auth-gcp: "false"
+      olm.skipRange: '>=1.1.0 <1.5.2'
```
gkurz:
```suggestion
      olm.skipRange: '>=1.1.0 <1.7.0'
```

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+    maintainers:
+    - email: support@redhat.com'
+      name: '''Red Hat'
+    maturity: beta
+    provider:
+      name: Red Hat
+relatedImages:
+- image: gcr.io/kubebuilder/kube-rbac-proxy@sha256:d99a8d144816b951a67648c12c0b988936ccd25cf3754f3cd85ab8c01592248f
+  name: ""
+- image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator-bundle:v1.6.0
```
gkurz:
```suggestion
- image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator-bundle:v1.7.0
```

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+  name: ""
+- image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:v1.6.0
+  name: ""
+- image: registry.redhat.io/openshift-sandboxed-containers/osc-cloud-api-adaptor-rhel9:latest
+  name: caa
```
gkurz: Why do previous images have `name: ""` ?

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+  name: podvm-builder
+- image: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9:latest
+  name: podvm-payload
+schema: olm.bundle
+---
+schema: olm.channel
+package: sandboxed-containers-operator
+name: stable
+entries:
+  - name: sandboxed-containers-operator.v1.6.0
```
gkurz:
```suggestion
  - name: sandboxed-containers-operator.v1.7.0
```

gkurz (commented):
Great work @tbuskey ! It just requires to bump the OSC version to 1.7.0 in the catalog index file. I'll try it out and approve.

catalog/index.yaml
gkurz: AFAIK we also need the following because of `olm.skipRange` :

```
    replaces: sandboxed-containers-operator.v1.6.0
```

See [KATA-2894](https://issues.redhat.com/browse/KATA-2894) for an explanation.

docs/DEVELOPMENT.md
```diff
@@ -12,82 +12,109 @@ install -m 755 operator-sdk_linux_amd64 ${SOME_DIR_IN_YOUR_PATH}/operator-sdk
...
+make opm                  # to get bin/opm
+rm -rf catalog && mkdir -p catalog
+bin/opm init  sandboxed-containers-operator --default-channel=${DEFAULT_CHANNEL} --description ./README.md --output yaml > catalog/index.yaml
+bin/opm render ${BUNDLE_IMG} --output=yaml >> catalog/index.yaml
+printf -- "---\n\
+schema: olm.channel\n\
+package: sandboxed-containers-operator\n\
+name: ${DEFAULT_CHANNEL}\n\
+entries:\n\
+  - name: sandboxed-containers-operator.v${VERSION}\n\
```
gkurz, line 99: As explained at https://github.com/openshift/sandboxed-containers-operator/pull/422/commits/55a01ed18c6249eb9a42f1e9c02862546f2c7f1c#r1673838632, please something like :
```
    replaces: sandboxed-containers-operator.v${PREVIOUS_VERSION}\n\
```

catalog/index.yaml
tbuskey: opm generate creates the index.yaml.  It starts at 1.6.0 and not something earlier

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+      name: '''Red Hat'
+    maturity: beta
+    provider:
+      name: Red Hat
+relatedImages:
+- image: gcr.io/kubebuilder/kube-rbac-proxy@sha256:d99a8d144816b951a67648c12c0b988936ccd25cf3754f3cd85ab8c01592248f
+  name: ""
+- image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator-bundle:v1.6.0
+  name: ""
+- image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:v1.6.0
```
gkurz:
```suggestion
- image: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator:v1.7.0
```

catalog/index.yaml
```diff
@@ -0,0 +1,243 @@
...
+  name: ""
+- image: registry.redhat.io/openshift-sandboxed-containers/osc-cloud-api-adaptor-rhel9:latest
```
gkurz: We try to avoid using the `latest` tag as explained in f8169dbedc131b60c1268c7757a0afd84523498b.

catalog/index.yaml
tbuskey: Changing the version

catalog/index.yaml
tbuskey: Changing the version

catalog/index.yaml
tbuskey: Changing the version

catalog/index.yaml
tbuskey: Adding

docs/DEVELOPMENT.md
tbuskey: Done

catalog/index.yaml
tbuskey: The `latest` versioning for the podvm probably comes from` config/peerpods/podvm`.
We will have to investigate the others

docs/DEVELOPMENT.md
```diff
@@ -12,82 +12,109 @@ install -m 755 operator-sdk_linux_amd64 ${SOME_DIR_IN_YOUR_PATH}/operator-sdk
...
+#### Overwrite/create catalog/index.yaml if needed
+If `IMAGE_TAG_BASE` was set for your private registry, `catalog/index.yaml` will need to be rewritten.  Before running `make catalog-build`, run this:
+
+```
+DEFAULT_CHANNEL="stable"  # from Makefile
+VERSION=1.7.0             # from Makefile
+BUNDLE_IMG=${IMAGE_TAG_BASE}-bundle:v${VERSION}
+make opm                  # to get bin/opm
+rm -rf catalog && mkdir -p catalog
+bin/opm init  sandboxed-containers-operator --default-channel=${DEFAULT_CHANNEL} --description ./README.md --output yaml > catalog/index.yaml
```
gkurz, line 92: I know this was discussed previously at https://github.com/openshift/sandboxed-containers-operator/pull/422#discussion_r1661202637 but something now looks wrong to me.

This `bin/opm init ...` has a dependency on `README.md` : it should be run again every time `README.md` is changed, not just once. Typical case for a rule in the Makefile.

Also I still don't understand why `catalog/index.yaml` should be committed in the repo.

@cpmeadors ?

Makefile
gkurz: This is the result of `make test` failing and not fixing the mode as per bef333e. It means that the SKIP_TESTS approach isn't a good workaround for the broken unit tests. Let's revert 48942c6 and ignore errors at https://github.com/openshift/sandboxed-containers-operator/blob/devel/Makefile#L128 instead. I'll create a PR for this.

gkurz:
Converted to draft to block merge until this PR is fully reviewed.

bpradipt:
@tbuskey since we rely on operator-sdk, one suggestion is to create a skeleton Makefile using operator-sdk and align your changes to that. Example you can following the instructions [here](https://sdk.operatorframework.io/docs/building-operators/golang/quickstart/#steps) to create a skeleton Makefile and compare it with what we have and add your changes.

For example, I see that latest Makefile has options to download all the required binaries including operator-sdk locally. Let's follow that and ignore my previous comment on (not) downloading the operator-sdk. It's better to align with operator-sdk generated skeleton as it helps when we migrate to newer versions of sdk. We already are facing issues with deviating from the sdk generated skeleton file `.gitignore`, so let's avoid deviating from the sdk generated skeleton as reference unless we have a really really strong reason.
Also looking at other operators and how they are building file based catalog will provide insights on what is needed.
Example - https://github.com/openshift/lvm-operator

Makefile
gkurz: Now that #429 is merged, please rebase and you can drop this commit.

docs/DEVELOPMENT.md
```diff
@@ -12,82 +12,109 @@ install -m 755 operator-sdk_linux_amd64 ${SOME_DIR_IN_YOUR_PATH}/operator-sdk
...
+- You should go to _app.ci_ in the above doc, login with SSO
+- Go to _copy login_ on your _username_ in the top right
+- You will login and see _view token_.  Click on that to see
+- **Your API token is**.  The next line is your full token
+- _podman login_ with your username and the token as your password should put the token in your AUTH_FILE
+
+#### Using the token
+If docker build does not automatically use the token in the default authfile, it can be forces:
+```shell
+export AUTH_FILE='~/.docker/config.json'
```
gkurz, line 32: From the `podman-login(1)` manual page :
```
The default path for reading and writing credentials is ${XDG_RUNTIME_DIR}/containers/auth.json.
Podman uses existing credentials if the user does not pass in a username.  Podman first searches
for  the  username and  password  in  the ${XDG_RUNTIME_DIR}/containers/auth.json,  if  they are
not valid, Podman then uses any existing credentials found in $HOME/.docker/config.json.
```

```suggestion
export AUTH_FILE="${XDG_RUNTIME_DIR}/containers/auth.json"
```

gkurz: Hmm... I gave a try without setting `AUTH_FILE` and it worked :
- `podman login` does add the credentials to `"${XDG_RUNTIME_DIR}/containers/auth.json"` 
- `podman build` does find the credentials there
I now suspect that `AUTH_FILE` isn't needed at all.

What were the steps that ended up in `docker build does not automatically use the token in the default authfile` ?

PR #424, opened 2024-07-01
Embed pause image

bpradipt:


bpradipt:
@savitaashture fyi

config/peerpods/podvm/aws-podvm-image-cm.yaml
```diff
@@ -33,3 +33,8 @@ data:
   # NVIDIA GPU vars
   NVIDIA_DRIVER_VERSION: "535"
   NVIDIA_USERSPACE_VERSION: "1.13.5-1"
+
+  # Pause image vars
+  PAUSE_IMAGE_REPO: "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256"
+  PAUSE_IMAGE_VERSION: "ac97982e845413b2b7377171d1b4c85d2b2cfda174e8251e3049d4843cea40e1"
```
snir911: would this be different for other architectures?

config/peerpods/podvm/aws-podvm-image-cm.yaml
```diff
@@ -33,3 +33,8 @@ data:
...
+  PAUSE_IMAGE_REPO_AUTH_FILE: "/tmp/regauth/auth.json"
```
snir911: Wouldn't it be preferred to hide these from the user, i.e. set default values withing the scripts and comment these variables out?
Alternatively, can we simply use by default the current pause image used normally by Openshift?

config/peerpods/podvm/lib.sh
```diff
@@ -61,13 +62,15 @@ function install_rpm_packages() {
...
 #"yq=https://github.com/mikefarah/yq/releases/download/v4.35.2/yq_linux_amd64.tar.gz"
+#"umoci=https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64"
 
 install_binary_packages() {
     # Define the required binary packages
     REQUIRED_BINARY_PACKAGES=(
         "packer=https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip"
         "kubectl=https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.14.9/openshift-client-linux.tar.gz"
         "yq=https://github.com/mikefarah/yq/releases/download/v4.35.2/yq_linux_amd64.tar.gz"
+        "umoci=https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64"
```
snir911, line 80: for later, maybe worth to validate hashs at some point

bpradipt: agree

config/peerpods/podvm/aws-podvm-image-cm.yaml
bpradipt: I agree.. If there is way to query the default pause for a release then we can set it in the operator, like we do for some other vars. Otherwise we can set it to a default (like above).

config/peerpods/podvm/aws-podvm-image-cm.yaml
bpradipt: I think so

config/peerpods/podvm/aws-podvm-image-cm.yaml
bpradipt: @snir911 I couldn't find a way to query this info. One option is to mount the `/etc/crio/crio.conf.d/00-default` and read it. But it'll increase the required permissions for the container so I prefer to avoid it. 
For now, my suggestion is to use this approach of setting the default in configmap. Since this is a generic pause image it should work with other OCP versions as well. Having these options in the configmap will make it easier to replace it - using smaller size pause image, different one for a specific OCP release etc

snir911: you can fetch it from the worker node mc, however, this indeed may be an overkill, i think it would be better to set it to some fixed value in the scripts (if not explicitly set) rather than having it in the cm. 
This will reduce confusion of users when customize cm image is required.

snir911: oh, downside will be it will require to explicitly set for different architectures...

bpradipt: So as you suggested, I'll set the default per arch (based on current info) in the script for now and leave it there..
Wdyt ?

bpradipt: Btw I didn't find the pause image details in the worker mc :-(

config/peerpods/podvm/aws-podvm-image-cm.yaml
bpradipt: The OCP pause image is multi-arch.

config/peerpods/podvm/lib.sh
```diff
@@ -61,13 +62,15 @@ function install_rpm_packages() {
 #"packer=https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip"
 #"kubectl=https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/4.14.9/openshift-client-linux.tar.gz"
 #"yq=https://github.com/mikefarah/yq/releases/download/v4.35.2/yq_linux_amd64.tar.gz"
+#"umoci=https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64"
```
snir911, line 72: we'll need a way to configure it once multiarch is supported

config/peerpods/podvm/lib.sh
```diff
@@ -187,6 +187,55 @@ function prepare_source_code() {
...
+# Download and extract pause container image
+# Accepts three arguments:
+# 1. pause_image_repo_url: The registry URL of the OCP pause image.
+# 2. pause_image_tag: The tag of the OCP pause image.
+# 2. auth_json_file (optional): Path to the registry secret file to use for downloading the image
+function download_and_extract_pause_image() {
+
+    # Set default values for the OCP pause image
+    # This image is multi-arch
+    PAUSE_IMAGE_REPO_DEFAULT="quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256"
```
snir911: nit: maybe define the defaults in the top of the file?

controllers/image_generator.go
```diff
@@ -308,6 +308,15 @@ func (r *ImageGenerator) createJobFromFile(jobFileName string) (*batchv1.Job, er
 		job.Spec.Template.Spec.Containers[0].Image = podvmBuilderImage
 	}
 
+	// If RELATED_PODVM_PAYLOAD_IMAGE environment variable is set, use it
+	// Otherwise, use the default podvm payload image
+	// There is only one initContainer in the job, so we don't need to check the container name
+	podvmPayloadImage := os.Getenv("RELATED_IMAGE_PODVM_PAYLOAD")
```
snir911, line 314: Would it work if it's being set in the podvm-image-cm? IIUC it's not, and i think it may be a bit confusing that this is image related configuration only but it's set differently

snir911 (approved):
LGTM, thanks! added some minor comments

config/peerpods/podvm/lib.sh
bpradipt: yeah, I thought about it but then not sure what made me put here ;) .. I think I was in two minds whether to put it at the top or in the function..

controllers/image_generator.go
bpradipt: The RELATED_* is also used for disconnected installs (useful for mirroring). Hence we use the same thing.

config/peerpods/podvm/lib.sh
bpradipt: Yeah, all the binaries are x86_64 for now and needs to be switched to multi-arch. I'll add a TBD

bpradipt:
Tested with custom operator and podvm-builder image to generate a podvm image with embedded pause image in Azure

PR #427, opened 2024-07-12
Update configuration-remote.toml to align with upstream changes

bpradipt:
Fixes: #[KATA-3155](https://issues.redhat.com//browse/KATA-3155)


cpmeadors:
How would one verify that this change?   Both positively with a use cases or with regression testing to make sure it didn't break anything.

bpradipt:
> How would one verify that this change?   Both positively with a use cases or with regression testing to make sure it didn't break anything.

This is the core config of kata-remote. So all your existing test cases (positive/negative) should work the same way as before.  
You can also take the raw configuration-remote.toml from this commit, apply it to a setup with latest shim and cloud-api-adaptor to test it.

snir911 (approved):

PR #429, opened 2024-07-15
Fix `make test` to unbreak `make clean`

gkurz:

`make test` is known to be failing since a long time. An issue is that it thus leaks a directory that prevents `make clean` to succeed because of restrictive file access modes.

**- What I did**

Separate the building of the test, which introduces the file access mode issue, from the actual test run. This allows to fully benefit from the access mode fixing introduced by PR #240.

**- How to verify it**

- `make test` # which will likely fail
- `make clean` # always succeed with this PR, fail otherwise

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,9 +125,11 @@ test: manifests generate fmt vet envtest ## Run tests.
 ifneq (, $(SKIP_TESTS))
 	@echo Skipping tests. Unset SKIP_TESTS to actually run them.
 else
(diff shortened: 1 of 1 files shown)


tbuskey (commented):
Does the k8s directory need to have chmod as well?
chmod -R +w $(LOCALBIN)/k8s # should do it

littlejawa (approved):
lgtm
Thanks @gkurz !

gkurz:
> Does the k8s directory need to have chmod as well? chmod -R +w $(LOCALBIN)/k8s # should do it

Why would it need it ? Do you have a case where this PR isn't enough for `make clean` to succeed @tbuskey ?

tbuskey:
I thought the bin/k8s directory was created read only.
/lgtm

gkurz:
> I thought the bin/k8s directory was created read only. /lgtm

It isn't. Thanks !

PR #434, opened 2024-07-18
podvm: allow setting custom agent-policy

snir911:
by having custom.rego file set as agent policy CM

kubectl create configmap agent-policy --from-file=<path/to/custom.rego> -n openshift-sandboxed-containers-operator

also use relative links is it essential when passed to packer

**do not merge**

**I failed to test yet due to (probably) unaltered issues**


config/peerpods/podvm/lib.sh
```diff
@@ -196,11 +196,18 @@ function prepare_source_code() {
...
+        cp /tmp/kata-opa/custom.rego "${podvm_dir}"/files/etc/kata-opa/custom.rego
+        ln -sf custom.rego  "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
+    elif [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]] ; then
+        echo "Setting custom agent policy to CoCo's recommended policy"
         sed 's/default SetPolicyRequest := true/default SetPolicyRequest := false/; s/default ExecProcessRequest := true/default ExecProcessRequest := false/' \
             "${podvm_dir}"/files/etc/kata-opa/default-policy.rego > "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego
-        ln -sf "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego  "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
+        ln -sf coco-default-policy.rego  "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
     fi
+    echo "\nCurrent Agent Policy:\n" && cat "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
```
bpradipt: Hmm afaik echo will not expand the escape sequences. `printf` needed ?

bpradipt:
Tested this successfully by building a new image
```
oc exec -it hello-b -- sh
error: Internal error occurred: error executing command in container: cannot enter container 6d9c0d50f7640003baed372f0804ada60e8f04ce8222a0de8ab6b6f48de6411c, with err rpc error: code = PermissionDenied desc = "ExecProcessRequest is blocked by policy: ": unknown
```

config/peerpods/podvm/lib.sh
bpradipt: While you are touching this file, it might be useful to split the sed command into multiline for easier readability. for example

```
  sed 's/default SetPolicyRequest := true/default SetPolicyRequest := false/;
   s/default ExecProcessRequest := true/default ExecProcessRequest := false/' \
        "${podvm_dir}"/files/etc/kata-opa/default-policy.rego > \
        "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego
```
Btw on second thoughts should SetPolicyRequest be set to true for now ?

bpradipt: So as default coco policy, we can disable `ReadStreamRequest` and `ExecProcessRequest` to start with and provide SetPolicy examples to enable/disable apis for trying out different scenario
Also we'll document (blog) the (wip) eventual solution with initData and measured disk. 
Sounds ok ?

gkurz: > Btw on second thoughts should SetPolicyRequest be set to true for now ?

Newbie question : what does SetPolicyResquest do ?

> So as default coco policy, we can disable `ReadStreamRequest` and `ExecProcessRequest` to start with and provide SetPolicy examples to enable/disable apis for trying out different scenario

IIUC this is calling for another PR that re-enables `SetPolicyRequest` and disables `ReadStreamRequest`, right ?

bpradipt: A new PR or at least a new commit explaining the rationale.

config/peerpods/podvm/lib.sh
```diff
@@ -197,7 +197,12 @@ function prepare_source_code() {
     fi
 
     # links must be relative
-    if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
+    if [[ -e /tmp/kata-opa/custom.rego ]] ; then
```
bpradipt: How about adding a folder in podvm with sample policy file and a README describing the details on how policy is handled in OSC (via the configmap etc) along with examples on using the custom policy  as well as another example of setting policy via the annotation?

config/peerpods/podvm/lib.sh
```diff
@@ -196,11 +196,14 @@ function prepare_source_code() {
...
     fi
 
+    # links must be relative
     if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
-        sed 's/default SetPolicyRequest := true/default SetPolicyRequest := false/; s/default ExecProcessRequest := true/default ExecProcessRequest := false/' \
+        sed 's/default ReadStreamRequest := true/default ReadStreamRequest := false/;
+            s/default ExecProcessRequest := true/default ExecProcessRequest := false/' \
             "${podvm_dir}"/files/etc/kata-opa/default-policy.rego > "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego
-        ln -sf "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego  "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
+        ln -sf coco-default-policy.rego "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
```
bpradipt, line 222: Hmm is the above correct?
the coco-default-policy.rego is in `"${podvm_dir}"/files/etc/kata-opa/`

gkurz: > Hmm is the above correct? the coco-default-policy.rego is in `"${podvm_dir}"/files/etc/kata-opa/`

Yes it is because both files are in the same directory. When the VFS will try to access the right most component of `"${podvm_dir}"/files/etc/kata-opa/default-policy.rego`  from `"${podvm_dir}"/files/etc/kata-opa/`, it will go to the exact content of the symlink, i.e. `coco-default-policy.rego` which exists in the directory.

The first argument to `ln -s` is basically arbitrary text that the VFS is expected to use when resolving the symlink. The `symlink(2)` syscall doesn't even require the target to exist.

config/peerpods/podvm/lib.sh
```diff
@@ -206,13 +206,22 @@ function prepare_source_code() {
     fi
 
     # links must be relative
-    if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
+    if [[ "${AGENT_POLICY}" ]]; then
+        echo "Custom agent policy is being set through the AGENT_POLICY value"
+        echo ${AGENT_POLICY} | base64 -d > "${podvm_dir}"/files/etc/kata-opa/custom.rego
+        if [[ $? == 0 ]] && grep -q "agent_policy" "${podvm_dir}"/files/etc/kata-opa/custom.rego; then # checks policy validity
```
gkurz, line 212: Isn't there a better way to validate the syntax than grepping ? Couldn't https://www.openpolicyagent.org/docs/latest/cli/#opa-check be used here ?

gkurz (approved):
Grep is really a poor man's tool to validate user input but we can start with this and improve later. Rest looks good to me.

Thanks @snir911 !

config/peerpods/podvm/lib.sh
snir911: right, my intention wasn't to actually validate the format rather to validate the output isn't ruined because of missing couple of characters when copied the encoded string.
This opa check is indeed an option, it will require to install the binary in the container, can certainly be done at future point.

config/peerpods/podvm/agent-policy/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Set Policy Via Pod Annotation
+
+As long as the `SetPolicyRequest` call was not disabled at image creation time, users set custom
+policy through annotation at pod creation time. To set policy through annotation, encode your policy
+file (e.g., [allow-all-except-exec-process.rego](allow-all-except-exec-process.rego)) in base64 format
+and set it to the `io.katacontainers.config.agent.policy` annotation.
+
+**note:** annotation policy will override any previous policy (as long `SetPolicyRequest` is allowed)
```
bpradipt:
```suggestion
**note:** annotation policy will override any previous policy (as long as `SetPolicyRequest` is allowed)
```

config/peerpods/podvm/lib.sh
```diff
@@ -205,11 +205,23 @@ function prepare_source_code() {
...
-    if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
-        sed 's/default SetPolicyRequest := true/default SetPolicyRequest := false/; s/default ExecProcessRequest := true/default ExecProcessRequest := false/' \
+    # links must be relative
+    if [[ "${AGENT_POLICY}" ]]; then
+        echo "Custom agent policy is being set through the AGENT_POLICY value"
+        echo ${AGENT_POLICY} | base64 -d > "${podvm_dir}"/files/etc/kata-opa/custom.rego
+        if [[ $? == 0 ]] && grep -q "agent_policy" "${podvm_dir}"/files/etc/kata-opa/custom.rego; then # checks policy validity
+            ln -sf custom.rego  "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
+        else
+            error_exit "Invalid AGENT_POLICY value set, expected 64 based valid agent policy, got: \"${AGENT_POLICY}\""
```
bpradipt:
```suggestion
            error_exit "Invalid AGENT_POLICY value set, expected base64 encoded valid agent policy, got: \"${AGENT_POLICY}\""
```

config/peerpods/podvm/lib.sh
```diff
@@ -205,11 +205,23 @@ function prepare_source_code() {
...
+	fi
+    elif [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
```
bpradipt, line 217: We don't want to allow custom policy when CONFIDENTIAL_COMPUTE_ENABLED is yes?

snir911: we don't? i think that if user is asking for certain policy explicitly we should accept it (it may be more restricted than the default)

bpradipt: Ok, custom agent policy gets priority, which is fair imho. Thanks

bpradipt (approved):
/lgtm
Thanks @snir911

Issue #435, opened 2024-07-18
Enhance the usability of the Libvirt provider for confidential container administrators


savitrilh:
## Summary

To improve the user experience for confidential container administrators to enable the confidential container functionality, sandbox container operators should utilize a specialized script designed for Libvirt provider for managing administrative tasks specific to confidential containers during the activation of Peerpod through the KataConfig configmap.

## Motivation

Currently, the OpenShift Sandbox Container(OSC) Operator manages the following tasks:

- This manages the lifecycle (install/configure/update) of sandboxed containers runtime (Kata containers) on OpenShift clusters.
- Build the PODVM QCOW2 image or pull the PODVM image from the registry, extract the PODVM image from a container wrapped QCOW2 image.
-  Uploads the operator built or pulled pre-built PODVM image to pre-created Libvirt volume. 

OSC operator handles above operations automatically but still it requires following manual operations to be executed by confidential container administrator:

**1. Generation of SSH Key(public-private key) in OCP cluster.**

`$ ssh-keygen -f ./id_rsa -N ""`

 **2. Copy the SSH public key to KVM Host(LPAR). So that worker node can perform Libvirt specific operations such as creating the peerpod on the LPAR(KVM host) and lifecycle management of the peerpod VM.**

`$ ssh-copy-id -i ./id_rsa.pub <KVM_HOST_IP_ADDRESS>`

 **3. Create the SSH secret object**

```
$ oc create secret generic ssh-key-secret \
    -n openshift-sandboxed-containers-operator \
    --from-file=id_rsa.pub=./id_rsa.pub \
    --from-file=id_rsa=./id_rsa
```

**4. Create Libvirt POOL and Volume**

**Set the name of the libvirt pool by running the following command:** 	      
`$ export LIBVIRT_POOL=<name_of_libvirt_pool_to_create>`

**Set the name of the libvirt volume by running the following command:**     
` $ export LIBVIRT_VOL_NAME=<name_of_libvirt_volume_to_create>`

**Set the path of the default storage pool location, by running the following command:**								
`$ export LIBVIRT_POOL_DIRECTORY=<name_of_target_directory> `
 
_Note:To ensure libvirt has read and write access permissions, use a subdirectory of libvirt’s storage directory. The default is /var/lib/libvirt/images/_		
							
**Create a libvirt pool by running the following command:** 				
[...]

PR #436, opened 2024-07-18
Adds script to auto create libvirt podvm config - #435

Saripalli-lavanya:
Added script to Automatically create libvirt pool and volume on KVM host and upon successful creation same script will create peer-pod secret, config map & ssh secret. - #435
**- What I did**

**- How to verify it**

### DIFF
--- a/config/peerpods/podvm/libvirt-config-manager.sh
+++ b/config/peerpods/podvm/libvirt-config-manager.sh
@@ -1,0 +1,380 @@
+#!/bin/bash
+
+ACTION=$1
+SSH_KEY_PASSPHRASE=$2
+
+# Define namespace and secret name
+NAMESPACE="openshift-sandboxed-containers-operator"
+SECRET_NAME="peer-pods-secret"
+
+# Retrieve KVM host details from secret ocp-libvirt-secret which will be defined by user/admin.
+if [ -z "$KVM_HOST_ADDRESS" ]; then
+    KVM_HOST_ADDRESS=$(
+        oc get secret ocp-libvirt-secret -n "$NAMESPACE" \
+        -o jsonpath="{.data.LPAR_IP}" | base64 --decode
+    )
+fi
+
+if [ -z "$KVM_HOST_USERNAME" ]; then
+    KVM_HOST_USERNAME=$(
+        oc get secret ocp-libvirt-secret -n "$NAMESPACE" \
+        -o jsonpath="{.data.LPAR_USER}" | base64 --decode
+    )
+fi
+
+if [ -z "$KVM_HOST_PASSWORD" ]; then
+    KVM_HOST_PASSWORD=$(
+        oc get secret ocp-libvirt-secret -n "$NAMESPACE" \
+        -o jsonpath="{.data.LPAR_PSWD}" | base64 --decode
+    )
+fi
+
+if [ -z "$HOST_KEY_CERTS" ]; then
+    HOST_KEY_CERTS=$(
+        oc get secret ocp-libvirt-secret -n "$NAMESPACE" \
+        -o jsonpath="{.data.HOST_KEY_CERTS}" | base64 --decode
+    )
+fi
+
+if [ -z "$REDHAT_OFFLINE_TOKEN" ]; then
+    REDHAT_OFFLINE_TOKEN=$(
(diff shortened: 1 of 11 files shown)


savitrilh:
@Saripalli-lavanya Add the script to cleanup when kataconfig is unconfigured

gkurz:
Hi @Saripalli-lavanya and @savitrilh ! Thanks for all the huge work you are doing :smile:

I'll just suggest that you break down the first commit into smaller logical and bisectable pieces. Even if this might look like extra work, it is largely beneficial for everyone :
- building a patch series incrementally allows contributors to spot bugs very early, often before even creating the PR
- it is way easier for reviewers to provide valuable feedback and speed up the acceptance of the PR
- subtle bugs can be found by bisecting, which is way more efficient than code inspection

Also, we don't have pre-merge CI currently : bugs will be found late. Being able to bisect can really make a difference for non-obvious ones.

Cheers !

controllers/image_generator.go
```diff
@@ -59,6 +59,8 @@ const (
 	LibvirtProvider               = "libvirt"
 	peerpodsImageJobsPathLocation = "/config/peerpods/podvm"
 	azureImageGalleryPrefix       = "PodVMGallery"
+	preConfigJobName              = "osc-podvm-pre-configuration"
```
bpradipt, line 62: On a generic note, I'm really not in favour of adding more jobs to the already complicated image generation code. 
Can't the jobs be run in a standalone fashion  (admin runs oc apply -f <job.yaml> or a script) to start with without code integration ?

savitrilh: @bpradipt  If we provide a builder image built with these scripts in the job, the OCP admin can run this job as a standalone process. We believe that the creation and deletion of ConfigMaps, Secrets, and Libvirt pools/volumes can be automated with OSC. We attempted to synchronize this with the Image Job and the code. Additionally, we enhanced it to be scalable so that other providers can also manage their ConfigMaps and Secrets by enhancing provider specific data.

bpradipt: Can't the configmap and secret updates be de-coupled from image generation ?

savitrilh: @bpradipt  Certainly possible to de-couple configmap and secret from image generation
This can be implemented as below:

File config/peerpods/podvm/libvirt-podvm-secret.yaml contains all user Input.
Add new Key/Flag in the config/peerpods/podvm/libvirt-podvm-secret.yaml file as below:
`IMAGE_GEN ="YES"`
This is an option parameter, by default it can be considered as `Yes` . If Admin user doesn't wish to generate image, then it can be set to "NO". Then it will not create image config map and pool, vol.
By default, it can create peer-pods-cm configmap and secrets.

If this option is fine, we can implement and drop that code changes

bpradipt: Let's hear what other maintainers have to say before you make any changes. 

The related issue is quite old and lots have changed since then. So I think some discussion is needed on the linked issue - https://github.com/openshift/sandboxed-containers-operator/issues/435

savitrilh: Thanks @bpradipt  

Please let me know where I need to initiate this discussion. I will initiate the discussion

config/peerpods/podvm/libvirt-config-manager.sh
```diff
@@ -0,0 +1,380 @@
...
+        "${KVM_HOST_USERNAME}@${kvm_host_address}" << EOF
+        sudo virsh pool-info "${libvirt_pool}" >/dev/null 2>&1
+        POOL_EXISTS=\$?
+        sudo virsh vol-info --pool "${libvirt_pool}" "${libvirt_vol_name}" \
+            >/dev/null 2>&1
+        VOL_EXISTS=\$?
+
+        if [ "\$POOL_EXISTS" -eq 0 ] && [ "\$VOL_EXISTS" -eq 0 ]; then
+            echo "A Libvirt pool named '${libvirt_pool}' with volume '${libvirt_vol_name}' already exists on the KVM host. Please choose a different name."
+            exit 0
```
littlejawa, line 254: There is no way to re-use an existing pool?
No big deal if it's a mandatory requirement, but if we can keep a little bit of flexibility, that would be nice - if only for development/test setups :-)

PR #438, opened 2024-07-19
Remove mount points

bpradipt:


config/peerpods/podvm/lib.sh
```diff
@@ -198,8 +198,12 @@ function prepare_source_code() {
 
     if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
         sed 's/default SetPolicyRequest := true/default SetPolicyRequest := false/; s/default ExecProcessRequest := true/default ExecProcessRequest := false/' \
-            "${podvm_dir}"/files/etc/kata-opa/default-policy.rego > "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego
-        ln -sf "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego  "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
+            "${podvm_dir}"/files/etc/kata-opa/default-policy.rego >"${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego
+        ln -sf "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
```
gkurz: These are just dropping some extra space characters, right ?

config/peerpods/podvm/lib.sh
```diff
@@ -198,8 +198,12 @@ function prepare_source_code() {
...
+
+        # Remove the root disk mounts from the podvm
+        rm -f "${podvm_dir}"/files/etc/systemd/system/{run-image,run-kata\\x2dcontainers}.mount
+        rm -f "${podvm_dir}"/files/etc/systemd/system/multi-user.target.wants/{run-image,run-kata\\x2dcontainers}.mount
```
gkurz: Can there be other mount files from some custom podvm layout ? Should'nt they be removed all the same, i.e. `find "${podvm_dir}"/files/etc/systemd/system --name *.mount --delete` ?

config/peerpods/podvm/lib.sh
bpradipt: Right.. Will wait for https://github.com/openshift/sandboxed-containers-operator/pull/434 to get merge and rebase this one.

config/peerpods/podvm/lib.sh
bpradipt: Hmm the image is generated with files that is coming from known sources and contents are known. So finding and deleting may not be needed. Wdyt ?

gkurz: If this list of known sources is authoritative, then it is fine. Maybe add a comment with a link to the CAA repo where people can check the list ?

c3d: Just curious, why the `\x2d` and not `-`? It looks weird.

c3d: Since the answer was given to me on Slack, here it is: https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#String%20Escaping%20for%20Inclusion%20in%20Unit%20Names

So `run-kata\x2dcontainers` is the `systemd` unit spelling for `run/kata-containers`.

bpradipt: Adding the systemd document link here for reference - https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#String%20Escaping%20for%20Inclusion%20in%20Unit%20Names

config/peerpods/podvm/lib.sh
```diff
@@ -208,18 +208,23 @@ function prepare_source_code() {
...
         echo "Setting custom agent policy to CoCo's recommended policy"
         sed 's/default ReadStreamRequest := true/default ReadStreamRequest := false/;
             s/default ExecProcessRequest := true/default ExecProcessRequest := false/' \
-            "${podvm_dir}"/files/etc/kata-opa/default-policy.rego > "${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego
+            "${podvm_dir}"/files/etc/kata-opa/default-policy.rego >"${podvm_dir}"/files/etc/kata-opa/coco-default-policy.rego
         ln -sf coco-default-policy.rego "${podvm_dir}"/files/etc/kata-opa/default-policy.rego
+
+        # Fix disk mounts for CoCo
+        remove_mount_units
+        create_overlay_mount_unit
```
snir911, line 229: This should be moved to a standalone conditional otherwise the mounts operation will be skipped if custom agent policy is set

config/peerpods/podvm/lib.sh
```diff
@@ -208,7 +208,7 @@ function prepare_source_code() {
     # links must be relative
     if [[ "${AGENT_POLICY}" ]]; then
         echo "Custom agent policy is being set through the AGENT_POLICY value"
-        echo ${AGENT_POLICY} | base64 -d >"${podvm_dir}"/files/etc/kata-opa/custom.rego
+        echo "${AGENT_POLICY}" | base64 -d >"${podvm_dir}"/files/etc/kata-opa/custom.rego
```
snir911, line 211: nit, I'd squash all the shellcheck and style fixes

config/peerpods/podvm/lib.sh
snir911: also, I'd add a comment in-code to describe the reasoning behind `create_overlay_mount_unit`

snir911 (approved):
LGTM, Thanks

PR #439, opened 2024-07-26
Make ssh-secret optional for podvm creation job

bpradipt:
The correct way to make it optional is to provide the optional keyword in the volumes spec.

Ref: https://kubernetes.io/docs/concepts/configuration/secret/#restriction-secret-must-exist

Fixes: #[KATA-3233](https://issues.redhat.com//browse/KATA-3233)


littlejawa (approved):
/lgtm
Thanks @bpradipt

PR #443, opened 2024-08-11
podvm: disable cloud-init unsafe modules for CoCo

snir911:
this is executed when CONFIDENTIAL_COMPUTE_ENABLED=yes or based on the EXPLICIT_adapt_cloud_init value, if set

in either case, this is currently disabled for the libvirt provider

Fixes: [KATA-3269](https://issues.redhat.com//browse/KATA-3269)

### DIFF
--- a/config/peerpods/podvm/lib.sh
+++ b/config/peerpods/podvm/lib.sh
@@ -229,6 +229,11 @@ function prepare_source_code() {
         create_overlay_mount_unit
     fi
 
+    # disable ssh and unsafe cloud-init modules
+    if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]] || [[ -n "$CUSTOM_CLOUD_INIT_MODULES" ]]; then
+       [[ "$CUSTOM_CLOUD_INIT_MODULES" != "no" ]] && [[ "$CLOUD_PROVIDER" != "libvirt" ]] && set_custom_cloud_init_modules
+    fi
+
     # Validate and copy HKD for IBM Z Secure Enablement 
     if [[ "$SE_BOOT" == "true" ]]; then
         if [[ -z "$HOST_KEY_CERTS" ]]; then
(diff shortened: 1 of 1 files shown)


config/peerpods/podvm/lib.sh
```diff
@@ -229,6 +229,11 @@ function prepare_source_code() {
         create_overlay_mount_unit
```
bpradipt, line 229: I think `adapt_cloud_init` can be called here ..

gkurz (commented):
@snir911 what's the `tmporary change` commit for ?

snir911:
tmporary change was to test wheter this is works with libvirt provider, currently thanks to @ajayvic it seems to be failing so i removed this patch for now
```
[  OK  ] Reached target Cloud-init target.
         Starting Agent Protocol Forwarder...
         Starting Record Runlevel Change in UTMP...
[  OK  ] Finished Record Runlevel Change in UTMP.
[FAILED] Failed to start Agent Protocol Forwarder.
See 'systemctl status agent-protocol-forwarder.service' for details.
[  OK  ] Started Confidential Data Hub TTRPC API Server.
```

config/peerpods/podvm/lib.sh
snir911: not an option ATM as this should be disabled for libvirt

bpradipt: How about something like the following ?
[[ "$CLOUD_PROVIDER" != "libvirt" ]] && adapt_cloud_init

bpradipt: If you want the ability to enable/disable the cloud-init modules explicitly then it makes sense to have it separately..
Can you please change the config param to CUSTOM_CLOUD_INIT_MODULES and the function name to set_custom_cloud_init_modules

config/peerpods/podvm/lib.sh
```diff
@@ -229,6 +229,11 @@ function prepare_source_code() {
...
     fi
 
+    # disable ssh and unsafe cloud-init modules
+    if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]] || [[ -n "$CUSTOM_CLOUD_INIT_MODULES" ]]; then
+       [[ "$CUSTOM_CLOUD_INIT_MODULES" != "no" ]] && [[ "$CLOUD_PROVIDER" != "libvirt" ]] && set_custom_cloud_init_modules
```
bpradipt, line 234: where is `CUSTOM_CLOUD_INIT_MODULES` set ?

bpradipt: Ok, so it's there only for disabling..

bpradipt (approved):
/lgtm

gkurz (approved):
Under the assumption that the content of `99_coco_only_allow.cfg` is correct, the patch looks good to me.

Thanks @snir911 !

PR #446, opened 2024-08-19
Sync release-1.7 with devel

beraldoleal:
Currently release-1.7 is 14 commits behind. I'm not sure how Greg was doing before, but the discussed plan was to sync this branch and then create the alpha.1 tag. This should trigger another 1.7 downstream build.

Is there any commit that should not be included?


bpradipt (approved):
/lgtm
Thanks @beraldoleal

cpmeadors (approved):
looks good

PR #451, opened 2024-08-30
Build fixes and sdk upgrade

bpradipt:
Misc changes to address the following:
1. Allow building of operator controller, bundle etc
~~2. Enable tracking changes to manifests. These are generated but need to be tracked so that we know the changes to rbac, csv, manager yaml etc either as part of code changes or sdk upgrade.~~

Removed the commit related to tracking the manifest changes. Will use a separate PR to handle it as it needs some discussions w.r.to which generated files needs tracking.


Makefile
```diff
@@ -76,20 +83,14 @@ space := $() $()
 comma := ,
 GOFLAGS := -tags=$(subst $(space),$(comma),$(strip $(BUILTIN_CLOUD_PROVIDERS)))
 
-#
-# If SKIP_TESTS is set, the test target will *not* run `go test`.
-# This is to be able to temporarily work around test failures when doing
-# local development.
-SKIP_TESTS =
```
pmores, line 83: Won't removing this risk breaking CI again, along with restoring an annoyance for devs who don't necessarily want to run the tests on each build?

bpradipt: If you see the `test` target is no longer called as part of other build targets. It's explicitly called. So it should be fine imho

Makefile
```diff
@@ -122,62 +123,75 @@ vet: ## Run go vet against code.
...
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
+# If you wish to build the manager image targeting other platforms you can use the --platform flag.
+# (i.e. docker build --platform linux/arm64). However, you must enable docker buildKit for it.
+# More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
-docker-build: test ## Build docker image with the manager.
-	docker build \
+docker-build: ## Build docker image with the manager.
```
bpradipt, line 155: @pmores check this build target. Earlier docker-build always called `test` target

bpradipt: On a safer side, do you recommend keeping SKIP_TESTS as-is?

gkurz: @bpradipt where does the `test` dep come from ? SDK ?

bpradipt: yes

bpradipt: I'm wrong. The test dep was infact added by me ;) 
```
c579d7f5 (Pradipta Banerjee      2021-09-08 03:39:56 -0400 146) docker-build: test ## Build docker image with the manager.
39574e79 (John Allberg           2023-03-09 21:10:45 +0100 147)         docker build \
39574e79 (John Allberg           2023-03-09 21:10:45 +0100 148)                 -t ${IMG} \
```

gkurz: Cool, let's drop it and match the SDK then, and drop the `SKIP_TESTS` stuff :smile:

Makefile
pmores: Ah alright, I missed that, sorry.

bpradipt:
@gkurz @pmores is this good to go ?

gkurz:
@bpradipt CI required test is failing at https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_sandboxed-containers-operator/451/pull-ci-openshift-sandboxed-containers-operator-devel-ci-index-openshift-sandboxed-containers-operator-bundle/1835572392877887488.

Older version `v1.25.3` of Operator SDK is being used instead of `v1.36.1`...

```
[1/2] STEP 13/15: RUN export ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac) OS=$(uname | awk '{print tolower($0)}') OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.25.3; curl -LO ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH}; mv operator-sdk_${OS}_${ARCH} operator-sdk; chmod +x operator-sdk
```

Could this come from https://github.com/bpradipt/sandboxed-containers-operator/blob/build-fix/bundle-custom.Dockerfile#L21 ?

cpmeadors:
> Could this come from https://github.com/bpradipt/sandboxed-containers-operator/blob/build-fix/bundle-custom.Dockerfile#L21 ?

I think we deprecated this file, but I am trying to remember why we had it and why we got rid of it.  I think ci-operator assumes that the container file is "Dockerfile" in the root of the repo.   This file is also out of date.  The solution one way or another is to have a container file that is up to date with downstream configuration and make sure ci-operator is using that one.

gkurz:
> > Could this come from https://github.com/bpradipt/sandboxed-containers-operator/blob/build-fix/bundle-custom.Dockerfile#L21 ?
> 
> I think we deprecated this file, but I am trying to remember why we had it and why we got rid of it. I think ci-operator assumes that the container file is "Dockerfile" in the root of the repo. This file is also out of date. The solution one way or another is to have a container file that is up to date with downstream configuration and make sure ci-operator is using that one.

https://github.com/openshift/release/blame/master/ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml#L18 still references `bundle-custom.Dockerfile` and this matches what is visible in the logs. This file should be updated.

gkurz:
> > > Could this come from https://github.com/bpradipt/sandboxed-containers-operator/blob/build-fix/bundle-custom.Dockerfile#L21 ?
> > 
> > 
> > I think we deprecated this file, but I am trying to remember why we had it and why we got rid of it. I think ci-operator assumes that the container file is "Dockerfile" in the root of the repo. This file is also out of date. The solution one way or another is to have a container file that is up to date with downstream configuration and make sure ci-operator is using that one.
> 
> https://github.com/openshift/release/blame/master/ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml#L18 still references `bundle-custom.Dockerfile` and this matches what is visible in the logs. This file should be updated.

FWIW CI passed with the addition of the following patch :

https://github.com/gkurz/sandboxed-containers-operator/commit/398a1880045fe3bd27a83c70abbd4dfe541bbdb0.patch

@bpradipt can you add it to this PR or I can probably do it as well ?

bpradipt:
> > > > Could this come from https://github.com/bpradipt/sandboxed-containers-operator/blob/build-fix/bundle-custom.Dockerfile#L21 ?
> > > 
> > > 
> > > I think we deprecated this file, but I am trying to remember why we had it and why we got rid of it. I think ci-operator assumes that the container file is "Dockerfile" in the root of the repo. This file is also out of date. The solution one way or another is to have a container file that is up to date with downstream configuration and make sure ci-operator is using that one.
> > 
> > 
> > https://github.com/openshift/release/blame/master/ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml#L18 still references `bundle-custom.Dockerfile` and this matches what is visible in the logs. This file should be updated.
> 
> FWIW CI passed with the addition of the following patch :
> 
> https://github.com/gkurz/sandboxed-containers-operator/commit/398a1880045fe3bd27a83c70abbd4dfe541bbdb0.patch
> 
> @bpradipt can you add it to this PR or I can probably do it as well ?

Thanks @gkurz, let me include as a separate commit in this PR

bpradipt:
@gkurz this actually brings back the discussion to the generated files. 
How about we track some limited generated files which are needed ?
- bundle.Dockerfile (use the same in the CI config as well) 
- rbac.yaml 

This will help us to know what changed when upgrading sdk (for bundle) or making changes to the code (for rbac)

gkurz:
> @gkurz this actually brings back the discussion to the generated files. How about we track some limited generated files which are needed ?
> 
> * bundle.Dockerfile (use the same in the CI config as well)
> * rbac.yaml
> 
> This will help us to know what changed when upgrading sdk (for bundle) or making changes to the code (for rbac)

I agree that we might have to reconsider our position with these files. I suggest we go on with this PR with my patch added and we address the case of generated files separately. Works with you @bpradipt ?

config/manager/kustomization.yaml
```diff
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/openshift_sandboxed_containers/openshift-sandboxed-containers-operator
-  newTag: 1.7.0
+  newTag: v1.7.0
```
gkurz, line 16: Do we want to keep this change ? It seems we didn't with previous Operator SDK updates.

bpradipt: In latest Makefile, the images are referenced with `v${VERSION}`. So I think we'll need this change no?

gkurz:
Fixes: https://issues.redhat.com/browse/KATA-3342

PR #455, opened 2024-09-04
Patch the CSV to handle breaking change for peer-pods

bpradipt:


config/manager/manager.yaml
```diff
@@ -88,6 +88,14 @@ spec:
               value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-builder-rhel9:1.7.0
             - name: RELATED_IMAGE_PODVM_PAYLOAD
               value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9:1.7.0
+            - name: RELEASE_VERSION
+              value: 1.7.0
```
gkurz, line 92: This could be inferred out of the version tag of one of the related images but I guess having an explicit env results in simpler code in the operator.

PR #458, opened 2024-09-11
Require CRI-O 1.28

gkurz:
Peer pods now require a feature that is only available in CRI-O 1.28 and newer. Let's make sure OSC cannot be deployed without this requirement being met. This means that users of OCP 4.14 and older will need to upgrade to 4.15 in order to install OSC.

The OSC 1.7.0 downstream release will use a bigger hammer to prevent installation. No real need to propagate this to the release-1.7 branch.


beraldoleal (approved):
LGTM, thanks

pmores (approved):
lgtm thanks @gkurz !

littlejawa (approved):
lgtm
Thanks!

PR #459, opened 2024-09-11
Configurable extension

gkurz:
This simplifies custom builds for OCP with respect to the CoreOS extension name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,11 @@ GOFLAGS := -tags=$(subst $(space),$(comma),$(strip $(BUILTIN_CLOUD_PROVIDERS)))
 # local development.
 SKIP_TESTS =
 
+#
+# The CoreOS extension that ships kata-containers
+#
+SANDBOXED_CONTAINERS_EXTENSION ?= kata-containers
+
 .PHONY: all
 all: build
 
@@ -240,10 +245,12 @@ $(YQ): $(LOCALBIN)
 	test -s $(LOCALBIN)/yq || GOBIN=$(LOCALBIN) go install github.com/mikefarah/yq/v4@latest
 
 .PHONY: bundle
-bundle: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
+bundle: manifests kustomize yq ## Generate bundle manifests and metadata, then validate generated files.
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle $(BUNDLE_GEN_FLAGS)
+	$(YQ) eval '(.spec.install.spec.deployments[].spec.template.spec.containers[].env.[] | select(.name == "SANDBOXED_CONTAINERS_EXTENSION") | .value) |= "$(SANDBOXED_CONTAINERS_EXTENSION)"' \
+		-i bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
 	operator-sdk bundle validate ./bundle
 
 .PHONY: bundle-build
--- a/Makefile
+++ b/Makefile
@@ -212,6 +212,7 @@ $(LOCALBIN):
 KUSTOMIZE ?= $(LOCALBIN)/kustomize
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
+YQ ?= $(LOCALBIN)/yq
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v4.5.5
@@ -232,6 +233,11 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
 	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+
+.PHONY: yq
+yq: $(YQ) ## Download yq locally if necessary.
+$(YQ): $(LOCALBIN)
+	test -s $(LOCALBIN)/yq || GOBIN=$(LOCALBIN) go install github.com/mikefarah/yq/v4@latest
 
 .PHONY: bundle
 bundle: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.


Makefile
```diff
@@ -240,10 +245,12 @@ $(YQ): $(LOCALBIN)
 	test -s $(LOCALBIN)/yq || GOBIN=$(LOCALBIN) go install github.com/mikefarah/yq/v4@latest
 
 .PHONY: bundle
-bundle: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
+bundle: manifests kustomize yq ## Generate bundle manifests and metadata, then validate generated files.
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle $(BUNDLE_GEN_FLAGS)
+	$(YQ) eval '(.spec.install.spec.deployments[].spec.template.spec.containers[].env.[] | select(.name == "SANDBOXED_CONTAINERS_EXTENSION") | .value) |= "$(SANDBOXED_CONTAINERS_EXTENSION)"' \
+		-i bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
```
pmores, line 253: I'd actually prefer doing the replacement in `config/manager/manager.yaml` before kustomizing instead of in the generated file after kustomizing since the difference between the template and the resulting generated file will end up konfusing someone I'm afraid.  But this works too.

gkurz: If the replacement is done in `config/manager/manager.yaml` then it will be dirty... which basically negates the benefit of this commit or I am missing something ?

pmores: There would still be benefit in setting this in `Makefile` and not having to know or find out where exactly you as a (new) developer need to make this change.  As I said, I'm fine with this way, too.

pmores (approved):
lgtm thanks @gkurz !

littlejawa (approved):
lgtm
Thanks @gkurz !

Makefile
gkurz: On the confusion side : people still need to explicitly set `SANDBOXED_CONTAINERS_EXTENSION`  to something != from the default value (`kata-containers`) to change the default behavior. They shouldn't be that much surprised.

bpradipt:
While technically this patch is fine, my worry is the more we deviate from a standard operator-sdk based Makefile, the harder it will be during sdk upgrades. This is from my personal experience when upgrading sdk.

An alternate is to set the extension name dynamically based on the environment by patching the CSV during start of the controller manager. 
This is just a suggestion and in no way blocking this PR

gkurz:
> While technically this patch is fine, my worry is the more we deviate from a standard operator-sdk based Makefile, the harder it will be during sdk upgrades. This is from my personal experience when upgrading sdk.
> 

Thanks for this comment @bpradipt , that I was kind of expecting :wink: . I'm pretty much ignorant on operator sdk... can you describe a bit more how an sdk upgrade could be impacted by this PR ?

> An alternate is to set the extension name dynamically based on the environment by patching the CSV during start of the controller manager. This is just a suggestion and in no way blocking this PR

This could be done as well but I'd prefer to explore non-runtime operator sdk friendly ways first.

bpradipt:
> > While technically this patch is fine, my worry is the more we deviate from a standard operator-sdk based Makefile, the harder it will be during sdk upgrades. This is from my personal experience when upgrading sdk.
> 
> Thanks for this comment @bpradipt , that I was kind of expecting 😉 . I'm pretty much ignorant on operator sdk... can you describe a bit more how an sdk upgrade could be impacted by this PR ?
> 
operator-sdk upgrades also brings in updated and new makefile targets, tool versions used etc.  The new makefile targets are helpful during development as well as CI (it's a separate thing that our CI is broken, but I'm hopeful it'll be fixed soon). 
A latest example is in this PR - https://github.com/openshift/sandboxed-containers-operator/pull/451/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52
The way I do it is to initialise a dummy operator project with new operator-sdk and manually reconcile the Makefile changes to the existing one. This gets tedious depending on the divergence. I have not found a better alternative for this exercise :-( . 

> This could be done as well but I'd prefer to explore non-runtime operator sdk friendly ways first.

Any specific reason for this preference? I was thinking removing this build time option will prevent issues when installing dev build catalog on OCP, especially w.r.to MCP going in bad state.

pmores:
> > While technically this patch is fine, my worry is the more we deviate from a standard operator-sdk based Makefile, the harder it will be during sdk upgrades. This is from my personal experience when upgrading sdk.
> 
> Thanks for this comment @bpradipt , that I was kind of expecting 😉 . I'm pretty much ignorant on operator sdk... can you describe a bit more how an sdk upgrade could be impacted by this PR ?

It's a lot like maintaining a private fork.  How hard it gets is entirely case-by-case.  I'm assuming that maintaining a change like this one across upgrades is more annoying than complex since it's a simple addition.

pmores:
> > This could be done as well but I'd prefer to explore non-runtime operator sdk friendly ways first.
> 
> Any specific reason for this preference? I was thinking removing this build time option will prevent issues when installing dev build catalog on OCP, especially w.r.to MCP going in bad state.

If this can be determined at runtime _reliably_ then that would be the better solution.  It's just one-off though, only works for this particular case, the broader topic of whether to track generated files remains the same.

bpradipt:
> If this can be determined at runtime _reliably_ then that would be the better solution. It's just one-off though, only works for this particular case, the broader topic of whether to track generated files remains the same.

We do detect OpenShift here - https://github.com/openshift/sandboxed-containers-operator/blob/devel/main.go#L125
So we can set "SANDBOXED_CONTAINERS_EXTENSION" env variable on detecting OpenShift and that should work. But then why we didn't do it earlier ? Is there no way to reliably detect OKD and OpenShift? I don't have the answers yet. And unless someone else knows about it, an alternate approach may take time.

On a side note, as I have mentioned earlier, I have no intention to block this PR. Just wanted to highlight the challenges I typically face when upgrading the sdk and using development catalog for testing ;)

The issues with upgrade, syncing manifest changes between upstream and downstream, broken CI etc needs separate attention anyway.  Unfortunately it's a mess and we'll need to find out practical solutions.

pmores:
> > If this can be determined at runtime _reliably_ then that would be the better solution. It's just one-off though, only works for this particular case, the broader topic of whether to track generated files remains the same.
> 
> We do detect OpenShift here - https://github.com/openshift/sandboxed-containers-operator/blob/devel/main.go#L125 So we can set "SANDBOXED_CONTAINERS_EXTENSION" env variable on detecting OpenShift and that should work. But then why we didn't do it earlier ?

Maybe simply just because no-one has thought of it...

> Is there no way to reliably detect OKD and OpenShift? I don't have the answers yet. And unless someone else knows about it, an alternate approach may take time.

How many flavours of OpenShift do we need to detect?  If it's just OCP and OKD, and we're already able to detect OCP, then perhaps we could just do `if controllers.IsOpenShift() then it's OCP else it's OKD`.

> The issues with upgrade, syncing manifest changes between upstream and downstream, broken CI etc needs separate attention anyway. Unfortunately it's a mess and we'll need to find out practical solutions.

Downstream needs to be able to perform full build including generated files generation.  I believe that's where problem is and where is should be solved.

bpradipt:
> How many flavours of OpenShift do we need to detect? If it's just OCP and OKD, and we're already able to detect OCP, then perhaps we could just do if controllers.IsOpenShift() then it's OCP else it's OKD.

I checked with okd folks and the suggested way is to look at the image in the clusterversion object. Example o/p from an okd cluster
```
oc get clusterversion version -o jsonpath='{.status.desired.image}'
quay.io/okd/scos-release:4.16.0-0.okd-scos-2024-08-21-155613
```

O/p from ocp cluster
```
quay.io/openshift-release-dev/ocp-release@sha256:2feae5767669c4773239e445d133541e2d384acec1624a94c1945a5794a09c51
```

We already read the clusterversion object https://github.com/openshift/sandboxed-containers-operator/blob/devel/controllers/utils.go#L189

So it should be possible to implement detection logic based on clusterversion object.

@gkurz does this sounds reasonable to you ?

gkurz:
> @gkurz does this sounds reasonable to you ?

My only concern would be future changes in the image URLs that could break the detection logic. But I guess we would catch that during development and fix the regexp. Let's give a try ! :smile:

bpradipt:
> > @gkurz does this sounds reasonable to you ?
> 
> My only concern would be future changes in the image URLs that could break the detection logic. But I guess we would catch that during development and fix the regexp. Let's give a try ! 😄

May be just search for `okd`. Should be safe ;)

pmores:
`okd` is just three letters, I'm afraid that could come up by chance in an irrelevant part of the URL.  How about checking for the `quay.io/okd` prefix?  That will only change if the OKD repo changes name or site.

gkurz:
Trying out the approach based on cluster image in #492 .

PR #460, opened 2024-09-16
Add support to use pre-built pod VM image for Azure

bpradipt:


config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -504,15 +530,127 @@ function create_image() {
...
+
+    # Set the IMAGE_VERSION and IMAGE_NAME
+    set_image_version_and_name
+
+    echo "Pulling the podvm image from the provided path"
+    IMAGE_SRC="/tmp/image"
+    EXTRACTION_DESTINATION_PATH="/image"
+    IMAGE_REPO_AUTH_FILE="/tmp/regauth/auth.json"
+
+    # Get the PODVM_IMAGE_TUPE and PODVM_IMAGE_SRC_PATH
```
snir911: nit: typo

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -504,15 +530,127 @@ function create_image() {
...
+    oci)
+        echo "Extracting the Azure image from the given path."
+
+        mkdir -p "${EXTRACTION_DESTINATION_PATH}" ||
+            error_exit "Failed to create the image directory"
+
+        extract_container_image "${PODVM_IMAGE_URL}" "${PODVM_IMAGE_TAG}" "${IMAGE_SRC}" "${EXTRACTION_DESTINATION_PATH}" "${IMAGE_REPO_AUTH_FILE}"
+
+        # Form the path of the podvm raw image.
+        PODVM_IMAGE_PATH="${EXTRACTION_DESTINATION_PATH}/rootfs${PODVM_IMAGE_SRC_PATH}"
```
snir911: i wonder if using "/rootfs/" would help to avoid wrong usage here

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -504,15 +530,127 @@ function create_image() {
...
+
+        # Check whether the podvm image is a valid vhd image or not.
+        validate_podvm_image "${PODVM_IMAGE_PATH}"
```
snir911: would it make sense to support also qcows for Azure? and for azure we'll convert it to vhd
currently qcow is allowed, but it will fail at some point if used

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -504,15 +530,127 @@ function create_image() {
...
+            --resource-group "${AZURE_RESOURCE_GROUP}" \
+            --gallery-name "${IMAGE_GALLERY_NAME}" \
+            --gallery-image-definition "${IMAGE_DEFINITION_NAME}" \
+            --gallery-image-version "${IMAGE_VERSION}" \
+            --os-vhd-uri "${VHD_URL}" \
+            --os-vhd-storage-account "${STORAGE_ACCOUNT_NAME}" \
+            --target-regions "${AZURE_REGION}" ||
+            error_exit "Failed to create the image version"
+
+        # Clean up
```
snir911, line 588: can the storage account can be deleted when image is in use?
Will the image-version, blob, etc will be cleaned on deletion?

snir911 (commented):
LGTM overall, added few comments

config/peerpods/podvm/azure-podvm-image-handler.sh
bpradipt: Qcow2 and/or raw also be possible. Do you know if qemu-img is available for install in the container ?

snir911: i'm pretty sure it's installed for the libvirt case, may not for the other providers

config/peerpods/podvm/azure-podvm-image-handler.sh
bpradipt: Once the image-version is created the storage account is not needed. I verified in on my setup.

config/peerpods/podvm/azure-podvm-image-handler.sh
bpradipt: I think the <dir>/rootfs is where the podvm image is available once extracted via umoci.

snir911: i meant only to the additional "/"

config/peerpods/podvm/lib.sh
```diff
@@ -395,16 +400,134 @@ function get_image_type_url_and_path() {
...
+    echo "Qcow2 disk name: ${qcow2disk}"
+    echo "Raw disk name: ${rawdisk}"
+    echo "VHD disk name: ${vhddisk}"
+
+    # Convert qcow2 to raw
+    qemu-img convert -f qcow2 -O raw "${qcow2disk}" "${rawdisk}" ||
+        error_exit "Failed to convert qcow2 to raw"
+
+    # Convert raw to vhd
+    resize_and_convert_raw_to_vhd_image "${rawdisk}"
```
snir911, line 461: why going through RAW? for the alignment?  ` -o subformat=fixed,force_size` doesn't suppose to handle that?

snir911 (approved):
LGTM thanks!, added one question

config/peerpods/podvm/lib.sh
bpradipt: Right.. 1MB alignment needs to be handled separately.

littlejawa (approved):
lgtm
Thanks @bpradipt

PR #461, opened 2024-09-26
metrics: add initial support for SLI/SLO metrics

beraldoleal:
Introduces high-level SLI metrics that can help derive SLOs.

Currently, the osc-monitor image only includes the kata-monitor binary. The idea is to reuse the same image to also include the metrics-server.

### DIFF
--- a/cmd/metrics/metrics.go
+++ b/cmd/metrics/metrics.go
@@ -1,0 +1,158 @@
+package main
+
+import (
+	"context"
+	"log"
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const runtimeClassName = "kata-remote"
+
+var (
+	runtimeClassAvailable = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_remote_runtimeclass_available",
+		Help: "Indicates if the " + runtimeClassName + " RuntimeClass is available (1) or not (0).",
+	})
+
+	kataConfigInstallationSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_config_installation_success",
+		Help: "Indicates if KataConfig installation is successful (1) or not (0).",
+	})
+
+	kataRemoteWorkloadFailureRatio = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_remote_workload_failure_ratio",
+		Help: "Percentage of " + runtimeClassName + " workloads that have failed.",
+	})
+
+	totalKataRemotePods = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_total_remote_pods",
+		Help: "Total number of " + runtimeClassName + " pods across all namespaces, regardless of their status.",
+	})
+
+	failedKataRemotePods = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_failed_remote_pods",
+		Help: "Total number of " + runtimeClassName + " pods across all namespaces, that status is != 'Running|Succeed'",
+	})
+)
+
+func collectMetricsData(clientset *kubernetes.Clientset, dynamicClient dynamic.Interface) {
+	// Defaults
+	runtimeClassAvailable.Set(0)
+	kataRemoteWorkloadFailureRatio.Set(0)
+	totalKataRemotePods.Set(0)
+	failedKataRemotePods.Set(0)
+
+	// Check if kata-remote runtime class is available
+	_, err := clientset.NodeV1().RuntimeClasses().Get(context.TODO(), runtimeClassName, metav1.GetOptions{})
+	if err == nil {
+		runtimeClassAvailable.Set(1)
+
+		// Fetch Pods for kata-remote workload metrics
+		pods, err := clientset.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			log.Printf("Error listing pods: %v", err)
+		} else {
+			totalPods := 0
+			successfulPods := 0
+			failedPods := 0
+			for _, pod := range pods.Items {
+				if pod.Spec.RuntimeClassName != nil && *pod.Spec.RuntimeClassName == runtimeClassName {
+					totalPods++
+					if pod.Status.Phase == "Running" || pod.Status.Phase == "Succeeded" {
+						successfulPods++
+					} else {
+						failedPods++
+					}
+				}
+			}
+
+			if totalPods > 0 {
+				kataRemoteWorkloadFailureRatio.Set(float64(failedPods) / float64(totalPods) * 100)
+			}
+
+			totalKataRemotePods.Set(float64(totalPods))
+			failedKataRemotePods.Set(float64(failedPods))
+		}
+	}
+
+	// Fetch KataConfig status
+	kataConfigGVR := schema.GroupVersionResource{
+		Group:    "kataconfiguration.openshift.io",
+		Version:  "v1",
+		Resource: "kataconfigs",
+	}
+	kataConfigs, err := dynamicClient.Resource(kataConfigGVR).List(context.TODO(), metav1.ListOptions{})
+	if err != nil || len(kataConfigs.Items) == 0 {
+		kataConfigInstallationSuccess.Set(0)
+	} else {
+		kataConfig := &kataConfigs.Items[0]
+		status, found, err := unstructured.NestedMap(kataConfig.Object, "status")
+		if err != nil || !found {
+			kataConfigInstallationSuccess.Set(0)
+		} else {
+			inProgress, _, _ := unstructured.NestedBool(status, "inProgress")
+			readyNodeCount, _, _ := unstructured.NestedInt64(status, "readyNodeCount")
+			totalNodeCount, _, _ := unstructured.NestedInt64(status, "totalNodeCount")
+			if !inProgress && readyNodeCount == totalNodeCount {
+				kataConfigInstallationSuccess.Set(1)
+			} else {
+				kataConfigInstallationSuccess.Set(0)
+			}
+		}
+	}
+}
+
+func getKubernetesClients() (*kubernetes.Clientset, dynamic.Interface, error) {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, nil, err
+	}
+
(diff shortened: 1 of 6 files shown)


cmd/metrics/metrics.go
```diff
@@ -0,0 +1,156 @@
...
+
+	clientset, dynamicClient, err := getKubernetesClients()
+	if err != nil {
+		log.Fatalf("Error setting up Kubernetes clients: %v", err)
+	}
+
+	http.Handle("/metrics", metricsHandler(clientset, dynamicClient))
+
+	log.Println("Starting OSC metrics server on port :8090")
+	log.Fatal(http.ListenAndServe(":8090", nil))
```
littlejawa: nit: Ideally, the address / port should configurable.

config/metrics/metrics-service.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: openshift-sandboxed-containers-operator
+  labels:
+    app: operator-metrics-server
+spec:
+  selector:
+    app: operator-metrics-server
+  ports:
+    - protocol: TCP
+      port: 8091
+      targetPort: 8091
```
littlejawa, line 14: If I get it right, this service definition won't work, because the go code is listening on 8090.

littlejawa (commented):
I'm not clear on the deployment side: 
If I understand correctly, the new metrics binary will be shipped in the same image as kata-monitor, but the image will be run twice: once for kata-monitor (with the daemonset defined by the operator's code), and a second time with the deployment files you're adding here?

I'm not sure about the benefit of using the same image in separate containers that do different things.
If it is the same image, why not run the two binaries in parallel from it? 
If we can't / don't want to have both binaries running from the same place, why do we make a single image ?

cmd/metrics/metrics.go
beraldoleal: Agree, this could be something for future PR, if you don't mind. This is just a basic implementation so we have something ready for the SRE team.

config/metrics/metrics-service.yaml
beraldoleal: Nice catch, changed the port on last minute. Fixed with a force push.

beraldoleal:
> I'm not clear on the deployment side: If I understand correctly, the new metrics binary will be shipped in the same image as kata-monitor, but the image will be run twice: once for kata-monitor (with the daemonset defined by the operator's code), and a second time with the deployment files you're adding here?

Yes. Currently kata-monitor binary is part of the "osc-monitor" image. And the idea is to reuse this image for "monitoring stuff". With both kata-monitor and the metrics server.

 > I'm not sure about the benefit of using the same image in separate containers that do different things. If it is the same image, why not run the two binaries in parallel from it? If we can't / don't want to have both binaries running from the same place, why do we make a single image ?

Even though they're in the same image, running them in separate containers makes it easier to manage and scale each one independently. This way, kata-monitor can continue running as a DaemonSet on every node, while the metrics server can be deployed separately as a Deployment, running only where needed.

Running both in the same container is possible and I considered this option too... but it can get tricky with managing processes, restarts, and resource allocation. By using the same image for both but keeping them in their own containers, we get the best of both worlds: fewer images to manage (specially on the downstream configuration), but still clear separation between the two services. Wdyt?

cmd/metrics/metrics.go
littlejawa: Sure :)

littlejawa:
> Running both in the same container is possible and I considered this option too... but it can get tricky with managing processes, restarts, and resource allocation. By using the same image for both but keeping them in their own containers, we get the best of both worlds: fewer images to manage (specially on the downstream configuration), but still clear separation between the two services. Wdyt?

I actually don't like the idea of running both processes from the same container :-) I only mentioned it to understand where the benefit is. So I'm fine running them separately.
But I tend to think that each container image should contain the things it needs to run, and nothing more - the simpler, the better. If we run them separately, then I'd be in favor of using separate images.

Now I understand this is to be balanced with the complexity of building/publishing/maintaining separate images, so I don't want to enforce that. Just feeling this is weird to deploy two applications in a single pod. 

I can go with it if other reviewers are OK.

beraldoleal:
> Now I understand this is to be balanced with the complexity of building/publishing/maintaining separate images, so I don't want to enforce that. Just feeling this is weird to deploy two applications in a single pod.

The separation won’t force us to use two applications in a single pod. Each will run in its own pod: kata-monitor as a DaemonSet and the metrics server as a Deployment.

cmd/metrics/metrics.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+const runtimeClassName = "kata"
```
gkurz: From the comment at line 48, this should be `"kata-remote"`.

cmd/metrics/metrics.go
```diff
@@ -0,0 +1,156 @@
...
+
+var (
+	runtimeClassAvailable = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_remote_runtimeclass_available",
+		Help: "Indicates if the kata-remote RuntimeClass is available (1) or not (0).",
```
gkurz:
```suggestion
		Help: "Indicates if the " + runtimeClassName + " RuntimeClass is available (1) or not (0).",
```

cmd/metrics/metrics.go
```diff
@@ -0,0 +1,156 @@
...
+
+func collectMetricsData(clientset *kubernetes.Clientset, dynamicClient dynamic.Interface) {
+	// Check if kata-remote runtime class is available
+	runtimeClassAvailable.Set(0)
+	_, err := clientset.NodeV1().RuntimeClasses().Get(context.TODO(), runtimeClassName, metav1.GetOptions{})
+	if err == nil {
+		runtimeClassAvailable.Set(1)
+	}
+
+	// Fetch Pods for kata-remote workload metrics
```
gkurz: No need to do that if `kata-remote` isn't available obviously.

cmd/metrics/metrics.go
```diff
@@ -0,0 +1,156 @@
...
+				if pod.Status.Phase == "Running" || pod.Status.Phase == "Succeeded" {
+					successfulPods++
+				} else {
+					failedPods++
+				}
+			}
+		}
+
+		if totalPods == 0 {
+			kataRemoteWorkloadSuccessRatio.Set(100)
```
gkurz: It is a bit questionable to report a metric in this case...

cmd/metrics/metrics.go
```diff
@@ -0,0 +1,156 @@
...
+			inProgress, _, _ := unstructured.NestedBool(status, "inProgress")
+			readyNodeCount, _, _ := unstructured.NestedInt64(status, "readyNodeCount")
+			totalNodeCount, _, _ := unstructured.NestedInt64(status, "totalNodeCount")
+			if !inProgress && readyNodeCount == totalNodeCount {
+				kataConfigInstallationSuccess.Set(1)
+			} else {
+				kataConfigInstallationSuccess.Set(0)
+			}
+		}
+	}
```
gkurz, line 111: This branch will also catch kata deployments with peer pods disabled. Is this intended ?

gkurz (commented):
> > Running both in the same container is possible and I considered this option too... but it can get tricky with managing processes, restarts, and resource allocation. By using the same image for both but keeping them in their own containers, we get the best of both worlds: fewer images to manage (specially on the downstream configuration), but still clear separation between the two services. Wdyt?
> 
> I actually don't like the idea of running both processes from the same container :-) I only mentioned it to understand where the benefit is. So I'm fine running them separately. But I tend to think that each container image should contain the things it needs to run, and nothing more - the simpler, the better. If we run them separately, then I'd be in favor of using separate images.
> 
> Now I understand this is to be balanced with the complexity of building/publishing/maintaining separate images, so I don't want to enforce that. Just feeling this is weird to deploy two applications in a single pod.
> 
> I can go with it if other reviewers are OK.

I guess it is acceptable to ship some extra metric logic in the existing image if it eases the work.

gkurz:
IIUC this is targeting peer pods exclusively. Maybe worth to mention it in the commit message.

cmd/metrics/metrics.go
beraldoleal: nice catch. Fixing with force push.

cmd/metrics/metrics.go
beraldoleal: I have inverted the logic to show the failure rate. Let me know if you are ok with it.

cmd/metrics/metrics.go
beraldoleal: Yes, if peer-pods is disabled the metrics will report accordingly until someone enable it.

cmd/metrics/metrics.go
beraldoleal: Solved.

cmd/metrics/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+
+	totalKataRemotePods = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_total_remote_pods",
+		Help: "Total number of kata-remote pods across all namespaces, regardless of their status.",
+	})
+
+	failedKataRemotePods = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_failed_remote_pods",
+		Help: "Total number of kata-remote pods across all namespaces, that status is != 'Running|Succeed'",
```
gkurz: ditto

cmd/metrics/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+
+	kataConfigInstallationSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_config_installation_success",
+		Help: "Indicates if KataConfig installation is successful (1) or not (0).",
+	})
+
+	kataRemoteWorkloadFailureRatio = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_remote_workload_failure_ratio",
+		Help: "Percentage of kata-remote workloads that have failed.",
```
gkurz: Sorry  I didn't mention this one :

```suggestion
		Help: "Percentage of " + runtimeClassName + " workloads that have failed.",
```

cmd/metrics/metrics.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+
+	totalKataRemotePods = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "kata_total_remote_pods",
+		Help: "Total number of kata-remote pods across all namespaces, regardless of their status.",
```
gkurz: ditto

gkurz (approved):
Just a couple of cosmetic changes left but rest LGTM.

Thanks @beraldoleal !

littlejawa (approved):
LGTM
Thanks @beraldoleal

beraldoleal:
@gkurz can we merge this one?

PR #463, opened 2024-10-03
Track files generated by `make bundle`

gkurz:
This adds the files generated by `make manifests` :

 config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
 config/rbac/role.yaml
 config/webhook/manifests.yaml

and `make bundle` :

 bundle/
 bundle.Dockerfile

This will ease Operator SDK upgrades.

Since the bundle is now tracked in the repo, the `bundle-custom.Dockerfile` file isn't needed anymore by CI. Drop it.

**- What I did**

**- How to verify it**


bpradipt (approved):
/lgtm
Thanks @gkurz 
Hopefully this will make the upstream CI green again and we can actually start adding some basic static tests as part of the upstream CI itself

bpradipt:
@gkurz I think we can merge this. I think the prow check failure is due to us pulling the latest envtest which needs golang 1.22. In my PR with Makefile changes, the envtest version is pinned and not using latest.

gkurz:
Fixes: https://issues.redhat.com/browse/KATA-3411

PR #468, opened 2024-10-23
Enable image signature check for CoCo

gkurz:
The kata-agent should manage the signature verification of container images pulled in the guest. It should be configured with two options:
- enable_signature_verification=true
- image_policy_file=<url>

While this can be easily achieved with annotations for regular kata pods, no mechanism exists for peer pods. In the meantime, let's hardcode the agent config directly in the podvm image for the CoCo case and have the kata-agent using it.

Fixes: https://issues.redhat.com/browse/KATA-3393

### DIFF
--- a/config/peerpods/podvm/lib.sh
+++ b/config/peerpods/podvm/lib.sh
@@ -244,6 +244,18 @@ function prepare_source_code() {
         fi
     fi
 
+    # Enable image signature check
+    if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
+	cat<<EOF>"${podvm_dir}"/files/etc/agent-config.toml
+server_addr = "unix:///run/kata-containers/agent.sock"
+guest_components_procs = "none"
+image_registry_auth = "file:///run/peerpod/auth.json"
+enable_signature_verification = true
+image_policy_file = "kbs:///default/security-policy/osc"
+EOF
+	sed -i 's,/run/peerpod/agent-config.toml,/etc/agent-config.toml,' \
+	    "${podvm_dir}"/files/etc/systemd/system/kata-agent.service
+    fi
 }
 
 # Download and extract the pause container image


config/peerpods/podvm/lib.sh
```diff
@@ -244,6 +244,18 @@ function prepare_source_code() {
         fi
     fi
 
+    # Enable image signature check
+    if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
+	cat<<EOF>"${podvm_dir}"/files/etc/agent-config.toml
+server_addr = “unix:///run/kata-containers/agent.sock”
```
wainersm: Hi @gkurz !

Could you fix the encoding of symbols like “ and ‘ (" and ')?

wainersm:
Hi @gkurz !

It LGTM; let's see if it works as expected :D

beraldoleal (approved):
LGTM, thanks

wainersm:
Hi @gkurz !

I reviewed the code and it's in align with the experiments I did upstream, it should work. Also our internals tests showed that the files were proper modified on the podvm image. So from my side:

/lgtm

wainersm (approved):

config/peerpods/podvm/lib.sh
```diff
@@ -244,6 +244,18 @@ function prepare_source_code() {
         fi
     fi
 
+    # Enable image signature check
+    if [[ "$CONFIDENTIAL_COMPUTE_ENABLED" == "yes" ]]; then
+	cat<<EOF>"${podvm_dir}"/files/etc/agent-config.toml
+server_addr = "unix:///run/kata-containers/agent.sock"
+guest_components_procs = "none"
+image_registry_auth = "file:///run/peerpod/auth.json"
+enable_signature_verification = true
```
snir911, line 253: What happens if image_policy_file doesn't exist? will it fail? should we allow some variable in the podvm-image-cm to skip signature check?

gkurz: IIUC agent will crash if the KBS doesn't have the policy. It is up to the KBS _owner_ to make sure there's a default container policy.

snir911 (approved):

config/peerpods/podvm/lib.sh
gkurz: Let's go without an extra knob for now as it isn't strictly needed for the upcoming release.

PR #469, opened 2024-10-29
alerts: adding prometheus rules for basic alerts

beraldoleal:
Let's start with those two alerts. The latest one is a basic alert, the easiest way to emit information about clusters with OSC installed for our internal dashboards. Only the first one will be converted into an incident report.

**- What I did**

New clusters will have those prometheus rules.

**- How to verify it**

Deploy the operator and check your alerts into your cluster UI.

**- Description for the changelog**[

New alerts: added basic Prometheus rules for OSC monitoring.

### DIFF
--- a/config/metrics/metrics-prometheus-rules.yaml
+++ b/config/metrics/metrics-prometheus-rules.yaml
@@ -1,0 +1,25 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: osc-alerts
+  namespace: openshift-sandboxed-containers-operator
+spec:
+  groups:
+  - name: osc_alerts
+    rules:
+    - alert: KataRemoteWorkloadFailureHigh
+      expr: kata_remote_workload_failure_ratio > 25
+      for: 30m
+      labels:
+        severity: warning
+      annotations:
+        summary: "High Kata Remote Workload Failure Ratio"
+        description: "The failure ratio of kata-remote workloads is above 25% for more than 30 minutes. This may indicate issues with the runtime or configuration."
+
+    - alert: kata_active_instance
+      expr: vector(1)
+      labels:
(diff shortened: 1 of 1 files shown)


beraldoleal:
@gkurz @bpradipt those are the two alerts that we came up after a few meetings with the ARO team. This will be consumed by Prometheus -> ARO-RP -> Geneva.

Flagging as draft because I'm still waiting some confirmation from them, but please feel free to comment on top this in meanwhile.

gkurz (approved):
I cannot really review the actual content as I don't know Prometheus at all.
Anyway, IIUC this is for our own use and it cannot affect the end user experience with core features of OSC. I see no harm in merging it.

config/metrics/metrics-prometheus-rules.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: PrometheusRule
+metadata:
+  name: osc-alerts
+  namespace: openshift-sandboxed-containers-operator
+spec:
+  groups:
+  - name: osc_alerts
+    rules:
+    - alert: KataRemoteWorkloadFailureHigh
+      expr: kata_remote_workload_failure_ratio > 0.25
```
littlejawa: I think kata_remote_workload_failure_ratio is a full percentage already.
So that would be > 25 I guess?

gkurz: Indeed :smile:

```
	kataRemoteWorkloadFailureRatio.Set(float64(failedPods) / float64(totalPods) * 100)
```

Good catch @littlejawa !

beraldoleal: good catch, thanks.

Fixed.

littlejawa (approved):
/lgtm
Thanks @beraldoleal

PR #474, opened 2024-11-06
backport: azure: use RHEL 9.4 as base image

snir911:
this is base image without disk encryption which suits for peer-pods, CoCo TDX and CoCo SEV-SNP
CONFIDENTIAL_COMPUTE_TYPE is removed as not used anymore

Signed-off-by: Snir Sheriber <ssheribe@redhat.com>
(cherry picked from commit 8492b92e73ef1bb0361b5e9e7363cce67f099967)

**- What I did**

**- How to verify it**


gkurz (approved):
Commit id did not change so we're good to go !

Thanks @snir911 !

PR #477, opened 2024-11-25
hack: fix handling of CLOUD_PROVIDER

gkurz:
Teach userVerification() to use the cloud provider requested by the user, which is not necessarily the same as the cluster infra. Output the cloud provider for peer pods and the cluster infra as two different messages for extra clarity.


snir911 (approved):
Thanks!

PR #485, opened 2024-12-03
dependency updates: move machine-config to latest

littlejawa:
- update machine-config to latest, to get dependency containers/image updated to 5.29.3
- update controller-gen to 0.16.5 for compatibility with the above
- update import statements and some function calls in .go files to take into account modifications in machine-config API


littlejawa:
This is also updating otelgrpc to 0.53
```
$ go mod graph | grep "otelgrpc"
k8s.io/apiextensions-apiserver@v0.31.1 go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc@v0.53.0
```

... and removes the dependency on goproxy
```
$ go mod graph | grep "goproxy"
$
```

Makefile
```diff
@@ -231,7 +231,7 @@ GOLANGCI_LINT = $(LOCALBIN)/golangci-lint-$(GOLANGCI_LINT_VERSION)
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.3.0
-CONTROLLER_TOOLS_VERSION ?= v0.14.0
+CONTROLLER_TOOLS_VERSION ?= v0.16.5
```
gkurz, line 234: The version of `controller-gen` is exposed in https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml#L6.
You need to run `make manifests` to regenerate them and a `make bundle` to propagate the change to the bundle.

littlejawa: Just updated

gkurz (approved):
/lgtm

Thanks @littlejawa

PR #487, opened 2024-12-05
KATA-3270: pod overhead for peer pods reduce actual worker capacity

balintTobik:
Peer pods' overhead reduce actual worker capacity, however most of the RAM/CPU allocated in cloud instance.

Fixes: [KATA-3270](https://issues.redhat.com/browse/KATA-3270)

**- What I did**

Operator create 'kata-remote' runtime class with a specific and lower overhead, like in upstream.

**- How to verify it**
1. create KataConfig for a peer pods
2. check 'kata-remote' runtime class created a lower overhead than 'kata' runtime class
3. 
Use separate pod's overhead related constants for kata and peer pods.

[EDIT from Greg: adapted description to final version]

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -78,15 +78,16 @@ const (
 	peerpodsCrioMachineConfigYaml       = "mc-50-crio-config.yaml"
 	peerpodsKataRemoteMachineConfig     = "40-worker-kata-remote-config"
 	peerpodsKataRemoteMachineConfigYaml = "mc-40-kata-remote-config.yaml"
-	kataRuntimeClassName                = "kata"
-	kataRuntimeClassCpuOverhead         = "0.25"
-	kataRuntimeClassMemOverhead         = "160Mi"
-	peerpodsRuntimeClassName            = "kata-remote"
-	peerpodsRuntimeClassCpuOverhead     = "0.25"
-	peerpodsRuntimeClassMemOverhead     = "120Mi"
 	// Use same Pod Overhead as upstream kata-deploy using, see
 	// https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/runtimeclasses/kata-qemu.yaml#L7
+	kataRuntimeClassName        = "kata"
+	kataRuntimeClassCpuOverhead = "0.25"
+	// We need a higher value than upstream (see https://github.com/openshift/sandboxed-containers-operator/pull/84)
+	kataRuntimeClassMemOverhead = "350Mi"
 	// https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/runtimeclasses/kata-remote.yaml#L7
+	peerpodsRuntimeClassName        = "kata-remote"
+	peerpodsRuntimeClassCpuOverhead = "0.25"
+	peerpodsRuntimeClassMemOverhead = "120Mi"
 )
 
 // +kubebuilder:rbac:groups=kataconfiguration.openshift.io,resources=kataconfigs;kataconfigs/finalizers,verbs=get;list;watch;create;update;patch;delete
@@ -702,7 +703,6 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
 				Name: runtimeClassName,
 			},
 			Handler: runtimeClassName,
-
 			Overhead: &nodeapi.Overhead{
 				PodFixed: corev1.ResourceList{
 					corev1.ResourceCPU:    resource.MustParse(cpuOverhead),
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -696,14 +694,16 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
 				Name: runtimeClassName,
 			},
 			Handler: runtimeClassName,
-			// Use same values for Pod Overhead as upstream kata-deploy using, see
-			// https://github.com/kata-containers/packaging/blob/f17450317563b6e4d6b1a71f0559360b37783e19/kata-deploy/k8s-1.18/kata-runtimeClasses.yaml#L7
-			Overhead: &nodeapi.Overhead{
+		}
+		// Use same values for Pod Overhead as upstream kata-deploy using, see
+		// https://github.com/kata-containers/packaging/blob/f17450317563b6e4d6b1a71f0559360b37783e19/kata-deploy/k8s-1.18/kata-runtimeClasses.yaml#L7
```
gkurz: This link points to the old kata-containers upstream repo which is now archived.

Equivalent link in the current repo is :

https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/runtimeclasses/kata-qemu.yaml#L7

Note that this overhead refers to the `kata-qemu` runtime class from upstream (named `kata` runtime class downstream), which is used for running the VMM on the cluster nodes. It accounts for running the kata shim + QEMU + virtiofsd.

In the case of peer pods, the overhead is just the kata shim since the VMM is running in the cloud. Upstream has a different overhead setting for the `kata-remote` runtime class :

https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/runtimeclasses/kata-remote.yaml#L7

These numbers might be questionable but regardless of that, we want to track upstream :
- we must keep keep copies of the upstream values for both kata and kata-remote cases in constants at the top of this file
- the comment above should be put along these constants for reference

controllers/openshift_controller.go
```diff
@@ -696,14 +694,16 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
...
+		if len(cpuOverhead) > 0 || len(memoryOverhead) > 0 {
```
gkurz: There will always be an overhead in practice. We will just adjust the numbers. This check is thus not needed.

controllers/openshift_controller.go
```diff
@@ -2279,8 +2279,8 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPodsMiscConfigs() error {
 		return err
 	}
 
-	// Create runtimeClass config for peer-pods
-	err = r.createRuntimeClass(peerpodsRuntimeClassName, peerpodsRuntimeClassCpuOverhead, peerpodsRuntimeClassMemOverhead)
+	// Create runtimeClass config for peer-pods without overhead
+	err = r.createRuntimeClass(peerpodsRuntimeClassName, "", "")
```
gkurz: As explained in the previous comments, we always have an overhead. We just want to use meaningful values for each case.

controllers/openshift_controller.go
```diff
@@ -79,8 +79,6 @@ const (
 	peerpodsKataRemoteMachineConfig     = "40-worker-kata-remote-config"
 	peerpodsKataRemoteMachineConfigYaml = "mc-40-kata-remote-config.yaml"
 	peerpodsRuntimeClassName            = "kata-remote"
-	peerpodsRuntimeClassCpuOverhead     = "0.25"
-	peerpodsRuntimeClassMemOverhead     = "350Mi"
```
gkurz, line 83: We will actually need to keep these constants but update the values. 

For consistency and improve clarity, we should create similar constants for the regular kata runtime class and use them at https://github.com/openshift/sandboxed-containers-operator/blob/devel/controllers/openshift_controller.go#L1172

See next comment for more details.

gkurz (commented):
Hi @balintTobik !

Thanks for working on that. I've commented and hopefully clarified what we want to achieved for now, i.e. use different overhead values for `kata` and `kata-remote`.

Also all commits should have a message and a valid Signed-off-by: tag.

Please address the comments in your local repo and force push to this PR.

controllers/openshift_controller.go
balintTobik: Hi @gkurz!
Thanks for clarifying me!

balintTobik:
Hi @gkurz!
Thank you for the review. I've updated the PR according to your comments.

gkurz (approved):
Hi @balintTobik ! Thanks for this work.

I've pushed another commit on top of your PR that does some trivial polishing and suggests a description for your commit. I will squash at merge time.

Now we'll ask for some feedback from @vvoronko.

littlejawa (approved):
/lgtm
Thanks!

gkurz:
I should have squashed the patches at merge time but I forgot... never mind.

PR #490, opened 2024-12-11
Build and deploy the metrics server

gkurz:
The metrics server is deployed alongside the OSC controller to collect SLI/SLO oriented metrics. It is currently assumed to be shipped in the OSC monitor image. This is suboptimal since the OSC repo doesn't build the OSC monitor image.

Build it and deploy it in the same container image as the controller. This is clearer and ensures a custom operator build also has a functional metrics server. This will also avoid the e2e check in the upstream CI to fail because the OSC monitor image isn't published yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@ COPY PROJECT PROJECT
 COPY go.mod go.mod
 COPY go.sum go.sum
 COPY main.go main.go
+COPY cmd/metrics cmd/metrics/
 COPY api api/
 COPY config config/
 COPY controllers controllers/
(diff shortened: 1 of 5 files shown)


gkurz:
CI is full green. This means that the e2e test was able to :
- build the OSC operator and OSC metrics server
- put them in the OSC operator image
- push the OSC operator image to the pipeline registry
- fully deploy the OSC operator and OSC metrics server from the pipeline registry (thanks to https://github.com/openshift/release/pull/59590)

Moved to draft so that @beraldoleal can jump in.

littlejawa (approved):
/lgtm

beraldoleal (approved):
Hi @gkurz, thanks for this. Is a little bit odd to me to have an "osc-monitor" image that monitor things (mostly kata), and have the metrics-server (that also monitor things) in a different image. Maybe we can rename the osc-monitor image in the future. 

But I'm ok with moving the metrics server here. We just need to be careful on how  1.9 release will fetch those changes, since CPaaS will not be used.

PR #491, opened 2024-12-13
Switch base image of OSC to rhel9

gkurz:
For the sake of being consistent with the builder which is rhel9 already.


gkurz:
This fixes the famous error of trying to run a rhel9-built binary on rhel8 :

```
/manager: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /manager)
```

littlejawa (approved):
/lgtm
Thanks!

PR #492, opened 2024-12-16
Automatic extension name

gkurz:
This teaches OSC to guess the right MCO extension to use for kata. Supersedes #459 .


bpradipt (approved):
/lgtm
nice.. This will save us the trouble of remembering to change the extension name for dev builds :-) ..

PR #494, opened 2025-01-06
baremetal-coco: Add the namespace to the default kbs address

fidencio:
Otherwise the address won't be discovered from another namespaces.


bpradipt (approved):
/lgtm

gkurz (approved):
/lgtm

Thanks @fidencio !

PR #495, opened 2025-01-08
AWS Snapshot isn't being deleted on kataconfig deletion

gsid100:
During KataConfig (with peer pods enabled) deletion, we delete the AMI
associated with the podvm image. This AMI is created by utilizing a snapshot and we make sure to delete the AMI on KataConfig deletion. This commit is for ensuring the snapshot is being deleted as well.

Fixes: https://issues.redhat.com/browse/KATA-3478

AWS Snapshots aren't being deleted during KataConfig deletion

**- What I did**
Updated the podvm-image-handler script to include snapshots deletion

**- How to verify it**
 - View available snapshot on current aws configuration
 - Install KataConfig
 - View newly added snapshots, associated with the KataConfig creation
 - Delete KataConfig
 - Verify the newly added snapshots are not longer present
 
Deleting AWS snapshots during KataConfig deletion

### DIFF
--- a/config/peerpods/podvm/aws-podvm-image-handler.sh
+++ b/config/peerpods/podvm/aws-podvm-image-handler.sh
@@ -343,9 +343,26 @@ function delete_ami_using_id() {
     # AMI_ID shouldn't be empty
     [[ -z "${AMI_ID}" ]] && error_exit "AMI_ID is empty"
 
+    # Retrieve the associated snapshots
+    SNAPSHOT_IDS_JSON_PATH='Images[].BlockDeviceMappings[].Ebs.SnapshotId'
+    SNAPSHOT_IDS=$(
+        aws ec2 describe-images \
+        --region "${AWS_REGION}" \
+        --image-ids "${AMI_ID}" \
+        --query "${SNAPSHOT_IDS_JSON_PATH}" \
+        --output text
+    ) || error_exit "Failed to get the ami's snapshot ids"
+
     # Delete the ami
     aws ec2 deregister-image --region "${AWS_REGION}" --image-id "${AMI_ID}" ||
         error_exit "Failed to delete the ami"
+
+    # Delete the associated snapshots
+    SNAPSHOT_IDS_LIST=($SNAPSHOT_IDS)
+    for SNAPSHOT_ID in "${SNAPSHOT_IDS_LIST[@]}"; do
+        aws ec2 delete-snapshot --region "${AWS_REGION}" --snapshot-id "${SNAPSHOT_ID}" ||
+            error_exit "Failed to delete snapshot '${SNAPSHOT_ID}'"
+    done
 
     # Remove the ami id annotation from peer-pods-cm configmap
     delete_ami_id_annotation_from_peer_pods_cm


gkurz:
Hi @gsid100 !

Thanks for working on this. Please make sure any commit in this PR has a one-line title separate from the description.

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -343,10 +343,26 @@ function delete_ami_using_id() {
...
+    )
+
     # Delete the ami
     aws ec2 deregister-image --region "${AWS_REGION}" --image-id "${AMI_ID}" ||
         error_exit "Failed to delete the ami"
 
+    # Delete the associated snapshots
+    for SNAPSHOT_ID in "${SNAPSHOT_ID_LIST}"; do
+        aws ec2 delete-snapshot --region "${AWS_REGION}" --snapshot-id "${SNAPSHOT_ID}" ||
+            error_exit "Failed to delete snapshot '${SNAPSHOT_ID}'"
```
snir911, line 364: This will not expand `${SNAPSHOT_ID}` because of the single quotes

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -343,10 +343,26 @@ function delete_ami_using_id() {
...
+        --query "${SNAPSHOT_ID_JSON_PATH}" \
+        --output text
+    )
+
     # Delete the ami
     aws ec2 deregister-image --region "${AWS_REGION}" --image-id "${AMI_ID}" ||
         error_exit "Failed to delete the ami"
 
+    # Delete the associated snapshots
+    for SNAPSHOT_ID in "${SNAPSHOT_ID_LIST}"; do
```
snir911: I suppose this usually returna a single snapshot id, were you able to verify the `parsing` and the `for loop` actually works when more than single image/blockDeviceMapping are returned?
Alternatively you can consider assuming there's always one it, take the first value and avoid iterating it

snir911 (commented):
Thanks LGTM overall, check my comments

config/peerpods/podvm/aws-podvm-image-handler.sh
gkurz: > This will not expand `${SNAPSHOT_ID}` because of the single quotes

The whole message sits in a double-quoted string, the single quotes will be preserved and not consumed by the shell in this case. Output will be something like :
```
Failed to delete snapshot 'foo'
```

snir911: Right! thanks, resolving this conversation

config/peerpods/podvm/aws-podvm-image-handler.sh
gkurz: > I suppose this usually returna a single snapshot id, were you able to verify the `parsing` and the `for loop` actually works when more than single image/blockDeviceMapping are returned?

Good catch @snir911 !

AFAICS `aws ec2 describe-images --output text`  generates a list of snapshot ids separated by spaces. The double-quotes prevent the parsing and cause SNAPSHOT_ID to get the full output of `aws ec2 describe-images` at the first iteration. This is shown below where a `break` is added to only iterate once :

```
$ for SNAPSHOT_ID in "$(aws ec2 describe-images --query Images[].BlockDeviceMappings[].Ebs.SnapshotId --output text)"; do echo $SNAPSHOT_ID ; break; done
snap-0789d845c06b2a400 snap-015586a114764b61a snap-0f9e16ccc5239314a ...
```

Same without the double-quotes :

```
$ for SNAPSHOT_ID in $(aws ec2 describe-images --query Images[].BlockDeviceMappings[].Ebs.SnapshotId --output text); do echo $SNAPSHOT_ID ; break; done
snap-0436adb617de71e4f
```

> Alternatively you can consider assuming there's always one it, take the first value and avoid iterating it

@snir911 Is it a strong assumption that you should get only one ? What if we get more ? Should this be considered as an unexpected error and logged ? If not, then I think the code should be able to handle multiple snapshot ids.

vvoronko: every AMI is created from a single snapshot, thus providing ami id as an identifier, describe command will return one value, except that - LGTM

gkurz (approved):
/lgtm

Thanks @gsid100 ! Let's wait for the green CI now.

gsid100:
Thanks :)

Eventually I decided to not rely on the assumption that an our podvm related AMI has one snapshot, as technically AMIs can be created with multiple snapshots - like when it's created from an instance with multiple volumes, such as one for the root and another for other data

I also listened to @snir911 advice and added an error exit in case retrieving the snapshots ids fails

snir911 (approved):
LGTM, thanks!

gkurz:
GH had a hiccup at merge time and incorrectly marked the PR as closed even though is was effectively merged, see https://github.com/openshift/sandboxed-containers-operator/commit/1467ead8cac0d8d20f0ebe9df6c628198332efb7

PR #502, opened 2025-01-22
baremetal-coco: Remove leftover instruction from the script

fidencio:
This was mistakenly added in order to try to enforce that the vsock_loopback module was loaded, and then ended up being removed as part of the series enabling DCAP for TDX.  However, this part slipped the developers and reviews (sorry!).


bpradipt (approved):
/lgtm

PR #508, opened 2025-01-27
docker build: make the build as user root

littlejawa:
When building with docker, the Dockerfile works fine. But it fails with podman, with a permission denied error. This is done on purpose, and requires either to build in a folder where the default user has permissions to write (like the $HOME folder), or to explicitely make the build as root.

We are using the later option here.

Also removing obsolete comments at the beginning of the Dockerfile.


gkurz (approved):
/lgtm

Thanks @littlejawa !

bpradipt (approved):
/lgtm

PR #512, opened 2025-01-28
Strenghten ODK/OCP detection heuristic

gkurz:
The desired image is enough to differentiate OKD and OCP when the cluster is deployed using default public repositories. It doesn't fly when pulling from custom repositories though, e.g. in some CI environments. For such cases, an option is to look for `ostree/rhcos` in the kernel command line, which is always exposed to containers.

Existing heuristic based on desired image is kept since it works for most deployments and it is recommended by the OKD team. An extra check targetting CentOS Stream CoreOS (SCOS) is added for completeness.

The environment based hidden knob is also kept in case it is needed again in the future.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -491,7 +492,7 @@ func (r *KataConfigOpenShiftReconciler) newMCPforCR() *mcfgv1.MachineConfigPool 
 
 func (r *KataConfigOpenShiftReconciler) getExtensionName() string {
 	// RHCOS uses "sandboxed-containers" as thats resolved/translated in the machine-config-operator to "kata-containers"
-	// FCOS however does not get any translation in the machine-config-operator so we need to
+	// FCOS/SCOS however does not get any translation in the machine-config-operator so we need to
 	// send in "kata-containers".
 	// Both are later send to rpm-ostree for installation.
 	//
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -507,7 +508,16 @@ func (r *KataConfigOpenShiftReconciler) getExtensionName() string {
...
 
-	// As RHCOS is rather special variant, use "kata-containers" by default, which also applies to FCOS
+	if strings.HasPrefix(clusterVersion.Status.Desired.Image, "quay.io/okd/scos-release") {
+		return "kata-containers" // SCOS
+	}
+
+	cmdline, err := ioutil.ReadFile("/proc/cmdline")
+	if err == nil && strings.Contains(string(cmdline), "ostree/rhcos") {
+		return "sandboxed-containers" // RHCOS
+	}
```
pmores, line 518: While I appreciate the reasoning I'd actually consider dropping the `clusterVersion.Status.Desired.Image` checks.  If they don't work then the OKD team endorsement seems of little value.

We can keep them if the kernel cmdline based check is not reliable either and the other checks cover the cases where it fails.

gkurz: > While I appreciate the reasoning I'd actually consider dropping the `clusterVersion.Status.Desired.Image` checks. If they don't work then the OKD team endorsement seems of little value.
> 
> We can keep them if the kernel cmdline based check is not reliable either and the other checks cover the cases where it fails.

My concern is I have absolutely no idea if any of these checks are  reliable in the long term. The only viable way I see to get out from this mess is to change the extension name to `kata-containers` in the MCO. Until then I'd tend to keep all the heuristics to increase the odds of doing the right guess.

pmores: Fair enough but let's make a solemn vow here that we'll change the RHCOS extension name ASAP, ideally pretty much immediately, shall we?  Otherwise I'd have to pity the poor newcomer soul trying to make sense of this cascade of checks. ;-)

pmores (approved):
lgtm thanks @gkurz !

controllers/openshift_controller.go
gkurz: > Fair enough but let's make a solemn vow here that we'll change the RHCOS extension name ASAP, ideally pretty much immediately, shall we? Otherwise I'd have to pity the poor newcomer soul trying to make sense of this cascade of checks. ;-)

I do solemnly swear that I will faithfully create a PR in the MCO repo, and will to the best of my ability, push, stand for and merge a new extension name for kata :stuck_out_tongue:

PR #513, opened 2025-01-28
must-gather: updates for the Dockerfile

littlejawa:
The Dockerfile for the must-gather image is very outdated. It actually fails to build because it's still based on centos:7, which repositories are not available anymore.

Since it doesn't seem to be used, I am suggesting these changes, which gets closer to what we actually need to release in Openshift.

Changes include:
- use ubi9 minimal for the base image
- do not install additional tools - they're not referenced by our gathering scripts, and they've never been installed in our currently released images.

NOTE: this Dockerfile will then be used by our Konflux instance to build the must-gather image.

### DIFF
--- a/must-gather/Dockerfile
+++ b/must-gather/Dockerfile
@@ -1,9 +1,9 @@
 FROM quay.io/openshift/origin-must-gather:latest as builder
 
-FROM centos:7
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.5
 
- # For gathering data from nodes
-RUN yum update -y && yum install iproute tcpdump pciutils util-linux nftables rsync -y && yum clean all
+# For gathering data from nodes
+RUN microdnf update -y && microdnf install tar rsync -y && microdnf clean all
 
 COPY --from=builder /usr/bin/oc /usr/bin/oc
 


must-gather/Dockerfile
```diff
@@ -1,9 +1,6 @@
 FROM quay.io/openshift/origin-must-gather:latest as builder
 
-FROM centos:7
-
- # For gathering data from nodes
-RUN yum update -y && yum install iproute tcpdump pciutils util-linux nftables rsync -y && yum clean all
```
snir911, line 6: could it be these are used by the `/usr/bin/gather` binary/script somehow? I'm in favor of the update but I'd sill verify it works first

littlejawa: I've verified in our gathering scripts, and in the `/usr/bin/gather` script, and none of them is referencing those tools.
Also note that our downstream build in CPaaS is not installing any of these, and that's what we've tested (and shipped) so far.
But yes, I will try to build and test that just in case :)

gkurz: There's also the original must-gather tool that was used to create ours :

https://github.com/openshift/must-gather

It seems to be relatively _well_ maintained compared to what we do :wink:

and the openshift variant that seems to be heavily maintained at :

https://github.com/openshift/must-gather

I think we should look at those before fixing our stuff.

> could it be these are used by the `/usr/bin/gather` binary/script somehow? I'm in favor of the update but I'd sill verify it works first

Yeah and I suspect it won't be functional because https://github.com/openshift/sandboxed-containers-operator/blob/devel/must-gather/collection-scripts/gather_sandboxed_containers_logs#L99 is assuming the `jq` command is available but I cannot find anything that would install it.

An option would be to use `registry.redhat.io/openshift4/ose-cli-rhel9` as base and add a `dnf install jq` line (tested), a bit like at https://github.com/openshift/must-gather/blob/master/Dockerfile.ocp .

littlejawa: Your two first links point to openshift/must-gather. I think the first one is https://github.com/kubevirt/must-gather :)

gkurz: > Your two first links point to openshift/must-gather. I think the first one is https://github.com/kubevirt/must-gather :)

oops bad copy/paste :wink:

littlejawa: Doing some tests, I'm getting errors for missing things... but not in the list of what I removed here.
I have errors for tar and xargs ( ! )
What I don't get is that I'm building from the same base image as what we do in CPaaS, so now I will try with our official release image, and compare notes...
I will update this PR as needed, but marking this as WIP for now. 
Thanks for raising this!

littlejawa: Updated again: I'm only adding the items that caused errors during my test.
Now the generated image can run without error.

spotlesstofu (approved):

PR #514, opened 2025-01-28
docker build: Use digit tag for Dockerfile UBI image

gsid100:
In order to improve the images update handling and utilize Konflux bot's automated image tags update functionality (e.g., https://github.com/openshift/sandboxed-containers-operator/pull/507), which depends on using digit tags for images, it was decided to replace the latest tag with a digit-based tag for the UBI image in the Dockerfile.

Changing the ubi9/ubi-minimal image to from 'latest' to '9.5-xyz'

Using digit tag for Dockerfile UBI image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN go mod vendor
 RUN make build
 
 # Use OpenShift base image
-FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.5-1736404155
 WORKDIR /
 COPY --from=builder /workspace/bin/manager .
 COPY --from=builder /workspace/bin/metrics-server .


gkurz (approved):
/lgtm

Thanks @gsid100 !

littlejawa:
@gsid100 - the base image doesn't seem to exist : the konflux build fails because of that, and I can't pull it either with `podman pull`.
From https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5?image=677f76976f42ae1c4ca9353e, it seems the latest ubi-minimal image is: registry.access.redhat.com/ubi9/ubi-minimal:9.5-1736404155

I think that's what we should be using

gkurz:
> @gsid100 - the base image doesn't seem to exist : the konflux build fails because of that, and I can't pull it either with `podman pull`. From https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5?image=677f76976f42ae1c4ca9353e, it seems the latest ubi-minimal image is: registry.access.redhat.com/ubi9/ubi-minimal:9.5-1736404155
> 
> I think that's what we should be using

Good catch @littlejawa ! Konflux is really helping here :smile: 

 Can you please fix that @gsid100 ?

littlejawa (approved):
/lgtm
Thanks  @gsid100 !

gkurz (approved):
/lgtm

Thanks @gsid100 !

Let's wait for the all green CI now.

PR #515, opened 2025-01-28
Snapshot deletion validation during KataConfig removal

gsid100:
During KataConfig (with peer pods enabled) deletion, we delete the AMI and the snapshot associated with the podvm image. In this PR I Included a validation for ensuring the Snapshot associated with the AMI was deleted properly

Lack of validation to ensure the snapshot was deleted

**- What I did**
Added validations to the 'aws-podvm-image-handler.sh' script

**- How to verify it**
 - Create KataConfig
 - Save the Snapshot ID associated with the AMI (by running aws cli commands)
 - Delete KataConfig
 - Examine the logs of the podvm delete pod and verify it shows "Snapshot '${SNAPSHOT_ID}' successfully deleted"
 
Validating aws ami snapshot is deleted properly during KataConfig removal

### DIFF
--- a/config/peerpods/podvm/aws-podvm-image-handler.sh
+++ b/config/peerpods/podvm/aws-podvm-image-handler.sh
@@ -331,7 +331,39 @@ function create_ami() {
 
 }
 
-# function to delete the ami
+# Function to delete the ami's snapshots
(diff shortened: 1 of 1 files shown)


snir911 (approved):
LGTM, thanks!

PR #517, opened 2025-01-30
Podvm builder dockerfile

littlejawa:
Creating a new Dockerfile dedicated to building the podvm-builder image in Konflux.
It seems we can't use the existing Dockerfile as it is, because of the subscription
management done in Konflux.
(see PR https://github.com/openshift/sandboxed-containers-operator/pull/516)

This Dockerfile is a clean copy of Dockerfile.podvm-builder, with the subscription
lines removed.
Also updating the source references to kata and cloud-api-adaptor to point to our forks under the Openshift org.

### DIFF
--- a/config/peerpods/podvm/Dockerfile.podvm-builder.openshift
+++ b/config/peerpods/podvm/Dockerfile.podvm-builder.openshift
@@ -1,0 +1,40 @@
+FROM registry.access.redhat.com/ubi9/ubi:9.5-1736404036
+
+# azure-podvm-image-handler.sh script under /scripts/azure-podvm-image-handler.sh
+# aws-podvm-image-handler.sh script under /scripts/aws-podvm-image-handler.sh
+# sources for cloud-api-adaptor under /src/cloud-api-adaptor
+# The podvm binaries are expected to be under /payload/podvm-binaries.tar.gz
+# Binaries like kubectl, packer and yq under /usr/local/bin will be installed by the scripts
+
+
+LABEL kata_src=https://github.com/kata-containers/kata-containers
+LABEL kata_src_commit=stable-3.7
+
+ARG ORG_ID
+ARG ACTIVATION_KEY
+
+RUN mkdir -p /scripts
+
+ADD lib.sh libvirt-podvm-image-handler.sh aws-podvm-image-handler.sh azure-podvm-image-handler.sh /scripts/
+
+RUN /scripts/azure-podvm-image-handler.sh -- install_rpms
+
+ARG CAA_SRC=https://github.com/confidential-containers/cloud-api-adaptor
+ARG CAA_REF=main
+ARG CERT_RPM
+
+
+ENV CAA_SRC=$CAA_SRC
+ENV CAA_REF=$CAA_REF
+ENV CERT_RPM=$CERT_RPM
+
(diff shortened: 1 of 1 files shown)


littlejawa:
@spotlesstofu - I'd like your comments on this one, and/or on the  #516 one.
The solution I'm going with (Adding a new Dockerfile) is not ideal, as it duplicates code, but I'm not sure how to deal with the subscription properly. Have you faced something like that with Trustee?

At the moment, my vote is to go with the duplicated Dockerfile, and if/when we find a better way, fix it to keep a single one. But any suggestion is welcome :-)

snir911:
Thanks @littlejawa , have you tried to comment this line out? if it's still identified by kuflux also as a comment I'd simply remove the line completely and add a comment mentioning "for unsubscribe systems subscribe using subscription manager as the first build step"  or something like that

littlejawa:
Yes, commenting it is enough.
I didn't dare removing it, but if it's an option, that would be nice. :-)
Should I reopen the other PR ?

snir911:
I'd leave it commented out, it's useful when you build on unsubscribed system

littlejawa:
I have re-opened #516 and commented the lines in the Dockerfile. It builds now.
If I get approvals and merge 516, we can close this one

config/peerpods/podvm/Dockerfile.podvm-builder.openshift
```diff
@@ -0,0 +1,40 @@
...
+
+LABEL kata_src=https://github.com/openshift/kata-containers
+LABEL kata_src_commit=osc-release
+
+ARG ORG_ID
+ARG ACTIVATION_KEY
+
+RUN mkdir -p /scripts
+
+ADD lib.sh libvirt-podvm-image-handler.sh aws-podvm-image-handler.sh azure-podvm-image-handler.sh /scripts/
```
beraldoleal, line 18: Probably this will conflict with #519 but I can handle on my side. Just adding a note so we (I) don't  forget :)

config/peerpods/podvm/Dockerfile.podvm-builder.openshift
```diff
@@ -0,0 +1,40 @@
...
+# The podvm binaries are expected to be under /payload/podvm-binaries.tar.gz
+# Binaries like kubectl, packer and yq under /usr/local/bin will be installed by the scripts
+
+
+LABEL kata_src=https://github.com/openshift/kata-containers
+LABEL kata_src_commit=osc-release
+
+ARG ORG_ID
+ARG ACTIVATION_KEY
+
```
beraldoleal, line 15: About the  subscription code that was here... I'm not familiar on how konflux handles this (maybe @spotlesstofu). 

But, I was wondering if something like this would work:

```
RUN if [[ -z "${KONFLUX_BUILD}" ]]; then \
    rm -f /etc/rhsm-host && rm -f /etc/pki/entitlement-host; \
    subscription-manager register --org=${ORG_ID} --activationkey=${ACTIVATION_KEY}; \
    fi
```

Not sure if the konflux already exposes any variable, but if not, maybe we can configure those variables within the tekton folder. This way we could avoid having two dockerfiles. Wdyt?

littlejawa:
Dropping in favor of #516

PR #523, opened 2025-02-03
Add file-based catalog for 4.17

spotlesstofu:
Add configuration to build the file-based catalog (FBC) on Konflux for OCP 4.17.

Includes changes by the Konflux bot, from #522.


spotlesstofu:
@littlejawa, @tbuskey, this pull is ready for review. Can you have a look?

I included the bare minimum to define the FBC and build it on Konflux, just for OCP 4.17.
After this is merged, we can work on the other OCP versions.

littlejawa:
Hey @spotlesstofu,
Thanks for looking into this !

Just to understand: we need to have a different Dockerfile for each OCP release, even if we are in single-stream mode?
We can't share the Dockerfile, even if we use different apps on the konflux side?

I'm also thinking about local builds - at the moment, we can use the Makefile to build the images, the bundle and the catalog.
But this is sqlite-based still.
Can we use the files you're adding here to build the catalog when doing a local build, or are they dedicated to konflux builds?

spotlesstofu:
> Just to understand: we need to have a different Dockerfile for each OCP release, even if we are in single-stream mode? We can't share the Dockerfile, even if we use different apps on the konflux side?

Konflux uses the base image tag (`v4.17`) to tell the OCP version to release to, so we need a different Dockerfile for each one.

> I'm also thinking about local builds - at the moment, we can use the Makefile to build the images, the bundle and the catalog. But this is sqlite-based still. Can we use the files you're adding here to build the catalog when doing a local build, or are they dedicated to konflux builds?

Yes, you can build the Dockerfile locally.

littlejawa:
/lgtm
Thanks @spotlesstofu

PR #530, opened 2025-02-13
baremetal-coco: Use kata-cc as runtimeclass name

fidencio:
Instead of having kata-tdx or kata-snp ones, as at the end of the day
what the user wants is to run their workload on a confidential
manner, regardless of what's under the hood.


bpradipt (approved):
/lgtm

snir911 (approved):
LGTM thanks!

PR #532, opened 2025-02-13
Re-Create the cloud-api-daemonset

littlejawa:
It has been removed when we dropped the code for peerpod config management.
Also removing a left over reference to peerpod config in kustomize script.

Note: this was done by copying the code [from cloud-api-adaptor](https://github.com/confidential-containers/cloud-api-adaptor/blob/6c159d568a29b25c548cd48b4de35f0196f751b3/src/peerpodconfig-ctrl/controllers/peerpodconfig_controller.go#L133) that used to create it, and was removed.

I'm doing creation/deletion at the places where we used to create/delete the peerpod config, so that it doesn't change behavior.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -29,6 +29,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	ignTypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/go-logr/logr"
@@ -90,6 +91,8 @@ const (
 	peerpodsRuntimeClassName        = "kata-remote"
 	peerpodsRuntimeClassCpuOverhead = "0.25"
 	peerpodsRuntimeClassMemOverhead = "120Mi"
+	// cloud-api-adaptor (CAA) daemonset name
+	caaDsName = "osc-caa-ds"
 )
 
 // +kubebuilder:rbac:groups=kataconfiguration.openshift.io,resources=kataconfigs;kataconfigs/finalizers,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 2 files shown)


littlejawa:
I've tested it with a custom build, it seems to work, but I didn't know how to make my custom build use the 1.9.0 images for peerpods, so I made it use 1.8.1 images instead... so the daemonset is created, but I can't tell if everything else works as it should.

snir911 (commented):
LGTM overall, the following was added recently to the ds definition and it's missing here:
https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/install/yamls/caa-pod.yaml#L25-L29

controllers/openshift_controller.go
```diff
@@ -324,6 +327,151 @@ func (r *KataConfigOpenShiftReconciler) removeLogLevel() error {
...
+		nodeSelector                 = r.getNodeSelectorAsMap()
+	)
+
+	dsLabelSelectors := map[string]string{
+		"name": caaDsName,
+	}
+
+	imageString := os.Getenv("RELATED_IMAGE_CAA")
+	if imageString == "" {
+		r.Log.Info("RELATED_IMAGE_CAA env var is unset or empty, cloud-api-adaptor pods will not run")
```
wainersm, line 350: Maybe Log.Error() as this is a fatal error.

wainersm:
hi @littlejawa , I'm not an expert here, so I only compared with the code removed upstream and looked for obvious logical errors. LGTM.

controllers/openshift_controller.go
littlejawa: Log.error is used to report an "error" object - which we don't have here.
This is actually consistent with the way we report the same issue when we create the Monitor daemonset (in the function below this one).

bpradipt:
I have been able to test this successfully on my ARO cluster.

bpradipt (approved):
/lgtm

controllers/openshift_controller.go
```diff
@@ -90,6 +91,8 @@ const (
 	peerpodsRuntimeClassName        = "kata-remote"
 	peerpodsRuntimeClassCpuOverhead = "0.25"
 	peerpodsRuntimeClassMemOverhead = "120Mi"
+	// cloud-api-adaptor (CAA) daemonset name
+	caaDsName = "osc-caa-daemon"
```
bpradipt: nit: It might be a good idea to name it as daemonset or ds to be clear when looking at the pod o/p in CLI
`s/osc-caa-daemon/osc-caa-ds`
This is not a blocker

PR #498, opened 2025-01-17
Remove peerpods config management.

littlejawa:
Newer version of cloud-api-adaptor removes the need to use peerpod config. This commit removes the associated code.

PR #533, opened 2025-02-13
Migrate PeerPodConfig Limit

beraldoleal:
Recent commit at cloud-api-adaptor removed PeerPodConfig logic. Also this operator does not use its CRD anymore. This will migrate the old config during an upgrade process to the new CM.

Related: #498
Fixes: rhjira#[KATA-3434](https://issues.redhat.com//browse/KATA-3434)

**- What I did**

Added the migration logic in the reconcile loop to ensure PEERPODS_LIMIT_PER_NODE is set in peer-pods-cm and clean up any existing PeerPodConfig CR.

**- How to verify it**

1. Deploy the updated operator.
2. Check PEERPODS_LIMIT_PER_NODE in peer-pods-cm:
```
kubectl get configmap peer-pods-cm -n openshift-sandboxed-containers-operator -o yaml | grep PEERPODS_LIMIT_PER_NODE
```
Ensure PeerPodConfig is deleted:
```
kubectl get peerpodconfigs -n openshift-sandboxed-containers-operator
```

(I'm not 100% sure if we should remove the cm tho)

Migrated PeerPodConfig.Limit to peer-pods-cm (PEERPODS_LIMIT_PER_NODE)

### DIFF
--- a/controllers/migrate.go
+++ b/controllers/migrate.go
@@ -1,0 +1,78 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controllers
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// migratePeerPodsLimit moves the PeerPodConfig "Limit" value to peer-pods-cm
+func (r *KataConfigOpenShiftReconciler) migratePeerPodsLimit() error {
+	peerPodConfig := &unstructured.Unstructured{}
+	peerPodConfig.SetAPIVersion("confidentialcontainers.org/v1alpha1")
+	peerPodConfig.SetKind("PeerPodConfig")
+
+	err := r.Client.Get(context.TODO(), client.ObjectKey{
+		Name:      peerpodConfigCrdName,
+		Namespace: OperatorNamespace,
+	}, peerPodConfig)
+
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.Log.Info("No PeerPodConfig found, skipping migration")
+			return nil
+		}
+		return err
+	}
+
+	limitValue, found, _ := unstructured.NestedString(peerPodConfig.Object, "spec", "limit")
+	if !found {
+		r.Log.Info("spec.limit not found, skipping migration, in favor of default value.")
+		r.Log.Info("Removing deprecated PeerPodConfig...")
+		return r.Client.Delete(context.TODO(), peerPodConfig)
+	}
+
+	configMap := &corev1.ConfigMap{}
+	err = r.Client.Get(context.TODO(), client.ObjectKey{
+		Name:      "peer-pods-cm",
+		Namespace: OperatorNamespace,
+	}, configMap)
+
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.Log.Info("No peer-pods-cm found, skipping migration")
+			r.Log.Info("Removing deprecated PeerPodConfig...")
+			return r.Client.Delete(context.TODO(), peerPodConfig)
+		}
+		return err
+	}
+
+	configMap.Data["PEERPODS_LIMIT_PER_NODE"] = limitValue
+	err = r.Client.Update(context.TODO(), configMap)
+	if err != nil {
+		return err
+	}
+
+	err = r.Client.Delete(context.TODO(), peerPodConfig)
+	if err != nil {
+		return err
+	}
+
+	r.Log.Info("Successfully migrated PeerPodConfig Limit to peer-pods-cm and deleted PeerPodConfig")
+	return nil
+}
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -131,6 +131,12 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
+	err = r.migratePeerPodsLimit()
+	if err != nil {
+		r.Log.Info("Failed to migrate PeerPodConfig limit", "err", err)
+		return ctrl.Result{}, err
+	}
+
 	err = r.processFeatureGates()
 	if err != nil {
 		r.Log.Info("Unable to process feature gates", "err", err)


beraldoleal:
@bpradipt @gkurz @snir911  not sure if this works for now or if you have a better approach in mind?

controllers/migrate.go
```diff
@@ -0,0 +1,74 @@
...
+		if k8serrors.IsNotFound(err) {
+			r.Log.Info("No PeerPodConfig found, skipping migration")
+			return nil
+		}
+		return err
+	}
+
+	limitValue, found, _ := unstructured.NestedString(peerPodConfig.Object, "spec", "limit")
+	// TODO: Not sure if we need to set the default value here, or just skip the
+	// upgrade.
```
littlejawa: If the peerpod config exists, we need to delete it, so I'd say we continue the migration using the default value

controllers/migrate.go
```diff
@@ -0,0 +1,74 @@
...
+		limitValue = "10"
+	}
+
+	configMap := &corev1.ConfigMap{}
+	err = r.Client.Get(context.TODO(), client.ObjectKey{
+		Name:      "peer-pods-cm",
+		Namespace: OperatorNamespace,
+	}, configMap)
+
+	if err != nil {
```
littlejawa, line 56: I feel like if the configmap doesn't exist, we should just skip migration like we do if peerpod config doesn't exist.
How I see it is that if the CM is not there, it means the install we're trying to upgrade is not complete, and we have nothing to do for migration. It's not really an error then.

controllers/migrate.go
```diff
@@ -0,0 +1,74 @@
...
+		return err
+	}
+
+	configMap.Data["PEERPODS_LIMIT_PER_NODE"] = limitValue
+	err = r.Client.Update(context.TODO(), configMap)
+	if err != nil {
+		return err
+	}
+
+	// TODO: Are we going to delete the object, or just flag it as migrated?
```
littlejawa: I think we should delete it. This object was created automatically by the previous version of our operator, so my expectation is that we should dispose of it if we don't need it anymore.

beraldoleal: My concern here is that I'm not sure if we are migrating all data from that object and we might loose data.

controllers/migrate.go
```diff
@@ -0,0 +1,74 @@
...
+			return nil
+		}
+		return err
+	}
+
+	limitValue, found, _ := unstructured.NestedString(peerPodConfig.Object, "spec", "limit")
+	// TODO: Not sure if we need to set the default value here, or just skip the
+	// upgrade.
+	if !found {
+		limitValue = "10"
```
bpradipt: default PEERPODS_LIMIT_PER_NODE is 10. So if `limit` key is not found we can we can skip setting PEERPODS_LIMIT_PER_NODE explicitly 
https://github.com/confidential-containers/cloud-api-adaptor/blob/b8d38b99732cbdd9b41970be4d60f308e1483c26/src/cloud-api-adaptor/cmd/cloud-api-adaptor/main.go#L127

controllers/migrate.go
bpradipt: Deleting should be fine. If you look at the spec other than the limit, rest of the params (secret, configmap etc) are already provided to the CAA pod - https://github.com/confidential-containers/cloud-api-adaptor/blob/v0.9.0/src/peerpodconfig-ctrl/api/v1alpha1/peerpodconfig_types.go#L24

controllers/migrate.go
littlejawa: Continued on https://github.com/openshift/sandboxed-containers-operator/pull/533/files#r1962921428

controllers/migrate.go
littlejawa: Closing my previous comment - I wasn't sure about the default on the CAA side.

beraldoleal:
@littlejawa @bpradipt thanks for the review, just force pushed with the changes.

bpradipt (approved):
/lgtm

controllers/migrate.go
```diff
@@ -0,0 +1,76 @@
...
+			r.Log.Info("No PeerPodConfig found, skipping migration")
+			return nil
+		}
+		return err
+	}
+
+	limitValue, found, _ := unstructured.NestedString(peerPodConfig.Object, "spec", "limit")
+	if !found {
+		r.Log.Info("spec.limit not found, skipping migration, in favor of default value.")
+		return nil
```
littlejawa: The peerpod config object still exists. We need to delete it, even if it doesn't contain the limit.

controllers/migrate.go
```diff
@@ -0,0 +1,76 @@
...
+	configMap := &corev1.ConfigMap{}
+	err = r.Client.Get(context.TODO(), client.ObjectKey{
+		Name:      "peer-pods-cm",
+		Namespace: OperatorNamespace,
+	}, configMap)
+
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			r.Log.Info("No peer-pods-cm found, skipping migration")
+			return nil
```
littlejawa: Same comment - the peerpod config object is still there and needs to be deleted.
Maybe we can delete it right after getting the limit from it, so that you can still get out early here.

beraldoleal: Somehow, it feels really bold to delete something before doing the actual migration! :)

Also, not sure if we should delete it here for safety reasons. If a user set a limit value there, and for some reason (not sure if it's possible) we don't find peer-pods-cm, we lose that information. I’d feel more cautious if we just skipped the migration in this case. wdyt?

beraldoleal:
@littlejawa I fixed one comment, but the second one I'm not sure if we should delete, vide my comment above.

controllers/migrate.go
littlejawa: I try to imagine what would happen if we leave it behind.
This is a reconcile loop, so every time it runs, it will find the peerpod config, fail to find the CM, and get out without deleting anything (logging weird messages along the way).
If at some point the user finally decide to create the  config map - with a limit that they set manually - then when our reconcile loop will run again, it will find the peerpod config, find the CM, and overwrite the newly created limit with a value that was set in a previous installation.
Is that really what we want?

I think that this migration step needs to get out cleanly, and run only once: if it finds a previous version, it should get out after deleting it.
Migrating the limit is a bonus - we try to do it, we log a message if we can't, but we delete the peerpodconfig anyway so that things are clean when we get out.

beraldoleal: I see your point, and it makes sense. But honestly, I’d rather have a service with repetitive logs pointing out an issue than a migration step that ends up losing user data and forcing users to re-enter it.

That said, I’m not an expert and don’t feel too strongly about it. Happy to go with whatever’s decided, just wanted to bring this up and hear what others think.

Tagging @bpradipt @pmores @gkurz for their thoughts.

pmores: How precious is the limit value?  Is it something that a user could reasonably determine by just going "let's say 10", or do they have to spend a week measuring, computing and contemplating stuff to arrive at it?

> I’d rather have a service with repetitive logs pointing out an issue than a migration step that ends up losing user data and forcing users to re-enter it.

Note that the logging isn't the main issue here, it's the risk of corrupting a newly entered value with an old pre-migration one as outlined by @littlejawa above.

I obviously agree that we need to be careful with data a user entered, too.  If there's no proper place to migrate the limit to we could consider storing it somewhere else, say an annotation on a KataConfig instance or some other place that's guaranteed to be there when this migration is in progress.  This would be just to preserve the value for the user and I'd only do it if the value is rather important or hard to get.

Otherwise I'd just document what the migration does, notifying users that if they want to migrate the limit then need to make sure the respective configmap exists ahead of time.

beraldoleal: Thanks @pmores for joining us...

> I obviously agree that we need to be careful with data a user entered, too. If there's no proper place to migrate the limit to we could consider storing it somewhere else, say an annotation on a KataConfig instance or some other place that's guaranteed to be there when this migration is in progress. This would be just to preserve the value for the user and I'd only do it if the value is rather important or hard to get.

I considered this as well, but it might be "complex" having the value in another place. Since is not hard to get this number, I would skip for now.

> Otherwise I'd just document what the migration does, notifying users that if they want to migrate the limit then need to make sure the respective configmap exists ahead of time.

Looks fine to me, something like: "You need to have the peer-pod-cm in place before migrating".

If there is no objections to this approach, I will update the PR.

littlejawa (approved):
/lgtm

PR #537, opened 2025-02-17
Upgrade operator-sdk to v1.39.1

### LINKED ISSUE #483: :warning: Action Required: Replace Deprecated gcr.io/kubebuilder/kube-rbac-proxy
## Description

:warning: **The image `gcr.io/kubebuilder/kube-rbac-proxy` is deprecated and will become unavailable.**
**You must move as soon as possible, sometime from early 2025, the GCR will go away.**

> _Unfortunately, we're unable to provide any guarantees regarding timelines or potential extensions at this time. Images provided under GRC will be unavailable from March 18, 2025, [as per announcement](https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr). However, `gcr.io/kubebuilder/`may be unavailable before this date due [to efforts to deprecate infrastructure](https://github.com/kubernetes/k8s.io/issues/2647)._

- **If your project uses `gcr.io/kubebuilder/kube-rbac-proxy`, it will be affected.**
  Your project may fail to work if the image cannot be pulled. **You must take action as soon as possible.**

- However, if your project is **no longer using this image**, **no action is required**, and you can close this issue.

## Using the image `gcr.io/kubebuilder/kube-rbac-proxy`?

[kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) was historically used to protect the metrics endpoint. However, its usage has been discontinued in [Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder). The default scaffold now leverages the [`WithAuthenticationAndAuthorization`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics/filters#WithAuthenticationAndAuthorization) feature provided by [Controller-Runtime](https://github.com/kubernetes-sigs/controller-runtime).

This feature provides integrated support for securing metrics endpoints by embedding authentication (`authn`) and authorization (`authz`) mechanisms directly into the controller manager's metrics server, replacing the need for (https://github.com/brancz/kube-rbac-proxy) to secure metrics endpoints.

### What To Do?

**You must replace the deprecated image `gcr.io/kubebuilder/kube-rbac-proxy` with an alternative approach. For example:**
- Update your project to use [`WithAuthenticationAndAuthorization`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/metrics/filters#WithAuthenticationAndAuthorization):
  > _You can fully upgrade your project to use the latest scaffolds provided by the tool or manually make the necessary changes. Refer to the [FAQ and Discussion](https://github.com/kubernetes-sigs/kubebuilder/discussions/3907) for detailed instructions on how to manually update your project and test the changes._
- Alternatively, replace the image with another trusted source at your own risk, as its usage has been discontinued in Kubebuilder.

**For further information, suggestions, and guidance:**
- 📖 [FAQ and Discussion](https://github.com/kubernetes-sigs/kubebuilder/discussions/3907)
- 💬 Join the Slack channel: [#kubebuilder](https://communityinviter.com/apps/kubernetes/community#kubebuilder).

> **NOTE:** _This issue was opened automatically as part of our efforts to identify projects that might be affected and to raise awareness about this change within the community. If your project is no longer using this image, **feel free to close this issue**._

We sincerely apologize for any inconvenience this may cause.

**Thank you for your cooperation and understanding!** :pray:

balintTobik:
Upgrade operator-sdk to v1.39.1 version.

**- What I did**
Update operator-sdk version in docs/DEVELOPMENT.md
Update versions in Makefile.
Remove codes related to deprecated kube-rbac-proxy.
Add new network-policy related changes.

**- How to verify it**

Upgrade operator-sdk to v1.39.1.
Closes: [KATA-3521](https://issues.redhat.com//browse/KATA-3521)


littlejawa:
As this PR is removing kube-rbac-proxy, it is also addressing #483

config/rbac/metrics_auth_role_binding.yaml
```diff
@@ -1,12 +1,12 @@
...
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: proxy-role
+  name: metrics-auth-role
 subjects:
-- kind: ServiceAccount
-  name: default
-  namespace: system
+  - kind: ServiceAccount
+    name: controller-manager
```
bpradipt: Who creates the `controller-manager` service account?
The manager.yaml doesn't specify `ServiceAccount` explicitly.

balintTobik: Hi @bpradipt! Thanks for the comment!
If I understand correctly this service account was introduced in v1.5.0 sdk version (https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.5.0/#gov3-add-a-systemcontroller-manager-serviceaccount-to-your-project)
Do you know the "default" was kept by purpose? Or I can change it to new "controller-manager" SA in other rolebinding as well?

gkurz (changes requested):
Hi @balintTobik ! Thanks for working on this.

A new operator SDK release can introduce incompatibilities that call for very specific changes. Every operator SDK release thus comes with a migration guide at https://sdk.operatorframework.io/docs/upgrading-sdk-version/ .

This PR is doing a 3 release jump. Please break this commit down into 3 per-release commits. Each commit should at least provide the link to the corresponding migration guide as in a9d0f2b7118410c5a4a344f9df0fad98dd77fa52 and any extra details that can help. This will greatly improve the reviewing experience and the overall maintainability.

pmores:
It would also be easier to review if it were clearer which changes are manual as opposed to generated by the upgrade.

balintTobik:
Thanks for the comments! 
@bpradipt - I've changed it to use "default" as other rolebindings.
@gkurz - I hope I've managed to change it as requested!
@pmores - Sorry, I don't understand what you meant. All these changes are manual changes, none of them are generated.

Makefile
```diff
@@ -50,11 +50,11 @@ endif
 
 # Set the Operator SDK version to use. By default, what is installed on the system is used.
 # This is useful for CI or a project to utilize a specific version of the operator-sdk toolkit.
-OPERATOR_SDK_VERSION ?= v1.36.1
+OPERATOR_SDK_VERSION ?= v1.38.0
```
gkurz: It is fine to skip v1.37.0 if it doesn't affect us, which seems to be the case from https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.37.0/, but it should be mentioned explicitly in the commit message for extra clarity.

config/manager/manager.yaml
```diff
@@ -59,7 +59,8 @@ spec:
         - command:
             - /manager
           args:
-            - --enable-leader-election
+            - --metrics-bind-address=127.0.0.1:8080
+            - --leader-elect
```
gkurz, line 63: This looks like this is fixing a bug from cbebd561bfa0a3967db447bb051cc451e4447e65 that turned `enable-leader-election` into `leader-elect`. This was a long long time ago so I am a bit surprised this wasn't discovered before. How did you spot it ? Code reading ? Hit some error ? Also, is `metrics-bind-address` needed here ?

Cc @bpradipt as author of cbebd561bfa0a3967db447bb051cc451e4447e65.

balintTobik: When I tested my change, I faced an issue, try to use a non-existing argument. I saw that the it's changed in clusterserviceversion.yaml during bundle creation. It's interesting, because when I build it without these changes (with 1.36.0 operator-sdk), it didn't changed this arguments. 
Maybe, because of removing/adding manifests to clusterserviceversion reload manager manifest as well...

Default, metrics-bind-address is ":8080", so I think it's needed. 
Also this value is used currently: https://github.com/openshift/sandboxed-containers-operator/blob/devel/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml#L373

Makefile
balintTobik: Thanks! I forget that.

config/rbac/metrics_auth_role_binding.yaml
bpradipt: @balintTobik I think we just didn't remove it since OSC operator was always installed in a specific namespace.

config/manager/manager.yaml
bpradipt: I wonder how it worked till now with the incorrect option in manager.yaml - https://github.com/openshift/sandboxed-containers-operator/blob/devel/main.go#L97-L98
Only possibility is that the config/manager/manager.yaml never got used.
Anyways, thanks @balintTobik for fixing this.

bpradipt (approved):
/lgtm
We should test the changes to ensure no breakage. Hope the upstream test will be sufficient.

littlejawa:
/hold
As discussed on slack: we want to make build 11 before merging this. 
We can continue the reviewing process, just putting the "hold" label to avoid merging it too soon.

wainersm (approved):
LGTM

littlejawa:
@balintTobik - It seems the kube-rbac-proxy image is still referenced in the bundle CSV file.
You probably need to run `make bundle` and commit the relevant modifications.

We also need to remove it from the manifest file (/config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml)

gkurz:
> @balintTobik - It seems the kube-rbac-proxy image is still referenced in the bundle CSV file. You probably need to run `make bundle` and commit the relevant modifications.
> 
> We also need to remove it from the manifest file (/config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml)

Just a reminder that we also need to convert the bundle in midstream.

docs/DEVELOPMENT.md
```diff
@@ -2,11 +2,11 @@
 
 ## Prerequisites
 - Golang - 1.22.x
-- Operator SDK version - 1.36.1
+- Operator SDK version - 1.39.1
 ```
 export ARCH=$(case $(uname -m) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(uname -m) ;; esac)
 export OS=$(uname | awk '{print tolower($0)}')
-export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.36.1
+export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/v1.39.1
```
gkurz, line 9: We don't need to download the SDK manually since 0a1028a134bdb4458df717a6aaedd7ab49cfbe8a but the documentation wasn't updated accordingly.

Please drop this block code entirely in a preparatory patch.

gkurz: We don't need to download the SDK manually since 0a1028a134bdb4458df717a6aaedd7ab49cfbe8a but the documentation wasn't updated accordingly.

Please drop this block code entirely in a preparatory patch (ideally first patch in the PR).

gkurz (approved):
/lgtm

Thanks @balintTobik !

Let's wait for CI to run and we'll merge this.

littlejawa (approved):
/lgtm

PR #541, opened 2025-02-18
FBC updates

spotlesstofu:
To complete KATA-3577.


spotlesstofu:
This is ready. I'll rebase and squash after #552 is merged. `v4.18` will come in a new pull.
@littlejawa @lmilleri if you could review please.

fbc/render.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env sh
+
+# Print what you're doing, exit on error.
+set -xe
+
+OCP_VERSIONS=$1
+
+test -n "$OCP_VERSIONS" || OCP_VERSIONS=$(ls v4.*)
+
+BUILD_REGISTRY="quay.io/redhat-user-workloads/ose-osc-tenant/trustee/"
```
littlejawa:
```suggestion
BUILD_REGISTRY="quay.io/redhat-user-workloads/ose-osc-tenant/"
```

.tekton/osc-operator-pull-request.yaml
```diff
@@ -7,8 +7,10 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
-      == "devel"
+    pipelinesascode.tekton.dev/on-cel-expression:
+      event == "pull_request" &&
+      target_branch == "devel" &&
+      files.all.exists(path, !path.matches('.tekton/.*fbc-.*|fbc/.*))
```
littlejawa, line 16: Do we need to do the same for the other pipelines?

littlejawa:
> I'll rebase and squash after https://github.com/openshift/sandboxed-containers-operator/pull/552 is merged

Maybe that's what you meant, but just to make sure:
Do not squash everything into a single commit. From what I see, I think you can group them by topic (duplicate, render, etc), and a separate one to add the generated files.
Introducing them as separate commits helps reviewing, but also in the future if we want to look up something specific about one of the scripts, we don't want to review one huge commit.

fbc/render.sh
```diff
@@ -0,0 +1,43 @@
...
+RELEASE_REGISTRY="registry.redhat.io/confidential-compute-attestation-tech-preview/"
```
lmilleri: replace confidential-compute-attestation-tech-preview?

.tekton/osc-operator-pull-request.yaml
spotlesstofu: no, the other pipelines start only when their files are modified ("allow list")

only the operator pipeline uses a "deny list"

spotlesstofu:
I addressed all the comments

littlejawa (approved):
/lgtm
Thanks @spotlesstofu

spotlesstofu:
Please merge when you see fit. I don't have write access.

PR #542, opened 2025-02-19
gcp: adding cco request

beraldoleal:
This commit adds a GCP CredentialsRequest following the existing AWS and Azure patterns. 
Fixes rhjira#KATA-3627.

### DIFF
--- a/config/peerpods/credentials-requests/credentials_request_gcp.yaml
+++ b/config/peerpods/credentials-requests/credentials_request_gcp.yaml
@@ -1,0 +1,63 @@
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
+metadata:
+  name: openshift-sandboxed-containers-gcp
+  namespace: openshift-cloud-credential-operator
+spec:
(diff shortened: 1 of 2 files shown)


bpradipt (approved):
/lgtm
It will be good to test this once before merge

snir911 (approved):
LGTM thanks, I'd test before merge

beraldoleal:
@snir911 @bpradipt, thanks for the review.

My Prow cluster couldn't fulfill the original request since computeAdmin needs the Backup API, which isn't enabled. I checked other compute.* roles and found a more restrictive set of permissions. I force-pushed the update and tested it with CCO, and the secret was created.

I'll test it further with Peer Pods to make sure everything works as expected.

snir911 (approved):

beraldoleal:
@bpradipt @snir911 I tested this, and it was able to create the VM using the credentials generated. We might need additional permissions in the future but for now I would say we are good to merge.

bpradipt (approved):

PR #552, opened 2025-02-25
Avoid unneeded rebuilds

littlejawa:
Modify the tekton files to avoid rebuilds in some situations:

Bundle: rebuild only if the bundle files are modified
Operator: rebuild only if the modification is not on the bundle, must-gather or podvm-builder files.


spotlesstofu (approved):
Love it!

PR #557, opened 2025-02-28
CSV update: use konflux repository for our images

littlejawa:
Changing the image references to use the konflux repo. This is a temporary measure, to validate konflux's nudging process, where the bundle file gets updated when the images are rebuilt.

We will put back the release registry URL when our release plan is set on the konflux side. Then konflux should be able to update the images with the release URL.

For reference: I am following the instructions from https://konflux-ci.dev/docs/advanced-how-tos/building-olm/#automating-updates-for-image-references-in-the-csv

### DIFF
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -378,15 +378,15 @@ spec:
                 - name: PEERPODS_NAMESPACE
                   value: openshift-sandboxed-containers-operator
                 - name: RELATED_IMAGE_KATA_MONITOR
-                  value: registry.redhat.io/openshift-sandboxed-containers/osc-monitor-rhel9:1.9.0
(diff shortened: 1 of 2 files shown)


littlejawa:
@spotlesstofu @gkurz - It's a pretty simple one, but I don't want to merge without a review :-)
Could you get a look ?

spotlesstofu (approved):

gkurz:
@littlejawa this change is reverted by running `make bundle` locally. Is it intended or should we patch https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/manager/manager.yaml as well ?

littlejawa:
> this change is reverted by running make bundle locally. 

@gkurz - Right, I need to update the manager.yaml file too.
I'm also adding another change: the metrics server image was not updated, and it needs to point to the same place.
I'm not sure Konflux will know how to look it up and update though - the document mention the operator's image and related images, but not secondary containers deployed from the same place... We will need to keep an eye out on this when we get CSV updates from konflux after a merge.

littlejawa:
Updated again with the current reference for the operator - Build was failing because the image didn't exist anymore.

beraldoleal (approved):
LGTM, thanks.

PR #563, opened 2025-03-04
azure: blob as source update

beraldoleal:
Per Microsoft's announcement, starting March 15, 2025, additional permission is required when using blobs as the source for Azure Compute Gallery (ACG) images. This updates our CCO request to include the required permission.

My understanding is that the Azure job that creates the image is building the podvm based on blob (VHD), so we need to update only the CCO request.

Fixes: rhjira#[KATA-3579](https://issues.redhat.com//browse/KATA-3579)

**- What I did**

**- How to verify it**

### DIFF
--- a/config/peerpods/credentials-requests/credentials_request_azure.yaml
+++ b/config/peerpods/credentials-requests/credentials_request_azure.yaml
@@ -14,3 +14,4 @@ spec:
       - role: Reader
       - role: Virtual Machine Contributor
       - role: Network Contributor
+      - role: Storage Account Contributor


snir911 (approved):
LGTM, thanks!

gkurz (commented):
Hi @beraldoleal !

> My understanding is that the Azure job that creates the image is
building the podvm based on blob (VHD),

The podvm script for azure currently supports building from VHD but it also supports building from scratch using packer. Are you suggesting that we shouldn't support building from scratch anymore ?

> so we need to update only the CCO request.

The jira also mentions to stop using an _old property_ source id, that we're using in the script at https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/azure-podvm-image-handler.sh#L785

Should something be done here ?

beraldoleal:
> The podvm script for azure currently supports building from VHD but it also supports building from scratch using packer. Are you suggesting that we shouldn't support building from scratch anymore ?

While I believe this could be deprecated in future versions, this is not what this PR is suggesting.  When building from packer (from scratch), we’re working with an image that’s already created (by packer), not a live VM being used as the source for an ACG Image version.

So it looks like we are not using this "from vm use case", _directly_. But this is new to me, and I might be wrong.

> The jira also mentions to stop using an old property source id, that we're using in the script at https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/azure-podvm-image-handler.sh#L785

> Should something be done here ?

I found `storageProfile.source.id` in our image deletion query, but swapping it for `storageProfile.source.virtualMachineId` didn’t return anything. That makes me think the query is pulling an image reference, not a VM source. So, unless I'm missing something, no update here is needed.

wainersm:
I know very few about this piece of code, so take my words with grain of salt.

It's not 100% clear to me what method packer uses to publish the image, but I bet it publishes from VM. I assume that packer somehow uses the CCO requested credentials, on which `Virtual Machine Contributor` role binding should satisfy the  `Users will require write permission on the Virtual Machine to create an ACG Image version` in [2].

As for `For Azure SDK, use the property [properties.storageProfile.source.virtualMachineId` in creation from VM [2], it should be addressed by packer if used. We are using the 2.0.2 version of the azure plug-in [3], which is quite old compared with latest 2.3.0 [4]. We might need to bump the plug-in version too... 

This PR seems to address the requirements for when publishing the image from a VHD, as `Users will require Microsoft.Storage/storageAccounts/listKeys/action (or) Storage Account Contributor on the storage account` stated in [2].

[1] https://github.com/openshift/sandboxed-containers-operator/blob/768d2192aac640733efeffa38bece0dc3412ce36/config/peerpods/credentials-requests/credentials_request_azure.yaml#L13
[2] https://learn.microsoft.com/en-us/azure/virtual-machines/shared-image-galleries?tabs=vmsource%2Cazure-cli#rbac-permissions-required-to-create-an-acg-image
[3] https://github.com/confidential-containers/cloud-api-adaptor/blob/main/src/cloud-api-adaptor/azure/image/rhel/azure-rhel.pkr.hcl#L4
[4] https://developer.hashicorp.com/packer/integrations/hashicorp/azure

littlejawa:
> > Should something be done here ?
> 
> I found `storageProfile.source.id` in our image deletion query, but swapping it for `storageProfile.source.virtualMachineId` didn’t return anything. That makes me think the query is pulling an image reference, not a VM source. So, unless I'm missing something, no update here is needed.

@beraldoleal - in the JIRA ticket mentioned above, for the blob creation case, there is this:

> If you're using the old property ‘[properties.storageProfile.[osDiskImage/dataDiskImages].source.Id,‘](https://learn.microsoft.com/rest/api/compute/gallery-image-versions/create-or-update?view=rest-compute-2023-10-02&%253Btabs=HTTP&tabs=HTTP#gallerydiskimagesource) you should move to the property ’[properties.storageProfile.osDiskImage.source.storageAccountId’](https://learn.microsoft.com/rest/api/compute/gallery-image-versions/create-or-update?view=rest-compute-2023-10-02&%3Btabs=HTTP&tabs=HTTP#gallerydiskimagesource). This property requires minimum api-version 2022-03-03.

You tried with `storageProfile.source.virtualMachineId`. Should we use something like: `properties.storageProfile.osDiskImage.source.storageAccountId` here instead?

beraldoleal:
> > > Should something be done here ?
> > 
> > 
> > I found `storageProfile.source.id` in our image deletion query, but swapping it for `storageProfile.source.virtualMachineId` didn’t return anything. That makes me think the query is pulling an image reference, not a VM source. So, unless I'm missing something, no update here is needed.
> 
> @beraldoleal - in the JIRA ticket mentioned above, for the blob creation case, there is this:
> 
> > If you're using the old property ‘[properties.storageProfile.[osDiskImage/dataDiskImages].source.Id,‘](https://learn.microsoft.com/rest/api/compute/gallery-image-versions/create-or-update?view=rest-compute-2023-10-02&%253Btabs=HTTP&tabs=HTTP#gallerydiskimagesource) you should move to the property ’[properties.storageProfile.osDiskImage.source.storageAccountId’](https://learn.microsoft.com/rest/api/compute/gallery-image-versions/create-or-update?view=rest-compute-2023-10-02&%3Btabs=HTTP&tabs=HTTP#gallerydiskimagesource). This property requires minimum api-version 2022-03-03.
> 
> You tried with `storageProfile.source.virtualMachineId`. Should we use something like: `properties.storageProfile.osDiskImage.source.storageAccountId` here instead?

```
# Current
$ az sig image-version show --resource-group "${AZURE_RESOURCE_GROUP}"     --gallery-name "${IMAGE_GALLERY_NAME}"     --gallery-image-definition "${IMAGE_DEFINITION_NAME}"     --gallery-image-version "${IMAGE_VERSION}"     --query "storageProfile.source.id" --output tsv
/subscriptions/<REMOVED>/resourceGroups/bleal-rg/providers/Microsoft.Compute/images/podvm-image-0.0.2024100258

# Suggestion
$ az sig image-version show --resource-group "${AZURE_RESOURCE_GROUP}"     --gallery-name "${IMAGE_GALLERY_NAME}"     --gallery-image-definition "${IMAGE_DEFINITION_NAME}"     --gallery-image-version "${IMAGE_VERSION}"     --query "storageProfile.osDiskImage.source.storageAccountId" --output tsv
```

wainersm (approved):
As discussed in Slack, despite having some loosing ends, this PR solves at least one of the issues associated with the upcoming azure changes. If it requires more fixes we will probably see only after the change is rolled out by azure as we cannot fully test it upfront.

I'm talking to much to say LGTM :)

gkurz (approved):
With the caveat that the validity of the `storageProfile.source.id` used during podvm image deletion is unsure, let's go forward with this.

PR #564, opened 2025-03-07
fix: correct behaviour of labelling script

butler54:
The Readme for the bare metal install script was incorrect. it did not provide the object type (a node) when trying to label the object
**- What I did**
Corrected the script in the readme
**- How to verify it**
Tested locally using oc against any cluster (e.g can be crc etc)

Updated bare metal installation helper scripts documentation correcting example code.

### DIFF
--- a/scripts/install-helpers/baremetal-coco/README.md
+++ b/scripts/install-helpers/baremetal-coco/README.md
@@ -65,7 +65,7 @@ The deployment sequence is described below:
   
   ```sh
   export NODENAME=<node>
-  oc label $NODE_NAME coco_bm=true
+  oc label node $NODENAME coco_bm=true
   export BM_NODE_LABEL="coco_bm=true"
   ```
 


beraldoleal (approved):
LGTM, thanks.

gkurz (commented):
The fix is trivial enough that it doesn't really require a description but please add a valid signed-off-by tag at least.

gkurz (approved):
/lgtm

Thanks @butler54 !

But please add the signed-off-by anyway :-)

butler54:
> /lgtm
> 
> Thanks @butler54 !
> 
> But please add the signed-off-by anyway :-)

Will do

littlejawa:
The build didn't trigger for some reason, but I know we've had trouble with it recently. Aside from that, all tests are passing.
The change is documentation only and doesn't affect the build, so I will merge it as is.

PR #571, opened 2025-03-13
Fix konflux build triggers

littlejawa:


gkurz (approved):
Good catch ! I just hope that konflux eventually adds some logic to detect such typos in tekton annotations and report them.

/lgtm

Thanks @littlejawa !

spotlesstofu (approved):

PR #572, opened 2025-03-13
Check the existence of the PeerPodConfig CRD

gkurz:
PR #533 added some logic to move away from the PeerPodConfig CRD. Unfortunately it doesn't cope with the case where the CRD doesn't exist at all in the cluster. This causes the controller to enter a reconcile failure loop with the following error.

Failed to migrate PeerPodConfig limit	{"err": "no matches for
 kind \"PeerPodConfig\" in version \"confidentialcontainers.org/v1alpha1\""}

If the CRD isn't defined, no need to migrate obviously.

Fast path fix without a jira :wink:

### DIFF
--- a/controllers/migrate.go
+++ b/controllers/migrate.go
@@ -17,6 +17,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


beraldoleal (approved):
LGTM, thanks @gkurz .

littlejawa (approved):
Thanks @gkurz !
/lgtm

PR #576, opened 2025-03-14
podvm: mark CAA_SRC and CAA_REF as optional in the config maps

littlejawa:
We are using our fork's v0.12.0-1.9 branch for this release.

Open question: what about [libvirt](https://github.com/openshift/sandboxed-containers-operator/blob/78fd45458a79c9a5de154ff5cdc0ad04a995b70e/config/peerpods/podvm/libvirt-podvm-image-cm.yaml#L13-L14)?
It's not looking at our fork, but directly upstream, and I don't know the support status for it in our OSC builds (aside from s390x). Should we update it for consistency, or is it a separate work?

Fixes: [KATA-3697](https://issues.redhat.com//browse/KATA-3697)

### DIFF
--- a/config/peerpods/podvm/aws-podvm-image-cm.yaml
+++ b/config/peerpods/podvm/aws-podvm-image-cm.yaml
@@ -18,12 +18,12 @@ data:
   # Pod VM sources
   # If changing the source, then ensure the respective payload binaries are available
   # for the new source
-  CAA_SRC: "https://gitlab.cee.redhat.com/osc/midstream/cloud-api-adaptor"
-  CAA_REF: "v0.8.0-1.6"
+  #CAA_SRC: "" ## set to cloud-api-adaptor sources URL for custom podvm builds
+  #CAA_REF: "" ## set to tag/branch for the cloud-api-adaptor code to use
 
   # Booleans
   INSTALL_PACKAGES: "no"
-  DOWNLOAD_SOURCES: "no"
+  DOWNLOAD_SOURCES: "no" ## set to "yes" for custom podvm builds. Need to set CAA_SRC and CAA_REF accordingly
   CONFIDENTIAL_COMPUTE_ENABLED: "no"
   DISABLE_CLOUD_CONFIG: "true"
   ENABLE_NVIDIA_GPU: "no"
--- a/config/peerpods/podvm/azure-podvm-image-cm.yaml
+++ b/config/peerpods/podvm/azure-podvm-image-cm.yaml
@@ -41,12 +41,12 @@ data:
   # Pod VM sources
   # If changing the source, then ensure the respective payload binaries are available
   # for the new source
-  CAA_SRC: "https://gitlab.cee.redhat.com/osc/midstream/cloud-api-adaptor"
-  CAA_REF: "v0.8.0-1.6"
+  #CAA_SRC: "" ## set to cloud-api-adaptor sources URL for custom podvm builds
+  #CAA_REF: "" ## set to tag/branch for the cloud-api-adaptor code to use
 
   # Booleans
   INSTALL_PACKAGES: "no"
-  DOWNLOAD_SOURCES: "no"
+  DOWNLOAD_SOURCES: "no" ## set to "yes" for custom podvm builds. Need to set CAA_SRC and CAA_REF accordingly
   CONFIDENTIAL_COMPUTE_ENABLED: "no"
   DISABLE_CLOUD_CONFIG: "true"
   ENABLE_NVIDIA_GPU: "no"
--- a/config/peerpods/podvm/gcp-podvm-image-cm.yaml
+++ b/config/peerpods/podvm/gcp-podvm-image-cm.yaml
@@ -7,8 +7,8 @@ data:
   # Pod VM sources
   # If changing the source, then ensure the respective payload binaries are available
   # for the new source
-  CAA_SRC: "https://github.com/confidential-containers/cloud-api-adaptor"
-  CAA_REF: "main"
+  #CAA_SRC: "" ## set to cloud-api-adaptor sources URL for custom podvm builds
+  #CAA_REF: "" ## set to tag/branch for the cloud-api-adaptor code to use
 
   IMAGE_BASE_NAME: "podvm-image"
   IMAGE_VERSION_MAJ_MIN: "0.0"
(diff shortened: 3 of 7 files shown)


config/peerpods/podvm/aws-podvm-image-cm.yaml
```diff
@@ -19,7 +19,7 @@ data:
   # If changing the source, then ensure the respective payload binaries are available
   # for the new source
```
bpradipt, line 20:
```Suggestion
# The CAA_SRC and CAA_REF build arguments are not used in downstream build
```

config/peerpods/podvm/azure-podvm-image-cm.yaml
```diff
@@ -42,7 +42,7 @@ data:
   # If changing the source, then ensure the respective payload binaries are available
   # for the new source
```
bpradipt, line 43:
```Suggestion
# The CAA_SRC and CAA_REF build arguments are not used in downstream build
```

config/peerpods/podvm/gcp-podvm-image-cm.yaml
```diff
@@ -7,8 +7,8 @@ data:
   # Pod VM sources
   # If changing the source, then ensure the respective payload binaries are available
   # for the new source
```
bpradipt, line 9:
```Suggestion
# The CAA_SRC and CAA_REF build arguments are not used in downstream build
```

gkurz:
Now that I realized that CAA_SRC and CAA_REF fall into the debugging/experimenting category, I think they should not be used at all in production code. They should not have default values either and force whoever uses them to provide valid values.

wainersm:
Hi @littlejawa ! Thanks for this!

I also don't know what to do with libvirt's reference either :(

One suggestion: update https://github.com/openshift/sandboxed-containers-operator/blob/devel/docs/DEVELOPMENT.md#updating-versions so that we don't leave these configmaps behind.

wainersm:
@gkurz @littlejawa If we decide for dropping these options from downstream then maybe we could mark with a tag similarly to `## OSC_VERSION` (for example, `## UPSTREAM-ONLY`) and let the downstream automation to parse and strip the properties.

Alternately, we could use `kustomize` (or any other deployment tool) to provide different configmaps according to the use case. Example, one for downstream, another for upstream, another for enabling debug...etc

config/peerpods/podvm/gcp-podvm-image-cm.yaml
```diff
@@ -7,8 +7,8 @@ data:
...
-  CAA_SRC: "https://github.com/confidential-containers/cloud-api-adaptor"
-  CAA_REF: "main"
+  #CAA_SRC: "" ## set to cloud-api-adaptor sources URL for custom podvm builds
+  #CAA_REF: "" ## set to tag/branch for the cloud-api-adaptor code to use
```
littlejawa, line 11: @beraldoleal - those variables are actually not used in the GCP case right?
It doesn't use a "DOWNLOAD_SOURCES" flag, I see no reference to CAA_SRC/CAA_REF in the script, and the script doesn't call the "download_source_code" function from libs.sh.
Should we remove them altogether?

littlejawa:
I'm changing to something a bit more complex, but that should make it clear for the future:
- make sure we don't look at the CAA_REF/CAA_SRC variables unless we are doing a custom podvm build (i.e: DOWNLOAD_SOURCES is set to "yes")
- comment the variables out, remove any default value, and add a comment that explains how to use them, and why

I think it clarifies the situation, and should prevent any further question (because I know I will forget everything about it in 6 months :-) )

littlejawa:
I'm probably not the only one, so let me share some thoughts:

The reason why I was confused with these variables being used downstream is probably this:
https://github.com/openshift/sandboxed-containers-operator/blob/3c859e45e6733ba1f44cfd82a3575993fb554141/config/peerpods/podvm/Dockerfile.podvm-builder#L36-L37

We are using the exact same variable names in the Dockerfile that is in the same folder as the scripts, and that we use for building the podvm builder image (not in CPaaS though, but that's another story).

Because of that, I was under the impression that the script was calling the Dockerfile, OR that the Dockerfile was running the script, or something like that... In other words: that there was a relation between the CAA_SRC of the Dockerfile, and the one from the script.

But that's wrong. The Dockerfile uses CAA_SRC and CAA_REF to checkout the code into the podvm-builder image (here : https://github.com/openshift/sandboxed-containers-operator/blob/3c859e45e6733ba1f44cfd82a3575993fb554141/config/peerpods/podvm/Dockerfile.podvm-builder#L50), and it copies the scripts in it without running them.

Then the scripts are called during podvm image creation, and by default it uses the code that the Dockerfile has put in there.

The variables in the Dockerfile are not propagated to the scripts - the scripts are run with the variable provided by the config map. So there is no relationship between the two sets of variables.

(sorry if that was obvious for you all, but I needed that much time to clarify it :-) And of course, if I'm *still* wrong, I would love to be enlightened :-p)

config/peerpods/podvm/gcp-podvm-image-cm.yaml
beraldoleal: @littlejawa gcp has no "from scratch" implementation at this point. So I'm tempted to say, yes.

config/peerpods/podvm/aws-podvm-image-cm.yaml
```diff
@@ -18,12 +18,12 @@ data:
...
-  CAA_SRC: "https://gitlab.cee.redhat.com/osc/midstream/cloud-api-adaptor"
-  CAA_REF: "v0.8.0-1.6"
+  #CAA_SRC: "" ## set to cloud-api-adaptor sources URL for custom podvm builds
+  #CAA_REF: "" ## set to tag/branch for the cloud-api-adaptor code to use
 
   # Booleans
   INSTALL_PACKAGES: "no"
-  DOWNLOAD_SOURCES: "no"
+  DOWNLOAD_SOURCES: "no" ## set to "yes" for custom podvm builds. Need to set CAA_SRC and CAA_REF accordingly
```
wainersm, line 26: Hi @littlejawa ! I think we should hide `DOWNLOAD_SOURCES` from final users too.

gkurz: Yes, IMO it could even be fully removed. In order to do `git clone` we just need `CAA_SRC`. If some other ref than the default branch is wanted then `CAA_REF` should be provided. No need to do any validation in this case.

config/peerpods/podvm/gcp-podvm-image-cm.yaml
gkurz: Let's drop the variables then. They'll eventually be added back when there's a case to build from source for GCP.

config/peerpods/podvm/aws-podvm-image-cm.yaml
bpradipt: My suggestion will be to just document the variables (and not remove anything right now), and sync with end-user documentation to ensure we explicitly state out config options that should not be touched.
Given we are looking at bootc based image build, things will change soon. So may be we should focus the effort there.

PR #580, opened 2025-03-17
GCP: Fixes to the upload image process

beraldoleal:


wainersm:
hi @beraldoleal !

Not a GCP expert but the changes here sounds reasonable.

/lgtm

config/peerpods/podvm/gcp-podvm-image-handler.sh
```diff
@@ -109,9 +108,7 @@ function create_image() {
 }
 
 function set_image_version_and_name() {
-  # Set the image version
-  # It should follow the Major(int).Minor(int).Patch(int)
-  IMAGE_VERSION="${IMAGE_VERSION_MAJ_MIN}.$(date +'%Y%m%d%S')"
+  IMAGE_VERSION="$(date +'%Y%m%d%S')"
```
bpradipt, line 111: You can add a comment here based on what is expected by GCP.
Like `Major(int).Minor(int).Patch(int)` is expected by Azure

littlejawa (approved):
Need to address the comment from Pradipta, but otherwise LGTM

config/peerpods/podvm/gcp-podvm-image-handler.sh
beraldoleal: @bpradipt since there is no version concept there, there is no expectations on the GCP side. This is a convention defined by us.

bpradipt: I see..

bpradipt (approved):
/lgtm

beraldoleal:
I will send an empty force push , just to see if the tasks will run again.

PR #583, opened 2025-03-19
controller: Ensure image creation job runs on an installed node

bpradipt:
This potentially limits the issues that may be caused due to node reboot due to Kata rpm installation
Ideally we need to handle the image creation (and subsequent lifecycle management) via a proper CRD and state management but we don't do it. Consider this as an interim fix till we have a proper image lifecycle management support.

Fixes: #[KATA-3573](https://issues.redhat.com//browse/KATA-3573)

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1428,7 +1428,18 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 			// RequeueNeeded
 			// ImageCreationStatusUnknown
 
-			status, err := ImageCreate(r.Client)
+			// Check if at least one node is installed with Kata
+			// If not, then we should not proceed with the image creation
+			// It's available r.kataConfig.Status.KataNodes.Installed
+
+			if len(r.kataConfig.Status.KataNodes.Installed) == 0 {
+				r.Log.Info("No nodes are installed with Kata. Skipping image creation")
+				return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
+			}
+
+			r.Log.Info("At least one node with Kata is installed. Proceeding to image creation")
+
+			status, err := ImageCreate(r.Client, r.kataConfig.Status.KataNodes.Installed[0])
 			switch status {
 			case ImageCreatedSuccessfully:
 				r.setInProgressConditionToPodVMImageCreated()
--- a/controllers/image_generator.go
+++ b/controllers/image_generator.go
@@ -136,7 +136,7 @@ func GetImageGenerator() *ImageGenerator {
 }
 
 // ImageCreate creates a podvm image for a cloud provider if not present
-func ImageCreate(c client.Client) (int, error) {
+func ImageCreate(c client.Client, nodeName string) (int, error) {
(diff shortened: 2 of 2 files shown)


beraldoleal (approved):
I could not test this , duo to cluster/credentials issues I'm debugging on my side. But LGTM. thanks.

littlejawa (approved):
/lgtm

controllers/openshift_controller.go
```diff
@@ -1428,7 +1428,18 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 			// RequeueNeeded
 			// ImageCreationStatusUnknown
 
-			status, err := ImageCreate(r.Client)
+			// Check if at least one node is installed with Kata
+			// If not, then we should not proceed with the image creation
+			// It's available r.kataConfig.Status.KataNodes.Installed
+
+			if len(r.kataConfig.Status.KataNodes.Installed) == 0 {
```
beraldoleal, line 1435: @bpradipt I was testing some deployments without this PR, and noticed the job didn't started before some nodes are ready. So, I got back to this PR, and I was looking at:

```
1357	if !isMcoUpdating {
```

And I got the impression this condition is skipping your code. What I'm missing?

beraldoleal: And I just had another deployment, that was the opposite of what I said above.. the job started right after installing kata-config... So something is causing the job to be delayed in some cases... :confused:

beraldoleal (approved):

controllers/openshift_controller.go
gkurz: > @bpradipt I was testing some deployments without this PR, and noticed the job didn't started before some nodes are ready. So, I got back to this PR, and I was looking at:
> 
> ```
> 1357	if !isMcoUpdating {
> ```
> 
> And I got the impression this condition is skipping your code. What I'm missing?

You're not missing anything :smile: . Nothing is supposed to happen as long as kata isn't fully deployed on all selected nodes. @pmores is digging some more. Until we get to the bottom of this, let's hold this PR.

beraldoleal: @gkurz, good to know, but keep in mind my second comment as well. Oddly yesterday I saw the case where the job started immediately after deploying the kata-config. :/

gkurz: > @gkurz, good to know, but keep in mind my second comment as well. Oddly yesterday I saw the case where the job started immediately after deploying the kata-config. :/

Sure, I don't challenge that. @vvoronko also shared a kataconfig with similar symptoms yesterday :

```
status:
  conditions:
  - lastTransitionTime: "2025-03-20T08:35:30Z"
    message: Creating Pod VM Image
    reason: PodVMImageJobRunning
    status: "True"
    type: InProgress
  kataNodes:
    installing:
    - ip-10-0-37-13.us-east-2.compute.internal
    nodeCount: 2
    readyNodeCount: 0
    waitingToInstall:
    - ip-10-0-22-123.us-east-2.compute.internal
  runtimeClasses:
  - kata
  waitingForMcoToStart: false
```

but since this isn't supposed to happen according to the code, this PR looks like a band-aid.

bpradipt:
This is no longer needed.

PR #584, opened 2025-03-19
podvm: Check for pre-existing image before image creation

bpradipt:
For OSC, we expect only one image to be created by the operator. The IMAGE_VERSION (for Azure) or AMI_VERSION (for AWS) uses clusterID as the unique id.  This helps to check if the image already exists before recreating the image again.

**- What I did**

**- How to verify it**

### DIFF
--- a/config/peerpods/podvm/azure-podvm-image-handler.sh
+++ b/config/peerpods/podvm/azure-podvm-image-handler.sh
@@ -755,22 +755,20 @@ function image_exists() {
 
     # Handle Azure command failure
     if [[ "${img_return_code}" -ne 0 ]]; then
-        error_exit "Error querying Azure for Image."
+        # Treating any error as non-existent image as unlike AWS, Azure does not return empty string for non-existent image
+        echo "Image does not exist in Azure."
+        return 1
     fi
 
     # Case 1: Image exists and matches the configmap
     if [[ "${image_id}" == "${latest_image_id}" ]]; then
         echo "Image (${image_id}) is up-to-date in configmap."
         return 0
-    # Case 2: No Image in Azure and no record in the configmap
-    elif [[ -z "${image_id}" && -z "${latest_image_id}" ]]; then
-        echo "No Image found in Azure, and no record in configmap."
-        return 1
-    # Case 3: Image missing in Azure but present in configmap
+    # Case 2: Image missing in Azure but present in configmap
     elif [[ -z "${image_id}" && -n "${latest_image_id}" ]]; then
         echo "No Image found in Azure, but configmap has record (${latest_image_id}). Image might have been deleted."
         return 1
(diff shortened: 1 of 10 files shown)


controllers/image_generator.go
```diff
@@ -745,15 +745,40 @@ func (r *ImageGenerator) createImageConfigMapFromFile() error {
...
+	// Also set the IMAGE_VERSION or AMI_VERSION
+
 	if r.provider == AzureProvider {
 		if cm.Data["IMAGE_GALLERY_NAME"] == "" {
 			image_gallery_name := azureImageGalleryPrefix + "_" + r.clusterId
 			cm.Data["IMAGE_GALLERY_NAME"] = image_gallery_name
 			igLogger.Info("Setting IMAGE_GALLERY_NAME", "image_gallery_name", image_gallery_name)
 		}
 
+		cm.Data["IMAGE_VERSION"] = imageVersion
```
beraldoleal, line 774: Since GCP the image version will be part of the name... and the name can't have DOTs. IIUC, we should add another if here to check if the provider == GCP and set  `cm.Data["IMAGE_VERSION"] = imageVersionWithoutdots` ?

beraldoleal, line 774: Since GCP the image version will be part of the name... and the name can't have DOTs. IIUC, we should add another if here to check if the provider == GCP and set  `cm.Data["IMAGE_VERSION"] = imageVersionWithoutdots` ?

beraldoleal (approved):
LGTM, thanks.

beraldoleal:
@bpradipt thanks for handling the gcp part, saved me some time. ;)

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -718,6 +737,52 @@ function delete_image_using_id() {
...
 }
 
+# Function to check if image already exists
+# This checks if IMAGE_VERSION exist in Azure
+# and the LATEST_IMAGE_ID annotation is set in the peer-pods-cm configmap
+# 0: Image exists and matches
+# 1: Image does not exist
+# 2: Image mismatch (caller should delete Image)
+# The required variables are assumed to be set
+function image_exists() {
```
beraldoleal, line 755: I missed this method under gcp. IIUC, most of it could be generic to all providers?

bpradipt: A portion of it yes, but querying is specific to individual provider. So I kept it for each provider..

beraldoleal: And where is the gcp implementation? I saw you changed the gcp-podvm-image-handler.sh but I could not find any image_exists method.

littlejawa (approved):
/lgtm

PR #585, opened 2025-03-19
gcp: PODVM_IMAGE_NAME is the right name

beraldoleal:
The annotation is LATEST_IMAGE_ID but the configmap value is _NAME. I would prefer to not touch on the metadata now, since the metadata.annotation is the same for all providers.


littlejawa (approved):
Looking at the commit that introduced it: it seems you were using the right name everywhere else, it's just this one.

/lgtm
Thanks @beraldoleal !

gkurz (approved):
/lgtm

Thanks @beraldoleal !

PR #586, opened 2025-03-20
bump dependencies: golang/x/crypto v0.35

littlejawa:
This requires bumping the version of go we use too - need 1.23.

golang 1.23 is not available in ubi9 yet, so I need to find another builder image.
We can use the `openshift-golang-builder:rhel_9_golang_1.23` image like we do downstream, but it requires login.

I ~will see how we can make Konflux access it, but~ wanted to discuss it here to see if there was concerns and/or other suggestion.

Edit: Konflux can pull this image without further config apparently.


littlejawa:
Looks like Konflux knows how to pull from brew.
The `ci/prow/check` job, on the other hand, seem to be using a different base image that still uses go 1.22.9 :-(
Waiting on the other tests to report - hopefully, the fact that they're still running is a good sign.

gkurz:
> Looks like Konflux knows how to pull from brew. The `ci/prow/check` job, on the other hand, seem to be using a different base image that still uses go 1.22.9 :-( Waiting on the other tests to report - hopefully, the fact that they're still running is a good sign.

https://github.com/openshift/release/blob/master/ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml#L11

Needs bumping as well :-)

Dockerfile
```diff
@@ -1,6 +1,5 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.22.9-1741020486 AS builder
+FROM brew.registry.redhat.io/rh-osbs/openshift-golang-builder:rhel_9_golang_1.23 AS builder
```
gkurz, line 1: And we're back at using a private image in upstream as before #274 :disappointed: ... well, maybe not a big deal since only RH seems to be contributing to this repo anyway.

An alternative could be to use some well known public builder, e.g. `quay.io/projectquay/golang:1.23` which is a mirror of the official golang builder at docker hub.

littlejawa:
Waiting for https://github.com/openshift/release/pull/63034 to be merged, so that the ci/prow/check test can pass

Dockerfile
littlejawa: Yeah, I'm still not sure what our position should be. I actually have trouble calling this repo "upstream" when I know it was designed and developed exclusively for Openshift, and nobody (that I know of) uses this code outside of it.
This repo is under the "Openshift" org in github. It produces an operator called "Openshift Sandboxed Containers Operator". Shaping the code to target Openshift and testing/building it with Openshift tools doesn't seem wrong.

As we're moving to setup CI/building pipelines on this repo, we need to produce images that are actually deliverable in Openshift. This limits us in the builder images we can use, because Openshift has some policies in place about the toolchain we use. And using ARG variables in the Dockerfile is not working because the release process need to have a hardcoded "FROM" URL (I suspect it is to enforce the policies, to keep control of what we build from?).

For test/debug work (which was the concern in the PR you mention) I think editing the Dockerfile to use a different builder image is acceptable. But for this repo, having a "private" builder image doesn't seem wrong to me.

gkurz (approved):
>  But for this repo, having a "private" builder image doesn't seem wrong to me.

A quick look at some other openshift operators shows that they use RH internal go builder images as well. This comforts your point :-)

/lgtm

Thanks @littlejawa !

PR #588, opened 2025-03-25
Dependency updates

littlejawa:
This PR contains several commits from Konflux, all grouped to make a single update of the go dependencies.

The original PRs are:
- #536
- #545
- #561 
- #562

I had to do some changes in addition to that, squashed into the commit that required it:
- controller-runtime update required a change to the webhook code, because of a deprecated type
- updating controller-tools requires us to update our Makefile too. This in turn results in a modified yaml file that I'm adding in the same commit.


c3d (approved):
Looks good to me, thanks @littlejawa

littlejawa:
Note: the controller-tool in go.mod goes from 0.14.0 to 0.17.2, but in our Makefile it moves only from 0.16.5 to 0.17.2
I think we missed an update at some point :-/

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #589, opened 2025-03-25
gcp: adding image check before creating image

beraldoleal:
As following other providers logic. A follow up to #584.

### DIFF
--- a/config/peerpods/podvm/gcp-podvm-image-handler.sh
+++ b/config/peerpods/podvm/gcp-podvm-image-handler.sh
@@ -105,6 +105,7 @@ function create_image() {
 function set_image_name() {
   # Set the image name
   IMAGE_NAME="${IMAGE_BASE_NAME}-${IMAGE_VERSION}"
+  echo "Image name: ${IMAGE_NAME}"
   export IMAGE_NAME
 }
 
@@ -113,6 +114,19 @@ function create_image_from_prebuilt_artifact() {
 
   # Set the IMAGE_NAME
   set_image_name
+
+  image_exists
+  image_status=$?
+
+  if [[ "${image_status}" -eq 0 ]]; then
+      echo "Image exists. Skipping creation"
+      return
+  elif [[ "${image_status}" -eq 2 ]]; then
+      echo "Deleting image version ${IMAGE_VERSION}, before recreating"
+      delete_image_using_id
+  fi
+
+  echo "Image does not exist. Proceeding to create the image"
 
   echo "Pulling the podvm image from the provided path"
   image_src="/tmp/image"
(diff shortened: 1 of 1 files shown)


beraldoleal:
@bpradipt @littlejawa, yesterday I force pushed this commit to the PR branch, I remember seeing that commit on that PR, for some reason, it looks like it got lost, maybe with another force-push. So my bad.

config/peerpods/podvm/gcp-podvm-image-handler.sh
```diff
@@ -114,6 +115,19 @@ function create_image_from_prebuilt_artifact() {
...
   set_image_name
 
+  image_exists
+  image_status=$?
+
+  if [[ "${image_status}" -eq 0 ]]; then
+      echo "Image exists. Skipping creation"
+      return
+  elif [[ "${image_status}" -eq 2 ]]; then
+      echo "Deleting image version, before recreating"
```
bpradipt:
```Suggestion
      echo "Deleting image version (${IMAGE_VERSION}), before recreating"
```

config/peerpods/podvm/gcp-podvm-image-handler.sh
```diff
@@ -188,6 +202,39 @@ function create_image_from_prebuilt_artifact() {
   echo "GCP image created successfully from prebuilt artifact"
 }
 
+function image_exists() {
+    echo "Checking if GCP image exists"
+
+    local image_path latest_image_id
+
+		# This command returns something like this:
```
bpradipt: nit: indentation

config/peerpods/podvm/gcp-podvm-image-handler.sh
```diff
@@ -188,6 +202,39 @@ function create_image_from_prebuilt_artifact() {
...
+		# https://www.googleapis.com/compute/v1/projects/my-project/global/images/my-image
+		# Getting the cutted version to fetch from projects.*
```
bpradipt:
```Suggestion
# Use `cut` to get the image-id
```

config/peerpods/podvm/gcp-podvm-image-handler.sh
```diff
@@ -188,6 +202,39 @@ function create_image_from_prebuilt_artifact() {
...
+    image_path=$(gcloud compute images describe "${IMAGE_NAME}-${IMAGE_VERSION}" \
+        --project="${GCP_PROJECT_ID}" \
+        --format='get(selfLink)' 2>/dev/null | cut -d/ -f6-)
```
bpradipt, line 215: What about handling the return code from `gcloud compute` ?

config/peerpods/podvm/gcp-podvm-image-handler.sh
```diff
@@ -188,6 +202,44 @@ function create_image_from_prebuilt_artifact() {
...
+    local image_path latest_image_id
+
+    # This command returns something like this:
+    # https://www.googleapis.com/compute/v1/projects/my-project/global/images/my-image
+    # Use `cut` to the image name
+    image_path=$(gcloud compute images describe "${IMAGE_NAME}-${IMAGE_VERSION}" \
+        --project="${GCP_PROJECT_ID}" \
+        --format='get(selfLink)' 2>/dev/null | cut -d/ -f6-)
+
+    if [[ $? -ne 0 ]]; then
```
bpradipt, line 217: Azure returns an error if image is missing, whereas AWS returns an empty response if image is missing.
What does GCP returns?

beraldoleal: Unfortunately, GCP doesn't return a specific exit code if the image is not present. It will show an error, but the return it will be 0. For that case we are going to hit one of the conditions following...

bpradipt (approved):
lgtm

config/peerpods/podvm/gcp-podvm-image-handler.sh
```diff
@@ -188,6 +202,44 @@ function create_image_from_prebuilt_artifact() {
...
 }
 
+function image_exists() {
+    echo "Checking if GCP image exists"
+
+    local image_path latest_image_id
+
+    # This command returns something like this:
+    # https://www.googleapis.com/compute/v1/projects/my-project/global/images/my-image
+    # Use `cut` to the image name
```
gkurz, line 212: It is obvious that `cut` is used below, it is less obvious what we want to drop.
```suggestion
    # we want to drop the leading "https://www.googleapis.com/compute/v1/"
```

config/peerpods/podvm/gcp-podvm-image-handler.sh
gkurz, line 217: AFAICT this branch is dead code.
```
greg@bahia:~$ image_path=$(false | cut -d/ -f6-)
greg@bahia:~$ echo $?
0
```

config/peerpods/podvm/gcp-podvm-image-handler.sh
```diff
@@ -188,6 +202,44 @@ function create_image_from_prebuilt_artifact() {
...
+        return 1
+    # Since GCP supports both cases, lets see if image_path is a suffix of
+    # latest_image_id.
+    elif [[ "${image_path}" == *"${latest_image_id}" ]]; then
+        echo "Image (${latest_image_id}) is up-to-date in configmap."
+        return 0
+    else
+        echo "Image mismatch: GCP image (${image_path}) different from ConfigMap (${latest_image_id})."
+        return 2
+    fi
```
gkurz, line 240: This `image_exists` function turns out to have non-trivial semantics, not just returning true or false. IMO it would be clearer to open-code it in its unique caller. This would spare the need for the exit status values.

gkurz: > gcp: adding image check before creating image
> As following other providers logic.
>
> Signed-off-by: Beraldo Leal <bleal@redhat.com>

The commit message would be a good place to describe the various cases that need to be handled BTW.

In the meantime, I went to #584 and I realized that you borrowed the convoluted `image_exists` pattern from there and it is already merged :disappointed: ... well, I guess I won't fight if everyone is apparently happy with that.

PR #593, opened 2025-03-28
webhook: Fix webhook creation

bpradipt:
Commit b6b0f93b53f3ebcebaf0f21113c9c89af0d5fadf updated to use CustomValidator. This also requires NewWebhookManagedBy to use WithValidator otherwise webhook will not start.

Kataconfig creation fails with 
```
Error "failed calling webhook "vkataconfig.kb.io": failed to call webhook: Post "[https://controller-manager-service.openshift-sandboxed-containers-operator.svc:443/validate-kataconfiguration-openshift-io-v1-kataconfig?timeout=10s](https://controller-manager-service.openshift-sandboxed-containers-operator.svc/validate-kataconfiguration-openshift-io-v1-kataconfig?timeout=10s)": dial tcp 10.128.2.23:9443: connect: connection refused" for field "undefined".
```
**- What I did**
Updated the NewWebhookManagedBy call to include `WithValidator`
**- How to verify it**
Kataconfig creation should work

Update the NewWebhookManagedBy call to include `WithValidator`

### DIFF
--- a/api/v1/kataconfig_webhook.go
+++ b/api/v1/kataconfig_webhook.go
@@ -38,8 +38,10 @@ var (
 func (r *KataConfig) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	clientInst = mgr.GetClient()
 
+	kataconfiglog.Info("SetupWebhookWithManager")
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
+		WithValidator(r).
 		Complete()
 }
 


littlejawa (approved):
/lgtm

Thanks @bpradipt - and sorry for introducing it :-(

bpradipt:
> /lgtm
> 
> 
> 
> Thanks @bpradipt - and sorry for introducing it :-( 

Np. I don't think there is anyway to know about this upfront unless we work with the same code frequently or CI catches it.
One thing is certain that we have to improve our upstream CI to catch these kinds of incompatibility issues.

gkurz:
> > /lgtm
> > Thanks @bpradipt - and sorry for introducing it :-(
> 
> Np. I don't think there is anyway to know about this upfront unless we work with the same code frequently or CI catches it. One thing is certain that we have to improve our upstream CI to catch these kinds of incompatibility issues.

We have absolutely no functional testing in upstream CI. 2 years old tech debt that was never considered a priority AFAICT.

api/v1/kataconfig_webhook.go
```diff
@@ -38,8 +38,10 @@ var (
 func (r *KataConfig) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	clientInst = mgr.GetClient()
 
+	kataconfiglog.Info("SetupWebhookWithManager")
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
+		WithValidator(r).
```
gkurz, line 44: How did you find out ? Can you share a link in the commit message for the sake of knowledge sharing ?

gkurz (approved):
Based on the example at https://github.com/kubernetes-sigs/controller-runtime/blob/v0.19.7/examples/builtins/main.go#L77

/lgtm

Thanks @bpradipt !

bpradipt:
> Based on the example at https://github.com/kubernetes-sigs/controller-runtime/blob/v0.19.7/examples/builtins/main.go#L77
> 
> 
> 

I will add this to the commit message. This is what I referred to for the solution. When kataconfig creation was failing due to webhook connection failure, I started looking at new commits, which then pointed me to controller-runtime upgrade. Usually I look at unit tests and examples to get an idea. 
The example pointed to the missing option in our code.

> /lgtm
> 
> 
> 
> Thanks @bpradipt !

bpradipt:
@gkurz @littlejawa I added the example in the commit message

PR #596, opened 2025-04-02
kata installation should not watch the 'worker' MCP state

pmores:
When the MCO migrates nodes between pools, the target pool enters the Updating state and has nodes gradually added to it as they are updated and rebooted while the source pool is drained by simply setting its machine count to the desired value at once and never leaves the Updated state.

Because the "worker" pool is the source pool in kata installation, it follows that nothing can be gained by watching its status.

Normally it doesn't hurt either but if "worker" is Updating already for some unrelated reason when kata installation is starting, it throws off the isInstallationInProgress detection and thus the whole installation flow.

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1267,7 +1267,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		r.setInProgressConditionToInstalling()
 	}
 
-	isInstallationInProgress := r.isMcpUpdating(machinePool) || (!isConvergedCluster && r.isMcpUpdating("worker"))
+	isInstallationInProgress := r.isMcpUpdating(machinePool)
 
 	// Create kata-oc MCP only if it's not a converged cluster
 	if !isConvergedCluster {
@@ -1318,11 +1318,6 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 	r.Log.Info("MCP updating state", "MCP name", machinePool, "is updating", isKataMcpUpdating)
 
 	isMcoUpdating := isKataMcpUpdating
-	if !isConvergedCluster {
-		isWorkerUpdating := r.isMcpUpdating("worker")
-		r.Log.Info("MCP updating state", "MCP name", "worker", "is updating", isWorkerUpdating)
-		isMcoUpdating = isKataMcpUpdating || isWorkerUpdating
-	}
 
 	if isMcoUpdating && r.getInProgressConditionValue() == corev1.ConditionFalse {
 		r.setInProgressConditionToUpdating()
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1267,8 +1267,6 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		r.setInProgressConditionToInstalling()
 	}
 
-	isInstallationInProgress := r.isMcpUpdating(machinePool)
-
 	// Create kata-oc MCP only if it's not a converged cluster
 	if !isConvergedCluster {
 		labelingChanged, err := r.updateNodeLabels()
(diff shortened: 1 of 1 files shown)


controllers/openshift_controller.go
```diff
@@ -1318,11 +1318,6 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 	r.Log.Info("MCP updating state", "MCP name", machinePool, "is updating", isKataMcpUpdating)
 
 	isMcoUpdating := isKataMcpUpdating
```
gkurz: This line ends up being the only user of `isKataMcpUpdating`. Some further simplification is possible :

```
@@ -1314,10 +1314,8 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
                }
        }
 
-       isKataMcpUpdating := r.isMcpUpdating(machinePool)
-       r.Log.Info("MCP updating state", "MCP name", machinePool, "is updating", isKataMcpUpdating)
-
-       isMcoUpdating := isKataMcpUpdating
+       isMcoUpdating := r.isMcpUpdating(machinePool)
+       r.Log.Info("MCP updating state", "MCP name", machinePool, "is updating", isMcoUpdating)
 
        if isMcoUpdating && r.getInProgressConditionValue() == corev1.ConditionFalse {
                r.setInProgressConditionToUpdating()
```

Either squashed here or pushed to a subsequent commit, as you prefer.

controllers/openshift_controller.go
```diff
@@ -1267,7 +1267,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		r.setInProgressConditionToInstalling()
 	}
 
-	isInstallationInProgress := r.isMcpUpdating(machinePool) || (!isConvergedCluster && r.isMcpUpdating("worker"))
+	isInstallationInProgress := r.isMcpUpdating(machinePool)
```
gkurz: Not strictly related to the goal of this PR but `isInstallationInProgress` doesn't look very useful actually. It is only used in one place in the non-converged cluster case, where the MCP is known to be `kata-oc`.

A subsequent commit or PR could maybe do :
```
@@ -1267,8 +1267,6 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
                r.setInProgressConditionToInstalling()
        }
 
-       isInstallationInProgress := r.isMcpUpdating(machinePool)
-
        // Create kata-oc MCP only if it's not a converged cluster
        if !isConvergedCluster {
                labelingChanged, err := r.updateNodeLabels()
@@ -1282,7 +1280,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
                if labelingChanged {
                        r.Log.Info("node labels updated")
 
-                       if !isInstallationInProgress {
+                       if !r.isMcpUpdating("kata-oc") {
                                r.Log.Info("Starting to wait for MCO to start")
                                r.kataConfig.Status.WaitingForMcoToStart = true
                        } else {
```

gkurz (approved):
The proposed change looks sane and will prevent the rest of the installation flow to be triggered because of external factors.

The extra simplifications that I suggest can be discussed and addressed by this PR but this isn't a strong requirement, so I'm approving right away.

Thanks @pmores !

controllers/openshift_controller.go
pmores: True, I'll keep only the `isMcoUpdating` name.

controllers/openshift_controller.go
pmores: I'd like to keep the `isInstallationInProgress` name since I feel it clarifies the actual meaning of the `r.isMcpUpdating(machinePool)` check which is specific to this context and quite possibly non-obvious to uninitiated.  I'd just move the variable definition right before its only user to make the connection rather clear.

gkurz (approved):
/lgtm

Thanks for the new changes @pmores !

bpradipt (approved):
/lgtm

PR #609, opened 2025-04-14
podvm: Restrict Azure image version is a 32-bit integer

bpradipt:
The Azure image version is of the form Major(int).Minor(int).Patch(int) format, where int is between 0 and 2,147,483,647 (inclusive). The previous code converted the cluster ID (which is in hex format) directly to an integer and used it as the major version number. However the converted value could be too large resulting in image creation failure. This change limits the major version to the max signed int value.  As for the Minor and Patch values, we use a fixed value of 0.1

Fixes: #[KATA-3741](https://issues.redhat.com//browse/KATA-3741)

### DIFF
--- a/controllers/image_generator.go
+++ b/controllers/image_generator.go
@@ -753,7 +753,12 @@ func (r *ImageGenerator) createImageConfigMapFromFile() error {
 	imageVersion := "0.0.1"
 	clusterIdInt, err := strconv.ParseUint(r.clusterId, 16, 32)
 	if err == nil {
-		imageVersion = fmt.Sprintf("%d.0.1", clusterIdInt)
+		// Ensure the version number is within Azure's allowed range (32-bit integer 0-2,147,483,647)
+		// Ref: https://learn.microsoft.com/en-us/rest/api/compute/gallery-image-versions/create-or-update
+		const maxVersionNumber = 2147483647
+		majorVersion := clusterIdInt % (maxVersionNumber + 1)
+		imageVersion = fmt.Sprintf("%d.0.1", majorVersion)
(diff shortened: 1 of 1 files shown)


wainersm (approved):
Hi @bpradipt , thanks for the quick fix!

wainersm:
Tom tested this fix in a real installation, but I've tested this fix with the following code:

```go
package main

import (
	"fmt"
	"strconv"
)

func main() {
	const maxVersionNumber = 2147483647
	//clusterId := "2532a10a-f9c0-463d-9b3f-4bac5b41ddc5"
	clusterId := "e50c4a4c-98a1-45c2-b0df-095acd8668af"

	// current
	clusterIdInt, err := strconv.ParseUint(clusterId[:8], 16, 32)
	if err != nil {
		fmt.Printf("%v", err)
	} else {
		fmt.Printf("Overflow? %t\n", clusterIdInt > maxVersionNumber)
		fmt.Printf("%d.0.1\n", clusterIdInt)
	}

	// Pradipta's fix
	majorVersion := clusterIdInt % (maxVersionNumber + 1)
	fmt.Printf("Overflow? %t\n", majorVersion > maxVersionNumber)
	fmt.Printf("%d.0.1\n", majorVersion)
}

```

When `clusterId := "2532a10a-f9c0-463d-9b3f-4bac5b41ddc5"`:

```
Overflow? false
624075018.0.1
Overflow? false
624075018.0.1
```
When `clusterId := "e50c4a4c-98a1-45c2-b0df-095acd8668af"`:

```
Overflow? true
3842787916.0.1
Overflow? false
1695304268.0.1
```

beraldoleal (approved):
LGTM, thanks.

bpradipt:
Thanks @wainersm @beraldoleal

PR #619, opened 2025-04-24
Enable hermetic build for operator

littlejawa:
Enable hermetic build for the controller manager.

**- What I did**
gomod dependencies are managed by cachi2/Hermeto, enabling it is pretty simple.
I had to work around the controller-gen problem (again). Taking inspiration from the way we do it in CPaaS, but trying to avoid using different Dockerfile/Makefile for local and CI builds.

I'm adding a new script that `make` will use in place of the regular controller-gen. This script uses "go run" to run the tool.
In the context of a local build, this results in downloading/building the file as usual.
In the context of a hermetic build, with the dependencies pre-fetched, it will build the tool based on the prefetched sources.

**- How to verify it**
For a local build: run `podman build .`
For CI: the Konflux task in this PR must pass.
You can look at the logs too:
- the "prefetch" step must pass and show that go dependencies were fetched
- the "build" step must have the line:
```
Build will be executed with network isolation
```

Fixes: [KATA-3796](https://issues.redhat.com//browse/KATA-3796)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,10 +17,6 @@ COPY controllers controllers/
 # See comments in the script itself for more details.
 COPY controller-gen bin/controller-gen-v0.17.2
 
-RUN go mod download
-# needed for docker build but not for local builds
-RUN go mod vendor
-
 RUN make build
 
 # Use OpenShift base image
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,10 @@ COPY api api/
 COPY config config/
 COPY controllers controllers/
 
+# Copy our controller-gen script to work around hermetic build issues
+# See comments in the script itself for more details.
+COPY controller-gen bin/controller-gen-v0.17.2
+
 RUN go mod download
 # needed for docker build but not for local builds
 RUN go mod vendor
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,9 +15,16 @@ COPY controllers controllers/
 
 # Copy our controller-gen script to work around hermetic build issues
 # See comments in the script itself for more details.
-COPY controller-gen bin/controller-gen-v0.17.2
+COPY controller-gen bin/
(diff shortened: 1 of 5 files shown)


beraldoleal (approved):
LGTM, thanks.

Dockerfile
```diff
@@ -13,6 +13,10 @@ COPY api api/
 COPY config config/
 COPY controllers controllers/
 
+# Copy our controller-gen script to work around hermetic build issues
+# See comments in the script itself for more details.
+COPY controller-gen bin/controller-gen-v0.17.2
```
c3d: When does that version number change? My understanding is that this needs to be kept consistent with go.mod, how do we document that? Also asking because controller-tools release v0.17.3 has been out for three weeks now.

c3d (commented):
For what it works, builds OK on macOS.

Also, I tried building with controller-tools v0.17.3, and it seems to work as well.

Dockerfile
littlejawa: Updating it is part of the work we do on each release cycle. It's been kind of a manual process up to now.

With Konflux, we are offered an automated dependency check with PRs that update our go.mod dependencies (that's from Mintmaker), so we might get it like that.
Updating it requires changing both go.mod and Makefile. And now this Dockerfile.

There's no clear documentation about it, maybe there is room for improvement...
At the very least, I can add a comment here.

littlejawa: @c3d - I have added a commit (affae46) that tries to remove the need to update all those files with controller-gen version.
Is it overkill?

It can build both on my laptop and in Konflux apparently.

c3d: I don't think it's overkill. I like it, as it encapsulates some knowledge in automation. So good for me.

c3d (approved):
Looks good to me.

PR #625, opened 2025-04-30
Enable hermetic build for podvm builder

littlejawa:
This PR enables hermetic builds for the podvm-builder.

There are two things to take into account here:
- pre-fetch RPMs, including some that are accessible only through a subscription
- get the source code for cloud-api-adaptor without relying on "git clone"

I'm also trying to avoid having separate Makefiles/Dockerfiles to maintain for CI and local builds.

Fixes: [KATA-3800](https://issues.redhat.com//browse/KATA-3800)

### DIFF
--- a/config/peerpods/podvm/redhat.repo
+++ b/config/peerpods/podvm/redhat.repo
@@ -1,0 +1,8942 @@
+#
+# Certificate-Based Repositories
+# Managed by (rhsm) subscription-manager
+#
+# *** This file is auto-generated.  Changes made here will be over-written. ***
+# *** Use "subscription-manager repo-override --help" if you wish to make changes. ***
+#
+# If this file is empty and this system is subscribed consider
+# a "yum repolist" to refresh available repos
+#
+
+[satellite-6.16-for-rhel-9-$basearch-rpms]
+name = Red Hat Satellite 6.16 for RHEL 9 $basearch (RPMs)
+baseurl = https://cdn.redhat.com/content/dist/layered/rhel9/$basearch/satellite/6.16/os
+enabled = 0
+gpgcheck = 1
+gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
+sslverify = 1
+sslcacert = /etc/rhsm/ca/redhat-uep.pem
+sslclientkey = /etc/pki/entitlement/1575057137318568460-key.pem
+sslclientcert = /etc/pki/entitlement/1575057137318568460.pem
+sslverifystatus = 1
+metadata_expire = 86400
+enabled_metadata = 0
+
+[ossm-3-for-rhel-9-$basearch-source-rpms]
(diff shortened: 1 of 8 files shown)


config/peerpods/podvm/rpms.in.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - qemu-img
+  - podman
+
+contentOrigin:
+  repofiles:
+    - ./ubi.repo
+    - ./redhat.repo
+
+arches:
+  - x86_64
```
beraldoleal, line 17: what about s390x?

beraldoleal (commented):
LGTM, just a question about s390x.

config/peerpods/podvm/redhat.repo
```diff
@@ -0,0 +1,6058 @@
+#
+# Certificate-Based Repositories
+# Managed by (rhsm) subscription-manager
+#
+# *** This file is auto-generated.  Changes made here will be over-written. ***
```
c3d, line 5: Two questions here:

1. If the file is auto-generated, is there any way to generate it as part of the build? Probably not if it's based on `subscription-manager`, but asking just in case.
2. It looks like this overshot by a mile or two. Do we really need JBoss or RHEL AI repositories for the podvm builder?

If we don't need that many repositories, is there an inexpensive way to prune it and trim it down a little?

This seems particularly true given that I don't find `layered` once in the `rpms.lock.yaml` file, and that seems to be in the URLs for many of these repos. So that seems to indicate we don't actually use them.

c3d (approved):
The `redhat.repo` file looks a bit "noisy" and "overweight" to me. But I assume this is just because is was easier to generate it automatically than to optimize it by hand.

Other than that, looks good to me.

config/peerpods/podvm/redhat.repo
littlejawa: This is a copy of the repo file from the base image we build from, we've just edited it to have modifiable arch references. 
We probably could truncate it, but here I just followed the documentation from https://konflux.pages.redhat.com/docs/users/building/activation-keys-subscription.html#hermetic-network-isolated-builds
Note that it ends with:
```
Finally, commit the rpms.in.yaml, rpms.lock.yaml and redhat.repo to source control.
Konflux will use these files to prefetch RPMs for hermetic builds.
```

I think we could keep only the repos we enable. But cleaning it everytime we need to update it seems cumbersome :-/

littlejawa: Correction: the file is actually generated by the 'subscription-manager register' command in the base image. And THEN we copy it in our repo.

config/peerpods/podvm/rpms.in.yaml
littlejawa: It took me some time, but I've just added them.
Note that s390x builds are not enabled yet - that why we didn't have an error. But I have added the RPMs in the lockfile so that when we enable the builds, "it should just work" (famous last words)

littlejawa:
Closing - I'm going to use submodules instead. See #678

PR #628, opened 2025-05-02
konflux: avoid building the operator when manager.yaml is updated

littlejawa:
We get automated PRs to update the digest of the images in our ClusterServiceVersion and in the manager.yaml file. These PRs are generated when one of the referenced images is updated. Now if modifying those files triggers another build, we can end up in a loop of builds/nudges.
This commit makes sure that that operator is not rebuilt when such PRs are merged.

An example PR where the operator is rebuilt can be seen in #627


littlejawa:
NOTE: I have reformatted the list to be on multiple lines for clarity, but the only addition is the last line.

littlejawa:
@spotlesstofu - FYI

spotlesstofu (approved):

PR #651, opened 2025-05-21
Image changes for 1.10

snir911:
**Each commit msg should be descriptive enough**

What this includes:
1. use prebuilt oci image (aiming to the dmverity qcow in it) for all providers (both CoCo & PP)
2. ~~support singed images for Azure~~
3. automate the AWS oci upload process completely 

**When oci image is available we should set RELATED_IMAGE_PODVM_OCI value and csv to be updated**

To test this it requires to update:
1. podvm-builder image
2. have a wrapped oci with a valid signed image
3. controller image and RELATED_* pointing correctly to the above images
alternately have an bundle & catalog that has it all embedded 

** reach me for prebuilt version **  

TODOs:
1. verify deletion
2. handle SecureBoot enablement in FG
3. verify in all providers CoCo + PeerPods (AWS, Azure, GCP, Libvirt? IBM?)


config/manager/manager.yaml
```diff
@@ -87,6 +87,8 @@ spec:
               value: quay.io/redhat-user-workloads/ose-osc-tenant/osc-podvm-builder@sha256:a35214a970ce7ff2729adbdad5aa7caf202bf131fafa702dd1b006f102779f07
             - name: RELATED_IMAGE_PODVM_PAYLOAD
               value: quay.io/redhat-user-workloads/ose-osc-tenant/osc-podvm-payload@sha256:c282c01cb7b7d9984c5b5a4a2b434a9dff4b5b63c125bc2ae129d818696c0e29
+            - name: RELATED_IMAGE_PODVM_COCO_URI # VALID? this is URI not an image!
```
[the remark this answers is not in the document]
bpradipt: s/RELATED_IMAGE_PODVM_COCO_URI/RELATED_IMAGE_PODVM_IMAGE_URI

config/peerpods/podvm/azure-podvm-image-cm.yaml
```diff
@@ -65,3 +65,6 @@ data:
...
...iq8obn+TZ+OZ/fXbsG4dlBMgttPK7KGtrd/95oRiLESQ3IHfqm/HXdhiXtMXlbpVBgMmW3zSeA0iBsZiO5n9240ljgY3QebtF7t3WlwBzVQaE1mro0+QUAIOih0ua5guVzGnQ1QFCQgchoPdd01zqUlXLTEIPJop4AQn9y3jV0RjE53d7c5l/F2Oj+HAAiUEeJ3EiFrqK2z8LUqSkLIlw42Ixp57B/O0/yTWyLzhM4AoHH/mhNrIbz9Dc3KohWn8+iiced0jqEnxFw4fxPfUOvNk6ZvFjz4N+WpCDWCrsRzOuS0gJd7pNRGGmJVyc4mk7FbtzgXgI/wBvLOuGhpGa+zlI5cnoLBqMxNN/wHTMkBRkHlBKECAwEAAaNTMFEwHQYDVR0OBBYEFMdJlU45cIgremkk+qeAUpk2b2pPMB8GA1UdIwQYMBaAFMdJlU45cIgremkk+qeAUpk2b2pPMA8GA1UdEwEB/wQFMAMBAf8wDQYJKoZIhvcNAQELBQADggIBAA18dX5s4uRsYDZVaglMWjRbCReoRLHy+HDIgz3pp/THpYNDm3ItQDyN7OgFhcfjoHhiA6OwD3BmF20KUD13HZNc01KgWSNLghZZKDy9pcIMM/vM+5WKwbP9QLbQg5MobUmW9nuWDQJ3/3MYzy754jEvzBPipi+d/AA1ml83TmnXDXTE00/ml5yis1t7lqZdNcqlfVxzmyn8E5t0Wlz7UqE7D6gykst+mQhIjs/C4WWuYdGEIzZQVALnSj3iexnJSZZ0S1szFuk3Tx3rqbxlutOipNg79kiREKnaP3L3++IkIY2oOc8gY2LKWSbqLPqtBwq+HrWEfu6WGJAKOUtEBvdOGmQR/di4u7qY+NG8xuGSdWiKLzeUDd6+7ipjWv4Bj9WJq1LLdqCZ8rpwzYKmXorp5w9h75YZLa2zPkP2WeDllEvOuRDmH2upvDGqlxsc2HMAR/2ux2OuzxsJWABx6Zf93xkdkjBDxEWAgEmnsRkjlDNkIhXbf3xI2zyKS4INpBHbQs41vKXJvr15FZ5DsWz0Rus1P8/IYfYYQFQofwUHbsn79vfqvNinSbdHA4K+gtoT+SKSMPKqw2aJ27z46ku3CQFajibY81E+fAq0mzbjnRZqa2gpsznEbQElEooevJrf0UVrDOVWyQSC1PPeMXgX51Ue/cGfw1Z33FEUIHhR"
```
bpradipt: What's the source of the key ?

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -445,6 +445,88 @@ function create_azure_image_from_scratch() {
     echo "Azure image created successfully from scratch"
 }
 
+function create_signed_image_version()
+{
+    echo "Creating Azure signed image version"
+    # UEFI_SIGNATURE_DATABASE_KEY is required to be set
+    [[ -n "${UEFI_SIGNATURE_DATABASE_KEY}" ]] && local UEFI_SDB="${UEFI_SIGNATURE_DATABASE_KEY}" || error_exit "UEFI_SIGNATURE_DATABASE_KEY is empty"
+    AZURE_DEPLOYMENT_TEMPLATE=/tmp/azure-deployment-template.json
```
bpradipt: if its only local lets use small letters to distinguish from global.
eg. local deployment_template = /tmp/azure-deployment-template.json

config/peerpods/podvm/azure-podvm-image-cm.yaml
esposem: Maybe this should be left empty. At the moment we have a locally generated key used for testing. We don't know which key process we will use, and we might use RHEL official public certificate, which *should* be available already in the clouds that run RHEL instances.
And as future plan the user should be able to add its own certificate, so this should be customizable.
So maybe this should be set as an optional configmap that the user can provide, and depending on which certificate we use default to a osc default cert configmap.

esposem:
LGTM but it's unlikely that is the end cert, and I am not sure why you add it in config/manager/manager.yaml and config/peerpods/podvm/azure-podvm-image-cm.yaml

config/peerpods/podvm/azure-podvm-image-cm.yaml
```diff
@@ -65,3 +65,6 @@ data:
...
```
beraldoleal: I noticed this cert is expired. Is this just an example? Should we improve the comments then?

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -443,6 +443,88 @@ function create_azure_image_from_scratch() {
     echo "Azure image created successfully from scratch"
 }
 
+function create_signed_image_version()
```
beraldoleal: IIUC, this method isn't signing the image, but "just" registering a pre-signed one.

If that's correct, maybe consider renaming the method and adjusting the comments/debug messages to reflect that more clearly, something like `register_signed_image_version`?

config/manager/manager.yaml
```diff
@@ -87,6 +87,10 @@ spec:
...
...iq8obn+TZ+OZ/fXbsG4dlBMgttPK7KGtrd/95oRiLESQ3IHfqm/HXdhiXtMXlbpVBgMmW3zSeA0iBsZiO5n9240ljgY3QebtF7t3WlwBzVQaE1mro0+QUAIOih0ua5guVzGnQ1QFCQgchoPdd01zqUlXLTEIPJop4AQn9y3jV0RjE53d7c5l/F2Oj+HAAiUEeJ3EiFrqK2z8LUqSkLIlw42Ixp57B/O0/yTWyLzhM4AoHH/mhNrIbz9Dc3KohWn8+iiced0jqEnxFw4fxPfUOvNk6ZvFjz4N+WpCDWCrsRzOuS0gJd7pNRGGmJVyc4mk7FbtzgXgI/wBvLOuGhpGa+zlI5cnoLBqMxNN/wHTMkBRkHlBKECAwEAAaNTMFEwHQYDVR0OBBYEFMdJlU45cIgremkk+qeAUpk2b2pPMB8GA1UdIwQYMBaAFMdJlU45cIgremkk+qeAUpk2b2pPMA8GA1UdEwEB/wQFMAMBAf8wDQYJKoZIhvcNAQELBQADggIBAA18dX5s4uRsYDZVaglMWjRbCReoRLHy+HDIgz3pp/THpYNDm3ItQDyN7OgFhcfjoHhiA6OwD3BmF20KUD13HZNc01KgWSNLghZZKDy9pcIMM/vM+5WKwbP9QLbQg5MobUmW9nuWDQJ3/3MYzy754jEvzBPipi+d/AA1ml83TmnXDXTE00/ml5yis1t7lqZdNcqlfVxzmyn8E5t0Wlz7UqE7D6gykst+mQhIjs/C4WWuYdGEIzZQVALnSj3iexnJSZZ0S1szFuk3Tx3rqbxlutOipNg79kiREKnaP3L3++IkIY2oOc8gY2LKWSbqLPqtBwq+HrWEfu6WGJAKOUtEBvdOGmQR/di4u7qY+NG8xuGSdWiKLzeUDd6+7ipjWv4Bj9WJq1LLdqCZ8rpwzYKmXorp5w9h75YZLa2zPkP2WeDllEvOuRDmH2upvDGqlxsc2HMAR/2ux2OuzxsJWABx6Zf93xkdkjBDxEWAgEmnsRkjlDNkIhXbf3xI2zyKS4INpBHbQs41vKXJvr15FZ5DsWz0Rus1P8/IYfYYQFQofwUHbsn79vfqvNinSbdHA4K+gtoT+SKSMPKqw2aJ27z46ku3CQFajibY81E+fAq0mzbjnRZqa2gpsznEbQElEooevJrf0UVrDOVWyQSC1PPeMXgX51Ue/cGfw1Z33FEUIHhR"
```
beraldoleal: Same concern here, iiuc, this cert is expired.

config/peerpods/podvm/ami-helper.sh
```diff
@@ -0,0 +1,303 @@
+#!/usr/bin/env bash
```
beraldoleal, line 1: IIUC, this seems to be non dependent/related to the signature commits. Unless I missed something, I would recommend to break into a different PR.

config/peerpods/podvm/extra_rbac.go
```diff
@@ -0,0 +1,4 @@
+package rbac
```
beraldoleal: IIUC, this could be squashed with the previous commit.

beraldoleal (commented):
Hi @snir911, thanks for this.


I left a few comments for you.

gkurz:
General comment before I go deeper : please put Signed-off-by tags in all commits @snir911 :wink:

controllers/image_generator.go
```diff
@@ -414,6 +414,7 @@ func (r *ImageGenerator) getPeerPodsCM() (*corev1.ConfigMap, error) {
 // azure-podvm-image-cm.yaml for Azure
 // aws-podvm-image-cm.yaml for AWS
 // libvirt-podvm-image-cm.yaml for Libvirt
+// gcp-podvm-image-cm.yaml for GCP
```
gkurz: There are quite a bunch of places that are missing such changes for GCP and libvirt as well (see `git grep podvm-image-cm`). I suggest you drop this commit and we create a separate PR for these trivial cleanups. Let's keep the current PR focused on the important changes :wink:

config/peerpods/podvm/azure-podvm-image-cm.yaml
gkurz: Does it make sense to provide a signature in the first place ?

I would make it
```suggestion
# UEFI Allowed Signature DB Key
#UEFI_SIGNATURE_DATABASE_KEY: "" # set to DER+base64 encoded key
```

config/manager/manager.yaml
snir911: it isn't , it's used for Azure, at the previous commented one my intention was to have it as example (and a place to keep it as this is the key we'll be using for now), i can remove the other one.

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -443,6 +443,88 @@ function create_azure_image_from_scratch() {
...
+{
+    echo "Creating Azure signed image version"
+    # UEFI_SIGNATURE_DATABASE_KEY is required to be set
+    [[ -n "${UEFI_SIGNATURE_DATABASE_KEY}" ]] && local uefi_sdb="${UEFI_SIGNATURE_DATABASE_KEY}" || error_exit "UEFI_SIGNATURE_DATABASE_KEY is empty"
```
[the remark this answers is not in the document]
gkurz: I agree with copilot but I'd make it

```suggestion
    local uefi_sdb="${UEFI_SIGNATURE_DATABASE_KEY}"

    if [[ -z "${uefi_sdb} ]]; then
        error_exit "UEFI_SIGNATURE_DATABASE_KEY is empty"
    fi
```

config/peerpods/podvm/azure-podvm-image-handler.sh
gkurz: I think that the `create` wording relates to that fact that this is creating an image version in Azure.

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -443,6 +443,88 @@ function create_azure_image_from_scratch() {
...
+        --resource-group "${AZURE_RESOURCE_GROUP}" \
+        --gallery-name "${IMAGE_GALLERY_NAME}" \
+        --gallery-image-definition "${IMAGE_DEFINITION_NAME}" \
+        --gallery-image-version "${IMAGE_VERSION}" \
+        --os-vhd-uri "${VHD_URL}" \
+        --os-vhd-storage-account "${STORAGE_ACCOUNT_NAME}" \
+        --target-regions "${AZURE_REGION}" ||
+        error_exit "Failed to create the image version"
+
+    echo "Azure image version created successfully"
```
gkurz: Maybe specify `unsigned image` for extra clarity ?

config/manager/manager.yaml
```diff
@@ -87,6 +87,10 @@ spec:
               value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-builder-rhel9@sha256:d635d2e688982986f6942b19948e0a874b093957045335c1009f0765aa9bd42e
             - name: RELATED_IMAGE_PODVM_PAYLOAD
               value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9@sha256:cacdcb85e4f421b394765fd4632e6accf53ab6dcf78afd285a1a56bc863dc714
+            - name: RELATED_IMAGE_PODVM_OCI
+              value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-oci-rhel9:1.10.0
```
gkurz: It looks weird that this image has a tag when all the other ones have digests. This seems to be related to konflux. Any insight @littlejawa ?

config/manager/manager.yaml
gkurz: @snir911 it is unclear why we're hard-coding this key... can you provide the full story ?

controllers/image_generator.go
```diff
@@ -834,12 +834,23 @@ func (r *ImageGenerator) updateImageConfigMap() error {
...
 
-		if err := r.client.Update(context.TODO(), cm); err != nil {
-			return err
+		if cm.Data["UEFI_SIGNATURE_DATABASE_KEY"] == "" && os.Getenv("OSC_SIGNATURE_DATABASE_KEY") != "" {
+			cm.Data["UEFI_SIGNATURE_DATABASE_KEY"] = os.Getenv("OSC_SIGNATURE_DATABASE_KEY")
+			igLogger.Info("Setting UEFI_SIGNATURE_DATABASE_KEY to default value", "UEFI_SIGNATURE_DATABASE_KEY", os.Getenv("OSC_SIGNATURE_DATABASE_KEY"))
 		}
 	}
 
+	if err := r.client.Update(context.TODO(), cm); err != nil {
```
gkurz, line 844: And now we can potentially do no-op updates. Not ideal but fortunately this is tolerated by k8s.

config/manager/manager.yaml
littlejawa: Images needs to be referenced by digest, so that they can be updated automatically by Konflux whenever they are rebuilt.
I think that this image doesn't exist yet, so we don't have a digest. I guess this is a placeholder then?
@snir911 - do you confirm?

config/peerpods/podvm/azure-podvm-image-handler.sh
snir911: Yes, this create refer to creating the image version, it's based on the previous function name if this is confusing i can change

config/peerpods/podvm/ami-helper.sh
snir911: It is related as now the helper script is called from within the podvm creation script, hence, some changes were required

config/peerpods/podvm/azure-podvm-image-cm.yaml
snir911: as mentioned, I can remove this from here..

config/manager/manager.yaml
snir911: Right, it should be pointing to the image @littlejawa is working on

config/manager/manager.yaml
snir911: dropped to changes in plan:)

Makefile
```diff
@@ -104,11 +104,11 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="{./api/v1,./cmd/manager,./cmd/metrics,./controllers}" output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="{./api/v1,./cmd/manager,./cmd/metrics,./controllers,./config/peerpods/podvm}" output:crd:artifacts:config=config/crd/bases
```
gkurz: For the records : as discussed in slack this would defeat https://github.com/openshift/sandboxed-containers-operator/pull/678/commits/bc47b2c5e13eba1d8c46a435d2e12a60eb7bdfa4#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R107 . The `+kubebuilder` annotation in `config/peerpods/podvm/extra_rbac.go` should be added somewhere else.

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -207,9 +207,28 @@ function bootc_to_ami() {
...
+function prepare_for_prebuilt_artifact() {
+    echo "Preparing infrastructure for prebuilt artifact"
+
+    [[ ! ${BUCKET_NAME} ]] && export BUCKET_NAME=${AMI_BASE_NAME}-${AMI_VERSION}
+
+    # Update the BUCKET_NAME in the peer-pods-cm configmap
+    kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator --type merge -p "{\"data\":{\"BUCKET_NAME\":\"${BUCKET_NAME}\"}}"
+
+    echo "Calling the helper script to extend credentials, create vmimport role and a bucket"
+    ./scripts/ami-helper.sh -i -b ${BUCKET_NAME} || error_exit "Failed to extend credentials, create vmimport role and a bucket using the helper script"
```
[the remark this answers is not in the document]
gkurz: Copilot's suggestion makes sense but you also need to drop `/scripts/` from the path.

controllers/credentials_controller.go
```diff
@@ -65,6 +65,9 @@ const (
 
 // TODO: reduce secret's RBAC if possible
 
+// the following is not required by this controller, it's required by the AWS podvm creation scripts (ami-helper.sh)
+//+kubebuilder:rbac:groups=cloudcredential.openshift.io,resources=credentialsrequests,verbs=create;delete;get;list
```
gkurz, line 69: Since we already have the following at line 64....

```
//+kubebuilder:rbac:groups=cloudcredential.openshift.io,resources=credentialsrequests,verbs=create;delete
```

... cannot we just add the `get` and `list` verbs ?

snir911: of course we can, it just for the logical splitting and the comment

config/peerpods/podvm/aws-podvm-image-handler.sh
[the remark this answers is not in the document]
gkurz: > Copilot's suggestion makes sense but you also need to drop `/scripts/` from the path.

Hmm... this is supposed to be called inside a container where all scripts were copied to `/scripts/` by https://github.com/openshift/sandboxed-containers-operator/pull/651/commits/74ca8827f212de33aafaca407b26eff6592bbad2#diff-d8022b161cdf836180dbb864cd51b3cbaacd0e143609a52b9d95fb70a9b84271R27, right ?

If so, please call this with an absolute path `/scripts/ami-helper.sh` for consistency with the rest of the scriptery.

```suggestion
    /scripts/ami-helper.sh -i -b ${BUCKET_NAME} || error_exit "Failed to extend credentials, create vmimport role and a bucket using the helper script"
```

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -251,13 +270,19 @@ function create_ami_from_prebuilt_artifact() {
         error_exit "Currently only OCI image unpacking is supported, exiting."
         ;;
     esac
+
+    echo "Deleting the bucket"
+    ./scripts/ami-helper.sh -i -d -b ${BUCKET_NAME} || error_exit "Failed to delete the bucket using the helper script"
```
gkurz:
```suggestion
    /scripts/ami-helper.sh -i -d -b ${BUCKET_NAME} || error_exit "Failed to delete the bucket using the helper script"
```

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -251,13 +270,19 @@ function create_ami_from_prebuilt_artifact() {
...
+
+    echo "Cleaning the credentials"
+    ./scripts/ami-helper.sh -i -c || error_exit "Failed to clean the credentials using the helper script"
```
gkurz:
```suggestion
    /scripts/ami-helper.sh -i -c || error_exit "Failed to clean the credentials using the helper script"
```

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -255,6 +255,8 @@ spec:
           verbs:
           - create
           - delete
+          - get
+          - list
```
gkurz, line 259: This change belongs to the previous patch where `config/rbac/role.yaml` was modified or update the bundle globally in a subsequent commit.

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -207,9 +207,28 @@ function bootc_to_ami() {
     bootc_image_builder_conversion "${container_image_repo_url}" "${image_tag}" "${auth_json_file}" "${run_args}" "${bib_args}"
 }
 
+function prepare_for_prebuilt_artifact() {
+    echo "Preparing infrastructure for prebuilt artifact"
+
+    [[ ! ${BUCKET_NAME} ]] && export BUCKET_NAME=${AMI_BASE_NAME}-${AMI_VERSION}
+
+    # Update the BUCKET_NAME in the peer-pods-cm configmap
+    kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator --type merge -p "{\"data\":{\"BUCKET_NAME\":\"${BUCKET_NAME}\"}}"
```
bpradipt: Should the bucket name be in `peer-pods-cm` or `azure-podvm-image-cm` configMap ?

snir911: good catch, yes, at this point it's better to be in the aws-podvm-image-cm

bpradipt: Yeah aws-podvm-image-cm ;) ..

gkurz (approved):
/lgtm

Thanks @snir911 !

bpradipt (approved):
/lgtm

PR #655, opened 2025-05-22
fbc: deprecate v1.6.0 and older

spotlesstofu:
deprecate all channels other than `stable` and all bundles older than `v1.7.0`.

don't remove the entries completely, to allow users to upgrade from unsupported versions

also update scripts and move the icon image to a separate file


spotlesstofu:
after removing the big icon base64 from the template files, i was able to use an AI assistant to generate the subsequent changes

littlejawa (approved):
/lgtm

PR #659, opened 2025-05-26
konflux: operator build triggers

littlejawa:
Modify the on-cel-expressions to fix triggering issues.
We need to prevent the operator from building on changes to the bundle, as otherwise we have a loop of builds/updates.

### DIFF
--- a/.tekton/osc-operator-pull-request.yaml
+++ b/.tekton/osc-operator-pull-request.yaml
@@ -10,11 +10,11 @@ metadata:
     pipelinesascode.tekton.dev/on-cel-expression: |
       event == "pull_request" &&
       target_branch == "devel" &&
-      files.all.exists(path, !path.matches('bundle/.*|.tekton/.*bundle.*yaml$')) &&
-      files.all.exists(path, !path.matches('must-gather/.*|.tekton/.*must-gather.*yaml$')) &&
-      files.all.exists(path, !path.matches('config/peerpods/podvm/.*|.tekton/.*podvm-builder.*yaml$')) &&
-      files.all.exists(path, !path.matches('.tekton/.*fbc-.*|fbc/.*')) &&
-      files.all.exists(path, !path.matches('config/manager/.*'))
+      !files.all.exists(path, path.matches('bundle/.*|.tekton/.*bundle.*yaml$')) &&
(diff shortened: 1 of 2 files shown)


littlejawa:
Example of a PR that triggers a rebuild of the operator because of the config/manager/manager.yaml change:
https://github.com/openshift/sandboxed-containers-operator/pull/654

.tekton/osc-operator-pull-request.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
       files.all.exists(path, !path.matches('must-gather/.*|.tekton/.*must-gather.*yaml$')) &&
       files.all.exists(path, !path.matches('config/peerpods/podvm/.*|.tekton/.*podvm-builder.*yaml$')) &&
       files.all.exists(path, !path.matches('.tekton/.*fbc-.*|fbc/.*')) &&
-      files.all.exists(path, !path.matches('config/manager/.*'))
+      files.all.exists(path, !path.matches('config/manager/manager.yaml'))
```
spotlesstofu: What about moving the `!` like below?

```suggestion
      !files.all.exists(path, path.matches('config/manager/.*'))
```

This should provide more predictable outputs. I tested it in [CEL Playground](https://playcel.undistro.io/).

spotlesstofu:
The enterprise contract fails because it now checks against the release policy, I merged that earlier today.
We can address the violations in a separate pull request.

.tekton/osc-operator-pull-request.yaml
littlejawa: It's certainly worth a try :)
Now I'm wondering if I should do the same to the other filters? They *seem* to work, but would that be more reliable if I move the `!` there too?

littlejawa:
> The enterprise contract fails because it now checks against the release policy, I merged that earlier today. We can address the violations in a separate pull request.

Right. I started looking into it - seems like I need to add some labels everywhere :-)

.tekton/osc-operator-pull-request.yaml
spotlesstofu: yes, I'd do the same on all of them

spotlesstofu (approved):

PR #661, opened 2025-05-27
konflux: fix enterprise contract violations

littlejawa:
- applying labels to our image (copied from our CPaaS builds)
- build the source image for every container

This is required to fix violations found by the enterprise contract job in Konflux

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -19,3 +19,23 @@ LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
 COPY bundle/manifests /manifests/
 COPY bundle/metadata /metadata/
 COPY bundle/tests/scorecard /tests/scorecard/
+
+# Red Hat labels
+LABEL io.k8s.display-name='OpenShift sandboxed containers operator'
+LABEL io.k8s.description='This operator manages the sandboxed-containers runtime'
+LABEL com.redhat.delivery.appregistry=''
+LABEL maintainer='support@redhat.com'
+LABEL name="openshift/sandboxed-containers-operator-bundle"
+LABEL com.redhat.component="osc-operator-bundle-container"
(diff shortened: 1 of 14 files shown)


bundle.Dockerfile
```diff
@@ -19,3 +19,26 @@ LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
...
+LABEL io.k8s.description='This operator manages the sandboxed-containers runtime'
+LABEL com.redhat.delivery.appregistry=''
+LABEL maintainer='support@redhat.com'
+LABEL name="openshift/sandboxed-containers-operator-bundle"
+LABEL com.redhat.component="osc-operator-bundle-container"
+LABEL io.openshift.maintainer.product='OpenShift Container Platform'
+LABEL io.openshift.maintainer.component='Sandboxed Containers'
+LABEL io.openshift.build.commit.id=${CI_UPSTREAM_COMMIT}
+LABEL io.openshift.build.source-location=${CI_UPSTREAM_URL}
+LABEL io.openshift.build.commit.url=${CI_UPSTREAM_URL}/commit/${CI_UPSTREAM_COMMIT}
```
spotlesstofu: these variables won't work in Konflux

spotlesstofu: we don't have them in Trustee, so you could remove them altogether

littlejawa: Sure :-( I replaced the "CI_VERSION" one everywhere, but missed those.

Now I'm wondering if there's a way to do it with dynamic values?
The best I've found is this: https://konflux.pages.redhat.com/docs/users/building/labels-and-annotations.html
Maybe I will try to do something like that.

Do you see a problem if we set the labels in the .tekton files rather than in the Dockerfile?

littlejawa: After discussion on slack, and comparing with what we do in trustee, I've removed those labels.

PR #662, opened 2025-05-27
podvm: use pause bundle from payload image

snir911:
As pause bundle is now included in the payload image as pause-bundle.tar.xz also rename the function to better reflect its current behavior of just extracting the bundle from the payload.

This change depend on https://github.com/openshift/cloud-api-adaptor/pull/59 and need to be tested together


bpradipt (approved):
/lgtm

PR #666, opened 2025-06-03
konflux: update bundle image references

littlejawa:
We are not getting updates for the bundle anymore from Konflux since we moved to registry.redhat.io.

Found that the repository was not the only thing to change: the name of the images themselves is different.
It needs to match what we have in the ReleasePlanAdmission.

Hopefully this will restore bundle updates.


littlejawa:
@spotlesstofu - did you do anything more to get bundle updates on the Trustee side?

spotlesstofu:
Not that I remember of. I'd ask someone who can look at the logs for the nudges bot.

PR #667, opened 2025-06-03
tekton: use remote fbc pipeline

spotlesstofu:
Enables easier maintenance, as long as we're sticking to the default pipeline definition.

Fixes failing enterprise contract because of missing tasks.


spotlesstofu:
@littlejawa could you please approve?

Since I also merged the FBC release configuration, we should see automatic stage releases after the merge.

littlejawa (approved):
/lgtm

Thanks @spotlesstofu

PR #675, opened 2025-06-05
tests: add mirrors for konflux images

littlejawa:
Konflux images are located in quay.io/redhat-user-workloads/ose-osc-tenant/[...] but our bundle references images from registry.redhat.io.

To be able to test our bundle with unreleased images, we need to set some mirrors in the test cluster to check the images from quay instead of registry.redhat.io.

This yaml provides the required mirrors, and needs to be applied to the test cluster.

### DIFF
--- a/.tekton/images-mirror-set.yaml
+++ b/.tekton/images-mirror-set.yaml
@@ -1,0 +1,61 @@
+apiVersion: config.openshift.io/v1
+kind: ImageTagMirrorSet
+metadata:
+  name: osc-registry
+spec:
+  imageTagMirrors:
+    - mirrors:
+      - quay.io/redhat-user-workloads/ose-osc-tenant/osc-monitor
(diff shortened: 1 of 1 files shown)


beraldoleal (approved):

tests/mirrors.yaml
gkurz, line 1: The build pipeline in konflux also needs some mirroring, at least the `ImageDigestMirrorSet` as mentioned in https://github.com/openshift/sandboxed-containers-operator/pull/676#discussion_r2140245503 . Maybe you can move this file to `.tekton/images-mirror-set.yaml` and document its use for tests ?

littlejawa: Moved/renamed it.
For the documentation, we're describing it in our team's documentation, not in this repo, so I'll do it separately.

gkurz (approved):
/lgtm

Thanks @littlejawa ! Let's hope it will do the job for the konflux pipelines as well.

PR #676, opened 2025-06-05
FBC update

spotlesstofu:
Update the FBC to the latest bundle.


fbc/v4.19/catalog/sandboxed-containers-operator/catalog.json
```diff
@@ -1624,7 +1624,7 @@
     "schema": "olm.bundle",
     "name": "sandboxed-containers-operator.v1.9.0",
     "package": "sandboxed-containers-operator",
-    "image": "registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:a37f38455aff326ef87ba4cef4c4c6268351f665265321f39d9e033de819cd69",
+    "image": "registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:1a0d9fc19062717b8a7779133d85584e271038211eddd4ddac6619b7ab2c8481",
```
littlejawa: This is replacing the 1.9.0 bundle with the latest one (which contains 1.10 code).
Is that right? Don't we want to add a 1.10 entry?

spotlesstofu: we do, and that has to change in the CSV first

littlejawa: Working on it: https://github.com/openshift/sandboxed-containers-operator/pull/677

fbc/update.sh
gkurz, line 1: It would be great to provide some details on this script in the commit message. Things like what it is useful for, how/when it should be used and any other interesting thing you can think of :smiley_cat: 

EDIT: maybe rename it to `update-digests.sh` for improved clarity.

fbc/update.sh
```diff
@@ -0,0 +1,67 @@
...
+# Check that a folder exists for the versions you set.
+ls $VERSIONS > /dev/null
+
+# Make script fail when API requests fail.
+alias curl="curl --fail"
+
+# Get the latest merge commit for the bundle. This is the string the bundle image is tagged with.
+get_tag () {
+    URL="https://api.github.com/repos/openshift/sandboxed-containers-operator/commits?per_page=1&path=bundle"
+    COMMIT=$(curl "$URL" | jq -r '.[0].sha')
```
gkurz: Please mention `curl` and `jq`  in https://github.com/openshift/sandboxed-containers-operator/blob/devel/fbc/README.md.

fbc/update.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# Check that a folder exists for the versions you set.
+ls $VERSIONS > /dev/null
+
+# Make script fail when API requests fail.
+alias curl="curl --fail"
+
+# Get the latest merge commit for the bundle. This is the string the bundle image is tagged with.
+get_tag () {
+    URL="https://api.github.com/repos/openshift/sandboxed-containers-operator/commits?per_page=1&path=bundle"
```
gkurz: It is generally good practice to use the `local` keyword and lowercase names for local variables.

fbc/update.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# Get the digest for a tagged image. Pass the <TAG> as the first argument.
+# Quay API docs are at https://docs.quay.io/api/swagger/#!/tag/listRepoTags.
+get_digest () {
+    TAG="$1"
+    URL="https://quay.io/api/v1/repository/redhat-user-workloads/ose-osc-tenant/osc-operator-bundle/tag?specificTag=$TAG"
+    curl -L "$URL" | jq -r '.tags[0].manifest_digest'
+}
+
+sed_digest () {
```
gkurz: This function doesn't have any user. Leftover ?

fbc/update.sh
```diff
@@ -0,0 +1,67 @@
...
+    DIGEST="$1"
+    sed -E -i "s/sha256:[0-9a-f]{64}/$DIGEST/"
+}
+
+# Replace any digest in the <FILE> with the new <DIGEST>.
+replace_digest () {
```
gkurz: Ditto.

fbc/README.md
```diff
@@ -10,8 +10,8 @@ Download the binary from [Github releases](https://github.com/operator-framework
 
 ## Update the FBC
 
-1. Update the digests in the template.
-1. Run `./render.sh` to update the actual catalog.
+1. Run `./update.sh [VERSION]` to update the digests in the template.
+1. Run `./render.sh [VERSION]` to update the actual catalog.
```
gkurz: Assuming this is supposed to be a numbered list, this should be `2.` and the line below be `3.`.

gkurz (commented):
The lack of commit messages makes it hard to review, especially for someone who is looking at the scripts for the first time. 

 I don't quite understand what the second commit `fbc: update render script` is trying to achieve.

Is the third commit `fbc: add v1.10.0` partly generated using the update and render scripts ? If so, make it clear in the commit message and maybe make it the last commit in the PR.

Finally, we do like to have Signed-off-by tags in all commits.

fbc/v4.15/catalog-template.yaml
```diff
@@ -83,6 +86,8 @@ entries:
     schema: olm.bundle
   - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:a37f38455aff326ef87ba4cef4c4c6268351f665265321f39d9e033de819cd69
     schema: olm.bundle
+  - image: quay.io/redhat-user-workloads/ose-osc-tenant/osc-operator-bundle@sha256:e76a1557dd47c701b15237eb264a5e3c91a39b4961e4f05a407b8be856bc8d1d
```
littlejawa: The enterprise contract for 4.15 doesn't like this reference to "quay" :-/
Can't tell why the others are failing (the logs are lost already)

gkurz: FWIW I could re-run the 4.19 check and it has logs at https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/ose-osc-tenant/applications/osc-fbc-4-19/pipelineruns/osc-fbc-4-19-enterprise-contract-69k6w/logs

Now trying to understand what went wrong.

gkurz:
```
Results:
✕ [Violation] test.no_failed_tests
  ImageRef: quay.io/redhat-user-workloads/ose-osc-tenant/osc-fbc-4-19@sha256:784f3596da512d0f3223528f003b84a8e1b6ba0063481fd2870da73b91448bf7
  Reason: The Task "fbc-fips-check-oci-ta" from the build Pipeline reports a failed test
  Term: fbc-fips-check-oci-ta
  Title: No tests failed
  Description: Produce a violation if any non-informative tests have their result set to "FAILED". The result type is configurable
  by the "failed_tests_results" key, and the list of informative tests is configurable by the "informative_tests" key in the rule
  data. To exclude this rule add "test.no_failed_tests:fbc-fips-check-oci-ta" to the `exclude` section of the policy
  configuration.
  Solution: There is a test that failed. Make sure that any task in the build pipeline with a result named 'TEST_OUTPUT' does not
  fail. More information about the test should be available in the logs for the build Pipeline.
```

This is the result of the failure below from `fbc-fips-check-oci-ta` in the build pipeline :

```
{"result":"FAILURE","timestamp":"2025-06-11T07:25:15+00:00","note":"Task osc-fbc-4-19-on-pull-request-cnhx5-fbc-fips-check-oci-ta failed: Could not render unreleased bundle image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:e76a1557dd47c701b15237eb264a5e3c91a39b4961e4f05a407b8be856bc8d1d. Make sure the image is accessible or a mirror is provided for the same in images-mirror-set.yaml","namespace":"default","successes":0,"failures":1,"warnings":0}
```

Suggested fix is to add mirroring :

```
The fbc-fips-check task uses the check-payload tool to verify if an unreleased operator bundle in an FBC
fragment image is FIPS compliant. It only scans operator bundle images which either claim to be FIPS
compliant by setting the features.operators.openshift.io/fips-compliant label to "true" on the bundle
image or require one of OpenShift Kubernetes Engine, OpenShift Platform Plus or OpenShift Container Platform subscriptions to run the operator on an Openshift cluster. This task extracts relatedImages from
all unreleased operator bundle images from your FBC fragment and scans them. In the context of FBC
fragment, an unreleased operator bundle image is the one that isn't currently present in the Red Hat
production Index Image (registry.redhat.io/redhat/redhat-operator-index). It is necessary for
relatedImages pullspecs to be pullable at build time of the FBC fragment. In order to resolve them, this
task expects a ImageDigestMirrorSet file located at .tekton/images-mirror-set.yaml of your FBC fragment
git repo. It should map unreleased registry.redhat.io pullspecs of relatedImages to their valid quay.io
pullspecs. If the ImageDigestMirrorSet is not provided, the task will attempt to process the
registry.redhat.io pullspecs as is and might fail.
```

This kind of relates to #675.

littlejawa: Updated #675 to provide that file. Once merged, we can rebase this PR to (hopefully) solve it

littlejawa: @spotlesstofu - if you rebase, the enterprise contracts for 4.16 and above should pass

spotlesstofu: last push addresses this (c5cb9fa)

fbc/v4.15/catalog/sandboxed-containers-operator/catalog.json
```diff
@@ -208,7 +208,7 @@
     "schema": "olm.bundle",
     "name": "sandboxed-containers-operator.v1.10.0",
     "package": "sandboxed-containers-operator",
-    "image": "quay.io/redhat-user-workloads/ose-osc-tenant/osc-operator-bundle@sha256:e76a1557dd47c701b15237eb264a5e3c91a39b4961e4f05a407b8be856bc8d1d",
+    "image": "quay.io/redhat-user-workloads/ose-osc-tenant/osc-operator-bundle@sha256:793181f8c3b03eefd22d9c51f87eb812a47f7c358ac223e434ed1d297e3c47a3",
```
littlejawa: It is still referencing quay, is that normal?

fbc/README.md
littlejawa: Agree with this.
Some of us read the documentation most of the time in its plain text version. Having consistent numbering here makes sense, even if the .md rendering eventually doesn't care and puts the right numbers.

littlejawa:
With @spotlesstofu's agreement, I've pushed some updates to this PR.
- rebased to take latest changes (including for enterprise contract) into account
- since the change to the "render.sh" script was merged in a separate PR (#711), this commit was removed
- ran "update.sh" and committed the change
- ran "render.sh" and committed the change
- squashed "bundle update" and "render update" commits for clarity

I'll look at failures and try to fix.

littlejawa:
FIPS check is failing with:
```
---- Failure Report
Operator Name,Executable Name,Status
osc-operator-container,/manager,go binary is not CGO_ENABLED
osc-operator-container,/metrics-server,go binary is not CGO_ENABLED
F0625 06:51:25.330869    1336 main.go:294] Error: run failed
check-payload scan failed for quay.io/redhat-user-workloads/ose-osc-tenant/osc-operator@sha256:8ec10c5ddbdc2c33744aded4a51cf6a369a0366dc1bb62cf6abb696dce117a28
```

We need to enable that flag. It was actually the case in our CPaaS builds, but we missed it :-(
I will do it as a separate PR: if I do it here, the check will not take the modified image into account. I need to fix it, build an image, and then update the bundle again here to make sure the tests takes the modification into account.

littlejawa:
Still have a failure in FIPS check, now it is:
```
---- Failure Report
Status
openssl library not present
F0625 09:43:39.042222     444 main.go:294] Error: run failed
check-payload scan failed for quay.io/redhat-user-workloads/ose-osc-tenant/osc-caa-webhook@sha256:abfbd658b8fa83492420ff276fe369388898a828ba18e060f33d13e54ae4353f
```

Not sure why openssl needs to be present. I don't see any reference to this in our previous builds (in CPaaS).

spotlesstofu:
> Not sure why openssl needs to be present.

@littlejawa for FIPS you need to link to the SSL libraries instead of bundling them in the main binary, I think Go has a flag to set that at build time

littlejawa:
> @littlejawa for FIPS you need to link to the SSL libraries instead of bundling them in the main binary, I think Go has a flag to set that at build time

I think that's the "CGO_ENABLED" flag? If so, we're using it already (and I just added to some other images).
The problem reported here, seems to be that the library is not found in our image.

From the konflux-users forum, I found people having the same problem because they don't use the right base image.
We're using "ubi-micro" as base image for osc-caa-webhook, and we use ubi-minimal for all others. I suspect that's why osc-caa-webhook is failing the test.

littlejawa:
Everything passed... exccept 4.16 build which got an infra issue.
I'll call it a win :-p

littlejawa:
@spotlesstofu - we're good on the Enterprise Contract side.
I did not address @gkurz's comments though.

littlejawa (approved):
/lgtm

littlejawa:
The e2e fail is expected. The "make test" fail is because the test pipeline is not in this branch - it needs a rebase.
Now the only reason for the operator to rebuild (and for "make test" to run) is because you changed the Makefile. It won't impact the result of the test. So I'm fine if we merge this PR as is.

PR #678, opened 2025-06-06
Enable hermetic build for podvm builder with submodules

littlejawa:
Trying to enable hermetic builds for podvm-builder using a submodule to cloud-api-adaptor sources.

Fixes: [KATA-3800](https://issues.redhat.com//browse/KATA-3800)


.tekton/osc-podvm-builder-push.yaml
```diff
@@ -101,7 +103,7 @@ spec:
       description: Skip checks against built image
       name: skip-checks
       type: string
-    - default: "false"
+    - default: "true"
```
c3d, line 106: Why this specific changes?

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "config/peerpods/podvm/cloud-api-adaptor"]
+	path = config/peerpods/podvm/cloud-api-adaptor
+	url = https://github.com/openshift/cloud-api-adaptor/
+	branch = osc-release
```
c3d, line 4: For reproducibility, shouldn't have an exact commit here? You mentioned that this causes conflicts during `go run`. I assume that would depend on the actual content, so maybe we want to make sure we know exactly which version we pick, not just a branch.

c3d (approved):
Overall looks good to me, but with two questions.

Still approving because this would not change with the answers to my questions.

.gitmodules
littlejawa: I think we would have it whatever the commit ID.
The fix is in this PR, not on the CAA repo. It consists in modifying the Dockerfile for the operator's build to remove the podvm-payload sources (including that submodule folder). Without it, there is no risk of conflict, whatever the version we use from cloud-api-adaptor.
So no, we don't need to pin a specific version.

.tekton/osc-podvm-builder-push.yaml
littlejawa: This is the flag that enables hermetic build (see the description below it).
That's maybe the confusion: the description for the flag is below it, not above ?

beraldoleal: I hate when the the json has as the first element something random :) I was reading the skip-checks instead of hermetic :)

c3d: So did I @beraldoleal 😉

c3d: As a matter of fact, I suggest [rewriting them](https://github.com/openshift/sandboxed-containers-operator/commit/a58d5427520eae434f943757ce9278b1ea46d0e9). @littlejawa would you mind adding that change to your PR?

littlejawa: I don't mind, but I'm not sure how long it will stay.
This file is initially generated by Konflux, and regularly updated by it (for tekton image references). And I've seen a couple times my refactoring to be reset by updates (typically the "on-cel-expression" at the top of the file).
Maybe it won't touch it, but I wouldn't bet :-)

littlejawa: cherry-picked

littlejawa: ok, I changed my mind and dropped it.
The commit changes just one of the two .tekton files involved in this PR, and the result makes the review even more confusing as it changes things totally unrelated to the topic.
If we want to do that, let's have a separate refactoring PR that deals with all our pipelines.

littlejawa:
The "ci/prow/check" job is having the same error that I had, with submodule's dependencies messing up with the operator's build.
I guess the check job needs to be modified in the same way, removing the podvm-builder folder from the sources before building? Where is that located?

bpradipt (approved):

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #698, opened 2025-06-16
konflux: Make description of entries more readable

c3d:
Some of the entries were beginning with `- default`. This is confusing during code review [1], where you tend to read that as belonging to the previous item instead of the current one. Move `default` after description, which puts description right after `-` to avoid that problem.

[1]: https://github.com/openshift/sandboxed-containers-operator/pull/678#discussion_r2132139897

### DIFF
--- a/.tekton/osc-must-gather-pull-request.yaml
+++ b/.tekton/osc-must-gather-pull-request.yaml
@@ -62,69 +62,69 @@ spec:
     - description: Source Repository URL
       name: git-url
       type: string
-    - default: ""
-      description: Revision of the Source Repository
+    - description: Revision of the Source Repository
+      default: ""
       name: revision
       type: string
     - description: Fully Qualified Output Image
       name: output-image
       type: string
-    - default: .
-      description: Path to the source code of an application's component from where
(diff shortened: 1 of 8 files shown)


littlejawa:
I agree with the change, but just want to reiterate a warning I did in another place: these files are generated by Konflux, and updated automatically by it. It may or may not restore the alphabetical ordering in a future update.
I am in favor of doing the change and hoping for the best, but if it is restored later, it's not my fault ! :-p

beraldoleal (approved):
This looks much better. I'm not sure if the same will hold true with Konflux, though. Unfortunately, this was previously overridden by some automation tools, and it was hard to maintain the correct order. Hopefully we won’t run into that again here.

beraldoleal (commented):
Any reason for not starting with the name?

beraldoleal:
> I agree with the change, but just want to reiterate a warning I did in another place: these files are generated by Konflux, and updated automatically by it. It may or may not restore the alphabetical ordering in a future update. I am in favor of doing the change and hoping for the best, but if it is restored later, it's not my fault ! :-p

Didn't saw this answer. And this answer my comment above :)

PR #700, opened 2025-06-17
osc-bundle: update pullspecs to use image indexes

littlejawa:
Enabling s390x means we need to make the bundle point to image indexes, each having a reference to the different arch-specific images.

The automated updates from Konflux were still updating only the x86_64 reference. This commits change them to the image indexes.

This PR is a prerequisite for #676

Fixes: [KATA-3907](https://issues.redhat.com//browse/KATA-3907)


littlejawa:
@spotlesstofu - When this PR is merged, we should have a bundle that references all s390x builds

littlejawa:
@dbkreling - FYI
That's the follow-up I was talking about.

spotlesstofu (approved):

PR #704, opened 2025-06-18
podvm-builder: fix the location of cloud-api-adaptor sources

littlejawa:
The COPY operation is putting the sources directly under /src. We want it to create a subfolder instead.

### DIFF
--- a/config/peerpods/podvm/Dockerfile.podvm-builder
+++ b/config/peerpods/podvm/Dockerfile.podvm-builder
@@ -47,7 +47,7 @@ RUN if [[ -n "$CERT_RPM" ]] ; then \
     fi
 
 
-COPY cloud-api-adaptor /src/
+COPY cloud-api-adaptor /src/cloud-api-adaptor
 
 ADD podvm-builder.sh /podvm-builder.sh
 


snir911 (approved):

config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -47,7 +47,7 @@ RUN if [[ -n "$CERT_RPM" ]] ; then \
     fi
 
 
-COPY cloud-api-adaptor /src/
+COPY cloud-api-adaptor /src/cloud-api-adaptor
```
gkurz, line 50: Can you mention in the commit message why this subfolder is needed ? Is is because some other code expects the CAA source to be there or something else ?

littlejawa: Sure. The sources are put in the container image along with scripts that will run a build when the container runs.
This is part of the image-building process that occurs in the remote hypervisor.
I will try to summarize that in the commit message

littlejawa:
Verified in the image built for this PR: the sources are in the right place

config/peerpods/podvm/Dockerfile.podvm-builder
littlejawa: Amended

bpradipt (approved):

c3d (approved):
LGTM with updated commit message.

PR #706, opened 2025-06-19
Make a test catalog

littlejawa:
Trying to find a way to automate the creation of a catalog.

This PR adds a new FBC build that creates a very simple catalog with only one image - our latest build.
I plan to add this as a component to our Konflux application (sandboxed-containers) and then try to use the nudging mechanism to make Konflux update the bundle reference every time a new bundle is built.

I'm not sure that's the right way, but if it works we should have a catalog built automatically for testing.
Obviously, this is not meant for release - all our other FBCs are here for that.

NOTE: about render.sh
I cherry-picked the commit for updating the render script.
I also had to modify that script slightly to make it skip numeric version check when we render the test catalog.

Since the default is to look only at version numbers, the script's behavior is to ignore the test-fbc catalog by default, which is fine with me: I want Konflux to update the references, I don't want us to render it on every build.

### DIFF
--- a/fbc/test-fbc/catalog/sandboxed-containers-operator/catalog.json
+++ b/fbc/test-fbc/catalog/sandboxed-containers-operator/catalog.json
@@ -1,0 +1,170 @@
+{
+    "schema": "olm.package",
+    "name": "sandboxed-containers-operator",
+    "defaultChannel": "stable",
+    "icon": {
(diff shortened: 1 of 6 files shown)


littlejawa:
/cc @spotlesstofu - I'm a bit confused about the process. I used "duplicate.sh", and I have a konflux pipeline running, but failing because there is no component on the Konflux side yet. Should I create the component? If so, this will create a separate PR with separate pipeline definitions isn't it?
Or was I supposed to create the component first?

fbc/test-fbc/catalog/sandboxed-containers-operator/catalog.json
```diff
@@ -0,0 +1,170 @@
...
+        }
+    ],
+    "relatedImages": [
+        {
+            "name": "",
+            "image": "registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:23eb79b96ef69c351b5bcb81bae0524e87aa964047759fbfe92d57b59d90cee2"
+        },
+        {
+            "name": "caa",
+            "image": "registry.redhat.io/openshift-sandboxed-containers/osc-cloud-api-adaptor-rhel9@sha256:2453b3f36547f5dde1f771df8c07eea6cd1467c07338d1598ee503d302db1e52"
```
littlejawa, line 147: I guess this is where my plan is failing - I can make the bundle updates trigger an update to the file, but if I want to update all the "relatedImages", I either have to make all components nudge this fbc (which is a nightmare to manage), or run "render.sh" as part of the Dockerfile?
I'll try to see if we can do something like that

littlejawa: Updated the PR with a change that runs render.sh during the build.
Works locally, at least.

littlejawa:
For reference: the catalog that I generated locally with the code from this PR is working on OCP 4.15.
I'm building with the 4.19 builder image, so I wanted to be sure. But since we're releasing the same images for all versions, the catalog just works.
That's the other benefit here: rather than forcing testers to use a different catalog for each OCP version, we can give them just one that works for all.
Of course, that could change in the future if we start releasing different things for different OCP versions, but we're not there yet :-)

littlejawa:
Continuing this work on #708

PR #707, opened 2025-06-19
Fix OSC version to 1.9.0 till 1.10.0 is GA

bpradipt:
Drive-by-fix: Typo in intel deviceplugin controller pod name


mythi (approved):
LGTM and double checked

PR #716, opened 2025-06-23
test-fbc: make sure only the test catalog is updated by bundle builds

littlejawa:
The test-fbc must be modified by automated PRs from Konflux whenever a new bundle is built. However by default, this nudging mechanism looks for references everywhere in the repository. The result is that when the bundle is updated, all our catalogs gets their files modified, with every single entry replaced by the new bundle's reference.

This commit adds an explicit list of files that need to be updated when the test-fbc is nudged: we want to have only this catalog template to be updated, and no other.


littlejawa:
See an example of a PR that updates too many files: #713

PR #731, opened 2025-06-25
osc-operator: build with CGO_ENABLED=1 to comply with FIPS checks

littlejawa:
The FIPS check was failing during the build, because our binaries were built with CGO_ENABLED=0.
Setting the flag to 1 to fix this.

Notes:
- somehow, this failure showed up only when we build the catalogs (?) As such, this PR should help fix #676 
- this flag was actually set to 1 already in our CPaaS build, so this is not a new change, I'm just fixing something that I missed earlier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,8 +139,8 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 
 .PHONY: build
 build: manifests generate fmt vet ## Build manager and metrics-server binaries.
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
/lgtm

snir911 (approved):

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #777, opened 2025-07-06
gcp: use disk.raw naming at disk upload

snir911:
otherwise it fails with:
The file inside the tar archive was named ''podvm.raw''. It should be named ''disk.raw''

**- What I did**

**- How to verify it**


beraldoleal (approved):
LGTM, thanks

PR #790, opened 2025-07-08
azure: add small debug msg after uploading vhd

beraldoleal:
Job gets stuck for a while after this step, let's just add a small msg here to keep users informed about what is going on.

**- What I did**

**- How to verify it**


snir911 (approved):

gkurz (approved):
/lgtm

Thanks @beraldoleal !

PR #797, opened 2025-07-09
cm-helper: add gcp support

snir911:
to test run the `pp-cm-helper.sh` script after osc operator is installed (before kataConfig)


beraldoleal (approved):
LGTM, thanks for working on this @snir911

PR #815, opened 2025-07-16
image: exclude s390x from using default prebuilt

snir911:


ajaypvictor (approved):
We have done a sanity test with this PR build on s390x and things looks green for us, once a catalog image is available with this PR build, we will test it thoroughly to ensure that various scenarios works as expected.
Thanks Snir.

controllers/image_generator.go
```diff
@@ -835,8 +836,8 @@ func (r *ImageGenerator) updateImageConfigMap() error {
 		igLogger.Info("Setting FIPS mode")
 	}
 
-	// set to default image
-	if cm.Data["PODVM_IMAGE_URI"] == "" && os.Getenv("RELATED_IMAGE_PODVM_OCI") != "" {
+	// set to default image which is currently not built for s390x
+	if cm.Data["PODVM_IMAGE_URI"] == "" && os.Getenv("RELATED_IMAGE_PODVM_OCI") != "" && runtime.GOARCH != "s390x" {
```
bpradipt, line 840: Do we need this conditional? My worry is with this change there won't be any way to test pre-built image for s390x without code change.
Can't it be ensured that RELATED_IMAGE_PODVM_OCI is empty for s390x?

snir911: why not? setting PODVM_IMAGE_URI will still work for s390x (this is what the hyper-protect instructions suggested now AFAIU)

bpradipt: Ah I forgot that PODVM_IMAGE_URI is used by s390x as well..

bpradipt (approved):
/lgtm

PR #821, opened 2025-07-21
bundle: set RELATED_IMAGE_PODVM_OCI

snir911:
This required also some configuration changes in order to be able to pull the image in-job

PR #824, opened 2025-07-22
osc-operator: build upon yaml changes

snir911:
some yaml are parsed at runtime, trigger operator build when these are modified:
controllers/utils.go:func parseJobYAML(yamlData []byte) (*batchv1.Job, error) { controllers/utils.go:func parseMachineConfigYAML(yamlData []byte) (*mcfgv1.MachineConfig, error) { controllers/utils.go:func parseCredentialsRequestYAML(yamlData []byte) (*ccov1.CredentialsRequest, error) { controllers/utils.go:func parseConfigMapYAML(yamlData []byte) (*corev1.ConfigMap, error) {

**- What I did**

**- How to verify it**

### DIFF
--- a/.tekton/osc-operator-pull-request.yaml
+++ b/.tekton/osc-operator-pull-request.yaml
@@ -11,10 +11,19 @@ metadata:
     pipelinesascode.tekton.dev/on-cel-expression: |
       event == "pull_request" &&
       target_branch == "devel" &&
-      files.all.exists(path, !path.matches('bundle/.*|.tekton/.*bundle.*yaml$|config/manager/.*')) &&
-      files.all.exists(path, !path.matches('must-gather/.*|.tekton/.*must-gather.*yaml$')) &&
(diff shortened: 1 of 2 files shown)


spotlesstofu:
Is `files.any` a thing? I'd say `files.all` like on the others.

spotlesstofu:
Could you point to which previous merge request triggered this?

snir911:
oops:D
> Could you point to which previous merge request triggered this?

The problem is the opposite, it wasn't triggered by https://github.com/openshift/sandboxed-containers-operator/pull/821

spotlesstofu:
I don't understand. The operator build ran when you merged #821. What is the desired behavior that is missing?

snir911:
can i see it? in that case why there's no following `chore(deps): update osc-operator`?

spotlesstofu:
I see #819 and #823 open. Please mind that if there is already a pull request open to update the same component, the bot updates the existing pull request, it doesn't open a new one.

PR #825, opened 2025-07-24
Daemonset deployment mode

Pacho20:
Introduce support for DaemonSet-based deployment mode for clusters running without the Machine Config Operator (MCO).

This feature enables an alternative deployment path by detecting the absence of the MachineConfig CRD and activating DaemonSet mode via a dedicated feature gate. It includes logic for deploying and managing DaemonSets on targeted nodes, using labels to control execution and track installation status.

There are still many TODOs in the implementation, and the list will be updated as development progresses.


controllers/daemonset_handler.go
```diff
@@ -0,0 +1,451 @@
...
+		r.Log.Info("Error checking if MC is available")
+		return err
+	}
+
+	if state == Enabled && !mcAvailable {
+		r.Log.Info("MC is not available, deployment mode will be set to DaemonSet")
+		r.DeploymentMode = DaemonSet
+	} else {
+		r.Log.Info("Deployment mode will be set to MachineConfig")
+		r.DeploymentMode = MachineConfig
```
spotlesstofu: What's the case for explicitly setting the DeploymentMode to MachineConfig?

Leaving it untouched may help avoid clashes with future reconcile code outside of this feature.

controllers/daemonset_handler.go
```diff
@@ -0,0 +1,1014 @@
...
+						IntVal: 1,
+					},
+				},
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: dsLabelSelectors,
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: "default", // TODO: Which service account should be used?
```
spotlesstofu: I see we already use this ServiceAccount in the operator. If it works, I'm OK with it.

controllers/daemonset_handler.go
```diff
@@ -0,0 +1,1014 @@
...
+	imageString, err := r.GetImageForComponent("cli")
+	if err != nil {
+		return "", err
+	}
+	if imageString == "" {
+		return "", fmt.Errorf("empty result for image name")
+	}
+
+	return imageString, nil
+}
```
spotlesstofu: `GetExtensionImage` and `GetCliImage` look the same. Can deduplicate?

controllers/daemonset_handler.go
```diff
@@ -0,0 +1,1014 @@
...
+    clear_status
+    set_status "installing"
+}
+
+set_status_waiting_for_reboot() {
+    clear_status
+    set_status "waiting_for_reboot"
+}
+
+# Initial wait: avoid doing anything if a previous staged update is pending
```
spotlesstofu: It may happen that an update is staged for other packages. It would be nice to check if kata is installed before requesting a reboot, and installing the package if it's not.

spotlesstofu (changes requested):
Thanks Patrik for working on this new feature.

I added a couple of inline comments about the logic.

Please move the shell scripts to a separate file and include them, or mount them as an immutable ConfigMap.

Nice to have: deduplicate code.

controllers/daemonset_handler.go
gcoon151: good suggestion. Before we start trying to deploy kata check to see if there's pending packages. We'll get stuck waiting on something that has nothing to do with kata.

controllers/daemonset_handler.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// Process the DaemonSet feature gate (FG).
+// This method is invoked by the reconcile loop at its initiation.
+// It examines the current state of the FeatureGate and adjusts the deployment mode (DaemonSet or MachineConfig)
+// based on the availability of the MachineConfig Add-on and the FeatureGateState.
+// If the MachineConfig Add-on is unavailable and the FeatureGateState is Enabled,
+// the deployment mode is set to DaemonSet. Otherwise, it defaults to MachineConfig.
+// If the FeatureGateState is Disabled, the deployment mode remains MachineConfig,
+// regardless of the MachineConfig Add-on's availability.
```
beraldoleal: Have you considered inverting the logic? For example: if the MachineConfig add-on is available and the feature gate is disabled, then we use MachineConfig (the current behavior); otherwise, if the feature gate is enabled, we always use the new DaemonSet mode, regardless of MC availability.

Just wondering if we really need to check MC availability when we're explicitly enabling DaemonSet mode.

controllers/fg_handler.go
```diff
@@ -12,11 +12,13 @@ const (
 	FgConfigMapName         = "osc-feature-gates"
 	ConfidentialFeatureGate = "confidential"
 	LayeredImageDeployment  = "layeredImageDeployment"
+	DaemonSetDeployment     = "daemonSetDeployment"
 )
 
 var DefaultFeatureGates = map[string]bool{
 	ConfidentialFeatureGate: false,
 	LayeredImageDeployment:  false,
+	DaemonSetDeployment:     false,
```
beraldoleal: Just to give you some visibility, there is a discussion going on that we are planning to rename "Feature Gates" to just "Config Option":

https://github.com/openshift/sandboxed-containers-operator/pull/827

So we are planing renaming the config map name, soon. Giving this is a new feature, I'm concerned the users will have to learn once and the change it latter.

@gkurz, @bpradipt ... maybe we could rename the config map as soon as possible, so this new config option would be already at the right place?

controllers/daemonset_handler.go
```diff
@@ -266,8 +269,58 @@ func (r *KataConfigOpenShiftReconciler) addPeerPodsConfigDaemonSet() error {
 }
 
 func (r *KataConfigOpenShiftReconciler) GetExtensionImage() (string, error) {
```
beraldoleal: Please, could you squash this commit with the previous one? This way, when reviewing individual commits we don't need to follow your flow.

controllers/daemonset_handler.go
```diff
@@ -203,6 +203,24 @@ func (r *KataConfigOpenShiftReconciler) processDaemonSetKataConfigInstallRequest
 		}
```
beraldoleal: Same here, squash with the 1e96e993911a3362c65aa2e38b7cccd6de899f0f

controllers/daemonset_handler.go
```diff
@@ -349,17 +349,24 @@ func (r *KataConfigOpenShiftReconciler) DaemonSetForKataInstall(imageString stri
 		kataInstallDsName       = "osc-rpm-install"
 
 		script = `
-sleep infinity
```
beraldoleal: Same here, please squash with 1e96e993911a3362c65aa2e38b7cccd6de899f0f

controllers/daemonset_handler.go
```diff
@@ -348,8 +348,32 @@ func (r *KataConfigOpenShiftReconciler) DaemonSetForKataInstall(imageString stri
 		nodeSelector            = r.getNodeSelectorAsMap()
```
beraldoleal: Same here, please squash with 1e96e993911a3362c65aa2e38b7cccd6de899f0f.

beraldoleal (commented):
Hey @Pacho20, thanks for this. Appreciated.

I left a few comments for you, but my first suggestion here, is to squash the commits that are refactoring and changing the same code base, so the commit-by-commit review would be easier.

config/peerpods/osc-configs-for-ds.yaml
```diff
@@ -0,0 +1,585 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: osc-configs
+  namespace: openshift-sandboxed-containers-operator
+#immutable: true
+data:
```
gcoon151: Should the installscript, 50-kata-remote, and configuration-remote.toml all move to their own configmaps so they don't have yaml inside yaml problems. I think I see a TODO comment that you want to do that though.

controllers/daemonset_handler.go
```diff
@@ -0,0 +1,1014 @@
...
+					ServiceAccountName: "default", // TODO: Which service account should be used?
+					NodeSelector:       nodeSelector,
+					HostPID:            true,
+					Containers: []corev1.Container{
+						{
+							Name:            "rpm-install",
+							Image:           imageString,
+							ImagePullPolicy: corev1.PullIfNotPresent,
+							SecurityContext: &corev1.SecurityContext{
+								// TODO: do we really need to run as root?
```
gcoon151: The answer here is yes. We're installing things on workers.

controllers/daemonset_handler.go
spotlesstofu: This is more of a future-looking design choice.

In a scenario where the feature comes enabled by default, it makes sense to also check for MachineConfig availability and select the deployment mode accordingly.

This logic conveys that the MachineConfig is a more mature mechanism and should be preferred when it's available.

beraldoleal: How would a user force the use of DaemonSet, regardless of whether the MC CRD is present or not?

Maybe I'm misinterpreting the intent here, but if this is just meant as a fallback, then I’d suggest renaming the feature to something like `DaemonSetDeploymentFallback` to avoid giving the impression that users are explicitly enabling DaemonSet mode.

Pacho20: If the MCO manages the nodes, we don’t want to interfere with that. Since we’re altering runtime configurations, there’s a risk that the DaemonSet could conflict with the MachineConfig.

Also, I don’t see a use case for choosing the DaemonSet installation over the MCO when the MCO is already being used in the cluster.

controllers/daemonset_handler.go
Pacho20: Done

controllers/daemonset_handler.go
Pacho20: Modified the script to also check the package list.

controllers/daemonset_handler.go
beraldoleal: Fair point about avoiding MCO conflicts. But if the feature is called “enable DaemonSet mode,” users will expect it to override MCO. Relying on docs isn’t enough. The name should make the intent clear on its own. So I’d suggest renaming it to avoid confusion.

controllers/fg_handler.go
beraldoleal: We discussed this today in a call, and we fine with changing this the future as it will be TP. 

Closing this thread, feel free to reopen if needed.

config/peerpods/osc-configs-for-ds.yaml
gcoon151: I see you moved these. This is handled. Thanks @Pacho20 !

controllers/daemonset_handler.go
```diff
@@ -81,9 +80,3 @@ func (r *KataConfigOpenShiftReconciler) isMCAvailable() (bool, error) {
 
 	return true, nil
 }
-
-
-func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, error) {
```
beraldoleal: This part of the commit should be squashed with 2c3da8db78973bba3f1479d551857690b9b6b65a

controllers/daemonset_handler.go
bpradipt: We may have a use case for forcing Daemonset mode - during testing, installing custom artefacts for dev-preview which cannot be installed via MC. Why not have a force/override mode as an option for the feature?

gcoon151: Just want to mention again that hypershift is the future on places like Azure and in hosted control planes (cloud providers) the masters have MCO so I think that will break the today code expectations of what works and what doesn't. I'm not sure if that influences how we want to handle this.

controllers/daemonset_handler.go
Pacho20: I squashed that part—hope it's okay this way

controllers/daemonset_handler.go
Pacho20: I modified the feature gate logic in commit c7afd44892de1be200793faf93809b6a7ceae6c9.

The feature gate now supports three values:
- `MachineConfig` (default)
- `DaemonSet`
- `DaemonSetFallback`

Use `DaemonSet` to explicitly force DaemonSet mode.

controllers/daemonset_handler.go
```diff
@@ -0,0 +1,122 @@
...
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Create enum to represent the state of the deployment mode
+type DeploymentMode int
```
beraldoleal: Since this file is called `daemonset_handler`, I feel most of the "DeploymentMode logic" isn't specific to the daemonset handler. So maybe move this out of this file?

controllers/fg_handler.go
```diff
@@ -40,22 +45,35 @@ const (
...
+				return nil, err
+			}
+			fgStatus.LayeredImageDeployment = layeredImageDeployment
+		}
+		if value, ok := cfgMap.Data[DeploymentModeConfig]; ok {
+			mode, err := ParseDeploymentModeOption(value)
+			if err != nil {
+				return nil, err
+			}
+			fgStatus.DeploymentModeOption = mode
```
beraldoleal: I agree with the new logic, but since it changes behavior (invalid values now error instead of falling back to defaults, for both bools and the custom type), I think we should keep the old semantics for this z-release and defer the stricter validation to a minor/major bump.

controllers/daemonset_handler.go
```diff
@@ -0,0 +1,122 @@
...
+// This method is invoked by the reconcile loop at its initiation.
+// It examines the current state of the FeatureGate and adjusts the deployment mode (DaemonSet or MachineConfig)
+// based on the selected DeploymentModeOption and the availability of the MachineConfig Add-on.
+//
+// The behavior is as follows:
+//   - If the mode is MachineConfigOption, the deployment mode is set to MachineConfig.
+//   - If the mode is DaemonSetOption, the deployment mode is forcibly set to DaemonSet, regardless of MC availability.
+//   - If the mode is DaemonSetFallbackOption, the deployment mode is set to DaemonSet only if the MachineConfig Add-on is unavailable.
+//     Otherwise, it defaults to MachineConfig.
+func (r *KataConfigOpenShiftReconciler) handleDaemonSetFeature(mode DeploymentModeOption) error {
```
beraldoleal: Now that this method is dealing with multiple modes, the method name is a bit misleading... I would suggest something like `resolveDeploymentMode` or `updateDeploymentMode`.

controllers/openshift_controller.go
```diff
@@ -146,6 +148,9 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
+	if r.DeploymentMode == DaemonSetMode {
+	}
+
```
beraldoleal: leftover?

beraldoleal: Oh, now that I'm reviewing 15c66c431323de800732efa7b66d29e3c6f05895 I see. You added an empty if on the first commit, and then added a return in a later commit. This needs to be squashed.

controllers/daemonset_reconcile.go
```diff
@@ -0,0 +1,1145 @@
+package controllers
```
beraldoleal, line 1: I'm concerned about this duplication here, and how we will keep maintaining both of them for a while.

Any reason for this?

beraldoleal (commented):
Hi @Pacho20, thanks for keeping pushing this PR. Appreciated. 

I left a few more comments for you, but also lets see if others join the party.

controllers/fg_handler.go
Pacho20: Oh, you're right. I'll modify the code to preserve the old semantics.

controllers/daemonset_handler.go
Pacho20: True, I'll rename it. I’m thinking of using something like `handleDeploymentModeFeature` to follow the naming convention of the feature gate handlers.

controllers/daemonset_handler.go
Pacho20: Valid point. It was originally called `daemonset_handler` because it handled the DaemonSet Feature Gate. Since that's no longer the case, I'll rename the file. The entire logic now relates to DeploymentMode.

controllers/daemonset_reconcile.go
Pacho20: Just to clarify, when you mention duplication, are you referring to the separate reconciliation logic paths, or something else?

beraldoleal: Yes, I'm talking about the "reconcile loop".

beraldoleal:
Also, please rebase, since go.sum got a conflict.

controllers/daemonset_reconcile.go
Pacho20: The logic of the reconcile function is deeply coupled with the MachineConfig CRD. It includes numerous checks and MachineConfig-specific operations. At the beginning of development, I chose to implement a separate reconcile loop to minimize changes to the existing codebase and avoid unintended side effects. I had no clear idea how the new implementation would look, but given the tight coupling, I assumed there wouldn’t be much overlap and that integration would be difficult.
Because of that, I didn’t attempt to merge the two reconcile loops early on. However, over the past few days, I’ve been exploring ways to make the solution more maintainable, since I agree that having two separate flows isn’t ideal. I just pushed some changes that merge parts of the logic and remove some duplication—these were relatively straightforward and safe to refactor. There are now shared function calls between the MachineConfig and DaemonSet deployment modes.
That said, the core logic still remains distinct, and there are conditional checks that determine which major function to call. There’s definitely room for further improvement. But tighter integration would require a deeper refactor and a better understanding of the original design. That’s not something I feel comfortable doing alone, given the complexity and potential impact. I’m also not sure if that level of integration is something we can—or should—tackle in this initial release.

controllers/daemonset_reconcile.go
```diff
@@ -0,0 +1,1086 @@
...
+		},
+	}, nil
+}
+
+// daemonSetForKataInstall creates a DaemonSet for installing or uninstalling Kata based on the specified action.
+// It uses two container images: one for the Kata binaries (extensionImageString) and another for kubectl/cli tools (cliImageString).
+// The DaemonSet's Pod contains three containers: one installs the binaries, one modifies the node labels, and one sets the runtime log level.
+func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDsAction) (*appsv1.DaemonSet, error) {
+	// This image contains the binaries for kata-containers
+	extensionImageString, err := r.GetExtensionImage()
```
bpradipt: Wouldn't it make sense to build a single container image with deployment artefacts?
- kata binaries
- CLI
- scripts
- ..

This way we don't need to create additional configMap with embedded script? Also will reduce to effort w.r.to setting up separate build pipelines.

controllers/daemonset_reconcile.go
```diff
@@ -0,0 +1,1086 @@
...
+					Containers: []corev1.Container{
+						{
+							Name:            "config-sync",
+							Image:           imageString,
+							ImagePullPolicy: corev1.PullIfNotPresent,
+							SecurityContext: &corev1.SecurityContext{
+								Privileged: &runPrivileged,
+								RunAsUser:  &runAsUser,
+							},
+							Command: []string{"/bin/bash", "/scripts/osc-configs-script.sh"},
```
bpradipt, line 520: See the comment https://github.com/openshift/sandboxed-containers-operator/pull/825/files#r2313413123

controllers/daemonset_reconcile.go
```diff
@@ -0,0 +1,1086 @@
...
+			},
+		},
+	}, nil
+}
+
+// addImmutableConfigMap adds or updates an immutable ConfigMap based on the provided YAML data.
+//
+// It returns false if no ConfigMap was created due to an error or
+// if the existing ConfigMap is identical to the provided YAML data.
+func (r *KataConfigOpenShiftReconciler) addImmutableConfigMap(path, yaml string) (bool, error) {
```
bpradipt: Is there a benefit of using the script from the configmap vs adding the script in the container image and run it through the usual static checks etc ?

controllers/openshift_controller.go
```diff
@@ -488,6 +513,54 @@ func (r *KataConfigOpenShiftReconciler) processDaemonsetForCAA() *appsv1.DaemonS
 	}
 }
 
+// Handles provider specific parts of the CAA Ds
+// Modifies the DaemonSet if needed
+func (r *KataConfigOpenShiftReconciler) processProviderConfigCAA(ds *appsv1.DaemonSet) error {
```
bpradipt: I don't think this is related to this PR of daemonset deployment mode or am I missing something ?

controllers/image_generator.go
```diff
@@ -54,12 +54,14 @@ const (
 	peerpodsCMAzureImageKey       = "AZURE_IMAGE_ID"
 	peerpodsCMGCPImageKey         = "PODVM_IMAGE_NAME"
 	peerpodsLibvirtImageKey       = "LIBVIRT_IMAGE_ID"
+	peerpodsCMIBMCloudImageKey    = "IBMCLOUD_PODVM_IMAGE_ID"
```
bpradipt: You can extract this code into a separate PR and get it merged earlier.
This was we will be focussing only on the DS deployment mode in this PR

controllers/deployment_mode_handler.go
```diff
@@ -0,0 +1,122 @@
...
+// The behavior is as follows:
+//   - If the mode is MachineConfigOption, the deployment mode is set to MachineConfig.
+//   - If the mode is DaemonSetOption, the deployment mode is forcibly set to DaemonSet, regardless of MC availability.
+//   - If the mode is DaemonSetFallbackOption, the deployment mode is set to DaemonSet only if the MachineConfig Add-on is unavailable.
+//     Otherwise, it defaults to MachineConfig.
+func (r *KataConfigOpenShiftReconciler) handleDeploymentModeFeature(mode DeploymentModeOption) error {
+	r.Log.Info("Feature gate", "featuregate", DeploymentModeConfig, "state", mode)
+
+	if mode == MachineConfigOption {
+		r.Log.Info("Deployment mode will be set to MachineConfig")
```
gkurz, line 74: Availability of the MachineConfig CRD should also be checked in this case IMO. It is preferred to detect early that the required config won't work over hitting an error later in the MCO-based reconcile logic.

controllers/daemonset_reconcile.go
gcoon151: @Pacho20 is new to IBM and openshift. Which image would you want to put these in? Also would we be able to make a new OCP image in time to release all this for September? We may need some help determining where to do this @bpradipt

controllers/daemonset_reconcile.go
gcoon151: I had suggested this from experience that yaml inside yaml can often have issues in ocp. The line wraps start to be interpreted by ocp when they should be running it verbatim inside the container.

controllers/openshift_controller.go
gcoon151: The reason we're building the daemonset code is to make it run on ROKS on IBM cloud which doesn't have MCO (it runs hypershift). This was something else that needed to be fixed to make that work. We can move to a different PR if desired, but this is a part of the whole thing we're wanting to make work in 1.10.2. Your call.

controllers/image_generator.go
gcoon151: Same as earlier comment. We can move to different PR but that will be a part of what we require and need in 1.10.2 by end of September. Your call @bpradipt

controllers/deployment_mode_handler.go
```diff
@@ -0,0 +1,122 @@
...
+	mc := &unstructured.Unstructured{}
+	mc.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   machineConfigGroup,
+		Version: machineConfigVersion,
+		Kind:    machineConfigKind,
+	})
+
+	// Attempt to GET any MachineConfig to verify if the GVK is known.
+	// If the CRD isn’t installed, it will return a NoMatchError.
+	err := r.Client.Get(context.Background(), client.ObjectKey{Name: machineConfigCrdName}, mc)
```
gkurz: It is rather confusing that you look for a MachineConfig object named like the MachineConfig CRD... Please use some dummy name instead.

controllers/deployment_mode_handler.go
```diff
@@ -0,0 +1,122 @@
...
+		// CRD is installed but no resource exists under that name
+		if k8serrors.IsNotFound(err) {
+			r.Log.Info("MachineConfig CRD is present")
+			return true, nil
+		}
+
+		return false, err
+	}
+
+	return true, nil
```
gkurz: MachineConfig CRD is also present in this case but it won't print the message... :-\

What about simplifying the return branches ? Something like :

```
	err := r.Client.Get(context.Background(), client.ObjectKey{Name: "foo"}, mc)
	if err == nil || k8serrors.IsNotFound(err) {
		r.Log.Info("MachineConfig CRD is present")
		return true, nil
	}

	if meta.IsNoMatchError(err) {
		r.Log.Info("MachineConfig CRD not found")
		return false, nil
	}

	return false, err
```

gkurz (commented):
Hi @Pacho20 !

Thanks for submitting this. I could only partially review the first commit so far, I'll try to devote some time everyday to make progress. Anyway, I've already some comments to share. PTAL.

controllers/image_generator.go
gkurz: I agree with @bpradipt. We will make sure that all required changes are there for 1.10.2.

The workflow will be :
- submit changes to the devel branch
- backport (cherry-pick) to the osc-release-v1.10 branch

beraldoleal: I agree with @bpradipt as well. I would say its better to split. Both can still target 1.10.2, but separate PRs simplify review and make those adjustments easier.

controllers/daemonset_handler.go
beraldoleal: Closing this thread, since @Pacho20  already implemented the modes. Feel free to reopen if needed.

controllers/daemonset_reconcile.go
bpradipt: I'm trying to understand this whole code as this is the key piece of creating the DS to install.
My understanding is that this code is fetching the ExtensionImage and the CLI image and using an install script in the config yaml for installation.

The approach is creative in a way that you rely on images already present (extension and CLI). 
Personally, using embedded script in yaml is something I would avoid. It makes it hard to maintain and yamls are brittle. 

I think a better approach is to create a separate DS image having the Kata rpm, required CLI, config tomls and install script and use this for installation. 

And my reasoning is that it helps us to use static scanning tool for the script which can be removed later with compiled binary and the ability to replace the daemonset image for different use cases. 

Now creating a new image is an extra work w.r.to setting up Konflux pipelines. 

Also there could be difference of opinion here w.r.to creating a separate image vs the existing method.

Let's hear from @beraldoleal @gkurz @littlejawa and others and decide accordingly.

gcoon151: Appreciate the explanation and will await further review from the others. The approach we took was specifically because we have no method of publishing an image with the script we need or modifying the openshift extensions image to have it. Very open to doing this however everyone feels is appropriate.

controllers/deployment_mode_handler.go
```diff
@@ -0,0 +1,122 @@
...
+
+	if mode == DaemonSetOption {
+		r.Log.Info("Deployment mode will be set to DaemonSet")
+		r.DeploymentMode = DaemonSetMode
+		return nil
+	}
+
+	mcAvailable, err := r.isMCAvailable()
+	if err != nil {
+		r.Log.Info("Error checking if MC is available")
```
c3d:
```suggestion
		r.Log.Info("Error checking if MachineConfig is available")
```

controllers/deployment_mode_handler.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+
+	if mode == DaemonSetFallbackOption && !mcAvailable {
+		r.Log.Info("MC is not available, deployment mode will be set to DaemonSet")
```
c3d:
```suggestion
		r.Log.Info("MachineConfig is not available, deployment mode will be set to DaemonSet")
```

controllers/daemonset_reconcile.go
```diff
@@ -89,20 +99,20 @@ func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, erro
...
-		// error which happens when returning from a reconciliation that
-		// deleted our KataConfig by removing its finalizer.  So if the
-		// finalizer is missing the actual KataConfig object is probably
-		// already gone from the cluster, hence the error.
-		if updateErr != nil && controllerutil.ContainsFinalizer(r.kataConfig, kataConfigFinalizer) {
-			r.Log.Info("Updating KataConfig failed", "err", updateErr)
-			return ctrl.Result{}, updateErr
-		}
-		return res, err
+			res, err := r.processKataConfigDeleteRequestDaemonSet()
```
gkurz: Space damage. Be sure to run `make fmt` for each individual commit. It is also a good practice to run `make vet` as well.

controllers/fg_handler.go
```diff
@@ -40,22 +45,38 @@ const (
...
+		DeploymentModeOption:   DefaultFeatureGates.DeploymentModeOption,
 	}
 
 	cfgMap := &corev1.ConfigMap{}
 	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: FgConfigMapName,
 		Namespace: OperatorNamespace}, cfgMap)
 	if err == nil {
-		for feature, value := range cfgMap.Data {
-			fgStatus.FeatureGates[feature] = value == "true"
+		if value, ok := cfgMap.Data[ConfidentialFeatureGate]; ok {
```
c3d, line 57: Changing the `FeatureGateStatus` from containing a map to containing explicit values is what forces you to copy-paste the same code again and again here. This scales OK with three fields, but if we add more, that function will need to be revisited every time. Do we want that?

controllers/fg_handler.go
```diff
@@ -82,7 +99,7 @@ func (r *KataConfigOpenShiftReconciler) processFeatureGates() error {
 	// Check which feature gates are enabled in the FG ConfigMap and
 	// perform the necessary actions
 	if r.kataConfig.Spec.EnablePeerPods {
-		if IsEnabled(fgStatus, ConfidentialFeatureGate) {
```
c3d, line 85: Even with a map `IsEnabled` could be a method so that you get `fgStatus.IsEnabled(Confidential)`, which I find more readable.

config/peerpods/osc-configs-for-ds.yaml
```diff
@@ -0,0 +1,585 @@
...
+  50-kata-remote: |
```
beraldoleal: IIU, as it is, this DS mode would work only for peer-pods?

Unless I'm missing something, I feel this could have some overlap with @gkurz's work, on moving things to RPM side.

@gkurz is my understanding right? Would that work be ready for 1.10.2 ? Or post 1.10.2?

controllers/daemonset_reconcile.go
```diff
@@ -144,10 +154,315 @@ func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, erro
...
+	if contains(r.kataConfig.GetFinalizers(), kataConfigFinalizer) {
+		// Get the list of pods that might be running using kata runtime
+		err := r.listKataPods()
+		if err != nil {
+			r.setInProgressConditionToBlockedByExistingKataPods(err.Error())
+			updErr := r.Client.Status().Update(context.TODO(), r.kataConfig)
+			if updErr != nil {
+				return ctrl.Result{}, updErr
+			}
+			r.Log.Info("Kata PODs are present. Requeue for reconciliation ")
```
c3d:
```suggestion
			r.Log.Info("Kata pods are present. Requeue for reconciliation ")
```

controllers/daemonset_reconcile.go
```diff
@@ -144,10 +154,315 @@ func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, erro
...
+
+	// Create kata uninstall ds if it doesn't exist
+	// Save whether it was just created or not
+	kataUninstallDs, err := r.daemonSetForKataInstall(UninstallKata)
+	if err != nil {
+		r.Log.Error(err, "Failed getting DaemonSet for Kata uninstallation")
+		return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+	}
+	if err := controllerutil.SetControllerReference(r.kataConfig, kataUninstallDs, r.Scheme); err != nil {
+		r.Log.Error(err, "Failed setting ControllerReference for kata uninstallation DS")
```
c3d:
```suggestion
		r.Log.Error(err, "Failed setting ControllerReference for kata uninstallation DaemonSet")
```

controllers/daemonset_reconcile.go
```diff
@@ -144,10 +154,315 @@ func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, erro
...
+			}
+			r.Log.Info("Kata PODs are present. Requeue for reconciliation ")
+			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+		}
+	}
+
+	r.setInProgressConditionToUninstalling()
+
+	// Delete the kata install ds if it exists
+	kataInstallDs, err := r.daemonSetForKataInstall(InstallKata)
```
c3d: Since DS is an acronym, suggest `kataInstallDS` or `kataInstallDaemonSet` at your preference. Note that `Ds` easily reads as _multiple `D`s_

controllers/daemonset_reconcile.go
```diff
@@ -144,10 +154,315 @@ func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, erro
...
+			r.Log.Error(err, "error when deleting kata install Daemonset, try again")
+			return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 15}, err
+		}
+	}
+
+	isUninstallDaemonSetJustCreated := false
+
+	// Create kata uninstall ds if it doesn't exist
+	// Save whether it was just created or not
+	kataUninstallDs, err := r.daemonSetForKataInstall(UninstallKata)
```
c3d: Suggest `kataUninstallDaemonsSet` (or `kataUninstallDS`)

controllers/daemonset_reconcile.go
```diff
@@ -144,10 +154,315 @@ func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, erro
...
+	if err != nil {
+		r.Log.Error(err, "Failed getting DaemonSet for Kata uninstallation")
+		return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+	}
+	if err := controllerutil.SetControllerReference(r.kataConfig, kataUninstallDs, r.Scheme); err != nil {
+		r.Log.Error(err, "Failed setting ControllerReference for kata uninstallation DS")
+		return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+	}
+
+	foundKataDs := &appsv1.DaemonSet{}
```
c3d: suggest `foundKataDaemonSet` or `foundKataDS`

controllers/daemonset_reconcile.go
```diff
@@ -144,10 +154,315 @@ func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, erro
...
+			}
+			isUninstallDaemonSetJustCreated = true
+
+			kataSelector, err := r.getKataConfigNodeSelectorAsSelector()
+			if err != nil {
+				r.Log.Error(err, "could not get kata node selector, skipping waiting to uninstall labeling")
+			} else {
+				err = labelNodes(r.Client, kataSelector, map[string]string{kataDsInstallationLabel: string(KataDsWaitingToUninstall)})
+				if err != nil {
+					r.Log.Error(err, "could not add label waiting to uninstall to kata nodes")
```
c3d:
```suggestion
					r.Log.Error(err, "could not add label while waiting to uninstall from kata nodes")
```

controllers/daemonset_reconcile.go
```diff
@@ -144,10 +154,315 @@ func (r *KataConfigOpenShiftReconciler) daemonSetDeployment() (ctrl.Result, erro
...
+			err = r.Client.Create(context.TODO(), kataUninstallDs)
+			if err != nil {
+				r.Log.Error(err, "error when creating kata uninstallation daemonset")
+				return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+			}
+			isUninstallDaemonSetJustCreated = true
+
+			kataSelector, err := r.getKataConfigNodeSelectorAsSelector()
+			if err != nil {
+				r.Log.Error(err, "could not get kata node selector, skipping waiting to uninstall labeling")
```
c3d: Sorry, I really have trouble parsing this error. What does "skipping waiting" mean in that context? Do you mean "skipping" waiting or "skipping to some later step", or "waiting for (labelling)", or some combination of the above?

```suggestion
				r.Log.Error(err, "could not get kata node selector, will skip labelling of nodes to uninstall")
```

controllers/daemonset_reconcile.go
beraldoleal: We should stay aligned with how we deploy across the project to avoid mixing methods. Since [there’s already an effort to move the crio/kata config into the kata RPM](https://github.com/openshift/sandboxed-containers-operator/pull/941), this DS could follow the same path, otherwise it risks becoming a separate “thing” that’s harder to maintain and ship in future versions. I’d also avoid introducing extra config maps, even if used only internally, since they tend to add confusion.

Now, about the scripts, we already have `scripts/install-helpers/`, so the needed files could go there, and the image could consume them at build time as other images do.

If we have time to deploy this image in our pipeline, that’s the option I’d vote for (unless we have a better alternative).

@littlejawa, how hard do you think it would be to add this new component/image in Konflux? From what I see, hermetic builds shouldn’t be a problem, but do you see any other enterprise contract breaking here?

controllers/openshift_controller.go
```diff
@@ -2556,27 +2720,6 @@ func (r *KataConfigOpenShiftReconciler) disablePeerPods() error {
...
-	}
-
-	// Delete the mutating webhook
-	err = r.deleteMutatingWebhookConfig()
-	if err != nil {
-		r.Log.Info("Error in deleting mutating webhook for peerpods", "err", err)
-		return err
-	}
-
 	return nil
```
beraldoleal: Appreciate the effort into joining them, thanks.

However, this commit its changing parts that you introduced in this PR. It should be squashed with the appropriate commits.

controllers/daemonset_reconcile.go
beraldoleal: Please vide my other commit, avoid adding new commits if you are changing the logic on previous commits you introduced.

beraldoleal (commented):
Hi @Pacho20, thanks again for this work.

The comments shows the importance of your work, This is a big feature and you’re making solid progress. Keep going.

controllers/fg_handler.go
beraldoleal: @c3d, right, we asked him to add DeploymentModeOption over Slack to cover what the code was already doing and to support the "force mode". Since we don’t have just bools anymore, the extra code here is expected.

We’re also [moving from feature gates to config options](https://github.com/openshift/sandboxed-containers-operator/blob/devel/docs/CONFIG-OPTIONS.md). I don’t love the duplication either, but for now I’d stick with this. If we add more gates later, we can always revisit with a table-driven/helper pattern.

controllers/daemonset_reconcile.go
littlejawa: It's not that hard to create and release a new image as part of our operator, as long as you can build it with a regular Dockerfile.
I think we're talking about a relatively simple image here? Not requiring special access rights, like dm-verity?

Is this image going to ship the RPM? I thought the install script was installing from repositories?

controllers/image_generator.go
beraldoleal: closing this thread, since the new provider is part of #950

controllers/daemonset_reconcile.go
beraldoleal: Closing this as it was removed from this PR.

beraldoleal: Wrong, thread, reopening.

controllers/openshift_controller.go
beraldoleal: Closing this as it was removed from this PR.

controllers/daemonset_reconcile.go
beraldoleal: > Is this image going to ship the RPM? I thought the install script was installing from repositories?

No, the idea is that the RPM would be fetched from the extension the same way we do today. The image would have only the scripts and some artifacts.

controllers/deployment_mode_handler.go
beraldoleal: @gkurz, this was fixed, I'm closing the thread, but feel free to reopen if needed.

controllers/deployment_mode_handler.go
beraldoleal: Done.

controllers/fg_handler.go
beraldoleal: I will close this thread, feel free to reopen if we came up with a better idea.

controllers/daemonset_reconcile.go
```diff
@@ -0,0 +1,1085 @@
...
+		r.kataConfig.Status.KataNodes.Uninstalling = append(r.kataConfig.Status.KataNodes.Uninstalling, node.GetName())
+	case string(KataUninstalled):
+	default:
+		r.kataConfig.Status.KataNodes.WaitingToUninstall = append(r.kataConfig.Status.KataNodes.WaitingToUninstall, node.GetName())
+	}
+
+	return nil
+}
+
+func (r *KataConfigOpenShiftReconciler) GetExtensionImage() (string, error) {
```
beraldoleal: I just missed the reason we have `GetExentionImage()` and `GetCliImage()` methods. Can't we avoid both methods and calling directly the `GetImageForComponent` ?

controllers/openshift_controller.go
```diff
@@ -165,7 +167,17 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
...
 		// is underway.
 		if r.kataConfig.GetDeletionTimestamp() != nil && !r.isInstalling() && !r.isUpdating() {
-			res, err := r.processKataConfigDeleteRequest()
+			var res ctrl.Result
+
+			switch r.DeploymentMode {
+			case MachineConfigMode:
+				res, err = r.processKataConfigDeleteRequest()
+			case DaemonSetMode:
+				res, err = r.processKataConfigDeleteRequestDaemonSet()
```
beraldoleal, line 176: What about the `DaemonSetFallback` ?

controllers/openshift_controller.go
```diff
@@ -181,7 +193,18 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
...
 		}
 
-		res, err := r.processKataConfigInstallRequest()
+		var res ctrl.Result
+
+		switch r.DeploymentMode {
+		case MachineConfigMode:
+			res, err = r.processKataConfigInstallRequest()
+		case DaemonSetMode:
+			res, err = r.processKataConfigInstallRequestDaemonSet()
```
beraldoleal, line 202: Same here, what about the `DaemonSetFallback`?

controllers/openshift_controller.go
```diff
@@ -1006,18 +1031,9 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
-		if err != nil {
-			r.setInProgressConditionToBlockedByExistingKataPods(err.Error())
-			updErr := r.Client.Status().Update(context.TODO(), r.kataConfig)
-			if updErr != nil {
-				return ctrl.Result{}, updErr
-			}
-			r.Log.Info("Kata PODs are present. Requeue for reconciliation ")
-			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
-		}
+	res, err := r.checkDeletionEligibility()
```
beraldoleal, line 1119: Nice move!

controllers/openshift_controller.go
```diff
@@ -2489,6 +2576,47 @@ func (r *KataConfigOpenShiftReconciler) disablePeerPods() error {
...
+		_ = r.deletePeerPodsMC()
+	}
+
+	if r.DeploymentMode == DaemonSetMode {
+		// We want to make sure that the osc-config-sync ds removal successfully started
+		// So we will try again in case of an error
+		err = r.deletePeedPodsConfigDaemonSet()
+		if err != nil {
+			return err
+		}
```
beraldoleal, line 2474: What about the fallback mode here ?

controllers/daemonset_reconcile.go
beraldoleal: I just saw your latest changes, and I'm happy with the split into methods and the reuse for both reconciles. It reduces a lot the duplication. And also, break our existing reconcile loop into smaller methods. Awesome.

I'm happy with this, so I'm closing this thread, @bpradipt feel free to reopen if you still see some room for improvements.

config/peerpods/osc-configs-for-ds.yaml
beraldoleal: OK, I just got this answer, the agreement so far is that we keep it here for 1.10.2, but we should move this to the RPM for 1.11.

So we can close this thread.

controllers/deployment_mode_handler.go
gcoon151: @gkurz we think we resolved this. Can you confirm?

controllers/deployment_mode_handler.go
gcoon151: @gkurz we think we resolved this. Can you confirm?

controllers/daemonset_reconcile.go
gcoon151: I think was fixed when it moved to here https://github.com/openshift/sandboxed-containers-operator/pull/950 the commit doesn't exist that has this error now.

controllers/openshift_controller.go
gcoon151: When the FG is set to DaemonSetFallback, this field is automatically set (L87 deployment_mode_handler.go) to either MachineConfigMode or DaemonSetMode depending on what is there. What's the suggestion?

controllers/openshift_controller.go
gcoon151: When the FG is set to DaemonSetFallback, this field is automatically set (L87 deployment_mode_handler.go) to either MachineConfigMode or DaemonSetMode depending on what is there. What's the suggestion?

controllers/openshift_controller.go
gcoon151: same as above. This is handled in deployment_mode_handler.go

controllers/daemonset_reconcile.go
Pacho20: We probably don't need the` GetCliImage()` function. I initially created both functions to allow for additional logic if needed. As you can see, `GetExtensionImage()` has a `TODO` comment because I wasn't sure whether this is the only extension image we need. Other extension images exist for different distros. If RHEL and CoreOS are sufficient, I can remove both functions and call `GetImageForComponent()` directly.

controllers/deployment_mode_handler.go
```diff
@@ -0,0 +1,123 @@
...
+	mc := &unstructured.Unstructured{}
+	mc.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   machineConfigGroup,
+		Version: machineConfigVersion,
+		Kind:    machineConfigKind,
+	})
+
+	// Attempt to GET any MachineConfig to verify if the GVK is known.
+	// If the CRD isn’t installed, it will return a NoMatchError.
+	err := r.Client.Get(context.Background(), client.ObjectKey{Name: machineConfigName}, mc)
```
gkurz: This is the only sensible user for `machineConfigName`. No deep to define a constant for that. Let's use the literal string `"dummy-machine-config"` directly here.

controllers/fg_handler.go
gkurz: Such a change should really happen in a preceding commit.

controllers/deployment_mode_handler.go
Pacho20: @gkurz this should now be resolved. Can you confirm?

controllers/fg_handler.go
Pacho20: I've created a new commit. Could you check if it's acceptable for you?

controllers/deployment_mode_handler.go
Pacho20: Done

controllers/deployment_mode_handler.go
gkurz: AFAICT the log still has `MC`.

controllers/fg_handler.go
gkurz: > I've created a new commit. Could you check if it's acceptable for you?

Yes, thanks !

controllers/deployment_mode_handler.go
Pacho20: Sorry, I don’t know how I missed that. It should be good now.

controllers/daemonset_reconcile.go
beraldoleal: Please, call directly.

controllers/openshift_controller.go
beraldoleal: I think I got it now. thanks. closing the thread.

beraldoleal:
Some required test are failing, but might be because of the conflict. Could you rebase, please?

controllers/daemonset_reconcile.go
Pacho20: Done.

scripts/kata-install/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN mkdir -p /files
+
+ADD 50-kata-remote configuration-remote.toml /files/
+
+RUN mkdir -p /scripts
+
+ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh /scripts/
+
+RUN curl -sSL "https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64" -o "/usr/local/bin/umoci" &&\
```
beraldoleal, line 11: Future issue: This will be an issue for 1.11.0 (when we migrate this to Konflux). We need to make the builds hermetic.

gkurz: It is possible to ask hermeto to prefetch arbitrary files for hermetic builds.

https://github.com/hermetoproject/hermeto/blob/main/docs/generic.md

spotlesstofu (approved):
We can merge

controllers/openshift_controller.go
```diff
@@ -213,9 +234,11 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 			}
 		}
 
-		err = r.processLogLevel(r.kataConfig.Spec.LogLevel)
-		if err != nil {
-			res = ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}
+		if r.DeploymentMode == MachineConfigMode {
```
gkurz, line 240: Can you clarify why this isn't wanted in `DaemonConfigMode` ?

controllers/openshift_controller.go
gkurz: Yes and it would be even nicer in a preparatory patch. There are far too many changes hitting `controllers/openshift_controller.go`... this makes the PR a lot harder to review.

controllers/openshift_controller.go
```diff
@@ -1181,15 +1195,9 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
 		}
 	}
 
-	ds := r.processDaemonsetForMonitor()
-	err = r.Client.Delete(context.TODO(), ds)
+	err = r.deleteDaemonsetForMonitor()
```
gkurz, line 1235: Please refactor the code in a separate commit.

controllers/openshift_controller.go
```diff
@@ -1199,74 +1207,119 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
-			r.setInProgressConditionToPodVMImageDeleted()
-			r.Log.Info("PodVM Image deleted successfully")
-
-		case UnsupportedPodVMImageProvider:
-			r.setInProgressConditionToPodVMImageUnsupportedProvider()
-			r.Log.Info("unsupported cloud provider, skipping image deletion")
-
-		case RequeueNeeded:
-			r.setInProgressConditionToPodVMImageDeleting()
+		err := r.deletePodVMImage()
```
gkurz: Please refactor the code in a separate commit.

controllers/openshift_controller.go
```diff
@@ -1199,74 +1207,119 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
-				// We requeue only if there is an error.
-				return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
-			}
-			// If there's no error, log and continue
-			r.Log.Info("Image deletion failed. Check logs for more details")
-
-		case ImageDeletionStatusUnknown:
-			r.setInProgressConditionToPodVMImageDeletionUnknown()
-			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+	err = r.deleteScc()
```
gkurz, line 1257: Please refactor the code in a separate commit.

gkurz (changes requested):
Hi @Pacho20 !

Since this affects the common code quite a lot, I want an easy review where I'm confident I can spot errors. The last commit mixes a lot of refactoring with the functional changes for the `DaemonSetMode`. Please move all the refactoring to preparatory commits.

gkurz:
> We can merge

I'd really prefer the last commit to be broken down before we merge.

Pacho20:
> > We can merge
> 
> I'd really prefer the last commit to be broken down before we merge.

@gkurz I've created a separate commit for the refactoring part. I hope this is what you had in mind.

controllers/openshift_controller.go
Pacho20: The processLogLevel method creates a ContainerRuntimeConfig managed by the MCO. At the end of the day, it just results in a single file being created in CRI-O's config folder. I added this file creation step to the installation DaemonSet's script so no need to call this method for the DaemonSet mode.

controllers/openshift_controller.go
```diff
@@ -1184,51 +1184,10 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
-				// We requeue only if there is an error.
-				return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
-			}
-			// If there's no error, log and continue
-			r.Log.Info("Image deletion failed. Check logs for more details")
-
-		case ImageDeletionStatusUnknown:
-			r.setInProgressConditionToPodVMImageDeletionUnknown()
+		err := r.deletePodVMImage()
+		if err != nil {
```
gkurz, line 1252: We can get there from the `RequeueNeeded` branch in `deletePodVMImage()` with `err` being `nil` and we won't requeue... `deletePodVMImage()` should return `(*ctrl.Result, error)` and you should test either one being not `nil`.

controllers/openshift_controller.go
```diff
@@ -1341,120 +1375,11 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
-		r.resetInProgressCondition()
-		err := r.createRuntimeClass(kataRuntimeClassName, kataRuntimeClassCpuOverhead, kataRuntimeClassMemOverhead)
-		if err != nil {
-			// Give sometime for the error to go away before reconciling again
-			return reconcile.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
-		}
-		r.Log.Info("create Scc")
-		err = r.createScc()
+		err = r.postKataInstallation()
 		if err != nil {
```
gkurz, line 1413: Same issue as with `deletePodVMImage()`.

controllers/openshift_controller.go
```diff
@@ -1184,51 +1184,13 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequest() (ctrl.R
...
-			// If there's no error, log and continue
-			r.Log.Info("Image deletion failed. Check logs for more details")
-
-		case ImageDeletionStatusUnknown:
-			r.setInProgressConditionToPodVMImageDeletionUnknown()
+		res, err := r.deletePodVMImage()
+		if res != nil {
+			return *res, err
+		}
+		if err != nil {
```
gkurz, line 1252: This cannot happen AFAICT : all paths that  could return an error also return a reschedule request.

controllers/openshift_controller.go
```diff
@@ -1344,120 +1378,13 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
...
-			return ctrl.Result{}, err
+		res, err := r.postKataInstallation()
+		if res != nil {
+			return *res, err
 		}
-		r.Log.Info("controller reference set for the monitor daemonset")
-
-		foundDs := &appsv1.DaemonSet{}
-		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: ds.Name, Namespace: ds.Namespace}, foundDs)
 		if err != nil {
```
gkurz, line 1413: Same remark as with `deletePodVMImage()` : all return paths of `postKataInstallation()` that return an error also return a reschedule request.

controllers/openshift_controller.go
Pacho20: Yes, that's true. It was more of a safety measure in case someone changes the methods. What would you suggest instead?

controllers/openshift_controller.go
gkurz: You're right that the current implementation of `processLogLevel()` does MCO stuff that isn't suitable for your needs but the DaemonSet code still requires and processes the log level, right ?

Ideally we should then have a `processLogLevelDaemonSet()` function to do that or alternatively call the original `processLogLevel()` from another location that isn't MCO specific.

This might generate too many changes at this stage though so I'm fine to keep the `if` for now. Please add a `// FIXME` comment that this should be revisited later.

Pacho20: I wanted to create a separate DaemonSet to follow the reconcile logic, but creating one just for a single file seemed excessive. The operator already creates at least four different DaemonSets. If that's not a concern, we could introduce a processLogLevelDaemonSet() function in a future iteration down the road.

controllers/openshift_controller.go
gkurz: I don't have an alternative to provide at this stage. We should carefully review all these return branches first.
So let's keep the safety measure around with a `// FIXME : dead code, revisit deletePodVMImage() return paths` comment.

controllers/openshift_controller.go
gkurz: and `// FIXME : dead code. Revisit ...` comment as well here.

beraldoleal (approved):
After addressing @gkurz's comments, LGTM.

Thanks again, @Pacho20. Well done.

scripts/kata-install/configuration-remote.toml
gkurz, line 1: Yet another instance of `scripts/kata-install/configuration-remote.toml` in the tree... fortunately we're working on moving the kata and CRI-O conf files to the RPM for OSC 1.11 since they are static.

scripts/kata-install/50-kata-remote
gkurz, line 1: As said in another comment, we'll get this file from the RPM in the future.

gkurz (approved):
/lgtm

Thanks for the very good work and patience @Pacho20 !

scripts/kata-install/configuration-remote.toml
Pacho20: This is needed for the 1.10 release, but I’m looking forward to deleting these files and updating the DaemonSet for 1.11.

gkurz:
Fixes https://issues.redhat.com/browse/KATA-3962

gkurz:
Fixes https://issues.redhat.com/browse/KATA-3963

PR #827, opened 2025-07-24
docs: FEATURE-GATES -> CONFIG-OPTIONS

beraldoleal:
Following community discussions, there's a desire to clarify the distinction between feature gates and config options. This update aligns the documentation with our intended direction for upcoming releases and avoids ambiguity.

Note: Some internal structs still use the old terminology and are not updated in this commit. We plan to address those separately.

We are updating our feature gates definition and from now on we are using the "config options" terminology. This will not change how users enable/disable features for this release, so probably we should not mention this change in the changelog.


bpradipt (approved):
/lgtm

gkurz (approved):
/lgtm

Thanks @beraldoleal !

beraldoleal:
@littlejawa, I tried to re-trigger the failing check, but not working. Safe to ignore since is not required?

littlejawa:
The error is about the dm-verity image. From what I can see in the log, the bundle in your branch references a pretty old version of it. That's why it's failing.
You would probably have it passing if your rebase.

Now since you're just changing docs, I don't think it's worth running through another full rebuild/testing (the prow tests take forever :-( ).
Let's ignore it.

PR #829, opened 2025-07-29
chore: update digest for osc-dm-verity-image

snir911:

**- What I did**

**- How to verify it**

### DIFF
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -400,7 +400,7 @@ spec:
                 - name: RELATED_IMAGE_PODVM_PAYLOAD
                   value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9@sha256:73a82ac1c252f068f9d046a36dfca7932202ad728f94fc575a332d8707650905
                 - name: RELATED_IMAGE_PODVM_OCI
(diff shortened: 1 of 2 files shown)


spotlesstofu (approved):

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -400,7 +400,7 @@ spec:
                 - name: RELATED_IMAGE_PODVM_PAYLOAD
                   value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9@sha256:73a82ac1c252f068f9d046a36dfca7932202ad728f94fc575a332d8707650905
                 - name: RELATED_IMAGE_PODVM_OCI
-                  value: registry.redhat.io/openshift-sandboxed-containers/osc-dm-verity-image@sha256:238ffe72da6977c57b93a329282b2134b60ad71037042dcc97ff609bfc6691eb
+                  value: registry.redhat.io/openshift-sandboxed-containers/osc-dm-verity-image@sha256:68c7f01fa01dd14da0d95f12945f3a79b9d7e1bd0d9afceb086bf4e364de9fd9
```
gkurz, line 403: Where does the SHA256 come from ? How can one verify this is the expected one ?

spotlesstofu: Look for a tag matching the latest git commit. That tag has this digest.

PR #836, opened 2025-07-30
Enable Confidential Containers on Baremetal

beraldoleal:
This adds initial support for Confidential on baremetal. Until now, Confidential was only supported by peerpods. With this change, we are introducing the Confidential feature to baremetal as well.
    
Our consensus so far, is to use the same config option "confidential" to enable Confidential features in both cases.
    
We are also reusing the same logic for KataConfigPoolSelector and checkNodeEligibility, just as we have been doing so far.

One aspect we are evaluating is enabling the MachineConfigs on all nodes alongside CRIO, regardless of whether the machine has support. The RuntimeClass would be responsible for filtering machines according to their capabilities.

RPM work is still needed.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -20,6 +20,14 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - secrets
   verbs:
   - create
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -1,25 +1,53 @@
 package controllers
 
 import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// When the feature is enabled, handleFeatureConfidential sets config maps to confidential values.
+const (
+	// kata-cc runtime class for CoCo BM
+	kataCCRuntimeClassName        = "kata-cc"
+	kataCCRuntimeClassCpuOverhead = "0.25"
+	kataCCRuntimeClassMemOverhead = "350Mi"
+
+	// TEE node labels
+	intelTDXNodeLabel = "intel.feature.node.kubernetes.io/tdx"
+	amdSNPNodeLabel   = "amd.feature.node.kubernetes.io/snp"
+
+	// RuntimeClass handlers for TEE
+	kataCCIntelHandler = "kata-cc-intel"
+	kataCCAmdHandler   = "kata-cc-amd"
+)
+
+// When the feature is enabled, handleFeatureConfidential configures confidential computing support.
 //
-// Changes the ImageConfigMap, so that the image creation job will create a confidential image.
-// This will happen at the first reconciliation loop, before the image creation job starts.
+// For peer pods: sets ImageConfigMap and peer pods configMap to enable confidential images and CVM support.
+// For baremetal: creates kata-cc runtime classes with TEE-specific handlers (Intel TDX or AMD SNP).
 //
-// Changes the peer pods configMap to enable confidential.
-// This will happen likely after several reconciliation loops, because it has prerequisites:
-//
-//   - Peer pods must be enabled in the KataConfig.
-//   - The peer pods config map must exist.
-//
-// When the feature is disabled, handleFeatureConfidential resets the config maps to non-confidential values.
+// When the feature is disabled, handleFeatureConfidential resets config maps and deletes runtime classes.
 func (r *KataConfigOpenShiftReconciler) handleFeatureConfidential(state FeatureGateState) error {
+	if r.kataConfig.Spec.EnablePeerPods {
+		if err := r.handleConfidentialPeerPods(state); err != nil {
+			return err
+		}
+	}
 
+	if err := r.handleConfidentialBaremetal(state); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// handleConfidentialPeerPods configures confidential computing for peer pods deployments.
+// It manages ImageConfigMap and peer pods configMap to control confidential images and CVM support.
+func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state FeatureGateState) error {
 	// ImageConfigMap
-
 	if err := InitializeImageGenerator(r.Client); err != nil {
 		return err
 	}
@@ -56,8 +84,6 @@ func (r *KataConfigOpenShiftReconciler) handleFeatureConfidential(state FeatureG
 		}
 	}
 
-	// peer pods config
-
 	// Patch peer pods configMap, if it exists.
 	var peerpodsCMData map[string]string
 	if state == Enabled {
@@ -75,4 +101,82 @@ func (r *KataConfigOpenShiftReconciler) handleFeatureConfidential(state FeatureG
 	}
 
 	return nil
+}
+
+// handleConfidentialBaremetal configures confidential computing for baremetal deployments.
+// It manages kata-cc runtime classes with TEE-specific handlers (Intel TDX or AMD SNP).
+func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state FeatureGateState) error {
+	if state == Enabled {
+		r.Log.Info("Creating " + kataCCRuntimeClassName + " runtime class for confidential containers")
+
+		handler, nodeLabel, err := r.computeTEEHandlerAndLabel()
+		if err != nil {
+			// If peer pods is enabled, just warn and skip baremetal coco
+			if r.kataConfig.Spec.EnablePeerPods {
+				r.Log.Info("WARNING: No TEE hardware detected, skipping baremetal confidential containers (peer pods CVM will handle confidential workloads)", "err", err)
+				return nil
+			}
+			// If peer pods disabled, this is an error - user wants baremetal coco but no hardware
+			r.Log.Info("failed to detect TEE platform", "err", err)
+			return err
+		}
+
+		// Create kata-cc runtime class restricted to the detected TEE subset
+		err = r.createRuntimeClass(kataCCRuntimeClassName, kataCCRuntimeClassCpuOverhead, kataCCRuntimeClassMemOverhead, handler, nodeLabel)
+		if err != nil {
+			r.Log.Info("Error creating "+kataCCRuntimeClassName+" runtime class", "err", err)
+			return fmt.Errorf("Error creating "+kataCCRuntimeClassName+" runtime class: %w", err)
+		}
+
+	} else {
+		r.Log.Info("Deleting " + kataCCRuntimeClassName + " runtime class for confidential containers")
+
+		// Delete kata-cc runtime class
+		err := r.deleteRuntimeClass(kataCCRuntimeClassName)
+		if err != nil {
+			r.Log.Info("Error deleting "+kataCCRuntimeClassName+" runtime class", "err", err)
+			return fmt.Errorf("Error deleting "+kataCCRuntimeClassName+" runtime class: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (r *KataConfigOpenShiftReconciler) computeTEEHandlerAndLabel() (string, string, error) {
+	selector, err := r.getKataConfigNodeSelectorAsSelector()
+	if err != nil {
+		return "", "", fmt.Errorf("failed to build node selector: %w", err)
+	}
+
+	nodes := &corev1.NodeList{}
+	listOpts := []client.ListOption{
+		client.MatchingLabelsSelector{Selector: selector},
+	}
+	if err := r.Client.List(context.TODO(), nodes, listOpts...); err != nil {
+		return "", "", fmt.Errorf("failed to list nodes: %w", err)
+	}
+
+	var hasIntelTDX bool
+	var hasAmdSNP bool
+	for _, n := range nodes.Items {
+		if v, ok := n.Labels[intelTDXNodeLabel]; ok && v == "true" {
+			hasIntelTDX = true
+		}
+		if v, ok := n.Labels[amdSNPNodeLabel]; ok && v == "true" {
+			hasAmdSNP = true
+		}
+	}
+
+	if hasIntelTDX && hasAmdSNP {
+		return "", "", fmt.Errorf("multiple TEE platforms detected; only one per cluster supported")
+	}
+
+	if hasIntelTDX {
+		return kataCCIntelHandler, intelTDXNodeLabel, nil
+	}
+	if hasAmdSNP {
+		return kataCCAmdHandler, amdSNPNodeLabel, nil
+	}
+
+	return "", "", fmt.Errorf("no TEE platform labels found (expected %s or %s)", intelTDXNodeLabel, amdSNPNodeLabel)
 }
--- a/controllers/migrate.go
+++ b/controllers/migrate.go
@@ -16,10 +16,13 @@ import (
 	"context"
 
 	corev1 "k8s.io/api/core/v1"
+	nodeapi "k8s.io/api/node/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // migratePeerPodsLimit moves the PeerPodConfig "Limit" value to peer-pods-cm
@@ -79,5 +82,34 @@ func (r *KataConfigOpenShiftReconciler) migratePeerPodsLimit() error {
 	}
 
 	r.Log.Info("Successfully migrated PeerPodConfig Limit to peer-pods-cm and deleted PeerPodConfig")
+	return nil
+}
+
+// ensureRuntimeClassFinalizers adds finalizers to existing runtime classes that don't have them
+func (r *KataConfigOpenShiftReconciler) ensureRuntimeClassFinalizers() error {
+	runtimeClasses := []string{
+		kataCCRuntimeClassName,
+		peerpodsRuntimeClassName,
+		kataRuntimeClassName,
+	}
+
+	for _, name := range runtimeClasses {
+		rc := &nodeapi.RuntimeClass{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Name: name}, rc)
+		if k8serrors.IsNotFound(err) {
+			continue
+		}
+		if err != nil {
+			return err
+		}
+
+		if !controllerutil.ContainsFinalizer(rc, runtimeClassFinalizerName) {
+			r.Log.Info("Adding finalizer to existing runtime class", "runtimeClass", name)
+			controllerutil.AddFinalizer(rc, runtimeClassFinalizerName)
+			if err := r.Client.Update(context.TODO(), rc); err != nil {
+				return err
+			}
+		}
+	}
 	return nil
 }
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -772,7 +772,7 @@ func (r *KataConfigOpenShiftReconciler) createDaemonsetForMonitor() error {
 	return nil
 }
 
-func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName string, cpuOverhead string, memoryOverhead string) error {
+func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName string, cpuOverhead string, memoryOverhead string, handler string, additionalNodeLabel string) error {
 
 	rc := func() *nodeapi.RuntimeClass {
 		rc := &nodeapi.RuntimeClass{
@@ -783,7 +783,7 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
 			ObjectMeta: metav1.ObjectMeta{
 				Name: runtimeClassName,
 			},
-			Handler: runtimeClassName,
+			Handler: handler,
 			Overhead: &nodeapi.Overhead{
 				PodFixed: corev1.ResourceList{
 					corev1.ResourceCPU:    resource.MustParse(cpuOverhead),
@@ -793,6 +793,11 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
 		}
 
 		nodeSelector := r.getNodeSelectorAsMap()
+
+		// Add additional node label if provided
+		if additionalNodeLabel != "" {
+			nodeSelector[additionalNodeLabel] = "true"
+		}
 
 		rc.Scheduling = &nodeapi.Scheduling{
 			NodeSelector: nodeSelector,
@@ -2197,7 +2202,7 @@ func (r *KataConfigOpenShiftReconciler) deleteScc() error {
 func (r *KataConfigOpenShiftReconciler) postKataInstallation() (*ctrl.Result, error) {
 	r.Log.Info("create runtime class")
 	r.resetInProgressCondition()
-	err := r.createRuntimeClass(kataRuntimeClassName, kataRuntimeClassCpuOverhead, kataRuntimeClassMemOverhead)
+	err := r.createRuntimeClass(kataRuntimeClassName, kataRuntimeClassCpuOverhead, kataRuntimeClassMemOverhead, kataRuntimeClassName, "")
 	if err != nil {
 		return &ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
 	}
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -834,6 +834,34 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
 	return nil
 }
 
+func (r *KataConfigOpenShiftReconciler) deleteRuntimeClass(runtimeClassName string) error {
+
+	foundRc := &nodeapi.RuntimeClass{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: runtimeClassName}, foundRc)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	if err := r.Client.Delete(context.TODO(), foundRc); err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	for i, name := range r.kataConfig.Status.RuntimeClasses {
+		if name == runtimeClassName {
+			r.kataConfig.Status.RuntimeClasses = append(r.kataConfig.Status.RuntimeClasses[:i], r.kataConfig.Status.RuntimeClasses[i+1:]...)
+			break
+		}
+	}
+
+	return nil
+}
+
 // "KataConfigNodeSelector" in the names of the following couple of helper
 // functions refers to the value of KataConfig.spec.kataConfigPoolSelector,
 // i.e. the original selector supplied by the user of KataConfig.
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -127,6 +127,12 @@ func (r *KataConfigOpenShiftReconciler) Reconcile(ctx context.Context, req ctrl.
 		return ctrl.Result{}, err
 	}
 
+	err = r.ensureRuntimeClassFinalizers()
+	if err != nil {
+		r.Log.Info("Failed to ensure runtime class finalizers", "err", err)
+		return ctrl.Result{}, err
+	}
+
 	err = r.processFeatureGates()
 	if err != nil {
 		r.Log.Info("Unable to process feature gates", "err", err)
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -781,7 +781,8 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
 				Kind:       "RuntimeClass",
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name: runtimeClassName,
+				Name:       runtimeClassName,
+				Finalizers: []string{runtimeClassFinalizerName},
 			},
 			Handler: handler,
 			Overhead: &nodeapi.Overhead{
--- a/controllers/runtimeclass_controller.go
+++ b/controllers/runtimeclass_controller.go
@@ -1,0 +1,139 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"time"
+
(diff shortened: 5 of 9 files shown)


beraldoleal:
@gkurz , while still a working in progress, feel free to validate this approach with our discussions. Also keep in mind MachineConfig files are just boilerplate at this moment as example.

beraldoleal:
@bpradipt @gkurz do you guys see any issues with this direction?

controllers/openshift_controller.go
```diff
@@ -2405,26 +2409,15 @@ func (r *KataConfigOpenShiftReconciler) createAuthJsonSecret() error {
...
-	}
-
-	//Create MachineConfig for kata-remote hyp config toml
-	kataConfigMachineConfigFilePath := filepath.Join(peerpodsMachineConfigPathLocation, peerpodsKataRemoteMachineConfigYaml)
-	err = r.createMcFromFile(kataConfigMachineConfigFilePath)
-	if err != nil {
-		r.Log.Info("Error in creating kata remote configuration.toml MachineConfig", "err", err)
-		return err
+// Create MachineConfigs from a list of YAML files
+func (r *KataConfigOpenShiftReconciler) createMachineConfigsFromFiles(yamlFiles []string) error {
```
gkurz: I don't find this function very useful actually... in the end it just calls `createMcFromFile` for a list of files that the caller must provide. What's the goal ?

If it is to reduce the number of LoC, I'd rather consolidate the `filepath.join` in `createMcFromFile` and add an  extra location path argument  and have a `createMcFromFile(someMachineConfigPathLocation, someMachineConfigYaml)` for all MC that need to be created instead of the array looping.

controllers/openshift_controller.go
```diff
@@ -2426,6 +2449,66 @@ func (r *KataConfigOpenShiftReconciler) createMachineConfigsFromFiles(yamlFiles
...
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "node.k8s.io/v1",
+				Kind:       "RuntimeClass",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: runtimeClassName,
+			},
+		}
+
+		err := r.Client.Delete(context.TODO(), rc)
```
gkurz: This isn't needed because `createRuntimeClass` does :

```
	// Set Kataconfig r.kataConfig as the owner and controller
	if err := controllerutil.SetControllerReference(r.kataConfig, rc, r.Scheme); err != nil {
		return err
	}
```

This causes the RuntimeClass object to be automatically deleted when KataConfig goes away.

controllers/openshift_controller.go
gkurz: I'm actually not a bit fan of having so many MachineConfigs to deploy the CRI-O config files. They have a fixed content and could be easily shipped with the RPM instead of all the boiler plate to do that from the controller.

controllers/openshift_controller.go
beraldoleal: For public visibility, we decided to remove the runtimeclass when confidential is disabled. This ensures users cannot deploy confidential pods. The same applies to kata-remote and any other runtimeclass except kata.

controllers/confidential_handler.go
```diff
@@ -57,19 +71,49 @@ func (r *KataConfigOpenShiftReconciler) handleFeatureConfidential(state FeatureG
...
+
+		// Create kata-cc-intel runtime class for Intel TDX
+		err := r.createRuntimeClass(kataCCIntelRuntimeClassName, kataCCIntelRuntimeClassCpuOverhead, kataCCIntelRuntimeClassMemOverhead, intelTDXNodeLabel)
+		if err != nil {
+			r.Log.Info("Error creating kata-cc-intel runtime class", "err", err)
+			return err
+		}
+
+		// Create kata-cc-amd runtime class for AMD SNP
+		err = r.createRuntimeClass(kataCCAMDRuntimeClassName, kataCCAMDRuntimeClassCpuOverhead, kataCCAMDRuntimeClassMemOverhead, amdSNPNodeLabel)
```
bpradipt: Do we really want to create arch specific runtimeClass or just use `kata-cc` as the runtimeClass with handlers as `kata-cc-{arch}` ?

beraldoleal: I’m ok with doing a single RC (kata-cc), but iiuc we can’t have multiple handlers since a RuntimeClass only takes one. So if we stick with one RC, the RPM would need to handle it by dropping all the configs and setting up the right configuration.toml (or symlink) depending on the platform.

Ex: on a TDX node it points to configuration-qemu-tdx.toml, on an SNP node it points to configuration-qemu-snp.toml. That way kata-cc always just uses the kata handler, and the node decides which TEE to run.

Users could still use node affinity if they want to pin a pod to Intel vs AMD.

bpradipt: > Ex: on a TDX node it points to configuration-qemu-tdx.toml, on an SNP node it points to configuration-qemu-snp.toml. That way kata-cc always just uses the kata handler, and the node decides which TEE to run.

Right. This was in my mind. The premise was that this is easier from an end-user standpoint deploying confidential workload in the cluster. Wdyt ?

beraldoleal: I'm ok with that approach. 

However, @gkurz mentioned something about having multiple runtimeclasses but also having a single one to be the "smart" one. Its still not clear to me, if we really need those, if the users can choose the nodes by using the node affinity labels.

So I will just wait a bit for his comments, to see if he has something against this.

gkurz: I'm now thinking that the operator should focus on a single "smart" runtime class `kata-cc` for a better end user experience. If the admin wants to add TEE-specific runtime classes, they can still do it manually since the RPM provides the CRI-O handlers. We can document how to do that.

beraldoleal: Thanks @gkurz . As we agreed today during the call, I will change this PR to use a single runtime class, However, since this is for the TP in November, we are assuming for now this will handle only single TEE per cluster, not multiple TEE platforms on the same cluster.

controllers/openshift_controller.go
```diff
@@ -880,7 +880,7 @@ func (r *KataConfigOpenShiftReconciler) createScc() error {
 	return nil
 }
 
-func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName string, cpuOverhead string, memoryOverhead string) error {
+func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName string, cpuOverhead string, memoryOverhead string, additionalNodeLabel string) error {
```
gkurz: Since the additional label will come from NFD, I'm wondering if this is the right way to go. Shouldn't we rather adapt our existing CheckNodeEligibility logic ?

beraldoleal: IIUC, if TEE label logic moves into CheckNodeEligibility, only TEE nodes are eligible for the Kata install.

With the current approach, the pool stays broad and the RuntimeClass narrows via TEE label, so kata, kata-remote and kata-cc can coexist.

Does that make sense? Or you have something else in mind?

beraldoleal:
@bpradipt @wainersm @gkurz ready for a second round. let me know if you have any comments.

controllers/openshift_controller.go
gkurz: I see CheckNodeEligibility as way to indicate that our runtime classes should take node properties into account, otherwise we assume that the admin is responsible to use kata on proper HW.

Each runtime class would have its specific selector :

- kata-remote : can run everywhere where kata was deployed (kata-oc)
- kata : can run on kata-oc hosts where KVM is available
- kata-cc: can run on kata-oc hosts where KVM  and TEE are available

We might have to come up with some NFD rules to make this fully functional.

beraldoleal: Yes, I think CheckNodeEligibility was added for the simpler (non‑confidential) case, so it probably needs a revisit now that TEE is in play.

My concern is that ... IIUC, currently it gates the kata installation pool (who gets node-role.kubernetes.io/kata-oc) AND, because our RuntimeClasses select kata-oc, it also constrains where RCs can run indirectly.

If we fold TEE into CheckNodeEligibility, won't we be shrinking the Kata installation footprint to TEE nodes only? Don't we need to think also in terms of kata installation as well?

controllers/openshift_controller.go
```diff
@@ -1481,6 +1509,33 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 	return ctrl.Result{}, nil
 }
 
+// Delete runtime class
+func (r *KataConfigOpenShiftReconciler) deletRuntimeClass(runtimeClassName string) error {
```
gkurz: Left over ?

controllers/confidential_handler.go
```diff
@@ -57,22 +77,98 @@ func (r *KataConfigOpenShiftReconciler) handleFeatureConfidential(state FeatureG
...
-	}
-	if err := updateConfigMap(r.Client, r.Log, peerpodsCMName, OperatorNamespace, peerpodsCMData); err != nil {
-		if k8serrors.IsNotFound(err) {
-			// When feature is Enabled: ConfigMap doesn't exist yet, will try again at the next reconcile run.
-			// Else: Nothing to do, feature is disabled and configMap doesn't exist.
-		} else {
-			return err
+		r.Log.Info("Deleting kata-cc runtime classes for confidential containers")
+
+		// Delete kata-cc runtime class
```
gkurz, line 134: I'm still a bit concerned that we can delete the runtime class even though it might still be used. Currently running pods will be unaffected but it will prevent new pods from being created using that runtime class and can cause instability during rollouts or scaling events in deployments and daemonsets... This badly breaks the end user experience IMO.

beraldoleal: I see the usability issue you mentioned. A few alternatives we could consider:

1. Add a pod validation webhook to reject pods using kata-cc when confidential is disabled. Not sure about the operational implications of adding another webhook though, specially for every new pod. I feel it adds some overhead we don't need.

2. Keep the runtime class but update the handler to "kata". I don't like this either, it silently masks the issue and users might think they're getting confidential computing when they're not.

3. Update the existing KataConfig validation webhook to block disabling confidential if there are pods running with kata-cc or kata-remote. This would force users to migrate workloads first.

Wdyt?

controllers/openshift_controller.go
beraldoleal: As discussed on slack, we can ignore this for now. closing the thread.

controllers/confidential_handler.go
beraldoleal: As we discussed today, I implemented the finalizer approach. Vide last commits and let me know if you spot any issues.

controllers/openshift_controller.go
gkurz: The function is still around...

controllers/openshift_controller.go
```diff
@@ -783,7 +783,7 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
 			ObjectMeta: metav1.ObjectMeta{
 				Name: runtimeClassName,
 			},
-			Handler: runtimeClassName,
+			Handler: handler,
```
gkurz, line 793: Maybe worth clarifying in the commit message the need to do this, e.g. "the upcoming kata-cc runtime class will be using different TEE-specific handlers"

controllers/confidential_handler.go
```diff
@@ -1,7 +1,27 @@
...
+
+const (
+	// kata-cc runtime class for CoCo
+	kataCCRuntimeClassName        = "kata-cc"
+	kataCCRuntimeClassCpuOverhead = "0.25"
+	kataCCRuntimeClassMemOverhead = "350Mi"
+
+	// TEE node labels
+	intelTDXNodeLabel = "intel.feature.node.kubernetes.io/tdx"
+	amdSNPNodeLabel   = "amd.feature.node.kubernetes.io/snp"
```
gkurz, line 20: Where do these labels come from ?

According to https://kubernetes-sigs.github.io/node-feature-discovery/v0.12/usage/features.html#cpu , if the NFD operator is deployed, we should get the following :

```
feature.node.kubernetes.io/cpu-security.tdx.enabled
feature.node.kubernetes.io/cpu-security.sev.snp.enabled	
```

beraldoleal: Yeah, I'm not sure if we have any good reason for not using them, and probably we should use the standard NFD labels, at some point.

However, right now, I'm just reusing the vendor-prefixed labels (`intel.feature.node.kubernetes.io/tdx`, and `amd.feature.node.kubernetes.io/snp`) that are already in our codebase (`scripts/install-helpers/baremetal-coco/nfd/`).

I would like to keep this PR using the same labels as we use already use today and let this discussion for another PR, if possible.

gkurz: That's fine but please add a comment that the strings should exactly match
https://github.com/openshift/sandboxed-containers-operator/blob/devel/scripts/install-helpers/baremetal-coco/nfd/intel-rules.yaml#L26 and https://github.com/openshift/sandboxed-containers-operator/blob/devel/scripts/install-helpers/baremetal-coco/nfd/amd-rules.yaml#L10 for now.

beraldoleal:
@bpradipt @gkurz , please let me know if we are ok to merge this one.

controllers/confidential_handler.go
```diff
@@ -1,7 +1,27 @@
...
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// kata-cc runtime class for CoCo
```
bpradipt: nit
`s/CoCo/CoCo BM`

gkurz:
In the log of commit 2 :
> All the MachineConfigs will be installed by default on all nodes

I guess you meant `All the CRI-O and kata-containers configurations will be installed by default on all nodes` since we're moving away from MachineConfigs.

controllers/confidential_handler.go
```diff
@@ -57,22 +77,98 @@ func (r *KataConfigOpenShiftReconciler) handleFeatureConfidential(state FeatureG
...
+	for _, n := range nodes.Items {
+		if v, ok := n.Labels[intelTDXNodeLabel]; ok && v == "true" {
+			hasIntelTDX = true
+		}
+		if v, ok := n.Labels[amdSNPNodeLabel]; ok && v == "true" {
+			hasAmdSNP = true
+		}
+	}
+
+	if hasIntelTDX && hasAmdSNP && !r.kataConfig.Spec.EnablePeerPods {
```
gkurz: Not sure to understand the goal of `!r.kataConfig.Spec.EnablePeerPods` here...

beraldoleal: You are right, makes no sense. I removed.

beraldoleal:
@gkurz @bpradipt sent the changes, please take a look when possible.

controllers/openshift_controller.go
```diff
@@ -782,6 +782,7 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				Name: runtimeClassName,
```
gkurz:
```suggestion
				Name:       runtimeClassName,
```

to silent `make fmt` .

controllers/openshift_controller.go
```diff
@@ -782,6 +782,7 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName stri
...
+				Finalizers: []string{"kataconfig.openshift.io/runtime-class-cleanup"},
```
gkurz:
```suggestion
				Finalizers: []string{runtimeClassFinalizerName},
````

File: controllers/migrate.go [hunk 13]
```diff
@@ -81,3 +84,32 @@ func (r *KataConfigOpenShiftReconciler) migratePeerPodsLimit() error {
...
+		rc := &nodeapi.RuntimeClass{}
+		err := r.Client.Get(context.TODO(), types.NamespacedName{Name: name}, rc)
+		if k8serrors.IsNotFound(err) {
+			continue
+		}
+		if err != nil {
+			return err
+		}
+
+		if !controllerutil.ContainsFinalizer(rc, "kataconfig.openshift.io/runtime-class-cleanup") {
```
gkurz:
```suggestion
		if !controllerutil.ContainsFinalizer(rc, runtimeClassFinalizerName) {
```

controllers/migrate.go
```diff
@@ -81,3 +84,32 @@ func (r *KataConfigOpenShiftReconciler) migratePeerPodsLimit() error {
...
+			r.Log.Info("Adding finalizer to existing runtime class", "runtimeClass", name)
+			controllerutil.AddFinalizer(rc, "kataconfig.openshift.io/runtime-class-cleanup")
```
gkurz:
```suggestion
			controllerutil.AddFinalizer(rc, runtimeClassFinalizerName)
```

config/rbac/role.yaml
gkurz, line 1: This change should be propagated to the bundle, i.e. `make bundle`.

controllers/confidential_handler.go
```diff
@@ -57,22 +77,98 @@ func (r *KataConfigOpenShiftReconciler) handleFeatureConfidential(state FeatureG
 	}
 
 	// peer pods config
+	if r.kataConfig.Spec.EnablePeerPods {
```
gkurz: All the lines above are also specific to peer pods AFAICT. It seems this `if` should move to line 40 for now.

gkurz:
I deployed this PR in my cluster and did the following :
- start a regular kata POD
- delete the kata runtime class
- observe that `oc delete` remains stuck
-  see the following trace in the logs of the controller :
```
2025-10-07T15:22:43Z	INFO	controllers.RuntimeClass	RuntimeClass is being deleted, checking for active pods	{"runtimeclass": "kata"}
2025-10-07T15:22:43Z	INFO	controllers.RuntimeClass	Found active pod using runtime class	{"pod": "example-fedora-kata", "namespace": "supergroug", "runtimeClass": "kata", "phase": "Running"}
2025-10-07T15:22:43Z	INFO	controllers.RuntimeClass	Runtime class still has active pods, waiting for them to be deleted manually or terminate naturally	{"runtimeclass": "kata"}
```
- delete the kata POD
- observe that `oc delete` completes
-  see the following trace in the logs of the controller :
```
2025-10-07T15:23:13Z	INFO	controllers.RuntimeClass	No active pods found, removing finalizer to allow deletion	{"runtimeclass": "kata"}
2025-10-07T15:23:13Z	INFO	controllers.RuntimeClass	Successfully removed finalizer, RuntimeClass will be deleted	{"runtimeclass": "kata"}
```

The finalizer-based logic to protect active runtime classes works as expected.

I could also verify the migration logic on a previous deployment :
```
2025-10-07T14:15:03Z	INFO	controllers.KataConfig	Adding finalizer to existing runtime class	{"runtimeClass": "kata"}
```

This is promising ! :smile:

controllers/confidential_handler.go
gkurz: No longer getting the errors from the image generator. Fixed.

gkurz (approved):
/lgtm

Thanks @beraldoleal !

controllers/confidential_handler.go
mythi: These `NodeFeatureRules` are important since they also register the necessary node resources that the runtimeclasses here need to add as the overhead values.

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
...
+		r.Log.Info("Creating kata-cc runtime classes for confidential containers")
+
+		handler, nodeLabel, err := r.computeTEEHandlerAndLabel()
+		if err != nil {
+			r.Log.Info("TEE detection failed", "err", err)
+			return fmt.Errorf("TEE detection failed: %w", err)
+		}
+
+		// Create kata-cc runtime class restricted to the detected TEE subset
+		err = r.createRuntimeClass(kataCCRuntimeClassName, kataCCRuntimeClassCpuOverhead, kataCCRuntimeClassMemOverhead, handler, nodeLabel)
```
mythi, line 125: in addition to CPU and Mem overhead, we consume "key ID" overhead resources, one for each CVM.

beraldoleal: I'm not sure if we should handle this inside the RuntimeClass or would it need to be a different mechanism. IIUC, it seems like a different type of resource.

Since this isn't handled today, I think it should be addressed in a separate PR.
Could you expand a bit on this key ID overhead?

gkurz: @mythi can you share a link to the definition of this "key ID" resource ?

mythi: Pod overhead accepts any kubernetes resource, including "extended resources". We have NFD configured to register cvm keys resources as node extended resources and they are used in pod overhead https://github.com/openshift/sandboxed-containers-operator/blob/c8c9e5eaf8c24d30bc5915cf83548f1e9b846b0c/scripts/install-helpers/baremetal-coco/install.sh#L532 This is the bookkeeping to ensure we don't run more TEE pods than what the node is capable of

gkurz: I suggest we merge this PR and we add support for this in a subsequent one. Is it ok for you @mythi ?

mythi: It's ok

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
 // When the feature is disabled, handleFeatureConfidential resets the config maps to non-confidential values.
 func (r *KataConfigOpenShiftReconciler) handleFeatureConfidential(state FeatureGateState) error {
```
wainersm, line 33: Hi @beraldoleal ! `handleFeatureConfidential()` doc string should be updated.

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
...
+		}
+		if err := updateConfigMap(r.Client, r.Log, peerpodsCMName, OperatorNamespace, peerpodsCMData); err != nil {
+			if k8serrors.IsNotFound(err) {
+				// When feature is Enabled: ConfigMap doesn't exist yet, will try again at the next reconcile run.
+				// Else: Nothing to do, feature is disabled and configMap doesn't exist.
+			} else {
+				return err
+			}
+		}
+	}
```
wainersm, line 101: It will handle CC on peer-pods `if r.kataConfig.Spec.EnablePeerPods` and if it reached to this point it should `return` otherwise it will continue the execution and try to run the code block following. In other words, it should not try to setup the `kata-cc` runtimeclass if peer pods. Did I get correct?

While in here, maybe I would split the handler code for baremetal and peer-pods in two functions (say `handle_baremetal()` and `handle_peerpods()`) and make the  `handleFeatureConfidential` only deciding which handler to call.

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
...
+		handler, nodeLabel, err := r.computeTEEHandlerAndLabel()
+		if err != nil {
+			r.Log.Info("TEE detection failed", "err", err)
+			return fmt.Errorf("TEE detection failed: %w", err)
+		}
+
+		// Create kata-cc runtime class restricted to the detected TEE subset
+		err = r.createRuntimeClass(kataCCRuntimeClassName, kataCCRuntimeClassCpuOverhead, kataCCRuntimeClassMemOverhead, handler, nodeLabel)
+		if err != nil {
+			r.Log.Info("Error creating kata-cc runtime class", "err", err)
```
wainersm: Use the constant `kataCCRuntimeClassName` in the log messages.

I don't know if it should be "runtime class" or "runtimeclass" or "runtimeClass". We just need to keep it consistent because people grep those logs for debugging problems.

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
...
+			}
+		}
+	}
 
-	// Patch peer pods configMap, if it exists.
-	var peerpodsCMData map[string]string
+	// Handle kata-cc runtime classes for baremetal environments
 	if state == Enabled {
-		peerpodsCMData = map[string]string{"DISABLECVM": "false"}
+		r.Log.Info("Creating kata-cc runtime classes for confidential containers")
```
wainersm: Use `kataCCRuntimeClassName` (see comment below)

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
...
+	selector, err := r.getKataConfigNodeSelectorAsSelector()
+	if err != nil {
+		msg := "Couldn't build KataConfig node selector for TEE detection"
+		r.Log.Info(msg, "err", err)
+		return "", "", fmt.Errorf("%s: %w", msg, err)
+	}
+
+	nodes := &corev1.NodeList{}
+	listOpts := []client.ListOption{
+		client.MatchingLabelsSelector{Selector: selector},
```
wainersm, line 153: `getKataConfigNodeSelectorAsSelector()` might return an empty map as seen in https://github.com/openshift/sandboxed-containers-operator/blob/devel/controllers/openshift_controller.go#L844 . Is it okay having `MatchingLabelsSelector` with a empty map?

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
...
+		}
+	}
+
+	if hasIntelTDX && hasAmdSNP {
+		msg := "Confidential install blocked: multiple TEE platforms detected in selected pool (intel TDX and AMD SNP); only one TEE per cluster is supported on baremetal"
+		r.Log.Info(msg)
+		return "", "", fmt.Errorf("%s", msg)
+	}
+
+	if hasIntelTDX && !hasAmdSNP {
```
wainersm: Checking `!hasAmdSNP` is not needed as both cannot be set at this point.

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
...
+		return kataCCIntelHandler, intelTDXNodeLabel, nil
+	}
+	if hasAmdSNP && !hasIntelTDX {
```
wainersm: Same here.

controllers/confidential_handler.go
```diff
@@ -18,61 +38,136 @@ import (
...
 		}
 	}
 
 	return nil
 }
+
+func (r *KataConfigOpenShiftReconciler) computeTEEHandlerAndLabel() (string, string, error) {
+	selector, err := r.getKataConfigNodeSelectorAsSelector()
+	if err != nil {
+		msg := "Couldn't build KataConfig node selector for TEE detection"
```
wainersm: In this function I've seen this pattern: log a message & return the error. However, the caller function will also log a message (that contain the message already logged) and return the error. It will produce duplicated log messages.

controllers/runtimeclass_controller.go
```diff
@@ -0,0 +1,130 @@
...
+
+// +kubebuilder:rbac:groups=node.k8s.io,resources=runtimeclasses,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
+
+// Reconcile handles RuntimeClass lifecycle, specifically finalizer cleanup
+func (r *RuntimeClassReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("runtimeclass", req.NamespacedName.Name)
+
+	runtimeClass := &nodeapi.RuntimeClass{}
+	err := r.Get(ctx, req.NamespacedName, runtimeClass)
```
wainersm, line 55: Sorry for my ignorance...don't we need to filter by runtimeclass that we want to reconcile (kata-cc, kata-remote...etc) only

wainersm:
@beraldoleal finished my review although I don't know much about the OSC operator (mainly the reconcile parts). So I don't know what is missing (if any) and/or what could be implemented better (if any).

controllers/confidential_handler.go
mythi: a follow-up PR would also need to cleanup `install.sh` for the parts added here, right?

controllers/confidential_handler.go
beraldoleal: > It will handle CC on peer-pods if r.kataConfig.Spec.EnablePeerPods and if it reached to this point it should return otherwise it will continue the execution and try to run the code block following. In other words, it should not try to setup the kata-cc runtimeclass if peer pods. Did I get correct?

From what we discussed before... peerpods and baremetal can coexist. Users can choose which to use via the RuntimeClass in their pod spec. We so need to handle both configs here.

> While in here, maybe I would split the handler code for baremetal and peer-pods in two functions (say handle_baremetal() and handle_peerpods()) and make the handleFeatureConfidential only deciding which handler to call.

That is a good suggestion. I will push the split.

controllers/confidential_handler.go
beraldoleal: Good catch, thanks. Fixing.

controllers/runtimeclass_controller.go
gkurz: Yes we should. Good catch @wainersm !

controllers/confidential_handler.go
beraldoleal: iiuc, an empty selector is valid here, it will return everything. And it just means the admin wants to install on all nodes cluster-wide.

We use the same pattern in `getNodesWithLabels()`.

controllers/confidential_handler.go
wainersm: > > It will handle CC on peer-pods if r.kataConfig.Spec.EnablePeerPods and if it reached to this point it should return otherwise it will continue the execution and try to run the code block following. In other words, it should not try to setup the kata-cc runtimeclass if peer pods. Did I get correct?
> 
> From what we discussed before... peerpods and baremetal can coexist. Users can choose which to use via the RuntimeClass in their pod spec. We so need to handle both configs here.
> 

Ok, I overlooked that. Even though, if you let the execution continue and it call `handler, nodeLabel, err := r.computeTEEHandlerAndLabel()`  it will fail if you have only peer-pods CoCo. 

> > While in here, maybe I would split the handler code for baremetal and peer-pods in two functions (say handle_baremetal() and handle_peerpods()) and make the handleFeatureConfidential only deciding which handler to call.
> 
> That is a good suggestion. I will push the split.

controllers/confidential_handler.go
beraldoleal: Good catch. Fixing..

controllers/runtimeclass_controller.go
beraldoleal: Yep. I would just suggest may we don' t use the class names, to avoid having this "hardcoded" here. We could just set the SetupWithManager to filter by the finalizer name, if its our finalizer then reconcile.

I just pushed this approach. Let me know what you guys think.

controllers/confidential_handler.go
beraldoleal: > Ok, I overlooked that. Even though, if you let the execution continue and it call handler, nodeLabel, err := r.computeTEEHandlerAndLabel() it will fail if you have only peer-pods CoCo.

You are right.

What I just pushed is the following logic when confidential is enabled and we fail to find TEE hardware:

* If peer pods is enabled, logs a warning and skips baremetal CoCo (because peer pods CVM will handle confidential workloads)
* If peer pods is disabled, returns an error (because admin explicitly requested confidential but there's no hardware to support it)

Does that makes sense for now? let me know what you think

wainersm: Hi @beraldoleal , yes, it makes sense to me. Thanks for addressing that too!

controllers/runtimeclass_controller.go
wainersm: Is https://github.com/openshift/sandboxed-containers-operator/pull/836/commits/9ff266d4d5829b49443c755a11bdecac722dd6a1#diff-2804d62dab597d3653f1d2640f5fa27727e60272d41de20713c4fea2fc981c19R134 that ensures it is filtered by the finalizer name?

wainersm (approved):
@beraldoleal , again, I'm not too versed on the operator but those changes looks good to me \o/

beraldoleal:
> @beraldoleal , again, I'm not too versed on the operator but those changes looks good to me \o/

"not too versed", but spotted a few good problems. thanks again.

beraldoleal:
@gkurz , @bpradipt... any other comments here? could we merge?

controllers/runtimeclass_controller.go
gkurz: Clever ! :smile:

bpradipt (approved):
/lgtm
Thanks @beraldoleal

beraldoleal:
Discussed with @gkurz and got green light on slack. merged. thanks.

PR #842, opened 2025-08-05
podvm: remove signature in skopeo copy

snir911:

**- What I did**

**- How to verify it**

### DIFF
--- a/config/peerpods/podvm/lib.sh
+++ b/config/peerpods/podvm/lib.sh
@@ -331,10 +331,11 @@ function extract_container_image() {
         error_exit "Usage: extract_container_image <container_image_repo_url> <image_tag> <dest_image> <destination_path> [registry_secret]"
 
     # Form the skopeo CLI. Add authfile if provided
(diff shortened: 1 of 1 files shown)


config/peerpods/podvm/lib.sh
```diff
@@ -332,9 +332,9 @@ function extract_container_image() {
 
     # Form the skopeo CLI. Add authfile if provided
     if [[ -n "${5}" ]]; then
-        SKOPEO_CLI="skopeo copy --authfile ${auth_json_file}"
+        SKOPEO_CLI="skopeo copy --remove-signatures --authfile ${auth_json_file}"
```
bpradipt, line 336: nit:
You can add a comment that "oci image destination supports signatures" hence the signatures if any are removed while copying

bpradipt (approved):
/lgtm
Just add the comment so that new readers will know why the signatures are removed

beraldoleal (approved):
LGTM, thanks.

PR #854, opened 2025-08-07
release 1.10.1: update the catalogs

littlejawa:
Updating the catalogs to use the new version and newest bundle image.


littlejawa:
NOTE: Enterprise contract failures are expected.
They are related to the failure of the FIPS check for the dm-verity image, reporting that the openssl lib is missing.
The dm-verity image being a scratch-based image embedding the actual VM image, it is not supposed to ship openssl.

This error is ignored in prod releases, but we didn't deactivate it for stage/build pipelines.

PR #879, opened 2025-08-19
OSC on Managed Identities Cluster Support

snir911:
**This pr add workload identities support for OSC in ARO clusters.**

**Testing catalog (OSC 1.11.1)**  : quay.io/redhat-user-workloads/ose-osc-tenant/osc-test-fbc:on-pr-7814d09d32e92712a49d3adc979c997e336c48dc

# OSC on Managed Identities Cluster Workflow:
1. Create an ARO [cluster with managed identities](https://learn.microsoft.com/en-us/azure/openshift/howto-create-openshift-cluster?pivots=aro-az-cli) or [enable Entra ID on existing self-managed cluster](https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/postinstallation_configuration/changing-cloud-credentials-configuration#enabling-entra-workload-id-existing-cluster_changing-cloud-credentials-configuration)
2. Create an identity in the RG you have created:
```
MID=created-identity 
az identity create --name $MID --resource-group $RESOURCEGROUP --location $LOCATION
CLIENT_ID=$(az identity show  --name $MID --resource-group $RESOURCEGROUP --query clientId -o tsv)
TENANT_ID=$(az identity show  --name $MID --resource-group $RESOURCEGROUP --query tenantId -o tsv)
SUB_ID=$(az account show --query id -o tsv)
```
3. Grant the required permission to the created identity
```
ROLES=("Reader" "Virtual Machine Contributor" "Network Contributor" "Storage Account Contributor" "Compute Gallery Artifacts Reader")
for r in "${ROLES[@]}"; do az role assignment create --assignee "$CLIENT_ID" --role "$r" --scope "/subscriptions/$SUB_ID"; done
```
4. Create an Azure federated identity credential that links the service account in OpenShift Container Platform to the Azure user-assigned managed identity
```
az identity federated-credential create  --name $MID-federation  --identity-name $MID --issuer $(oc get authentication.config.openshift.io cluster -o jsonpath='{.spec.serviceAccountIssuer}') --subject system:serviceaccount:openshift-sandboxed-containers-operator:default -g $RESOURCEGROUP --audiences openshift
```
5. Configure the [default workers subnet](https://docs.redhat.com/en/documentation/openshift_sandboxed_containers/1.11/html/deploying_red_hat_openshift_sandboxed_containers/deploying-osc_azure-osc#outbound-connections_azure-osc) (make sure to use your RG, not the cluster’s one as instucted)
6. Install custom OSC catalog (quay.io/redhat-user-workloads/ose-osc-tenant/osc-test-fbc:on-pr-7814d09d32e92712a49d3adc979c997e336c48dc)
7. install operator though the UI, in subscription page you'll be required to provide the identity values `az identity show  --name $MID --resource-group $RESOURCEGROUP`
8. Set peer-pods-cm with the the RG used above
9. install kataConfig with peerpods enabled.

## Main Test Cases:
1. Standard Peer pods workflow can be fully executed on a managed-identities ARO cluster
2. No regression standard use cases (CCO created secret and provided peer-pods-secert)

# TODO after MERGE
**render.sh should be executed after this PR (and updated bundle PR) is merged in order to create the working catalogs**

### DIFF
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -13,7 +13,7 @@ metadata:
         }
       ]
     capabilities: Seamless Upgrades
-    createdAt: "2026-01-08T14:32:12Z"
+    createdAt: "2026-02-16T11:35:51Z"
     features.operators.openshift.io/cnf: "false"
     features.operators.openshift.io/cni: "false"
     features.operators.openshift.io/csi: "false"
@@ -21,9 +21,9 @@ metadata:
     features.operators.openshift.io/fips-compliant: "false"
     features.operators.openshift.io/proxy-aware: "false"
     features.operators.openshift.io/tls-profiles: "false"
-    features.operators.openshift.io/token-auth-aws: "false"
-    features.operators.openshift.io/token-auth-azure: "false"
-    features.operators.openshift.io/token-auth-gcp: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
+    features.operators.openshift.io/token-auth-azure: "true"
+    features.operators.openshift.io/token-auth-gcp: "true"
     olm.skipRange: '>=1.1.0 <1.11.2'
     operatorframework.io/suggested-namespace: openshift-sandboxed-containers-operator
     operators.openshift.io/valid-subscription: '["OpenShift Container Platform", "OpenShift
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -451,6 +451,9 @@ spec:
                 - mountPath: /root/.ssh/
                   name: ssh
                   readOnly: true
+                - mountPath: /var/run/secrets/openshift/serviceaccount
+                  name: bound-sa-token
+                  readOnly: true
               securityContext:
                 runAsNonRoot: true
                 seccompProfile:
@@ -481,6 +484,12 @@ spec:
                   defaultMode: 384
                   optional: true
                   secretName: ssh-key-secret
+              - name: bound-sa-token
+                projected:
+                  sources:
+                  - serviceAccountToken:
+                      audience: openshift
+                      path: token
       - label:
           app: operator-metrics-server
         name: operator-metrics-server
--- a/config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -20,9 +20,9 @@ metadata:
     features.operators.openshift.io/fips-compliant: "false"
     features.operators.openshift.io/proxy-aware: "false"
     features.operators.openshift.io/tls-profiles: "false"
-    features.operators.openshift.io/token-auth-aws: "false"
-    features.operators.openshift.io/token-auth-azure: "false"
-    features.operators.openshift.io/token-auth-gcp: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
+    features.operators.openshift.io/token-auth-azure: "true"
+    features.operators.openshift.io/token-auth-gcp: "true"
     olm.skipRange: '>=1.1.0 <1.11.2'
     operatorframework.io/suggested-namespace: openshift-sandboxed-containers-operator
     operators.openshift.io/valid-subscription: '["OpenShift Container Platform", "OpenShift
--- a/config/peerpods/podvm/azure-podvm-image-handler.sh
+++ b/config/peerpods/podvm/azure-podvm-image-handler.sh
@@ -35,7 +35,10 @@ function verify_vars() {
 
     # Ensure that the Azure specific values are set
     [[ -z "${AZURE_CLIENT_ID}" ]] && error_exit "AZURE_CLIENT_ID is empty"
-    [[ -z "${AZURE_CLIENT_SECRET}" ]] && error_exit "AZURE_CLIENT_SECRET is empty"
+    # Either AZURE_CLIENT_SECRET or AZURE_FEDERATED_TOKEN_FILE must be set
+    if [[ -z "${AZURE_CLIENT_SECRET}" && -z "${AZURE_FEDERATED_TOKEN_FILE}" ]]; then
+        error_exit "Either AZURE_CLIENT_SECRET or AZURE_FEDERATED_TOKEN_FILE must be set"
+    fi
     [[ -z "${AZURE_SUBSCRIPTION_ID}" ]] && error_exit "AZURE_SUBSCRIPTION_ID is empty"
     [[ -z "${AZURE_TENANT_ID}" ]] && error_exit "AZURE_TENANT_ID is empty"
 
@@ -132,11 +135,32 @@ function login_to_azure() {
     echo "Logging in to Azure"
     # If any error occurs, exit the script with an error message
 
-    az login --service-principal \
-        --user="${AZURE_CLIENT_ID}" \
-        --password="${AZURE_CLIENT_SECRET}" \
-        --tenant="${AZURE_TENANT_ID}" ||
-        error_exit "Failed to login to Azure"
+    if [[ -n "${AZURE_CLIENT_SECRET}" ]]; then
+        # Use client secret authentication
+        echo "Using client secret authentication"
+        az login --service-principal \
+            --user="${AZURE_CLIENT_ID}" \
+            --password="${AZURE_CLIENT_SECRET}" \
+            --tenant="${AZURE_TENANT_ID}" ||
+            error_exit "Failed to login to Azure"
+    elif [[ -n "${AZURE_FEDERATED_TOKEN_FILE}" ]]; then
+        # Use federated token authentication
+        echo "Using federated token authentication"
+
+        # Read the token from the file
+        [[ ! -f "${AZURE_FEDERATED_TOKEN_FILE}" ]] && error_exit "Federated token file ${AZURE_FEDERATED_TOKEN_FILE} does not exist"
+
+        FEDERATED_TOKEN=$(cat "${AZURE_FEDERATED_TOKEN_FILE}") ||
+            error_exit "Failed to read federated token from ${AZURE_FEDERATED_TOKEN_FILE}"
+
+        [[ -z "${FEDERATED_TOKEN}" ]] && error_exit "Federated token is empty"
+
+        az login --service-principal \
+            --user="${AZURE_CLIENT_ID}" \
+            --federated-token="${FEDERATED_TOKEN}" \
+            --tenant="${AZURE_TENANT_ID}" ||
+            error_exit "Failed to login to Azure using federated token"
+    fi
 
     echo "Selecting subscription"
     az account set --subscription ${AZURE_SUBSCRIPTION_ID} ||
--- a/controllers/credentials_controller.go
+++ b/controllers/credentials_controller.go
@@ -93,8 +93,8 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	} else if err != nil {
 		r.Log.Info("error in getting peer-pods secret", "err", err)
 		return ctrl.Result{Requeue: true}, nil
-	} else if !isSecretOwned(peerPodsSecret) { // Either STS workflow or user created secret
-		r.Log.Info("unowned peer-pods-secret exist, skipping CCO workflow...")
+	} else if !isCCOFlowSecret(peerPodsSecret) { // not a CCO created secret, shouldn't reach here
+		r.Log.Info("unexpected unowned peer-pods-secret exist, skipping CCO secret mapping flow...")
 		return ctrl.Result{}, nil
 	}
 
@@ -171,7 +171,7 @@ func (r *SecretReconciler) ccoDataMapping(ccoSecretData map[string][]byte) map[s
 	return peerPodsSecretData
 }
 
-func isSecretOwned(secret *corev1.Secret) bool {
+func isCCOFlowSecret(secret *corev1.Secret) bool {
 	if secret == nil {
 		return false
 	}
@@ -351,7 +351,7 @@ func (kh *KataConfigHandler) skipCredentialRequests() bool {
 		if !(k8serrors.IsNotFound(err) || k8serrors.IsGone(err)) {
 			kh.reconciler.Log.Info("failed to get peer-pods Secret skipping peer-pods Secret check", "error", err)
 		}
-	} else if !isControllerGenerated(peerPodsSecret) {
+	} else if !isCCOFlowSecret(peerPodsSecret) {
 		kh.reconciler.Log.Info("peerPodsSecret has already been created by the user, skipping...")
 		return true
 	}
--- a/controllers/credentials_controller.go
+++ b/controllers/credentials_controller.go
@@ -79,7 +79,7 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	ccoSecret := &corev1.Secret{}
 	if err := r.Client.Get(context.TODO(), req.NamespacedName, ccoSecret); err != nil {
 		if k8serrors.IsNotFound(err) {
-			r.Log.Info("cco-secret not found")
+			r.Log.Info("no cco-secret has been found")
 			return ctrl.Result{}, nil
 		} else {
 			r.Log.Info("error in getting cco-secret", "err", err)
@@ -89,27 +89,23 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	peerPodsSecret, err := getPeerPodsSecret(r.Client)
 	if k8serrors.IsNotFound(err) {
-		peerPodsSecret = r.newOwnedPeerPodsSecret(ccoSecret)
+		r.Log.Info("peer-pods-secret is not found, trying to map cco-secret")
 	} else if err != nil {
 		r.Log.Info("error in getting peer-pods secret", "err", err)
 		return ctrl.Result{Requeue: true}, nil
-	}
-
-	// skip if peer-pods-secret was created by the user
-	if !isControllerGenerated(peerPodsSecret) {
-		r.Log.Info("peerPodsSecret has been created by the user, skipping...")
+	} else if !isSecretOwned(peerPodsSecret) { // Either STS workflow or user created secret
+		r.Log.Info("unowned peer-pods-secret exist, skipping CCO workflow...")
 		return ctrl.Result{}, nil
 	}
 
-	if _, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, peerPodsSecret, func() error {
-		r.secretMapping(peerPodsSecret, ccoSecret)
-		return nil
-	}); err != nil {
+	peerpodsData := r.ccoDataMapping(ccoSecret.Data)
+	labels := map[string]string{labelCredentialsRequest: labelCredentialsRequestValue}
+	if err := r.createOrUpdateSecret(context.TODO(), peerPodsSecretName, OperatorNamespace, peerpodsData, ccoSecret, labels); err != nil {
 		r.Log.Info("error in creating or updating peer-pods secret", "err", err)
 		return ctrl.Result{Requeue: true}, nil
 	}
 
-	r.Log.Info("cco-secret created and mapped to peer-pods secret", "CCO Secret", ccoSecret.GetName(), "Peer-Pods Secret", peerPodsSecret.GetName())
+	r.Log.Info("cco-secret created and mapped to peer-pods secret", "CCO Secret", ccoSecret.GetName(), "Peer-Pods Secret", peerPodsSecretName)
 	return ctrl.Result{}, nil
 }
 
@@ -144,8 +140,8 @@ func secretsFilterPredicate() predicate.Predicate {
 	}
 }
 
-// map ccoSecret fields to peer-pods compatible fields and set to peerPodsSecret
-func (r *SecretReconciler) secretMapping(peerPodsSecret *corev1.Secret, ccoSecret *corev1.Secret) {
+// map ccoSecret fields to peer-pods compatible fields
+func (r *SecretReconciler) ccoDataMapping(ccoSecretData map[string][]byte) map[string][]byte {
 	ccoToPp := map[string]string{
 		"aws_access_key_id":     "AWS_ACCESS_KEY_ID",
 		"aws_secret_access_key": "AWS_SECRET_ACCESS_KEY",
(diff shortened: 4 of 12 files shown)


bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
         }
       ]
     capabilities: Seamless Upgrades
-    createdAt: "2026-01-08T14:32:12Z"
+    createdAt: "2025-08-18T11:19:03Z"
```
gkurz: This seems to indicate that you're patching the bundle manually... what about running `make bundle` ?

fbc/v4.17/catalog/sandboxed-containers-operator/catalog.json
```diff
@@ -377,7 +377,7 @@
                     "features.operators.openshift.io/proxy-aware": "false",
                     "features.operators.openshift.io/tls-profiles": "false",
                     "features.operators.openshift.io/token-auth-aws": "false",
-                    "features.operators.openshift.io/token-auth-azure": "false",
+                    "features.operators.openshift.io/token-auth-azure": "true",
```
gkurz: Same remark here... I see that v4.20 is missing. The FBC files are supposed to be rendered by the `render.sh` script (see https://github.com/openshift/sandboxed-containers-operator/blob/devel/fbc/README.md#update-the-fbc).

gkurz (commented):
> fbc: enable token-auth-azure
> to mark operator as STS supported

What's STS ?

Can you please add missing Signed-off-by to some commits in this PR ? :wink:

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -13,15 +13,15 @@ metadata:
         }
       ]
     capabilities: Seamless Upgrades
-    createdAt: "2025-08-18T11:19:03Z"
+    createdAt: "2025-09-01T07:52:17Z"
```
gkurz: Same as the previous patch.

I suggest you squash patches 1 and 2 :
- add changes in `config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml`
- `make bundle`
- https://github.com/openshift/sandboxed-containers-operator/blob/devel/fbc/README.md#update-the-fbc

config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
     features.operators.openshift.io/fips-compliant: "false"
     features.operators.openshift.io/proxy-aware: "false"
     features.operators.openshift.io/tls-profiles: "false"
-    features.operators.openshift.io/token-auth-aws: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
     features.operators.openshift.io/token-auth-azure: "true"
     features.operators.openshift.io/token-auth-gcp: "false"
```
gkurz: > fbc: add auth-token to all provider

AFAICS this is only adding AWS and not GCP, the `all provider`  wording is misleading.

>requried because of OCPBUGS-61183

It would so nice if you could provide relevant details from this ticket so that everyone doesn't need to go there.

controllers/credentials_controller.go
```diff
@@ -93,8 +93,8 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	} else if err != nil {
 		r.Log.Info("error in getting peer-pods secret", "err", err)
 		return ctrl.Result{Requeue: true}, nil
-	} else if !isSecretOwned(peerPodsSecret) { // Either STS workflow or user created secret
-		r.Log.Info("unowned peer-pods-secret exist, skipping CCO workflow...")
+	} else if !isCCOFlowSecret(peerPodsSecret) { // Either STS flow or user created secret
```
gkurz: > isCCOFlow

Can you provide some clarifications in a commit message ?

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
snir911: it was using make bundle but i committed only the relevant changes (it removes some labels and image tags also which IIRC i was told not to commit this before)

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
snir911: 2 may be reverted at some point, hence, i made it in separate

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -35,7 +35,10 @@ function verify_vars() {
 
     # Ensure that the Azure specific values are set
     [[ -z "${AZURE_CLIENT_ID}" ]] && error_exit "AZURE_CLIENT_ID is empty"
-    [[ -z "${AZURE_CLIENT_SECRET}" ]] && error_exit "AZURE_CLIENT_SECRET is empty"
+    # Either AZURE_CLIENT_SECRET or AZURE_FEDERATED_TOKEN_FILE must be set
+    if [[ -z "${AZURE_CLIENT_SECRET}" && -z "${AZURE_FEDERATED_TOKEN_FILE}" ]]; then
```
wainersm, line 39: Hi @snir911 , I've a couple of questions for these changes in the podvm image builder, let me write it all here:

* podvm image builder isn't used, it is? I thought we had replace it with the dmverity-enabled podvm image, so that the code in config/peerpods/podvm is dead
* suppose the podvm image builder is still used, my next questions are:
* Why read the federated token from a file (AZURE_FEDERATED_TOKEN_FILE) instead of passing its value as a secret (e.g like `AZURE_CLIENT_SECRET`)?
* AZURE_FEDERATED_TOKEN_FILE should be exported in the environmet just like the other variables. In https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/osc-podvm-create-job.yaml#L53 I see a list of configmaps and secrets it will read and export. So in what configmap or secret `AZURE_FEDERATED_TOKEN_FILE` is defined?
* Still in https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/osc-podvm-create-job.yaml , where is the `AZURE_FEDERATED_TOKEN_FILE` is mounted in the container so that it can be read?

snir911: > podvm image builder isn't used, it is? I thought we had replace it with the dmverity-enabled podvm image, so that the code in config/peerpods/podvm is dead 

Partly correct, it is still used only the upload the image to the relevant cloud provider, the rest should be removed, hopefully it will be removed for 1.12

I'll share with you an internal doc explaining this but also I'll try to add a design doc in here.

wainersm: Hi @snir911 ,

I think I now understand the flow. As I started the review from that change, I missed the part where the controller creates the `peer-pods-secrets` that contains the `AZURE_FEDERATED_TOKEN_FILE`.

I will mark this thread resolved as it doesn't make sense what I said previously.

controllers/credentials_controller.go
```diff
@@ -159,40 +156,196 @@ func (r *SecretReconciler) secretMapping(peerPodsSecret *corev1.Secret, ccoSecre
...
+		r.Log.Error(err, "Failed to create/update peer-pods-secret for STS workflow")
+		return false, err
+	}
+
+	r.Log.Info("Successfully created/updated peer-pods-secret for STS workflow")
+	return true, nil
+}
+
+// setupPeerPodsCredentials handles the complete credential setup flow for peer-pods.
+// Priority order: User-created → STS workflow → CCO workflow
```
wainersm: Just a nit: I don't like the special chars like → that Claude/Cursor adds on text. If you remove these then you would make me very happy :)

controllers/credentials_controller.go
```diff
@@ -159,40 +156,196 @@ func (r *SecretReconciler) secretMapping(peerPodsSecret *corev1.Secret, ccoSecre
...
+}
+
+// checkAndSetupSTSWorkflow checks if STS (Security Token Service) workflow environment variables
+// are set and creates/updates the peer-pods-secret accordingly.
+// Currently supports Azure only.
+// Returns true if STS workflow is detected and secret was created/updated successfully.
+// Returns false if no STS environment variables are found or if there's an error.
+func (r *SecretReconciler) checkAndSetupSTSWorkflow(ctx context.Context) (bool, error) {
+	// STS environment variables (from OLM/web console installation)
+	// Reference: https://github.com/openshift/enhancements/pull/1800
```
wainersm: Hi @snir911 , I need some clarifications....

IIUC the variables `CLIENTID`, `TENANTID` and `SUBSCRIPTIONID` are set by the user installing the OSC operator via Web Console installation.

I see here, and elsewhere, a reference to https://github.com/openshift/enhancements/pull/1800, which still Work-in-progress. What's this enhancement meant for? Is it related with the OLM installation method?

snir911: Sure,
The OLM integration is already integrated in OSC, once cluster is in Workload Identity/Federated Identity mode the user will be prompted to provide those when deploying the subscription (in the web console, i assume it's enforced also in the CLI case).
Previously, operators were required to consume these variables and create a CredentialRequest for the CCO. In response, the CCO would create a secret containing those values. The operator then had to detect the secret’s creation and use it for API access. However, the proposed enhancement simplifies this flow by having the operator create the secret directly.

controllers/peerpods.go
```diff
@@ -236,6 +236,43 @@ func (r *KataConfigOpenShiftReconciler) processProviderConfigCAA(ds *appsv1.Daem
 			}
 		}
 
+		return nil
+	case AzureProvider:
+		// Only add bound-sa-token volume for Azure federated identity if using STS flow
+		// Having SUBSCRIPTIONID defined implies STS flow is enabled
+		if os.Getenv("SUBSCRIPTIONID") != "" {
```
wainersm: In https://github.com/openshift/sandboxed-containers-operator/pull/879/changes#diff-39a31e999077a96847da6bf59c79fcad91a44cb57f6683d71b321b961375e931R247 it's checking for clientid, tenantid and subscriptionid to determine whether STS flow or not. I think it should do the same checking here for coherence.

controllers/credentials_controller.go
wainersm: Suppose that I want to install OSC operator in command-line, when I create the subscription I simply need to do the following?

```
kind: Subscription
metadata:
name: ...
spec:
 config:
   env:
   - name: CLIENTID
     value: "<client ID >"
   - name: TENANTID
     value: "<tenant ID >"
   - name: SUBSCRIPTIONID
     value: "<subscription ID >"
```

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -13,17 +13,17 @@ metadata:
...
     features.operators.openshift.io/cni: "false"
     features.operators.openshift.io/csi: "false"
     features.operators.openshift.io/disconnected: "true"
     features.operators.openshift.io/fips-compliant: "false"
     features.operators.openshift.io/proxy-aware: "false"
     features.operators.openshift.io/tls-profiles: "false"
-    features.operators.openshift.io/token-auth-aws: "false"
-    features.operators.openshift.io/token-auth-azure: "false"
-    features.operators.openshift.io/token-auth-gcp: "false"
+    features.operators.openshift.io/token-auth-aws: "true"
```
wainersm, line 24: I don't think we should enable STS on AWS and GCP because the support is not yet implemented in OSC. It might confuse users or even force them installing the operator via web console to fill unused fields.

controllers/credentials_controller.go
```diff
@@ -159,40 +156,196 @@ func (r *SecretReconciler) secretMapping(peerPodsSecret *corev1.Secret, ccoSecre
...
+		secretData["AZURE_FEDERATED_TOKEN_FILE"] = []byte(tokenPath)
+		labels[labelSTS] = "azure"
+	} else {
+		// No STS credentials found
+		return false, nil
+	}
+
+	// Create or update the peer-pods-secret with STS credentials
+	if err := r.createOrUpdateSecret(ctx, peerPodsSecretName, OperatorNamespace, secretData, nil, labels); err != nil {
+		r.Log.Error(err, "Failed to create/update peer-pods-secret for STS workflow")
```
wainersm: Use `peerPodsSecretName` instead of spelling "peer-pods-secret"

controllers/credentials_controller.go
```diff
@@ -159,40 +156,196 @@ func (r *SecretReconciler) secretMapping(peerPodsSecret *corev1.Secret, ccoSecre
...
+		return false, err
+	}
+
+	r.Log.Info("Successfully created/updated peer-pods-secret for STS workflow")
```
wainersm: Use `peerPodsSecretName` instead of spelling "peer-pods-secret"

controllers/credentials_controller.go
```diff
@@ -159,40 +156,196 @@ func (r *SecretReconciler) secretMapping(peerPodsSecret *corev1.Secret, ccoSecre
...
+		return false, err
+	}
+
+	// 2. If secret exists, skip (regardless of who created it)
+	if peerPodsSecret != nil {
+		kh.reconciler.Log.Info("peer-pods-secret already exists, skipping credential setup")
+		return false, nil
+	}
+
+	// 3. Try STS workflow first (check environment variables)
```
wainersm, line 295: I suggest to log a msg like "Attempting STS workflow for credential setup"

controllers/credentials_controller.go
```diff
@@ -89,27 +90,23 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		r.Log.Info("peer-pods-secret is not found, trying to map cco-secret")
 	} else if err != nil {
 		r.Log.Info("error in getting peer-pods secret", "err", err)
 		return ctrl.Result{Requeue: true}, nil
-	}
-
-	// skip if peer-pods-secret was created by the user
-	if !isControllerGenerated(peerPodsSecret) {
-		r.Log.Info("peerPodsSecret has been created by the user, skipping...")
+	} else if !isCCOFlowSecret(peerPodsSecret) { // Either STS flow or user created secret
```
wainersm: I need some help to understand this `Reconcile()` entirely. The documentation of the method implies it is a handler for the CCO secret only. Also in https://github.com/openshift/sandboxed-containers-operator/pull/879/changes#diff-39a31e999077a96847da6bf59c79fcad91a44cb57f6683d71b321b961375e931L81 it seems to return if the `cco-secret` is not found.

controllers/credentials_controller.go
```diff
@@ -210,23 +363,21 @@ func (kh *KataConfigHandler) Create(ctx context.Context, event event.CreateEvent
...
 	if event.ObjectNew.(*kataconfigurationv1.KataConfig).Spec.EnablePeerPods {
-		if err := kh.createCredentialsRequests(); err != nil {
-			kh.reconciler.Log.Info("error in creating credentialsRequests", "err", err)
+		if _, err := kh.setupPeerPodsCredentials(ctx); err != nil {
+			kh.reconciler.Log.Error(err, "error setting up peer-pods credentials")
 		}
 	} else {
-		if err := kh.deleteCredentialsRequests(); err != nil {
-			kh.reconciler.Log.Info("error in deleting credentialsRequests", "err", err)
+		if _, err := kh.teardownPeerPodsCredentials(ctx); err != nil {
```
wainersm: I didn't understand why we need this `else` conditional. The kataConfig is updated and it knows it's not to `EnablePeerPods` so why to attempt to delete the credentials?

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
snir911: See the commit msg, there's a bug, it doesn't work as expected if i don't set these

controllers/credentials_controller.go
snir911: to support update that disables peerpods, if the function idempotent i think it's valid, I'll revisit this

controllers/credentials_controller.go
snir911: Good catch, I should have update the filter to handle all secrets, i think i have missed committing this, i wonder how it worked without it... will revisit this

controllers/credentials_controller.go
wainersm: My concern is to run this teardown function for updates on kataconfig that are unrelated with peer pods. For example, suppose that peer pods is already enabled and the user update the kataconfig to change the value of `logLevel`.

controllers/credentials_controller.go
snir911: I had another look, it should never reach here, the reconcile loop only role is to map the cco created secret in case cco-secret has been created, having said that, i will keep this condition there're user manipulation that could cause reaching here.
I'll update the code comments to make it clearer

controllers/credentials_controller.go
snir911: but it will be executed only if peerpods is disabled.. so it should be harmful

controllers/credentials_controller.go
snir911: no straight forward way to integrate in the string itself AFAIU

controllers/credentials_controller.go
snir911: no straight forward way to integrate in the string itself AFAIU

controllers/credentials_controller.go
snir911: actually in second thought, i'm not sure we want this to be printed to Azure users that has not clue what this STS is all about

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -132,11 +135,34 @@ function login_to_azure() {
...
+            --tenant="${AZURE_TENANT_ID}" ||
+            error_exit "Failed to login to Azure"
+    elif [[ -n "${AZURE_FEDERATED_TOKEN_FILE}" ]]; then
+        # Use federated token authentication
+        echo "Using federated token authentication"
+
+        # Read the token from the file
+        [[ ! -f "${AZURE_FEDERATED_TOKEN_FILE}" ]] && error_exit "Federated token file ${AZURE_FEDERATED_TOKEN_FILE} does not exist"
+
+        FEDERATED_TOKEN=$(cat "${AZURE_FEDERATED_TOKEN_FILE}") ||
```
bpradipt, line 153: Does az cli allows reading the token directly from the file instead of reading it first in a var?

controllers/credentials_controller.go
```diff
@@ -89,27 +89,23 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		r.Log.Info("unowned peer-pods-secret exist, skipping CCO workflow...")
 		return ctrl.Result{}, nil
 	}
 
-	if _, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, peerPodsSecret, func() error {
-		r.secretMapping(peerPodsSecret, ccoSecret)
-		return nil
-	}); err != nil {
+	peerpodsData := r.ccoDataMapping(ccoSecret.Data)
+	labels := map[string]string{labelCredentialsRequest: labelCredentialsRequestValue}
```
bpradipt, line 101: There is a comment in line 56 which mentions that this label - labelCredentialsRequest is not used anymore and should be dropped. But it's used here.

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
gkurz: `make bundle` bumps `createdAt:` and your PR is going backwards in time. I usually look at that field to know how old is the latest `make bundle`. Please re-run it.

Here's a what I usually do to only update the CSV :
```
# Set IMG to what we already have in the bundle
IMG=$(yq .spec.install.spec.deployments[0].spec.template.spec.containers[0].image bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml) make bundle
# Only keep changes in the CSV
git restore bundle.Dockerfile config/manager/kustomization.yaml config/metrics/kustomization.yaml
```

gkurz:
> > fbc: enable token-auth-azure
> > to mark operator as STS supported
> 

AFAICS fist commit still has the same changelog.

> What's STS ?
> 

Ok this is https://en.wikipedia.org/wiki/Security_token_service. It would be great to have some details on what this means in Azure in the commit message, for the sake of not having to google around.

> Can you please add missing Signed-off-by to some commits in this PR ? 😉

Also the commit title is misleading since it doesn't relate to the FBC. Please make it `Enable Security Token Service on Azure`.

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
gkurz: We won't revert changes to the bundle. A new PR will simply set these fields to `false` when OCPBUGS-61183 is fixed and we'll re-generate the bundle accordingly.

And I really prefer these fixes to be embedded in the previous commit that requires them to be functional IIUC.

controllers/image_generator.go
```diff
@@ -614,7 +614,7 @@ func (r *ImageGenerator) validatePeerPodsConfigs() error {
 	// aws ConfigMap Keys
 	awsConfigMapKeys := []string{"AWS_REGION", "AWS_SUBNET_ID", "AWS_VPC_ID", "AWS_SG_IDS", "CLOUD_PROVIDER"}
 	// azure Secret Keys
-	azureSecretKeys := []string{"AZURE_TENANT_ID", "AZURE_CLIENT_ID", "AZURE_CLIENT_SECRET", "AZURE_SUBSCRIPTION_ID"}
+	azureSecretKeys := []string{"AZURE_TENANT_ID", "AZURE_CLIENT_ID", "AZURE_SUBSCRIPTION_ID"}
```
gkurz: What about the case where workload identity isn't used ? Isn't `AZURE_CLIENT_SECRET` required in this case ?

config/peerpods/podvm/azure-podvm-image-handler.sh
snir911: i didn't see such option for federated token

controllers/image_generator.go
snir911: Yes, it is, i can add `if (os.Getenv("CLIENTID") != "" && os.Getenv("TENANTID") != "" && os.Getenv("SUBSCRIPTIONID") to understand whether this is STS, or to add logic for both cases, however, i find it a bit out of scope for image-generator, in any case these are only the essentials and checked again in the job.

controllers/credentials_controller.go
snir911: i removed the comment, i planned to avoid using the label and eventually kept it

bpradipt (approved):
/lgtm

config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -132,11 +135,34 @@ function login_to_azure() {
...
+
+        [[ -z "${FEDERATED_TOKEN}" ]] && error_exit "Federated token is empty"
+
+        az login --service-principal \
+            --user="${AZURE_CLIENT_ID}" \
+            --federated-token="${FEDERATED_TOKEN}" \
+            --tenant="${AZURE_TENANT_ID}" ||
+            error_exit "Failed to login to Azure using federated token"
+    else
+        error_exit "Neither AZURE_CLIENT_SECRET nor AZURE_FEDERATED_TOKEN_FILE is set"
```
gkurz: AFAICT this is already handled in `verify_vars()` ... do we really need to add another user for these variable names ?

snir911: Correct, not really used, just a generic fallback, I'd keep it just in case, but i can also removed if you prefer.

wainersm:
Hi @snir911 ,

I don't have further comments, thanks for addressing all my concerns. There are smalls improvements here and there (e.g. log messages), but I think it follows the current standard of the operator.

Based on the docs that you shared with me, the implementation is correct. I am saying that because I am not an expert on the area, so based my reviews on these docs.

/lgtm

controllers/image_generator.go
gkurz: > Yes, it is, i can add `if (os.Getenv("CLIENTID") != "" && os.Getenv("TENANTID") != "" && os.Getenv("SUBSCRIPTIONID") to understand whether this is STS, or to add logic for both cases, however, i find it a bit out of scope for image-generator, in any case these are only the essentials and checked again in the job.

What is the logic that decides where user provided values should be validated ?

snir911: i meant to something like:

```
azureSecretKeys := []string{"AZURE_TENANT_ID", "AZURE_CLIENT_ID", "AZURE_CLIENT_SECRET", "AZURE_SUBSCRIPTION_ID"}
azureFederatedKeys := []string{"AZURE_TENANT_ID", "AZURE_CLIENT_ID", "AZURE_FEDERATED_TOKEN_FILE", "AZURE_SUBSCRIPTION_ID"}
...
if checkKeysPresentAndNotEmpty(peerPodsSecret.Data, azureSecretKeys) || checkKeysPresentAndNotEmpty(peerPodsSecret.Data, azureFederatedKeys)
```

config/peerpods/podvm/azure-podvm-image-handler.sh
gkurz: This is dead code actually... but I understand your point. Another option could be to drop the redundant check in `verify_vars()`.

controllers/image_generator.go
gkurz: This would work but it doesn't quite answer my question...  we can validate user provided inputs upfront in the controller or later in the job itself, or both as the current code seems to be doing. I suspect this should be reworked at some point for clarity.

Until then I'd go with the current status quo of validating in both places.

snir911: @gkurz, done, i got you question, validating is needed in both case to :
1. support running the job as standalone job/script
2. avoid executing job that will surely fail in advance.

config/peerpods/podvm/azure-podvm-image-handler.sh
snir911: removed (to match previous code also )

PR #881, opened 2025-08-19
Delete kata-remote runtime class when peerpods gets disabled

gkurz:
We want peerpods to be fully turned off if enablePeerPods transitions from true to false. From a user POV this means that the kata-remote runtime class shouldn't be available anymore. 

Adapt the KataConfig install path to support this transition.

Fixes: https://issues.redhat.com/browse/KATA-4109

**- What I did**

**- How to verify it**


controllers/peerpods.go
```diff
@@ -0,0 +1,482 @@
+/*
```
bpradipt, line 1: @gkurz nice.. Long due..

bpradipt (approved):
/lgtm

gkurz:
Converted back to draft as it needs some more testing on my side before we merge.

controllers/peerpods.go
```diff
@@ -552,6 +552,57 @@ func (r *KataConfigOpenShiftReconciler) processPeerPods(enable bool) (ctrl.Resul
 
 		// Reset the in progress condition
 		r.resetInProgressCondition()
+	} else {
```
beraldoleal: For semantic clarity, I would recommend to split this into `enablePeerpods()` and `disablePeerpods()`.

Calling `processPeerpods(false)` doesn’t make the intent obvious without checking the signature, and separate methods would also trim down the lines here.

beraldoleal: And I just noticed we have already a disablePeerpods(). So maybe move that there? Anyway, take this as nit suggestion, if isn't trivial to fix, that is fine.

controllers/peerpods.go
```diff
@@ -570,6 +570,17 @@ func (r *KataConfigOpenShiftReconciler) processPeerPods(enable bool) (ctrl.Resul
...
+		if err != nil {
+			r.setInProgressConditionToBlockedByExistingKataPods(err.Error())
+			updErr := r.Client.Status().Update(context.TODO(), r.kataConfig)
+			if updErr != nil {
+				return ctrl.Result{}, updErr
+			}
+			r.Log.Info("Kata Peer PODs are present. Requeue for reconciliation ")
+			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+		}
+
```
beraldoleal, line 627: Just to check: if a pod is still running, do we keep requeuing forever? That might leave us in a “blocked” state with noisy logs?

Also, since the RuntimeClass stays in place, would new pods still be scheduled?

beraldoleal (commented):
Hi @gkurz , thanks for this. Appreciate the refactoring part. Always good to see those.

I left a nit suggestion and some questions for you.

controllers/peerpods.go
gkurz: I'll look into that.

controllers/peerpods.go
gkurz: Yes for all questions.

Rationale is that it is recommended to first delete pods and possibly reschedule them to some other runtime class before deleting the runtime class they're currently using. This requires the KataConfig owner (admin) to sync with the pod owners (users).

beraldoleal: Thanks. We will just need to document this properly to let admins know this is not a kill switch.

gkurz:
@beraldoleal @bpradipt PTAL again.

Notable changes :
- stop using `sigs.k8s.io/controller-runtime/pkg/reconcile` directly (cleanup)
- separate enable and disable path as asked by @beraldoleal 
- fix return path for reschedule requests

I'm leaving the PR in draft mode as I'm still having problem testing this in my setup...

controllers/openshift_controller.go
```diff
@@ -333,261 +318,6 @@ func (r *KataConfigOpenShiftReconciler) removeLogLevel() error {
...
-							},
-						},
-					},
-				},
-			},
-		},
-	}
-}
-
-func (r *KataConfigOpenShiftReconciler) processDaemonsetForMonitor() *appsv1.DaemonSet {
```
pmores, line 491: This function is not peer pods though, is it?

controllers/peerpods.go
pmores: This behaviour is not new BTW.

gkurz:
After discussing with @snir911 , another thing to do is to make sure the credentials controllers doesn't delete the peer pods secret *before* the podvm image deletion is started. Doable with finalizers.

controllers/openshift_controller.go
gkurz: Oops you're right... I fumbled my cut-n-paste obviously :wink:. Good catch @pmores !

controllers/peerpods.go
gkurz: > This behaviour is not new BTW.

Yes this behavior is already [documented](https://docs.redhat.com/fr/documentation/openshift_sandboxed_containers/1.10/html/deploying_openshift_sandboxed_containers/uninstalling-osc) for the full uninstallation of kata (AKA delete `KataConfig`). This change applies the same logic for the uninstallation of peer pods only.

PR #885, opened 2025-08-21
osc-release-v1.10: update the tekton files with new IDs

littlejawa:
For the pipeline to trigger on the new branch, we need the tekton files to be updated so that they reference the new Konflux object names, branch name, and quay.io image location.


dbkreling (approved):
LGTM

PR #888, opened 2025-08-21
[1.10] osc-operator-bundle: use quay.io for the repository

littlejawa:
As we did for the initial onboarding: konflux nudge PRs won't happen if we reference registry.redhat.io until we have a release plan in place.

In the meantime, I'm reverting the references to the quay.io repo so that we can have working test images.


littlejawa:
/hold
I think the right move is to deploy the ReleasePlan. Looking into it.

PR #903, opened 2025-08-25
coco: approve InstallPlans via Subscription

beraldoleal:
The install was hanging because the script tried to match and approve an InstallPlan by a hardcoded CSV (e.g., v1.9.0), while the catalog often resolved a different CSV (e.g., v1.10.1).

We now approve the InstallPlan referenced directly by the Subscription, which works regardless of the resolved CSV, and we stop pinning startingCSV so the channel picks the right version automatically. If needed, you can still pin to a specific version via OSC_OPERATOR_CSV.

### DIFF
--- a/scripts/install-helpers/baremetal-coco/README.md
+++ b/scripts/install-helpers/baremetal-coco/README.md
@@ -58,7 +58,16 @@ The deployment sequence is described below:
 
 ## Install OSC operator GA release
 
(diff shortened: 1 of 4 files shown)


niteeshkd:
It works for me. Thanks @beraldoleal

bpradipt (approved):
/lgtm
Thanks @beraldoleal

PR #941, opened 2025-09-01
Drop the peerpod machine configs

gkurz:
These files are now shipped by the RPM.

https://issues.redhat.com/browse/KATA-3969


gkurz:
This change breaks peer pods with the current RPM. Let's wait for an official build of the kata 3.21 RPM to be available.

/hold

bpradipt (approved):
/lgtm
Thanks @gkurz

gkurz:
/unhold

It is better to drop these now because the manual installation of a kata 3.21 RPMs end up conflicting with the MCO view of the OS and degrades the node.

PR #944, opened 2025-09-02
Fix cleanup error: we need write auth on the new folder created for kubebuilder assets

gkurz:
Re-apply bef333e8 that was mistakenly removed by #451


beraldoleal (approved):
LGTM, thanks.

bpradipt (approved):
/lgtm

littlejawa:
The enterprise contract failure seems unrelated. We started to have these errors since yesterday evening, and I'm not sure where they come from. I'm looking into this.

littlejawa:
Confirmed from the Konflux side: there are some issues on the enterprise contract that raise false positive errors on some tests. This is being worked on by the Konflux team. In the meantime, we can ignore this error.

PR #950, opened 2025-09-03
ibmcloud: initial support for IBM Cloud as a provider

Pacho20:
Made modifications to support IBM Cloud as a cloud provider. Changes include:
- Adding IBM Cloud to the list of supported providers.
- Updating the image generator with new return values and an error message to reflect that IBM Cloud only supports prebuilt images; in-cluster image building is currently not supported.
- Introducing a new function to apply IBM-specific configuration to the CAA DaemonSet.

This PR depends on PR #825, as the Operator only works using the DaemonSet deployment mode on IBM Cloud.

### DIFF
--- a/config/peerpods/podvm/podvm-builder.sh
+++ b/config/peerpods/podvm/podvm-builder.sh
@@ -31,6 +31,11 @@ function install_gcp_deps() {
   /scripts/gcp-podvm-image-handler.sh -- install_binaries
 }
 
+# Function to install IBM Cloud deps
+function install_ibmcloud_deps() {
+  /scripts/ibmcloud-podvm-image-handler.sh -- install_cli
+  /scripts/ibmcloud-podvm-image-handler.sh -- install_binaries
+}
 
 # Function to check if peer-pods-cm configmap exists
 function check_peer_pods_cm_exists() {
@@ -381,6 +386,10 @@ libvirt)
   ;;
 gcp)
   install_gcp_deps
+  ;;
+ibmcloud)
+  echo "IBM Cloud doesn't support this feature, please provide IBMCLOUD_PODVM_IMAGE_ID in your peer-pods-cm"
+  exit 1
   ;;
 *)
   echo "CLOUD_PROVIDER is not set to azure or aws or libvirt"
--- a/controllers/image_generator.go
+++ b/controllers/image_generator.go
@@ -680,18 +680,8 @@ func (r *ImageGenerator) validatePeerPodsConfigs() error {
 
 func checkKeysPresentAndNotEmpty(data interface{}, keys []string) bool {
 	// Convert the input map to a map[string]string
-	var strMap map[string]string
-
-	switch v := data.(type) {
-	case map[string]string:
-		strMap = v
-	case map[string][]byte:
-		strMap = make(map[string]string)
-		for key, value := range v {
-			strMap[key] = string(value)
-		}
-	default:
-		// Unsupported type
+	strMap := toStrMap(data)
+	if strMap == nil {
 		return false
 	}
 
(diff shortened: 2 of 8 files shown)


bpradipt:
@Pacho20 thanks for splitting this out. I took a quick look and here are some general guidelines that will be helpful:
- Take a look at https://github.com/openshift/sandboxed-containers-operator/blob/devel/controllers/image_generator.go to understand the support added for existing cloud-providers. Just add only the absolute min for supporting IBMCloud provider. Don't tie DaemonSet deployment with this PR. The daemonset specific changes will go in the DS PR. This PR is bare minimal to support IBMCloud provider (doesn't matter if IBMCloud doesn't support MCO). The hardening w.r.to DS approach comes in the other PR.
- Add an image config file like for other providers under - https://github.com/openshift/sandboxed-containers-operator/tree/devel/config/peerpods/podvm (ibmcloud-podvm-image-cm.yaml)
- Add an handler file like for other providers under - https://github.com/openshift/sandboxed-containers-operator/tree/devel/config/peerpods/podvm  (ibmcloud-podvm-image-handler.sh)
- operator-built or pre-built image is set here based on PODVM_IMAGE_URI - https://github.com/openshift/sandboxed-containers-operator/blob/02ef23f9c897c0667c2f65edabf5584a9ef2b339/config/peerpods/podvm/podvm-builder.sh#L45
So ensure that ibmcloud-podvm-image-cm.yaml has PODVM_IMAGE_URI set before creating the job in this method- https://github.com/openshift/sandboxed-containers-operator/blob/devel/controllers/image_generator.go#L811. So imho this should be enough.

controllers/image_generator.go
```diff
@@ -680,18 +700,8 @@ func (r *ImageGenerator) validatePeerPodsConfigs() error {
...
-	case map[string]string:
-		strMap = v
-	case map[string][]byte:
-		strMap = make(map[string]string)
-		for key, value := range v {
-			strMap[key] = string(value)
-		}
-	default:
-		// Unsupported type
+	strMap := toStrMap(data)
```
gkurz, line 703: Ideally, you should have introduced this helper function in a preceding patch.

gcoon151: @gkurz do you want us to move this to another PR?

gkurz: Another PR would be overkill. Just a separate commit to introduce `checkAnyKeyPresentWithValue()` and `toStrMap()` should be ok.

The idea is to have smaller commits :
- it is way easier to review several small patches over a big monolithic one
- `git bisect` is more efficient

Pacho20: I've created a separate commit—let me know if there's anything else you think should be changed.

bpradipt (approved):
/lgtm
The refactor and addition looks good. Thanks @Pacho20 
Let's wait for @gkurz feedback and we should be good to go

controllers/image_generator.go
gkurz: It is usually recommended to write down some context in the commit log, e.g.

```
image_generator: add function checkAnyKeyPresentWithValue

We already have a helper function that checks that all keys from a given set are present with
a non-nil value. Now add a new helper that checks if any key out of of a given set is present with
a non-nil value. This will be required to support IBM Cloud authentication.

Factor out the map conversion code to a function that can be used by the two check functions.
```

controllers/image_generator.go
```diff
@@ -708,6 +698,49 @@ func checkKeysPresentAndNotEmpty(data interface{}, keys []string) bool {
...
+	}
+
+	// Iterate over the list of keys and check if any key exists with a non-empty value.
+	for _, key := range keys {
+		value, ok := strMap[key]
+		if ok && value != "" {
+			return true // Found a key with a non-empty value.
+		}
+	}
+
```
gkurz, line 740: All this boiler plate code of converting the map and iterating over it could be itself factored out to a separate function and you'd pass it the `if` block as a closure. This can be done later, not strictly required now.

controllers/image_generator.go
Pacho20: Extended the commit message based on your suggestion.

config/peerpods/podvm/podvm-builder.sh
```diff
@@ -317,8 +322,8 @@ function delete_podvm_image() {
     if [ "${UPDATE_PEERPODS_CM}" == "yes" ]; then
       kubectl patch configmap peer-pods-cm -n openshift-sandboxed-containers-operator --type merge -p "{\"data\":{\"LIBVIRT_IMAGE_ID\":\"\"}}"
     fi
-
-    ;;
+    
+  ;;
```
gkurz: White space damage... 
```suggestion

  ;;
```

gkurz (approved):
With the space damage fixed in the second commit

/lgtm

Since CI already passed, we should be able to merge this quickly.

Thanks @Pacho20 !

config/peerpods/podvm/podvm-builder.sh
Pacho20: Fixed.

bpradipt (approved):

PR #973, opened 2025-09-04
[1.10] osc-podvm-builder: update the builder image to latest ubi9.6

littlejawa:


littlejawa:
Mintmaker is doing the same (#974), closing

PR #1019, opened 2025-09-10
Update FBC files to the 1.10.2 release

dbkreling:
Add information for 1.10.2 upcoming release.
Update bundle existing reference with the one from the snapshot.
Add files created after running "fbc/render.sh" script.


littlejawa (approved):
/lgtm

PR #1021, opened 2025-09-11
test-fbc: set the catalog's tag to 1.11.x-timestamp

littlejawa:
Keeping the tag to 1.10.1 is confusing when bugs are reported from this catalog. Using "1.11.x" to mark the fact that this is not an official release version, but an intermediate build.

Also, this commit does not change the version strings everywhere else, it is just about the test catalog.

### DIFF
--- a/.tekton/osc-test-fbc-push.yaml
+++ b/.tekton/osc-test-fbc-push.yaml
@@ -309,7 +309,7 @@ spec:
       - name: IMAGE_DIGEST
         value: $(tasks.build-image-index.results.IMAGE_DIGEST)
       - name: ADDITIONAL_TAGS
-        value: ["latest", "1.10.1-$(tasks.clone-repository.results.commit-timestamp)"]
+        value: ["latest", "1.11.0-$(tasks.clone-repository.results.commit-timestamp)"]
       runAfter:
       - build-image-index
       taskRef:


littlejawa:
@tbuskey FYI - to differentiate the next release from the 1.10 work

.tekton/osc-test-fbc-push.yaml
```diff
@@ -309,7 +309,7 @@ spec:
       - name: IMAGE_DIGEST
         value: $(tasks.build-image-index.results.IMAGE_DIGEST)
       - name: ADDITIONAL_TAGS
-        value: ["latest", "1.10.1-$(tasks.clone-repository.results.commit-timestamp)"]
+        value: ["latest", "1.11.x-$(tasks.clone-repository.results.commit-timestamp)"]
```
tbuskey: Shouldn't this be 1.11.0?  Or does the x get translated before the tag?

I'm expecting [0-9]*.[0-9]*.[0-9]*-[0-9]* in prow to extract the expected version to check against the csv in a test
I can modify though

gkurz: Nothing translates the `x` in the catalog name AFAICT. @littlejawa made this change _just_ to avoid the confusion on the `Affected Version` field like in https://issues.redhat.com/browse/KATA-4132.
You can either relax the check in your test (e.g. don't check the patch level) or the catalog name can be changed to be `1.11.0-....`. Up to you to decide @tbuskey and @littlejawa.

littlejawa: I'm fine both ways. I used "x" because I want to convey the idea that this is not an official release, and will never be. It reflects ongoing code changes. I could have named it "1.11.next" for instance.
I will put it to "1.11.0", we all know this is not released yet, so it works. But then it means we'll need to bump that tag as soon as 1.11.0 gets out, to avoid confusion again.

gkurz: Let's go that way for now to unblock Tom. We'll need a more global thinking about when/where to bump versions during the development window.

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #1023, opened 2025-09-12
[1.10] osc 1.10.3: bump version strings

littlejawa:
Preparing for 1.10.3: bump version string now so there's no doubt about what we're testing


config/manifests/bases/sandboxed-containers-operator.clusterserviceversion.yaml
gkurz, line 1: The changes in this file should be propagated to the CSV within the bundle directory.

littlejawa: Forgot about it :-( 
Done

littlejawa:
must-gather build fails for s390x because it can't find the builder image we use.
It works for amd64 though, so something's wrong with the repos?

littlejawa:
We're using the "latest" tag on our builder, I suspect we're hurting an issue with image availability following a recent update.
Will retry later, no urgency on this PR anyway.

must-gather/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.redhat.io/openshift4/ose-must-gather:latest as builder
+FROM registry.redhat.io/openshift4/ose-must-gather:v4.15 as builder
```
gkurz: There's also a rhel9 version at https://catalog.redhat.com/en/software/containers/openshift4/ose-must-gather-rhel9/652809b17e435c863687fed2. Shouldn't we use that since we use ubi9 as base image ?

littlejawa: Yes, that's probably better.
I'm changing that, using "v4.19" as the tag. Hopefully, Mintmaker will pick it up and start giving us PRs to update it when it changes - this would help control if/when we want to change our builder image, instead of just taking "latest" and have arbitrary build failures.

gkurz:
> ...
> This commit modifies the builder image to use "ose-must-gather-rhel9",
> and uses the tag "v4.19" rather than "latest".
>
> Signed-off-by: Julien Ropé <jrope@redhat.com>
>
> Signed-off-by: Julien Ropé <jrope@redhat.com>

Nit: double sob tag.

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #1027, opened 2025-09-12
Remake bundle in docker

gkurz:
When bumping OSC versions or any other change that affects the final CSV, we need to run `make bundle`. This is a manual action that is error prone. Also `make bundle` always do extra changes that we don't want to merge (e.g. changing image digests into tags or reverting wanted changes in bundle related files).

Add all the logic to call `make bundle` from a container :
- add `controller-gen` as done in the operator build for hermetic builds
- add `kustomize` as a dep in go.mod and do like `controller-gen` for hermetic builds
- add  `operator-sdk` using hermeto's generic fetcher

The OSC operator image is now stored in the `config/manager/manager.yaml` file for reference. It is monitored by konflux and passed as `IMG` to `make bundle`. This requires the `yq` command. Use the same workaround as `controller-gen` to honor hermetic builds in konflux.


bundle.Dockerfile
```diff
@@ -1,3 +1,41 @@
...
+COPY cmd cmd/
+COPY config config/
+COPY controllers controllers/
+
+# Copy our controller-gen script to work around hermetic build issues
+# See comments in the script itself for more details.
+COPY controller-gen bin/
+
+# get the version of controller-gen in an env variable for reusing
+RUN echo "export CONTROLLER_TOOLS_VERSION=$(grep controller-tools go.mod | awk '{print $2}')" > controller-tools-ver.env
```
littlejawa:
```suggestion
RUN echo "export CONTROLLER_TOOLS_VERSION=$(grep controller-tools go.mod | awk '{print $2}')" > bundle-build.env
```

bundle.Dockerfile
```diff
@@ -1,3 +1,41 @@
...
+# rename the script to use the same version as defined in our go.mod file
+RUN . ./controller-tools-ver.env && mv bin/controller-gen bin/controller-gen-$CONTROLLER_TOOLS_VERSION
+
+COPY kustomize bin/
+
+# copy our yq script for hermetic builds
+COPY yq bin/
+
+# save the current version of the operator image in an env variable for reusing
+RUN echo "export IMG=$(./bin/yq .spec.install.spec.deployments[0].spec.template.spec.containers[0].image bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml)" > operator-image.env
```
littlejawa:
```suggestion
RUN echo "export IMG=$(./bin/yq .spec.install.spec.deployments[0].spec.template.spec.containers[0].image bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml)" >> bundle-build.env
```

bundle.Dockerfile
```diff
@@ -1,3 +1,41 @@
...
+
+# make sure 'make bundle' uses the right version of controller-gen and preserves the operator image
+RUN . ./controller-tools-ver.env && . ./operator-image.env && make bundle
```
littlejawa:
```suggestion
RUN . ./bundle-build.env && make bundle
```

littlejawa (commented):
I think we can use a single env file for the parameters we give to "make"

gkurz:
> I think we can use a single env file for the parameters we give to "make"

Heh I did that already locally :smile:

littlejawa (approved):
/lgtm
Thanks @gkurz !

gkurz:
I need to check the resulting bundle is as expected.

/hold

gkurz:
Too complex.

PR #1046, opened 2025-09-17
osc-must-gather: change builder image

gkurz:
Using "latest" is causing an issue in our build. For some reason the s390x build can't find the "latest" tagged image. Also, we were still using "ose-must-gather", while there is an "ose-must-gather-rhel9" image available that's probably better suited for our release.

This commit modifies the builder image to use "ose-must-gather-rhel9", and uses the tag "v4.19" rather than "latest".


littlejawa (approved):
/lgtm

PR #1052, opened 2025-09-18
[1.10] Cherry picks for labels

littlejawa:
This is a backport of the changes made on the 'devel' branch for image labels:
- #1042 
- #1043 
- #1044 
- #1047


gkurz (approved):
I could check that the cherry-picks are identical to the original changes (only visible differences are `version=` lines in the context which is fine).

/lgtm

Thanks @littlejawa !

Now let's hope that the on-push pipeline rebuilds everything needed :crossed_fingers:

PR #1058, opened 2025-09-18
controller: Add overhead annotation using pod mutating webhook

### LINKED ISSUE #1057: Implement pod-mutating web hook to pass overhead to Kata runtime
## Overview

This proposal outlines how to add support for the `io.katacontainers.config.hypervisor.memory_overhead` annotation in the OpenShift Sandboxed Containers Operator. This annotation allows external operators to specify VM memory overhead, which is crucial for proper cgroup management and [preventing OOM issues](https://github.com/kata-containers/kata-containers/issues/6129).

## Background

The Kata Containers runtime will soon supports a memory overhead compensation mechanism that requires the `io.katacontainers.config.hypervisor.memory_overhead` annotation to be set on pods. This annotation works in conjunction with the existing RuntimeClass PodFixed overhead to ensure proper memory accounting.

## Current State

The operator currently:
- Creates RuntimeClass objects with PodFixed overhead (CPU and memory)
- Uses hardcoded memory overhead values: 350Mi for kata, 120Mi for kata-remote
- Does not inject any Kata-specific annotations on pods

### 1. Add Memory Overhead Configuration to KataConfigSpec

Add a new field to the KataConfigSpec to allow users to configure memory overhead:

```go
// KataConfigSpec defines the desired state of KataConfig
type KataConfigSpec struct {
    // ... existing fields ...
    
    // MemoryOverheadMB specifies the memory overhead in MB for Kata containers
    // This value will be used to set the io.katacontainers.config.hypervisor.memory_overhead annotation
    // +optional
    // +kubebuilder:default:=350
    MemoryOverheadMB *int32 `json:"memoryOverheadMB,omitempty"`
}
```

### 2. Create a Pod Mutation Webhook

Create a new mutating admission webhook that:
- Intercepts pod creation/update events
- Checks if the pod uses a Kata runtime class
- Injects the memory overhead annotation based on the KataConfig

#### 3.1 Webhook Structure

```go
// +kubebuilder:webhook:path=/mutate-pods-v1,mutating=true,failurePolicy=fail,sideEffects=None,groups="",resources=pods,verbs=create;update,versions=v1,name=mpods.kb.io,admissionReviewVersions=v1

type PodMutator struct {
    client.Client
    Log logr.Logger
}

func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
    // Implementation to inject memory overhead annotation
}
```

#### 3.2 Annotation Injection Logic

```go
func (m *PodMutator) injectMemoryOverheadAnnotation(pod *corev1.Pod) error {
    // Check if pod uses Kata runtime class
    if !m.usesKataRuntime(pod) {
        return nil
    }
    
    // Get memory overhead from KataConfig
    memoryOverhead, err := m.getMemoryOverheadFromKataConfig()
    if err != nil {
        return err
    }
    
    // Set annotation
    if pod.Annotations == nil {
        pod.Annotations = make(map[string]string)
    }
    
    pod.Annotations["io.katacontainers.config.hypervisor.memory_overhead"] = fmt.Sprintf("%d", memoryOverhead)
    
    return nil
}
```

#### 3.3 Runtime Class Detection

```go
func (m *PodMutator) usesKataRuntime(pod *corev1.Pod) bool {
    runtimeClassName := pod.Spec.RuntimeClassName
    if runtimeClassName == nil {
        return false
    }
    
    // Check against known Kata runtime classes
    kataRuntimeClasses := []string{"kata", "kata-remote"}
    for _, rtc := range kataRuntimeClasses {
        if *runtimeClassName == rtc {
            return true
        }
    }
    
    return false
}
```

#### 4.1 Default Values

- Default memory overhead: 350MB (matching current RuntimeClass overhead; this is [likely to change soon](https://github.com/kata-containers/kata-containers/issues/6533))
- Configurable per KataConfig instance
- Fallback to hardcoded values if not specified

#### 4.2 Validation

- Memory overhead must be positive
- Maximum reasonable limit (e.g., 4GB) to prevent abuse
- Validation webhook to ensure consistency

c3d:
In order for the Kata runtime to correctly deal with hot plugging and set a VM size that is compatible with the host cgroup size [1], we need to pass the known overhead (from the RuntimeClass) to the runtime using an annotation.

This code implements a mutating web hook to modify the pod so that the annotation is added when the runtime class is Kata.

Fixes: #1057 

When hot-plugging memory in Kata (typically because the container spec provides a memory limit), the host-side cgroup and the maximum VM size currently become inconsistent, because the host-side cgroup is defined as "pod overhead + container limit" whereas the actual maximum VM size is "default VM size + container limit".

The in-guest cgroup is insufficient to constrain the memory usage sufficiently, because it does not account for things like kernel buffers. This leads to situations where we end up hitting a host-side cgroup out-of-memory, which kills the VM in a nasty way: we lose connexion to the kata runtime itself, and therefore things like container exit status, etc.

**- What I did**

This adds a pod mutating web hook that injects an annotation that [the runtime can use](https://github.com/kata-containers/kata-containers/pull/11824) to adjust the amount of memory being hot-plugged to avoid the issue.

**- How to verify it**

There is some local testing of the operator's pod mutation logic which can be exercised by running the run_pod_overhead_mutators.sh

Add pod-mutating web hook to inject memory overhead annotation for the kata runtime

Assisted-by: Cursor with gpt5 model (test generation)


c3d:
I don't have the proper privileges to change labels on this repo, but this is still work-in-progress. The actual interaction with the runtime remains to be tested.

api/v1/kataconfig_webhook.go
```diff
@@ -61,16 +61,28 @@ func (r *KataConfig) ValidateCreate(ctx context.Context, obj runtime.Object) (ad
...
 
-	if len(kataConfigList.Items) == 1 {
-		return nil, fmt.Errorf("A KataConfig instance already exists, refusing to create a duplicate")
+	if r.Spec.MemoryOverheadMB != nil {
+		if *r.Spec.MemoryOverheadMB < 60 {
+			return nil, fmt.Errorf("memoryOverheadMB must be at least 60MB")
+		}
+		if *r.Spec.MemoryOverheadMB > 4096*1024 {
+			return nil, fmt.Errorf("memoryOverheadMB must be at most 4GB")
+		}
```
gkurz: Can you clarify how you came up with these numbers ?

api/v1/kataconfig_webhook.go
```diff
@@ -85,6 +97,15 @@ func (r *KataConfig) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.
...
 
+	if r.Spec.MemoryOverheadMB != nil {
+		if *r.Spec.MemoryOverheadMB < 60 {
+			return nil, fmt.Errorf("memoryOverheadMB must be at least 60MB")
+		}
+		if *r.Spec.MemoryOverheadMB > 4096*1024 {
+			return nil, fmt.Errorf("memoryOverheadMB must be at most 4GB")
+		}
+	}
+
```
gkurz, line 90: Can you move the duplicate code block to some helper ?

config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
```diff
@@ -124,6 +124,12 @@ spec:
                 description: Sets log level on kata-equipped nodes.  Valid values
                   are the same as for `crio --log-level`.
                 type: string
+              memoryOverheadMB:
+                default: 350
+                format: int32
+                maximum: 2048
+                minimum: 1
```
gkurz: Why do the min and max don't match what's tested in the code above ?

config/manager/manager.yaml
```diff
@@ -86,7 +97,7 @@ spec:
             - name: RELATED_IMAGE_PODVM_BUILDER
               value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-builder-rhel9@sha256:517b1d76dad553871c745a315aa73c341f4672029704d50992a3eb663945e6dc
             - name: RELATED_IMAGE_PODVM_PAYLOAD
-              value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9@sha256:211a4291cd13302a18ae2766a8e680a2593450ce31f119e78ac3367c5858104a 
+              value: registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9@sha256:211a4291cd13302a18ae2766a8e680a2593450ce31f119e78ac3367c5858104a
```
gkurz: Current value in the `devel` branch is :

```
registry.redhat.io/openshift-sandboxed-containers/osc-podvm-payload-rhel9@sha256:9c07280746952c7bd4ea811cd15e6dc2e35e0f6c2015e21436dfeaa418ec5799 
```

More generally,  these lines with image digests are _maintained_ by mintmaker. You should never have to modify them in a PR, so please just drop this change.

This being said, there's indeed a trailing space... even in the current HEAD of `devel`, Cc @littlejawa .

gkurz: It seems that this space has been here for a very long time and mintmaker PRs gladly honor this formatting nit :wink:

gkurz: Let's fix that in a separate PR.

controllers/pod_mutator.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+const (
+	// Kata memory overhead annotation
+	KataMemoryOverheadAnnotation = "io.katacontainers.config.hypervisor.memory_overhead"
+	KataMemoryOverheadDefault    = 350
+
+	// Known Kata runtime classes
+	KataRuntimeClass       = "kata"
+	KataRemoteRuntimeClass = "kata-remote"
```
gkurz: No need to duplicate strings that we already have in the main controller file. The controller maintains a list of the known kata runtime class in the KataConfig.

https://github.com/openshift/sandboxed-containers-operator/blob/devel/api/v1/kataconfig_types.go#L51

controllers/pod_mutator_test.go
```diff
@@ -0,0 +1,802 @@
...
+		},
+		{
+			name: "Pod with Kata-remote runtime class should get annotation",
+			pod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-pod-remote",
+					Namespace: "default",
+				},
+				Spec: corev1.PodSpec{
+					RuntimeClassName: stringPtr("kata-remote"),
```
gkurz, line 74: Maybe use https://github.com/openshift/sandboxed-containers-operator/blob/devel/controllers/peerpods.go#L45 ?

And fix other sites.

controllers/pod_mutator_test.go
```diff
@@ -0,0 +1,802 @@
...
+	}{
+		{
+			name: "Pod with Kata runtime class should get annotation",
+			pod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-pod",
+					Namespace: "default",
+				},
+				Spec: corev1.PodSpec{
+					RuntimeClassName: stringPtr("kata"),
```
gkurz, line 45: Maybe use https://github.com/openshift/sandboxed-containers-operator/blob/devel/controllers/openshift_controller.go#L78 ?

And fix other sites.

api/v1/kataconfig_types.go
```diff
@@ -44,6 +44,12 @@ type KataConfigSpec struct {
 	// +optional
 	// +kubebuilder:default:=false
 	EnablePeerPods bool `json:"enablePeerPods"`
+
+	// +optional
+	// +kubebuilder:default:=350
```
gkurz: Is this taken from https://github.com/openshift/sandboxed-containers-operator/blob/devel/controllers/openshift_controller.go#L81 ?

controllers/pod_mutator_test.go
gkurz: You might need to rebase to get the changes from #1069.

api/v1/kataconfig_types.go
gkurz: If the answer is yes, should they stay in sync ?

gkurz:
> I don't have the proper privileges to change labels on this repo, but this is still work-in-progress. The actual interaction with the runtime remains to be tested.

Converting the PR to draft is the way to go.

api/v1/kataconfig_webhook.go
c3d: Added comments:
```
                // The Linux kernel was tested not to boot reliably with less than 60MB                                                                                                                                                                                  
                if *r.Spec.MemoryOverheadMB < 60 {                              │
                        return nil, fmt.Errorf("memoryOverheadMB must be at least 60MB")
                }                                                               │
                // Kata configures the VM with 2G by default, so overhead twice as big is likely an error                                                                                                                                                                
                if *r.Spec.MemoryOverheadMB > 4096*1024 {                       │
                        return nil, fmt.Errorf("memoryOverheadMB must be at most 4GB")
                }                                                               │
```

api/v1/kataconfig_webhook.go
c3d: Done

config/crd/bases/kataconfiguration.openshift.io_kataconfigs.yaml
c3d: Oversight.

config/certmanager/certificate.yaml
```diff
@@ -6,15 +6,15 @@ apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
   name: selfsigned-issuer
-  namespace: system
+  namespace: openshift-sandboxed-containers-operator
```
gkurz: What's the reason for that ? This comes from Operator SDK AFAIK. Not sure it should be changed...

config/manager/manager.yaml
```diff
@@ -12,7 +12,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: system
+  namespace: openshift-sandboxed-containers-operator
```
gkurz: Same remark. This is rendered with `openshift-sandboxed-containers-operator` when updating the bundle.

config/manager/manager.yaml
```diff
@@ -55,30 +55,41 @@ spec:
...
+              readOnly: true
           envFrom:
             - secretRef:
                 name: peer-pods-secret
                 optional: true
             - configMapRef:
                 name: peer-pods-cm
                 optional: true
-          image: controller:latest
+          image: quay.io/c3d/openshift-sandboxed-containers-operator:v1.10.1-memory-overhead
```
gkurz: Same here, this is rendered in the final bundle. And we're not going to merge a reference to your personal repo, do we ? :wink:

api/v1/kataconfig_types.go
```diff
@@ -44,6 +44,12 @@ type KataConfigSpec struct {
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=2048
+	MemoryOverheadMB *int32 `json:"memoryOverheadMB,omitempty"`
```
bpradipt: Is updating CRD really needed? 
We have consciously avoided updating KataConfig CRD till now and instead using feature configmap. 
If we are updaing the CRD then we need to handle migration for installed CRDs and probably version it as well.

config/certmanager/certificate.yaml
c3d: This was for testing purpose, should not be committed. Dropping it.

config/manager/manager.yaml
c3d: Same reason, reverting that change too.

config/manager/manager.yaml
c3d: No, all this was for testing purpose and should not have been committed ;-)

api/v1/kataconfig_types.go
c3d: This is a very valid point. I will rework

bpradipt:
With K8s 1.34 there is MutatingAdmissionPolicy which provides an in-process alternative compared to a mutating webhook.
https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/

Just to keep in mind going forward

gkurz:
@c3d FYI there's some ongoing work in this area upstream but it is for runtime-rs only.

https://github.com/kata-containers/kata-containers/pull/12735

PR #1082, opened 2025-09-24
Bump to 1.11.0

gkurz:
Provide a script to bump the OSC version in the whole repo.

Bump to 1.11.0.


Dockerfile
```diff
@@ -41,7 +41,7 @@ RUN getent group nonroot || groupadd -o -g 499 nonroot
 # Red Hat labels
 LABEL name="openshift-sandboxed-containers/osc-rhel9-operator" \
 cpe="cpe:/a:redhat:confidential_compute_attestation:1.10::el9" \
-version="1.10.1" \
+version="1.11.0" \
```
littlejawa: I'm thinking we may want to keep the container LABEL version to "1.11", without the "z" version, so that an image doesn't seem tied to a specific "z" version.
This allows us to re-release the same images in a subsequent release, if only a subset of images needed to be modified, without making people confused, thinking they're not using the right one.

This is nitpicking honestly, because it really seems that this "version" label is not used anywhere :-/

gkurz: > I'm thinking we may want to keep the container LABEL version to "1.11", without the "z" version, so that an image doesn't seem tied to a specific "z" version. 

But an image can also be tied to a specific "z" version in case of bug fix, can't it ?

>This allows us to re-release the same images in a subsequent release, if only a subset of images needed to be modified, without making people confused, thinking they're not using the right one.

If the goal is to avoid to release an image that has same payload, would it be possible to re-label the existing image in this case ?

> This is nitpicking honestly, because it really seems that this "version" label is not used anywhere :-/

The "version" label is _just_ a user visible piece of information. It seems to be frequently present in images, e.g. ubi9.

dbkreling: I understand and agree in **not** having discrepant versions available to customers with different "z" digit when they refer to the same codebase. But I want to confirm the customers will be able to differentiate their current installation version.

Images contain 3 different tags visible in [the customer portal](https://catalog.redhat.com/en/software/containers/openshift-sandboxed-containers/osc-rhel9-operator/63eb68805620225430ca2c97?container-tabs=gti&gti-tabs=registry-tokens) (_latest_, _version-with-timestamp_ and _git_short_sha_). If the customers are able to compare their current installation with the latest released image available, I think we definitely should go with the two digit (X.Y) pattern for the version. If not, we should think how we could differentiate them.

must-gather/Dockerfile
```diff
@@ -24,7 +24,7 @@ COPY node-gather/node-gather-ds.yaml /etc/
 # Red Hat labels
 LABEL name="openshift-sandboxed-containers/osc-must-gather-rhel9" \
 cpe="cpe:/a:redhat:confidential_compute_attestation:1.10::el9" \
```
dbkreling, line 26: Should we bump 1.10 to 1.11 in this line?

Dockerfile
```diff
@@ -41,7 +41,7 @@ RUN getent group nonroot || groupadd -o -g 499 nonroot
 # Red Hat labels
 LABEL name="openshift-sandboxed-containers/osc-rhel9-operator" \
 cpe="cpe:/a:redhat:confidential_compute_attestation:1.10::el9" \
```
dbkreling, line 43: Same here, should we bump the cpe label here to 1.11?

scripts/bump-osc-version.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
```
dbkreling, line 6: Should we add a snippet with the usage of the script here, or else point the usage instructions in the development.md file to connect the two files?

dbkreling (commented):
Dropping some questions I was curious about if they should be changed now, later or ever.

dbkreling:
Oh, also, we have a reference here that was not updated. 
[config/peerpods/podvm/bootc/Containerfile.rhel#L2 ](https://github.com/gkurz/sandboxed-containers-operator/blob/11d0b7744b146db89a12705f5a8d125293b83c5d/config/peerpods/podvm/bootc/Containerfile.rhel#L2)
Is it relevant?

Dockerfile
littlejawa: The LABEL we see here is attached to the image. You can see it when inspecting the container image itself. I don't think we can re-label the image without re-building it.
It is different than a tag that we can apply/modify on an existing image.
We will be tagging images with x.y.z when we release them, even if they were already released before, but the LABEL could stay the same.

must-gather/Dockerfile
littlejawa: Eventually, yes.
But the cpe is defined by our ReleasePlanAdmission version, and this gets updated separately (and potentially closer to the release).
I think this should get updated along with the ReleasePlanAdmission, not with the rest of our version bumps.

littlejawa: For reference: 
cpe definition: (based on ReleasePlanAdmission version)
https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/main/prodsec/ose-osc.yaml#L3

Prod ReleasePlanAdmission version:
https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/main/config/stone-prd-rh01.pg1f.p1/product/ReleasePlanAdmission/ose-osc/osc-prod.yaml#L19

Stage ReleasePlanAdmission version:
https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/main/config/stone-prd-rh01.pg1f.p1/product/ReleasePlanAdmission/ose-osc/osc-stage.yaml#L19

Dockerfile
littlejawa: same answer :-)

Dockerfile
gkurz: Ok. I'll convert the version labels to be x.y only in a preparatory commit in the same PR.

gkurz:
New version drops patch level in version labels, adds usage to the script and a new OSC_RELEASE tag for the bootc container file.

littlejawa (approved):
/lgtm
Thanks @gkurz !

PR #1090, opened 2025-09-25
mirror-set: update the file to use both default and 1.10 mirrors

littlejawa:
We can have multiple mirrors defined in the image-mirrorset.yaml
With this change, we can use the same file for 1.10 and devel images tests.

Tested with the test-fbc from devel and 1.10 on the same cluster.


tbuskey:
I didn't know it could have multiple!
@wainersm we need to use this file in Prow instead of our own!

littlejawa:
@tbuskey @wainersm - I've tested it, but please double (triple) check if possible :-)

PR #1096, opened 2025-09-29
Update fbc versions to v1.10.3

dbkreling:
Bump versions and references for 1.10.3 release.


fbc/v4.18/catalog-template.yaml
```diff
@@ -53,6 +53,9 @@ entries:
       - name: sandboxed-containers-operator.v1.10.2
         replaces: sandboxed-containers-operator.v1.10.1
         skipRange: '>=1.1.0 <1.10.2'
+      - name: sandboxed-containers-operator.v1.11.2
```
littlejawa:
```suggestion
      - name: sandboxed-containers-operator.v1.10.3
```

fbc/v4.18/catalog-template.yaml
```diff
@@ -53,6 +53,9 @@ entries:
...
+        replaces: sandboxed-containers-operator.v1.10.3
```
littlejawa:
```suggestion
        replaces: sandboxed-containers-operator.v1.10.2
```

fbc/v4.18/catalog-template.yaml
```diff
@@ -53,6 +53,9 @@ entries:
...
+        skipRange: '>=1.1.0 <1.11.3'
```
littlejawa:
```suggestion
        skipRange: '>=1.1.0 <1.10.3'
```

littlejawa (approved):
/lgtm

.tekton/osc-fbc-4-19-push.yaml
```diff
@@ -38,7 +38,7 @@ spec:
     - name: name
       value: fbc-builder
     - name: bundle
-      value: quay.io/konflux-ci/tekton-catalog/pipeline-fbc-builder:devel@sha256:484d26ab153b67f4a85ba3f5ba43dccd26a4de71486b6eebdb3ea7ccae8aa57b
+      value: quay.io/konflux-ci/tekton-catalog/pipeline-fbc-builder:devel@sha256:8be144915833bf4d092191566882656c2e6ccadb9974f5ce2388f81d4b15a595
```
gkurz: Why is SHA `8be144915833bf4d092191566882656c2e6ccadb9974f5ce2388f81d4b15a595` while it is `e2ffda5b697f421ea98132aa16ad80b84c9998caa65dd527e72bc64a6b735e7d` everywhere else ?

littlejawa: Forgot it... But it doesn't count, as I still had issues with the new version of the task.
On the next commit, I ditch everything and use a locally defined pipeline (copied from the test-fbc one) instead of the remote version, and it seems to be working better.
I will remove that commit, and clean up the whole thing.

littlejawa:
Remaining errors are expected: the Enterprise Contract complains about the FIPS check error about dm-verity not shipping openssl - which is expected, as dm-verity just ships the VM image.
We have an exception for that on "prod" releases, but not on "stage" - we need to address that.
For now, I consider the checks as successful.

PR #1134, opened 2025-10-07
osc-operator: remove controller-gen pre-fetch workaround

littlejawa:
"go tool" is a feature introduced with go v1.24 [0], that lets developpers declare a tool that is needed for their project, and manage its dependency along with any other.

This removes the need to make a blank import from our main.go file. We can also remove the script that calls "go run" for downstream builds, as now we can use "go tool" even within the Makefile.

[0] https://go.dev/doc/modules/managing-dependencies#tools

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ COPY controllers controllers/
 COPY controller-gen bin/
 
 # get the version of controller-gen in an env variable for reusing
-RUN echo "export CONTROLLER_TOOLS_VERSION=$(grep controller-tools go.mod | awk '{print $2}')" > controller-tools-ver
+RUN echo "export CONTROLLER_TOOLS_VERSION=$(grep -m 1 controller-tools go.mod | awk '{print $2}')" > controller-tools-ver
 
 # rename the script to use the same version as defined in our go.mod file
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -223,7 +223,7 @@ $(LOCALBIN):
 ## Tool Binaries
 KUBECTL ?= kubectl
 KUSTOMIZE ?= $(LOCALBIN)/kustomize
-CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
+CONTROLLER_GEN ?= go tool sigs.k8s.io/controller-tools/cmd/controller-gen
```
gkurz: What about _just_ calling that in https://github.com/openshift/sandboxed-containers-operator/blob/devel/controller-gen#L15 ? This would prevent the Makefile to diverge too much from Operator SDK.

littlejawa: Yeah. I was hoping we could clean things up, but I guess you're right. As long as we use operator-sdk, we need to keep things compatible :-(

gkurz (approved):
/lgtm

Thanks @littlejawa !

Makefile
gkurz: This still is a great improvement ! :smile:

PR #1174, opened 2025-10-14
Drop trailing space in manager.yaml

gkurz:
The trailing space at the end of the podvm payload image URI was there
forever. Since this line gets updated by mintmaker for every new podvm
payload image build, it can cause conflicts in someone's PR if they happen
to remove the trailing space on the way.

Clean this now in a dedicated PR.


gkurz:
Merging this trivial space change right away.

PR #1177, opened 2025-10-14
Delete image config map on KataConfig deletion

gkurz:
In the default scenario, the podvm image config map is created by the operator. If the a podvm base OCI image is available, it is set in `PODVM_IMAGE_URI` item during KataConfig creation.

The config map is left untouched on KataConfig deletion. If a new KataConfig is created afterwards, it will end up silently using the OCI image from the previous deployment.

Have the KataConfig to own the podvm image config map when it is created by the controller. This will ensure it is fully removed when KataConfig is deleted.

### DIFF
--- a/controllers/image_generator.go
+++ b/controllers/image_generator.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	kataconfigurationv1 "github.com/openshift/sandboxed-containers-operator/api/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -37,6 +38,7 @@ import (
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 /*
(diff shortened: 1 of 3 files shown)


gkurz:
Fixes https://issues.redhat.com/browse/KATA-4162

bpradipt (approved):
/lgtm
Thanks @gkurz

snir911 (approved):
Thanks @gkurz ! 
This LGTM overall, having said that, I'd consider marking the whole CM as generated and delete it completely when it's marked or even use OwnerReference (to kataConfig?) at creation and get it deleted automatically.

controllers/image_generator.go
```diff
@@ -905,6 +912,38 @@ func (r *ImageGenerator) updateImageConfigMap() error {
 
 }
 
+// Method to reset the default image in the image configmap
+// azure-podvm-image-cm.yaml for Azure
+// aws-podvm-image-cm.yaml for AWS
+// libvirt-podvm-image-cm.yaml for Libvirt
```
snir911: BTW, we have GCP now also :)

gkurz:
> Thanks @gkurz ! This LGTM overall, having said that, I'd consider marking the whole CM as generated and delete it completely when it's marked or even use OwnerReference (to kataConfig?) at creation and get it deleted automatically.

Great suggestion @snir911 ! I'll give a try at this approach right away.

/hold

controllers/image_generator.go
gkurz: Yeah... and I'm pretty sure I saw other things missing in this area. Tech debt for later :wink:

gkurz:
Now using `SetControllerReference` and it does the job. PTAL again @bpradipt and @snir911 .

controllers/image_generator.go
```diff
@@ -848,6 +851,11 @@ func (r *ImageGenerator) createImageConfigMapFromFile() error {
 		igLogger.Info("Setting IMAGE_VERSION", "image_version", imageVersion)
 	}
 
+	// To ensure that it gets deleted along with the KataConfig
+	if err := controllerutil.SetControllerReference(kataConfig, cm, scheme); err != nil {
```
snir911: I think you should be able to call r.client.Scheme() instead of passing it

gkurz: Oh yes you're right. Let me do that.

snir911 (approved):
LGTM, thanks !

gkurz:
Failures of the EC check in konflux are unrelated and will be ignored here.

PR #1192, opened 2025-10-16
Update operator sdk to 1.41.1

littlejawa:
Updating operator-sdk from 1.39.1 to 1.41.1.

Preliminary steps: adding some code that we are supposed to modify, but doesn't exist in our repo.
Also fixing some update steps that were missing in previous commits.

1.39.1 => 1.40: the number of changes is big, so I'm keeping each modification separated, based on the titles in https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.40.0/

1.40 => 1.41.1: doing it in a single commit as it is easier.


littlejawa:
Note: 69dcc908e9b874234cf846907055b0adbbb623f4 is breaking the build. 
When I run "make bundle" on it, I get an error in kustomize, which apparently can't apply the patch from `config/default/manager_webhook_patch.yaml`.
This is why I made 0aeea5a66f70a75adc6f2b18129576f62f95dc1e, restoring the content of this file and fixing the build - this is not a solution, it just unblocked me to finish the upgrade.

Everything else seems to work - I have run "go mod tidy && make && make bundle" on every separate commit, and the generated files are not changing anymore. This kustomize issue is the last problem I need to fix.

littlejawa:
Fixes: [KATA-3600](https://issues.redhat.com//browse/KATA-3600)

gkurz:
Bumping Operator SDK this late is too risky for the release. We can do this post-release and maybe release a OSC 1.11.1 if needed. For now, make it clear we shouldn't merge this.

/hold

PR #1209, opened 2025-10-21
fbc: create catalog for OCP 4.20 release (OSC 1.10)

dbkreling:
Create catalog template for OCP v4.20.

Created with help of Cursor, based on the documentation found in [add-a-new-openshift-version](https://github.com/openshift/sandboxed-containers-operator/tree/devel/fbc#add-a-new-openshift-version).

For further info, see [KATA-4269](https://issues.redhat.com/browse/KATA-4269)

### DIFF
--- a/.tekton/fbc-pipeline.yaml
+++ b/.tekton/fbc-pipeline.yaml
@@ -283,7 +283,7 @@ spec:
     - name: IMAGE_DIGEST
       value: $(tasks.build-image-index.results.IMAGE_DIGEST)
     - name: ADDITIONAL_TAGS
-      value: ["latest", "1.11.0-$(tasks.clone-repository.results.commit-timestamp)"]
+      value: ["latest", "1.10.3-$(tasks.clone-repository.results.commit-timestamp)"]
     runAfter:
     - build-image-index
     taskRef:
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -36,7 +36,7 @@ metadata:
     operatorframework.io/arch.amd64: supported
     operatorframework.io/arch.s390x: supported
     operatorframework.io/os.linux: supported
-  name: sandboxed-containers-operator.v1.10.3
+  name: sandboxed-containers-operator.v1.11.0
 spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -36,7 +36,7 @@ metadata:
     operatorframework.io/arch.amd64: supported
     operatorframework.io/arch.s390x: supported
     operatorframework.io/os.linux: supported
-  name: sandboxed-containers-operator.v1.11.0
+  name: sandboxed-containers-operator.v1.10.3
 spec:
   apiservicedefinitions: {}
   customresourcedefinitions:
--- a/config/samples/deploy.yaml
+++ b/config/samples/deploy.yaml
@@ -36,4 +36,4 @@ spec:
   name: sandboxed-containers-operator
   source: my-operator-catalog
   sourceNamespace: openshift-marketplace
-  startingCSV: sandboxed-containers-operator.v1.11.0  ## OSC_VERSION
+  startingCSV: sandboxed-containers-operator.v1.10.3  ## OSC_VERSION
--- a/config/samples/deploy.yaml
+++ b/config/samples/deploy.yaml
@@ -36,4 +36,4 @@ spec:
   name: sandboxed-containers-operator
   source: my-operator-catalog
   sourceNamespace: openshift-marketplace
-  startingCSV: sandboxed-containers-operator.v1.10.3  ## OSC_VERSION
+  startingCSV: sandboxed-containers-operator.v1.11.0  ## OSC_VERSION
--- a/fbc/v4.20/catalog/sandboxed-containers-operator/catalog.json
+++ b/fbc/v4.20/catalog/sandboxed-containers-operator/catalog.json
@@ -1,0 +1,2477 @@
+{
+    "schema": "olm.package",
+    "name": "sandboxed-containers-operator",
+    "defaultChannel": "stable",
+    "icon": {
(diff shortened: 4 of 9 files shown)


fbc/v4.20/catalog/sandboxed-containers-operator/catalog.json
```diff
@@ -0,0 +1,2481 @@
...
+        },
+        {
+            "name": "",
+            "image": "registry.redhat.io/openshift-sandboxed-containers/osc-rhel9-operator@sha256:b71ab5667a3afb8aba1567f479534176d5b044158d27b5a37a0f7d62da92976e"
+        }
+    ]
+}
+{
+    "schema": "olm.bundle",
+    "name": "sandboxed-containers-operator.v1.11.0",
```
gkurz: Is it correct to expose the v1.11 bundle at this point ? Asking because of the following failure in the konflux pipeline :

```
time="2025-10-22T08:48:48Z" level=fatal msg="failed to load or rebuild cache: failed to rebuild cache: build package index: process package \"sandboxed-containers-operator\": package \"sandboxed-containers-operator\", bundle \"sandboxed-containers-operator.v1.11.0\" not found in any channel entries"
```

dbkreling: You're correct, downgrading them to v1.10.3

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -36,7 +36,7 @@ metadata:
     operatorframework.io/arch.amd64: supported
     operatorframework.io/arch.s390x: supported
     operatorframework.io/os.linux: supported
-  name: sandboxed-containers-operator.v1.11.0
+  name: sandboxed-containers-operator.v1.10.3
```
gkurz: Hmm... downgrading the bundle version looks wrong. This seems to indicate that this PR should rather happen in the `osc-release-v1-10` branch since we currently don't have anything new to publish on `devel`.

dbkreling: +1. I will pause the work on this branch and apply the changes on `osc-release-v1-10` and continue there.

I will not close this one because we will eventually push the changes to devel.

Changing it to Draft for the moment.

gkurz: Maybe replying this PR in the 1.10 branch isn't the way to go. My actual concern is the downgrade of OSC in the `bundle` and `config` directories... why is this needed ?

.tekton/fbc-pipeline.yaml
```diff
@@ -283,7 +283,7 @@ spec:
     - name: IMAGE_DIGEST
       value: $(tasks.build-image-index.results.IMAGE_DIGEST)
     - name: ADDITIONAL_TAGS
-      value: ["latest", "1.11.0-$(tasks.clone-repository.results.commit-timestamp)"]
+      value: ["latest", "1.10.3-$(tasks.clone-repository.results.commit-timestamp)"]
```
gkurz, line 286: This change is fine because at this point we're not ready to push 1.11.0 to any official fbc catalog.

.tekton/osc-test-fbc-push.yaml
```diff
@@ -312,7 +312,7 @@ spec:
       - name: IMAGE_DIGEST
         value: $(tasks.build-image-index.results.IMAGE_DIGEST)
       - name: ADDITIONAL_TAGS
-        value: ["latest", "1.11.0-$(tasks.clone-repository.results.commit-timestamp)"]
+        value: ["latest", "1.10.3-$(tasks.clone-repository.results.commit-timestamp)"]
```
gkurz: I would say this change isn't needed as we don't need to create a test catalog.

fbc/test-fbc/catalog-template.yaml
```diff
@@ -7,9 +7,9 @@ entries:
     name: sandboxed-containers-operator
     schema: olm.package
   - entries:
-      - name: sandboxed-containers-operator.v1.11.0
-        replaces: sandboxed-containers-operator.v1.10.3
-        skipRange: '>=1.1.0 <1.11.0'
+      - name: sandboxed-containers-operator.v1.10.3
```
gkurz: Test catalogs are for the current version in `devel` , which is 1.11.0. We should just ignore them in this PR.

dbkreling:
The enterprise contract for 4.20 is failing because of fips-check, and the enterprise contract for the OSC are failing because of the recently added TDX images (failing because of unauthorized CentOS repo).
I believe we should be good to merge.

FYI: @littlejawa @spotlesstofu @gkurz ?

.tekton/fbc-pipeline.yaml
spotlesstofu: This tag is for internal use only, mostly to aid manual testing. There is no need to downgrade it.

spotlesstofu (approved):

go.mod
```diff
@@ -150,7 +150,7 @@ require (
 	github.com/opencontainers/image-spec v1.1.1 // indirect
 	github.com/opencontainers/runtime-spec v1.2.1 // indirect
 	github.com/openshift/client-go v0.0.0-20251015124057-db0dee36e235 // indirect
-	github.com/openshift/library-go v0.0.0-20251016115829-235f9f1a273e // indirect
+	github.com/openshift/library-go v0.0.0-20251021141706-f489e811f030 // indirect
```
littlejawa, line 153: My expectation was that the PR contains only the new FBC files. Why do you need this change in this PR?

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
littlejawa: Resolving the conversation: This change is removed from the PR.

.tekton/fbc-pipeline.yaml
littlejawa: Right, this is not exposed to customers as far as I can tell. On the other hand, this downgrade doesn't hurt, and it keep things consistent. We can revert that when we do 1.11 for real.

littlejawa:
Looks ok to me. Comparing with other FBCs shows no difference - this is shipping the same images as all other catalogs, just for 4.20.
Just one question about the go.mod change, otherwise LGTM

gkurz:
> Looks ok to me. Comparing with other FBCs shows no difference - this is shipping the same images as all other catalogs, just for 4.20. Just one question about the go.mod change, otherwise LGTM

@littlejawa are the FIPS checks expected to fail in EC ?

littlejawa:
> > Looks ok to me. Comparing with other FBCs shows no difference - this is shipping the same images as all other catalogs, just for 4.20. Just one question about the go.mod change, otherwise LGTM
> 
> @littlejawa are the FIPS checks expected to fail in EC ?

Unfortunately, yes.
The EC for FBCs is checking that there was no errors in FIPS during build, and we do have FIPS errors on dm-verity, because FIPS check looks for openssl-libs in the image - and dm-verity is just a scratch image embedding the VM qcow file.
We have an exception to ignore that error on "prod" builds, but not on "stage". And we are in the process of duplicating that exception for "stage" so that we get rid of this error (but it is taking time).

.tekton/fbc-pipeline.yaml
gkurz: > On the other hand, this downgrade doesn't hurt, and it keep things consistent. We can revert that when we do 1.11 for real.

That's exactly what I meant @littlejawa ! :grin:

go.mod
gkurz: I agree that I don't quite expect this PR to change `go.mod` at all...

dbkreling: Rebasing my branch on devel got rid of the unwanted commit. Force-pusing it at the expense of losing @spotlesstofu's approval.

littlejawa (approved):
/lgtm

littlejawa:
EC is failing because of the FIPS check - as expected.
All other catalogs are passing because the FIPS check was skipped for them, as there is actually no change in the released bundle.

config/samples/deploy.yaml
```diff
@@ -36,4 +36,4 @@ spec:
   name: sandboxed-containers-operator
   source: my-operator-catalog
   sourceNamespace: openshift-marketplace
-  startingCSV: sandboxed-containers-operator.v1.10.3  ## OSC_VERSION
+  startingCSV: sandboxed-containers-operator.v1.11.0  ## OSC_VERSION
```
dbkreling, line 39: /retest branch:create_fbc_v4.20

dbkreling (commented):
/retest branch:create_fbc_v4.20

PR #1236, opened 2025-10-27
Update Dockerfile to support multiple architectures

gcoon151:
The umoci was hard coded to amd64. Diversifying the dockerfile. Disclaimer: I haven't tested this yet. Just wanted to put a proposal out here.

s390 team noticed daemonset install container had resources that wouldn't run on s390. 

**- What I did**
Added a TARGETARCH to specify platform.

**- How to verify it**
build container and run on different platforms. 

expand daemonset installer container to support s390x, ppc64le, arm64 (beyond the original amd64)

### DIFF
--- a/scripts/kata-install/Dockerfile
+++ b/scripts/kata-install/Dockerfile
@@ -1,15 +1,26 @@
 FROM registry.access.redhat.com/ubi9/skopeo:9.6-1760517870
 
-RUN mkdir -p /files
+ARG UMO_VERSION=0.4.7
+ARG TARGETARCH
+
+RUN mkdir -p /files /scripts
 
 ADD 50-kata-remote configuration-remote.toml /files/
-
-RUN mkdir -p /scripts
-
 ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh /scripts/
 
-RUN curl -sSL "https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64" -o "/usr/local/bin/umoci" &&\
-    chmod +x "/usr/local/bin/umoci"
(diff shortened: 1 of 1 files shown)


scripts/kata-install/Dockerfile
```diff
@@ -1,15 +1,26 @@
...
+# fetch the right umoci binary for the target arch
+RUN set -eux; \
+    case "${TARGETARCH}" in \
+      amd64)    UMO_SUFFIX=amd64 ;; \
+      arm64)    UMO_SUFFIX=arm64 ;; \
+      ppc64le)  UMO_SUFFIX=ppc64le ;; \
+      s390x)    UMO_SUFFIX=s390x ;; \
+      *) echo "Unsupported TARGETARCH: ${TARGETARCH}"; exit 1 ;; \
+    esac; \
+    curl -fsSL "https://github.com/opencontainers/umoci/releases/download/v${UMO_VERSION}/umoci.${UMO_SUFFIX}" \
```
gkurz, line 20: This might be okay for now but we'll have to get `umoci` from an official image in the RH repository when we go GA.

gcoon151: My searches turned up empty for an official umoci from RH. Happy to update with wherever we want to pull that from.

littlejawa: I had a look, there is no packaging.
I think there are two ways out of it:
- pre-fetch the binaries themselves, from the github repo (which means declaring each binary with path and checksum, and let the CI job pull them before the build starts)
- build umoci as part of the container image build

I didn't dig to search for alternatives, those are the most obvious. We can decide what option we take when we get to it.

gkurz:
Superseded by #1416 .

PR #1238, opened 2025-10-28
Fix looping caused by LastTransitionTime updates

Pacho20:
The reconciliation loop was being triggered unnecessarily due to two main issues:
1. The kataConfig.Status.KataNodes list was not sorted, causing changes in item order between reconciliations.
1. The LastTransitionTime field was always updated, even when the rest of the status remained unchanged.

These issues led to frequent status updates and redundant reconciliations.

**- What I did**
Added sorting to the kataConfig.Status.KataNodes list to ensure consistent ordering.
Introduced the IsKataConfigStatusChanged method to compare two KataConfigStatus objects while ignoring LastTransitionTime.
Updated the reconciliation logic to only update the KataConfig if meaningful changes are detected (excluding LastTransitionTime).

**- How to verify it**
Deploy the operator and observe the reconciliation behavior.
Observe reconciliation loops are no longer triggered by unchanged status or reordered lists.

Fix reconciliation loop caused by unsorted KataNodes and unnecessary LastTransitionTime updates.

### DIFF
--- a/controllers/daemonset_reconcile.go
+++ b/controllers/daemonset_reconcile.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"sort"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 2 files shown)


gkurz (approved):
Great thanks for the quick analysis and fix @Pacho20 !

/lgtm

Fixes: https://issues.redhat.com/browse/KATA-4287

PR #1294, opened 2025-11-06
Set INITDATA for non-confidential peer pods

snir911:
Follow up to this change:
https://github.com/openshift/cloud-api-adaptor/pull/218

### DIFF
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -73,13 +73,13 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 
 				// Patch ImageConfigMap.
 				imageConfigMapData := map[string]string{"CONFIDENTIAL_COMPUTE_ENABLED": "yes"}
-				if err := updateConfigMap(r.Client, r.Log, ig.getImageConfigMapName(), OperatorNamespace, imageConfigMapData); err != nil {
+				if err := updateConfigMap(r.Client, r.Log, ig.getImageConfigMapName(), OperatorNamespace, imageConfigMapData, nil); err != nil {
 					return err
 				}
 			} else {
 				// Patch ImageConfigMap.
 				imageConfigMapData := map[string]string{"CONFIDENTIAL_COMPUTE_ENABLED": "no"}
-				if err := updateConfigMap(r.Client, r.Log, ig.getImageConfigMapName(), OperatorNamespace, imageConfigMapData); err != nil {
+				if err := updateConfigMap(r.Client, r.Log, ig.getImageConfigMapName(), OperatorNamespace, imageConfigMapData, nil); err != nil {
 					if k8serrors.IsNotFound(err) {
 						// Nothing to do, feature is disabled and configMap doesn't exist.
 					} else {
@@ -97,7 +97,7 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 	} else {
 		peerpodsCMData = map[string]string{"DISABLECVM": "true"}
 	}
-	if err := updateConfigMap(r.Client, r.Log, peerpodsCMName, OperatorNamespace, peerpodsCMData); err != nil {
+	if err := updateConfigMap(r.Client, r.Log, peerpodsCMName, OperatorNamespace, peerpodsCMData, nil); err != nil {
 		if k8serrors.IsNotFound(err) {
 			// When feature is Enabled: ConfigMap doesn't exist yet, will try again at the next reconcile run.
 			// Else: Nothing to do, feature is disabled and configMap doesn't exist.
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -28,6 +28,11 @@ const (
 	// Extended resources for TEE
 	intelTDXExtendedResource = "tdx.intel.com/keys"
 	amdSNPExtendedResource   = "sev-snp.amd.com/esids"
+
+	// INITDATA value for non-confidential peer pods, this is required in order
+	// to override the default restrictive CoCo agent policy
+	// created from sourced plaintxt: cat config/peerpods/default-non-cc-initdata.toml | gzip | base64 -w0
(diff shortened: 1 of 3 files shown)


controllers/confidential_handler.go
```diff
@@ -28,6 +28,10 @@ const (
 	// Extended resources for TEE
 	intelTDXExtendedResource = "tdx.intel.com/keys"
 	amdSNPExtendedResource   = "sev-snp.amd.com/esids"
+
+	// INITDATA value for non-confidential peer pods, this is required in order
+	// to override the default restrictive CoCo agent policy
+	defaultInitdata = "H4sIAAAAAAAAA42UTU/DMAyG7/0VVS+7TUNwQEg7jG3AJMaqdmNH5LVeG5HGJXHGyq8npRLiQNodcrGfOH79EZAFacFlFU7DyJRwfXsTBSfURpBqTZPx1XgSBUEQ1SRF1ow1FhQ5z2g0CmrI3qHA0B3Fbx0QBDkewUoOZ3k+S+IXFEV5IG0S/LBoOLybhqwt/sXST6h97rkkgynnQnkJqpsHIdHr1wiMc1IMQqHux1JQ+YHOPmjhjJqaAWp5xizWlKHxqn5EXiNrkfURm816eXKl9SKtdYFOl/SGeRaGV4pRH8Hl00clZNlPrLF6Iq6lLe4bp+3gLfdGSVfmebxzN3xMDNYMdyS2Uq4qN1w+IEHIU3Z98z6UYEWn4Zc6LGWQ+Co0CzqatASNa7KKvSVJ0CDmiZsFqhZ46sFsNZxDitz1003hVlRe0Y6LfzbNC4hCgRyYP6dV83BOLbXVkAlV9DBsLohE9UCgLTf7dsuN+PKK39X5JbvcYcu6xAo1yP42dvDvjvRj/UuyB8EDhd+73xb/H9r2S/0GHJexXZAFAAA="
```
bpradipt: May be a reference to the plaintext initdata in source tree will be good for some dev looking at in future ?

controllers/confidential_handler.go
```diff
@@ -94,8 +98,12 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 	var peerpodsCMData map[string]string
 	if state == Enabled {
 		peerpodsCMData = map[string]string{"DISABLECVM": "false"}
+		// TODO: make sure we clear INITDATA value if not default
```
gkurz: IIUC the discussion I had with @snir911 on slack, the issue is that if KataConfig is created first, the other branch of the `if` will add the unsecure `defaultInitdata` to the CM and we don't want that to creep into the podvm when the confidential FG is enabled later on, but at the same time we don't want to clear out an INITDATA provided by the user. Right ?

This is an ownership problem. If the controller is allowed to set the initdata, it means it owns it and it will always take precedence on whatever the user puts there. The user provided initdata should come from some other config and the controller would be able to reconcile appropriately :
- non coco peer pods : always put `defaultInitdata` since policies are out of scope for this use case
- coco peer pods : set the user provided initdata if present or clear it out

controllers/confidential_handler.go
gkurz: Since this for non-confidential peer pods only, what about naming the constant `permissiveInitData` or something like that ?

controllers/utils.go
```diff
@@ -224,6 +224,15 @@ func updateConfigMap(client client.Client, logger logr.Logger, cmName string, na
...
 
+	// Remove keys that match the values in keysToRemove map
+	for key, valueToMatch := range keysToRemove {
+		if currentValue, exists := configMap.Data[key]; exists && currentValue == valueToMatch {
+			logger.Info("updateConfigMap - removing key", "namespace", namespace, "name", cmName, "key", key)
+			delete(configMap.Data, key)
+			update = true
+		}
+	}
+
```
bpradipt, line 235: What's the purpose of removing a key from existing configmap ?

snir911: See the next commit, this is needed in order to clear the permissive INITDATA in case it exist when CoCo fg is being enabled.
AFAIU enabling/disabling fg after installation is not supported, however, it's better to have this just to make sure we don't get unwanted pp configuration for CoCo

bpradipt: Ok.

bpradipt (approved):

bpradipt:
@snir911 since this is still WIP, I was thinking if it makes sense to add the default registry cred to the initdata based on the cluster secret as part of this PR ?

Ref: https://github.com/confidential-containers/guest-components/blob/main/confidential-data-hub/example.config.toml#L121

controllers/confidential_handler.go
```diff
@@ -92,12 +97,19 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 
 	// Patch peer pods configMap, if it exists.
 	var peerpodsCMData map[string]string
+	var keysToRemove map[string]string
 	if state == Enabled {
 		peerpodsCMData = map[string]string{"DISABLECVM": "false"}
+		// Remove INITDATA if it matches the default value
+		keysToRemove = map[string]string{"INITDATA": defaultNonCCInitdata}
```
gkurz, line 104: This will silently go against the intent of the user if they want to set such a permissive INITDATA on purpose... is it something we want ?

As said earlier, this is an ownership problem. It doesn't look right to tell the user they can set INITDATA if OSC is likely to overwrite it. It looks like user input should go to a dedicated CM or at least to a separate CUSTOM_INITDATA entry in the peerpods CM.

Cc @pmores for insights.

snir911: > This will silently go against the intent of the user if they want to set such a permissive INITDATA on purpose... is it something we want ? 

It's very unlikely to collide as CoCo users will normally be required to set in the initdata not only the agent policy but also aa/cdh toml, having said that, i can add a unique comment in the file "This initdata was created by the osc-operator" or something like that.

>  It doesn't look right to tell the user they can set INITDATA if OSC is likely to overwrite it.

Bare in mind we set it only for the non-CoCo users while initdata is officially supported only in CoCo. (FWIW, the concept of dummy (empty) initdata for pp is used also upstream).

At this point i find this solution fairly decent, however, I agree it's an ownership problem and open to better option if someone have it.

gkurz: This ownership problem requires some redesign that we might not be able to perform this close to the release. Let's add a `This initdata was created by the osc-operator` marker inside the OSC generated one in the meantime then.

gkurz (approved):
/lgtm

Thanks @snir911 !

controllers/confidential_handler.go
pmores: In general I believe (as I said before) that every k8s resource should have a single owner.  It shouldn't be that different items of the same resource have different owners, let alone that that a single item has multiple owners, the reason being that it can get confusing and frustrating rather quickly.

In particular, I unfortunately don't have enough context to propose a specific solution here but I do believe that if we let users set something we should do our best to avoid overwriting it later.  I understand that the semantics of the current situation seemingly makes incidents unlikely.  I still feel there should be ways to do better - e.g. if I understand correctly the setting in question here only makes sense under certain circumstances.  This sounds like something to be enforced by a schema.

PR #1334, opened 2025-11-11
controllers: support addon artifact images

ANJANA-ARK:
Add support for installing custom kernel/initrd artifacts from container images in DaemonSet mode. Introduces osc-kata-addons-install.sh script and ConfigMap-based configuration (kata-addon-artifacts).

Reuses extract_container_image from lib.sh. Minimal changes to existing flows. Provider-specific config updates (configuration-se.toml, configuration-tdx.toml, etc.) based on ConfigMap provider field.

### DIFF
--- a/config/baremetal/kata-addon-artifacts.yaml
+++ b/config/baremetal/kata-addon-artifacts.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kata-addon-artifacts
+  namespace: openshift-sandboxed-containers-operator
+data:
+  addonImage: "quay.io/openshift_sandboxed_containers/kata-se-artifacts:v1.0"
+  
+  # Path within the container image where kernel is located
+  kernelPath: "/artifacts/kernel/vmlinuz"
--- a/controllers/daemonset_reconcile.go
+++ b/controllers/daemonset_reconcile.go
@@ -7,6 +7,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 
@@ -582,6 +583,35 @@ func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDaemo
 	volumeMounts := r.volumeMountsForRegistries()
 	volumes := r.volumesForRegistries()
 
+	kataInstallEnv := []corev1.EnvVar{
+		{
+			Name: "NODE_NAME",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "spec.nodeName",
+				},
+			},
+		},
+		{
+			Name:  "EXTENSION_IMAGE",
+			Value: extensionImageString,
+		},
+		{
+			Name:  "CLI_IMAGE",
+			Value: cliImageString,
+		},
+		{
+			Name:  "NODE_LABEL",
+			Value: kataInstallationDaemonSetLabel,
+		},
+	}
+
+	// Add addon environment variables if ConfigMap exists
+	addonEnvVars := r.getAddonEnvVars()
+	if addonEnvVars != nil {
+		kataInstallEnv = append(kataInstallEnv, addonEnvVars...)
+	}
+
 	return &appsv1.DaemonSet{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "apps/v1",
@@ -624,28 +654,7 @@ func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDaemo
 							Command:      []string{"/bin/bash", "/scripts/osc-kata-install.sh"},
 							Args:         []string{string(action)},
 							VolumeMounts: volumeMounts,
-							Env: []corev1.EnvVar{
-								{
-									Name: "NODE_NAME",
-									ValueFrom: &corev1.EnvVarSource{
-										FieldRef: &corev1.ObjectFieldSelector{
-											FieldPath: "spec.nodeName",
-										},
-									},
-								},
-								{
-									Name:  "EXTENSION_IMAGE",
-									Value: extensionImageString,
-								},
-								{
-									Name:  "CLI_IMAGE",
-									Value: cliImageString,
-								},
-								{
-									Name:  "NODE_LABEL",
-									Value: kataInstallationDaemonSetLabel,
-								},
-							},
+							Env:          kataInstallEnv, // MODIFIED: Use the env array with addons
 						},
 						{
 							Name:            "set-log-level",
@@ -670,6 +679,46 @@ func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDaemo
 			},
 		},
 	}, nil
+}
+
+// getAddonEnvVars retrieves addon configuration from ConfigMap and returns environment variables
+func (r *KataConfigOpenShiftReconciler) getAddonEnvVars() []corev1.EnvVar {
+    var cm corev1.ConfigMap
+    if err := r.Client.Get(context.Background(),
+        types.NamespacedName{Name: "kata-addon-artifacts", Namespace: OperatorNamespace},
+        &cm,
+    ); err != nil {
+        if !errors.IsNotFound(err) {
+            r.Log.Error(err, "Failed to get addon ConfigMap")
+        }
+        return nil
+    }
+
+    data := cm.Data
+
+    // --- Mandatory variables ---
+    mandatory := []struct {
+        key, env string
+    }{
+        {"addonImage", "ADDON_IMAGE"},
+        {"kernelPath", "ADDON_KERNEL_PATH"},
+    }
+
+    var envs []corev1.EnvVar
+    for _, v := range mandatory {
+        val := data[v.key]
+        if val == "" {
+			continue
+        }
+        envs = append(envs, corev1.EnvVar{Name: v.env, Value: val})
+    }
+
+    r.Log.Info("Addon artifacts configured",
+        "image", data["addonImage"],
+        "kernelPath", data["kernelPath"],
+    )
+
+    return envs
 }
 
 func (r *KataConfigOpenShiftReconciler) volumesForRegistries() []corev1.Volume {
--- a/scripts/kata-install/Dockerfile
+++ b/scripts/kata-install/Dockerfile
@@ -6,10 +6,9 @@ ADD 50-kata-remote configuration-remote.toml /files/
 
 RUN mkdir -p /scripts
 
-ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh /scripts/
+ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh osc-kata-addons-install.sh /scripts/
 
 RUN curl -sSL "https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64" -o "/usr/local/bin/umoci" &&\
     chmod +x "/usr/local/bin/umoci"
 
 CMD ["/scripts/osc-kata-install.sh"]
- 
--- a/scripts/kata-install/osc-kata-addons-install.sh
+++ b/scripts/kata-install/osc-kata-addons-install.sh
@@ -1,0 +1,137 @@
+#!/bin/bash
+#
+# osc-kata-addons-install.sh
+# Install addon artifacts (kernel) from container images
+# Reuses functions from lib.sh
+#
+
+set -e
+
+# Source the shared library functions
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${SCRIPT_DIR}/lib.sh"
+
+#######################################
+# Update the configuration file
+# Arguments:
+#   $1: kernel path (empty if not installed)
+# Returns:
+#   0 on success
+#######################################
+update_config() {
+    local kernel_path="$1"
+    
+    local config_file="/etc/kata-containers/kata-se/configuration.toml"
+    
+    if [ ! -f  chroot /host "$config_file" ]; then
+        echo "Warning: Configuration file not found: $config_file"
+        return 1
+    fi
+    
+    echo "Updating configuration: $config_file"
+    
+    # Backup config
+    chroot /host cp "$config_file" "${config_file}.backup-$(date +%s)"
+    
+    # Update kernel if provided
+	chroot /host sed -i "s|^\(kernel[[:space:]]*=[[:space:]]*\)\".*\"|\1\"$kernel_path\"|g" "$config_file"
+	echo "  Updated kernel: $kernel_path"
+
+    return 0
+}
+
+#######################################
+# Install addon artifacts
+# Reads from environment variables
+# Returns:
+#   0 on success
+#######################################
+install_addons() {
+    local addon_image="${ADDON_IMAGE:?}"
+    
+    echo "Installing addon artifacts from: $addon_image"
+    
+    local kernel_src="${ADDON_KERNEL_PATH:?}"
+
+    if [ -z "$kernel_src" ]; then
+        echo "ERROR: ADDON_KERNEL_PATH is mandatory but was not provided." >&2
+        exit 1
+    fi
+    
+    # Standard installation directory
+    local install_dir="/host/etc/kata-containers"
+    local temp_dir="/tmp/kata-addons-$$"
+    local auth_file="/tmp/regauth/auth.json"
+    
+    mkdir -p "$install_dir"
+    mkdir -p "$temp_dir"
+    
+    local kernel_installed=""
+    local kernel_path=""
+    
+    # Extract and install kernel
+	echo "Extracting kernel from: $kernel_src"
+	
+	# Reuse extract_container_image from lib.sh
+	if extract_container_image "$addon_image" "$kernel_src" "$temp_dir" "$auth_file"; then
+		local kernel_file=$(basename "$kernel_src")
+		if [ -f "$temp_dir/$kernel_file" ]; then
+			kernel_installed="$install_dir/$kernel_file"
+			kernel_path="/etc/kata-containers/$kernel_file"
+			cp "$temp_dir/$kernel_file" "$kernel_installed"
+			chmod 644 "$kernel_installed"
+			echo "Kernel installed: $kernel_installed"
+
+			# Print checksum of the installed kernel
+			echo "Kernel image $kernel_file checksum (sha256):"
+			sha256sum "$kernel_installed"
+		fi
+	fi
+
+    
+    # Cleanup
+    rm -rf "$temp_dir"
+    
+    update_config "$kernel_path"
+
+    echo "Addon installation completed"
+    return 0
+}
+
+#######################################
+# Uninstall addon artifacts
+# Returns:
+#   0 on success
+#######################################
+uninstall_addons() {
+    local install_dir="/etc/kata-containers"
+    echo "Uninstalling addon artifacts"
+    
+    local kernel_src="${ADDON_KERNEL_PATH:?}"
+    # Remove installed artifacts
+    chroot /host rm -f "$install_dir/$(basename "$kernel_src")"
+     
+    # Restore config backup
+    local config_file="/etc/kata-containers/kata-se/configuration.toml"
+    local backup=$(ls -t "${config_file}.backup-"* 2>/dev/null | head -1)
+    [ -n "$backup" ] && [ -f "$backup" ] && chroot /host cp "$backup" "$config_file"
+    
+    echo "Addon artifacts uninstalled"
+    return 0
+}
+
+# Main execution
+action=${1:-install}
+
+case "$action" in
+    install)
+        install_addons
+        ;;
+    uninstall)
+        uninstall_addons
+        ;;
+    *)
+        echo "Usage: $0 {install|uninstall}"
+        exit 1
+        ;;
+esac
--- a/scripts/kata-install/osc-kata-install.sh
+++ b/scripts/kata-install/osc-kata-install.sh
@@ -227,6 +227,9 @@ main() {
 
 		install
 
+		# Install addon artifacts, if ADDON_IMAGE is present
+		[ -n "${ADDON_IMAGE:-}" ] && /scripts/osc-kata-addons-install.sh install
+
 		sleep infinity
 		;;
 	uninstall)
@@ -235,6 +238,9 @@ main() {
 		#/osc-log-level.sh "$action"
 
 		#/osc-configs-script.sh "$action"
+
+		# Call addon uninstaller if configured
+		[ -n "${ADDON_IMAGE:-}" ] && /scripts/osc-kata-addons-install.sh uninstall
 
 		uninstall
 		;;
--- a/scripts/kata-install/osc-kata-install.sh
+++ b/scripts/kata-install/osc-kata-install.sh
@@ -8,7 +8,23 @@ source "$(dirname "$0")"/lib.sh
 
 set -xeuo pipefail
 
-PACKAGES="capstone daxctl-libs edk2-ovmf ipxe-roms-qemu kata-containers libfdt libpmem libpng librdmacm ndctl-libs pixman qemu-img qemu-kvm-common qemu-kvm-core seabios-bin seavgabios-bin virtiofsd"
+ARCH=$(chroot /host uname -m)
+
+case "$ARCH" in
+    x86_64)
+        PACKAGES="capstone daxctl-libs edk2-ovmf ipxe-roms-qemu kata-containers libfdt libpmem libpng librdmacm ndctl-libs pixman qemu-img qemu-kvm-common qemu-kvm-core seabios-bin seavgabios-bin virtiofsd"
+        ;;
+
+    s390x)
+        PACKAGES="capstone kata-containers libfdt libpng pixman qemu-img qemu-kvm-common qemu-kvm-core virtiofsd"
+        ;;
+
+    *)
+        echo "ERROR: unsupported architecture: $ARCH" >&2
+        exit 1
+        ;;
+esac
+
 
 # label the node with the passed state
 label_node() {
--- a/scripts/kata-install/osc-kata-install.sh
+++ b/scripts/kata-install/osc-kata-install.sh
@@ -112,39 +112,38 @@ install() {
 	# If nothing to install, exit early
 	if [[ ${#install_rpms[@]} -eq 0 ]]; then
 		set_status_installed
-		sleep infinity
+	else
+		# Set installation status to installing
+		set_status_installing
+
+		# Prepare working directory
+		mkdir -p /host/tmp/extensions/
+
+		# Copy only needed RPMs
+		for rpm_path in "${install_rpms[@]}"; do
+			cp "$rpm_path" /host/tmp/extensions/
+		done
+
+		# Build install command
+		install_cmd="rpm-ostree install"
+		for pkg in "${uninstall_list[@]}"; do
+			install_cmd+=" --uninstall=$pkg"
+		done
+		for rpm_path in "${install_rpms[@]}"; do
+			rpm_filename=$(basename "$rpm_path")
+			install_cmd+=" /tmp/extensions/$rpm_filename"
+		done
+
+		# Run install inside chroot
+		echo "Running in chroot: $install_cmd"
+		chroot /host bash -c "$install_cmd"
+
+		# Clean up temp dir
+		rm -rf /host/tmp/extensions/
+
+		# Wait again: rpm-ostree install stages changes, requiring a reboot
+		wait_for_reboot_clear
 	fi
-
-	# Set installation status to installing
-	set_status_installing
-
-	# Prepare working directory
-	mkdir -p /host/tmp/extensions/
-
-	# Copy only needed RPMs
-	for rpm_path in "${install_rpms[@]}"; do
-		cp "$rpm_path" /host/tmp/extensions/
-	done
-
-	# Build install command
-	install_cmd="rpm-ostree install"
-	for pkg in "${uninstall_list[@]}"; do
-		install_cmd+=" --uninstall=$pkg"
-	done
-	for rpm_path in "${install_rpms[@]}"; do
-		rpm_filename=$(basename "$rpm_path")
-		install_cmd+=" /tmp/extensions/$rpm_filename"
-	done
-
-	# Run install inside chroot
-	echo "Running in chroot: $install_cmd"
-	chroot /host bash -c "$install_cmd"
-
-	# Clean up temp dir
-	rm -rf /host/tmp/extensions/
-
-	# Wait again: rpm-ostree install stages changes, requiring a reboot
-	wait_for_reboot_clear
 }
 
 uninstall() {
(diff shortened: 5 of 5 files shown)


config/baremetal/kata-addon-artifacts.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kata-addon-artifacts
+  namespace: openshift-sandboxed-containers-operator
+data:
+  addonImage: "quay.io/<username>/kata-se-artifacts:v1.0"
```
beraldoleal: username?

config/baremetal/kata-addon-artifacts.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
```
beraldoleal, line 2: are we introducing this new cm or should we use kataconfig? @gkurz any comments here?

config/baremetal/kata-addon-artifacts.yaml
```diff
@@ -0,0 +1,20 @@
...
+  provider: "se"
+  
+  # Optional: Path within the container image where kernel is located
+  kernelPath: "/artifacts/kernel/vmlinuz"
+  
+  # Optional: Path within the container image where initrd is located  
+  initrdPath: "/artifacts/initrd/kata-containers-initrd.img"
+  
+  # Define the OSC version
+  version: "v1.11"
```
beraldoleal: Do you mind expanding a bit on the comments why this version is needed? In which situation we could have OSC version installed X and here using Y?

scripts/kata-install/osc-kata-install.sh
```diff
@@ -211,18 +210,38 @@ main() {
...
+		if [ -n "${ADDON_IMAGE:-}" ]; then
+  		  echo "Running osc-kata-addons-install.sh inside the host chroot..."
+   		  mkdir -p /host/tmp/scripts
+    		  cp /scripts/* /host/tmp/scripts/
+    		  chroot /host /bin/bash -c "/tmp/scripts/osc-kata-addons-install.sh install"
+		fi
+
+		sleep infinity
 		;;
+	upgrade)
```
Pacho20: Could you explain how this would be executed in the script? Currently, the controller cannot set upgrade as the action for the DaemonSet. Do you have plans to extend the controller to handle that?

controllers/daemonset_reconcile.go
```diff
@@ -672,6 +681,97 @@ func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDaemo
...
+	// Try to get the addon ConfigMap
+	configMap := &corev1.ConfigMap{}
+	err := r.Client.Get(ctx, types.NamespacedName{
+		Name:      "kata-addon-artifacts",
+		Namespace: OperatorNamespace,
+	}, configMap)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// ConfigMap not found - addon not configured
```
beraldoleal: Do plan to add a specific log msg here? If not, this if is unnecessary.

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,227 @@
...
+    return 0
+}
+
+#######################################
+# Upgrade addon artifacts
+# Returns:
+#   0 on success
+#######################################
+upgrade_addons() {
+    local version_file="/usr/share/kata-containers/.addon-version"
```
Pacho20: Both the installation and uninstallation use `"/etc/kata-containers/.addon-version"`. I don’t see `/usr` used anywhere else. I’m not very familiar with Kata - does it create this file? Otherwise, the upgrade would always call `install_addons`.

controllers/daemonset_reconcile.go
```diff
@@ -672,6 +681,97 @@ func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDaemo
...
+		// Extract version from image tag if not provided
+		version = extractVersionFromImage(addonImage)
+	}
+	envVars = append(envVars, corev1.EnvVar{
+		Name:  "ADDON_VERSION",
+		Value: version,
+	})
+
+	return envVars
+}
```
beraldoleal: I haven't tested my proposal, but maybe a less verbose method, with two structures for optional and mandatory?

```
	var cm corev1.ConfigMap
	if err := r.Client.Get(context.Background(),
		types.NamespacedName{Name: "kata-addon-artifacts", Namespace: OperatorNamespace},
		&cm,
	); err != nil {
		if !errors.IsNotFound(err) {
			r.Log.Error(err, "Failed to get addon ConfigMap")
		}
		return nil
	}

	data := cm.Data

	// --- Mandatory variables ---
	mandatory := []struct {
		key, env string
	}{
		{"addonImage", "ADDON_IMAGE"},
	}

	var envs []corev1.EnvVar
	for _, v := range mandatory {
		val := data[v.key]
		if val == "" {
			return nil // mandatory missing → addon not configured
		}
		envs = append(envs, corev1.EnvVar{Name: v.env, Value: val})
	}

	// --- Optional variables ---
	optional := []struct {
		key, env string
	}{
		{"provider", "ADDON_PROVIDER"},
		{"kernelPath", "ADDON_KERNEL_PATH"},
		{"initrdPath", "ADDON_INITRD_PATH"},
		{"version", "ADDON_VERSION"},
	}

	for _, v := range optional {
		val := data[v.key]
		if v.key == "version" && val == "" {
			val = extractVersionFromImage(data["addonImage"])
		}
		if val != "" {
			envs = append(envs, corev1.EnvVar{Name: v.env, Value: val})
		}
	}

	r.Log.Info("Addon artifacts configured", "image", data["addonImage"])
	return envs
```

config/baremetal/kata-addon-artifacts.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: ConfigMap
+metadata:
+  name: kata-addon-artifacts
+  namespace: openshift-sandboxed-containers-operator
+data:
+  addonImage: "quay.io/<username>/kata-se-artifacts:v1.0"
+  
+  # Required: Provider type (se, tdx, snp, etc.)
+  # This determines which config file to update: kata-<provider>/configuration.toml
+  provider: "se"
```
gkurz: AFAIK this is for IBM SE only and it is unlikely to change anytime soon. Let's keep it simple and drop that.

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,227 @@
...
+        else
+            # Add after commented image line or kernel line
+            sed -i "/^# image[[:space:]]*=/a initrd = \"$initrd_path\"" "$config_file"
+            if ! grep -q "^initrd[[:space:]]*=" "$config_file"; then
+                sed -i "/^kernel[[:space:]]*=/a initrd = \"$initrd_path\"" "$config_file"
+            fi
+        fi
+        echo "  Updated initrd: $initrd_path"
+    fi
+
```
gkurz, line 39: It seems that all this logic could be replaced with a simple config drop in. Something like :

```
cat<<EOF>/etc/kata-containers/kata-se/config.d/coco-guest.toml
[hypervisor.qemu]
kernel = "${kernel_path}"
initrd = "${initrd_path}"
EOF
```

config/baremetal/kata-addon-artifacts.yaml
gkurz: Good question. CM is definitely more flexible and easier to put in place but offers no validation, nor versioning. So I would say it depends on how far you look. If the main goal is to release fast at the expense of stability and maintainability, then go CM. If the goal is to have rock solid data structures that provide consistency and stability for the long term, then go CR.

At this point, it seems that the project is clearly driven by short term so I'd stick to CM.

beraldoleal: Hi @gkurz , thanks for chiming in..

I get were you came from... but I’m still unsure why this needs a separate ConfigMap instead of extending KataConfig.

We can release fast with a ConfigMap, but that creates a second configuration surface with no validation, no upgrade guarantees, and no clear lifecycle. KataConfig already defines the operator’s configuration API, and adding new fields there isn´t additive? iiuc, it doesn’t require a new CRD version, right?

One more point: is this DP, TP, or GA code? If it’s not GA, I’m fine keeping it as a ConfigMap for now (Running the risk of us keeping this forever: ( ). But if it’s GA, we should think carefully before asking users to edit config in two different places and losing the CRD benefits. This tends to come back to us sooner or later. We can still move fast if we do things in the right place.

gkurz: > iiuc, it doesn’t require a new CRD version, right?

Not familiar enough with CRDs to know when versioning is mandatory.

> One more point: is this DP, TP, or GA code? If it’s not GA, I’m fine keeping it as a ConfigMap for now (Running the risk of us keeping this forever: ( ). 

This is for bm coco on IBM-z, thus TP.

config/baremetal/kata-addon-artifacts.yaml
ANJANA-ARK: I've changed it to quay.io/openshift_sandboxed_containers/kata-se-artifacts:v1.0 as sample image

scripts/kata-install/osc-kata-addons-install.sh
ANJANA-ARK: As we remove the upgrade option, I've removed this as well

scripts/kata-install/osc-kata-addons-install.sh
ANJANA-ARK: As per my understanding, https://github.com/kata-containers/kata-containers/blob/main/src/runtime/config/configuration-qemu-se.toml.in this will be present in /etc/kata-containers/kata-se as I'm changed just the kernel/initrd in that file

scripts/kata-install/osc-kata-install.sh
ANJANA-ARK: Hi @Pacho20 
We planned to add an upgrade option so the controller could compare the running OSC version and perform a version upgrade. Since the DaemonSet controller currently does not support upgrade actions, for now I’ll remove the upgrade option to avoid confusion.
Do you happen to know when support for upgrade actions might be added to the controller?

config/baremetal/kata-addon-artifacts.yaml
ANJANA-ARK: We added versions so that we'll use it in upgrade function to upgrade OSC, but now as discussed above https://github.com/openshift/sandboxed-containers-operator/pull/1334#discussion_r2527488195 I've removed upgrade option, so I've removed this param

config/baremetal/kata-addon-artifacts.yaml
beraldoleal: Ok, thanks @gkurz . 

@ANJANA-ARK we can close this thread, but I would suggest you to consider keeping that in mind for for the GA.

scripts/kata-install/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
+FROM registry.access.redhat.com/ubi9/skopeo:9.6-1756959203
 
 RUN mkdir -p /files
 
 ADD 50-kata-remote configuration-remote.toml /files/
 
 RUN mkdir -p /scripts
 
-ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh /scripts/
+ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh osc-kata-install.sh osc-kata-addons-install.sh /scripts/
```
beraldoleal: osc-kata-install.sh is duplicated here.

beraldoleal (commented):
Hi @ANJANA-ARK , thanks for the changes, I closed all my previous threads, but I left one minor comment for you. Besides that and the others comments, lgtm.

config/baremetal/kata-addon-artifacts.yaml
ANJANA-ARK: sure

scripts/kata-install/Dockerfile
ANJANA-ARK: Thanks for pointing out I've changed it 👍

config/baremetal/kata-addon-artifacts.yaml
ANJANA-ARK: Hi @gkurz , I've removed this param

ANJANA-ARK:
Hi @gkurz @beraldoleal @Pacho20 I've addressed the review comments, kindly review. Thank  you!

config/baremetal/kata-addon-artifacts.yaml
ajaypvictor: Hi @gkurz @bpradipt @beraldoleal,
As per the initial design plan for this feature, idea was to use this for other providers as well maybe for dev-test, by hardcoding this variable, we are restricting this feature only for IBM SE, as this suggestion requires hardcoding of the SE specific config toml file on the code itself.
Could you please confirm if this approach is fine?
Thanks..

gkurz: Other TEEs are fully handled by the kata-containers RPM. This PR only exists because IBM SE artifacts cannot be shipped the same way. Hardcoding is the way to go.

ajaypvictor (approved):
LGTM, Thanks Anjana..

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,139 @@
...
+# Source the shared library functions
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${SCRIPT_DIR}/lib.sh"
+
+#######################################
+# Update provider-specific configuration file
+# Arguments:
+#   $1: kernel path (empty if not installed)
+#   $2: initrd path (empty if not installed)
```
gkurz: It looks like a leftover.

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,139 @@
...
+        return 1
+    fi
+    
+    echo "Updating configuration: $config_file"
+    
+    # Backup config
+    chroot /host cp "$config_file" "${config_file}.backup-$(date +%s)"
+    
+    # Update kernel if provided
+    if [ -n "$kernel_path" ]; then
```
gkurz: Shouldn't `kernel_path` be mandatory ?

beraldoleal (approved):
Except from @gkurz comments, lgtm. thanks.

scripts/kata-install/osc-kata-addons-install.sh
ANJANA-ARK: Thanks for pointing out, I've removed this

scripts/kata-install/osc-kata-addons-install.sh
ANJANA-ARK: Yes, I've changed this 👍

ANJANA-ARK:
Hi @gkurz @Pacho20 , I've addressed the review comments can you please review, thank you!

scripts/kata-install/osc-kata-install.sh
Pacho20: I don’t have any information about adding an upgrade action to the controller. Nevertheless, the installation action is designed to handle upgrades by checking whether the latest versions of the packages are installed. If they aren’t, it reinstalls them with the newer versions.

gkurz (commented):
This part of the commit log is outdated. Please drop it.

```
Provider-specific config updates (configuration-se.toml, configuration-tdx.toml, etc.) based on ConfigMap provider field.
```

scripts/kata-install/Dockerfile
```diff
@@ -1,15 +1,14 @@
-FROM registry.access.redhat.com/ubi9/skopeo:9.6-1760517870
+FROM registry.access.redhat.com/ubi9/skopeo:9.6-1756959203
```
gkurz: This is wrong on several aspects :
- you should never have to downgrade a base image like that as it can degrade the health index of the resulting image
- unless you really need a new base image in your PR, avoid adding an unrelated change like this

According to https://catalog.redhat.com/en/software/containers/ubi9/skopeo/61959869ce09ff98ccc36338?image=68b91a5b1c978d120a3f0f1d and https://catalog.redhat.com/en/software/containers/ubi9/skopeo/61959869ce09ff98ccc36338?image=68b91a5b1c978d120a3f0f1d, both images have a B grade... we should upgrade to something newer, but this will happen in a separate PR.

In the meatime, please drop this change.

controllers/daemonset_reconcile.go
```diff
@@ -672,6 +682,73 @@ func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDaemo
...
+			return nil // mandatory missing → addon not configured
+		}
+		envs = append(envs, corev1.EnvVar{Name: v.env, Value: val})
+	}
+
+	// --- Optional variables ---
+	optional := []struct {
+		key, env string
+	}{
+		{"kernelPath", "ADDON_KERNEL_PATH"},
```
gkurz: Why is the kernel path optional ? Isn't providing a kernel the justification for this PR ?

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,140 @@
...
+#
+
+set -e
+
+# Source the shared library functions
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${SCRIPT_DIR}/lib.sh"
+
+#######################################
+# Update provider-specific configuration file
```
gkurz: Please drop the `provider` wording everywhere. This is only for IBM SE.

scripts/kata-install/osc-kata-addons-install.sh
gkurz: You dropped the check in the caller but this still looks optional to me... I'll reword my question : shouldn't the script just fail if no kernel is provided ?

scripts/kata-install/osc-kata-addons-install.sh
gkurz: I was suggesting to leverage https://github.com/kata-containers/kata-containers/blob/main/src/runtime/README.md#drop-in-configuration-file-fragments instead of modifying the existing configuration file.  This can be improved later.

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,140 @@
...
+    local addon_image="${ADDON_IMAGE:-}"
+    
+    if [ -z "$addon_image" ]; then
+        echo "No addon image configured"
+        return 0
+    fi
+    
+    echo "Installing addon artifacts from: $addon_image"
+    
+    local kernel_src="${ADDON_KERNEL_PATH:-}"
```
gkurz: It looks like a good place to bail out if no kernel path is provided.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -109,42 +109,41 @@ install() {
 		fi
 	done
 
-	# If nothing to install, exit early
```
gkurz, line 112: why dropping the comment ?

scripts/kata-install/osc-kata-install.sh
```diff
@@ -109,42 +109,41 @@ install() {
...
 	if [[ ${#install_rpms[@]} -eq 0 ]]; then
-		set_status_installed
-		sleep infinity
```
gkurz, line 115: why dropping the `sleep infinity` ?

scripts/kata-install/osc-kata-install.sh
```diff
@@ -109,42 +109,41 @@ install() {
...
+	    	set_status_installed
```
gkurz: Broken indentation.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -109,42 +109,41 @@ install() {
...
+
+    	    	# Run install inside chroot
+    	    	echo "Running in chroot: $install_cmd"
+   	    	chroot /host bash -c "$install_cmd"
+
+    	   	# Clean up temp dir
+   	    	rm -rf /host/tmp/extensions/
+
+    		# Wait again: rpm-ostree install stages changes, requiring a reboot
+   		wait_for_reboot_clear
```
gkurz: Same here, if you really need to move the lines in the `else` branch, please fix indentation.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -166,7 +165,23 @@ uninstall() {
...
 	# Uninstall extensions from the node
-	chroot /host /bin/bash -c "rpm-ostree uninstall $PACKAGES"
+
+	LAYERED=$(chroot /host rpm-ostree status --json \
+  | sed -n '/"requested-local-packages"/,/]/p' \
+  | tr -d '\n' \
+  | sed 's/.*\[\(.*\)\].*/\1/' \
+  | tr ',' '\n' \
+  | sed 's/"//g; s/^[ \t]*//; s/[ \t]*$//;/^$/d'
+)
```
gkurz: A comment to explain the intent behind this `sed` black magic would be greatly appreciated.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -166,7 +165,23 @@ uninstall() {
...
+
+	for pkg in $PACKAGES; do
+    		if echo "$LAYERED" | grep -q "$pkg"; then
+        		echo "Uninstalling $pkg..."
+        		chroot /host rpm-ostree uninstall "$pkg"
+   	 	else
+        		echo "Skipping $pkg (not layered)"
+   		 fi
+	done
```
gkurz: What's the justification for this change ? Please add comments here and in the commit changelog.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -211,18 +226,25 @@ main() {
 		#/osc-configs-script.sh "$action"
 
 		install
+
+		# Install addon artifacts, if ADDON_IMAGE is present
+		[ -n "${ADDON_IMAGE:-}" ] && /scripts/osc-kata-addons-install.sh install
+
+		sleep infinity
```
gkurz, line 233: Is this the `sleep infinity` that got removed in the other change ?

scripts/kata-install/osc-kata-install.sh
```diff
@@ -211,18 +226,25 @@ main() {
...
 
 		#/osc-configs-script.sh "$action"
+    		# Call addon uninstaller if configured
+		[ -n "${ADDON_IMAGE:-}" ] && /scripts/osc-kata-addons-install.sh uninstall
 
 		uninstall
 		;;
 	*)
-		echo "Usage: $0 {install|uninstall}"
+        echo "Usage: $0 {install|uninstall}"
```
gkurz: This change doesn't seem to be relevant for this PR.

scripts/kata-install/osc-kata-addons-install.sh
ANJANA-ARK: sure I'll make this changes in another PR

beraldoleal (commented):
While we don't have an explicit rule that we should have >1 approval, its a good practice to solve all reviewer comments (or find an agreement) before merging.

I'm removing my LGTM, just to avoid this getting merged by accident... We are aware of the importance of this PR but I still see a few open comments from @gkurz that needs to be addressed before this gets in. @ANJANA-ARK  could you please fix them so we can merge this ASAP?

scripts/kata-install/osc-kata-install.sh
ANJANA-ARK:
```
install

# Install addon artifacts, if ADDON_IMAGE is present
[ -n "${ADDON_IMAGE:-}" ] && /scripts/osc-kata-addons-install.sh install

sleep infinity
```

here I've added sleep infinity after the execution of osc-kata-addons-install.sh script, if  not when install is called it goes to sleep infinity and rest of the script is not getting executed

scripts/kata-install/osc-kata-install.sh
ANJANA-ARK: sure

scripts/kata-install/osc-kata-install.sh
ANJANA-ARK: sure

scripts/kata-install/osc-kata-install.sh
ANJANA-ARK: sure, I'll add a comment. This is to get the list of "layered" packages (requested-local-packages) from rpm-ostree. https://github.com/openshift/sandboxed-containers-operator/pull/1334#discussion_r2563650335

scripts/kata-install/osc-kata-install.sh
ANJANA-ARK: sure I'll add in commit changelog, 
in the script we have `PACKAGES="capstone daxctl-libs edk2-ovmf ipxe-roms-qemu kata-containers libfdt libpmem libpng librdmacm ndctl-libs pixman qemu-img qemu-kvm-common qemu-kvm-core seabios-bin seavgabios-bin virtiofsd"` 
here  few are not applicable for s390x, when the package is not present uninstall fails stating the package is not present. So using the layered packages sed command we're taking the packages which are present core OS and if any package which is given in `PACKAGES ` is not present here we're just skipping those, an alternative to sed command is using jq but we need to install jq on the RHCOS since we didn't want to install a new package from the script I've used sed command.

scripts/kata-install/osc-kata-install.sh
ANJANA-ARK: yes

scripts/kata-install/osc-kata-addons-install.sh
ANJANA-ARK: sure I'll add a check

scripts/kata-install/Dockerfile
ANJANA-ARK: changed this

controllers/daemonset_reconcile.go
ANJANA-ARK: sure I've changed this

scripts/kata-install/osc-kata-addons-install.sh
ANJANA-ARK: sure

scripts/kata-install/osc-kata-addons-install.sh
gkurz: Since this is checked in the caller, the `if` isn't needed anymore.

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,140 @@
...
+}
+
+#######################################
+# Install addon artifacts
+# Reads from environment variables
+# Returns:
+#   0 on success
+#######################################
+install_addons() {
+    local addon_image="${ADDON_IMAGE:-}"
```
gkurz: Since this is mandatory, make it `${ADDON_IMAGE:?}`. The script will print `ADDON_IMAGE: parameter null or not set` and exit 1.

scripts/kata-install/osc-kata-addons-install.sh
gkurz: Same here.

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,140 @@
...
+    local auth_file="/tmp/regauth/auth.json"
+    
+    mkdir -p "$install_dir"
+    mkdir -p "$temp_dir"
+    
+    local kernel_installed=""
+    local kernel_path=""
+    
+    # Extract and install kernel
+    if [ -n "$kernel_src" ]; then
```
gkurz: With the check above, the `if` isn't needed anymore.

scripts/kata-install/osc-kata-addons-install.sh
```diff
@@ -0,0 +1,140 @@
...
+#######################################
+# Uninstall addon artifacts
+# Returns:
+#   0 on success
+#######################################
+uninstall_addons() {
+    local install_dir="/etc/kata-containers"
+    echo "Uninstalling addon artifacts"
+    
+    local kernel_src="${ADDON_KERNEL_PATH:-}"
```
gkurz: AFAICT this is mandatory as well, `"${ADDON_KERNEL_PATH:?}"` and you can drop `[ -n "$kernel_src" ]` below, that becomes redundant.

scripts/kata-install/osc-kata-install.sh
gkurz: This is typically a change that should be made in a preparatory patch :
1. a commit that moves `sleep infinity` to the caller at the right place, without changing behavior
2. current commit then only needs to focus on calling the new script

scripts/kata-install/Dockerfile
```diff
@@ -6,10 +6,9 @@ ADD 50-kata-remote configuration-remote.toml /files/
 
 RUN mkdir -p /scripts
 
-ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh /scripts/
+ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh osc-kata-addons-install.sh /scripts/
 
 RUN curl -sSL "https://github.com/opencontainers/umoci/releases/download/v0.4.7/umoci.amd64" -o "/usr/local/bin/umoci" &&\
```
beraldoleal, line 11: @ANJANA-ARK please, help me to understand something...

This is something your PR is not touching, but the amd64 got my attention here... 

umoci v0.4.7 only has amd64 binaries. Is this intended to work only on x86_64 clusters? Or is this work focused on s390x as well? Will this work?

scripts/kata-install/osc-kata-install.sh
ANJANA-ARK: Sure I've changed it to 2 commits

scripts/kata-install/Dockerfile
ANJANA-ARK: Hi Beraldo,
Thanks for bringing this up..
We had a discussion on this when we started this development item and the suggestion was to enable Konflux for this daemonset image builds for GA and in the interim to share a s390x image built based on the PR once it is merged. 
For dev-test we have manually built a diff image and tested it, once the PR is merged, will share the container image built on s390x.
https://issues.redhat.com/browse/KATA-4307
Thanks..

beraldoleal: Ok, just to make sure you are aware. Maybe you could consider adapting this Docker file to handle both archs.

gkurz: This is definitely needed for the release, in another PR of course.

scripts/kata-install/osc-kata-install.sh
gkurz: Please make the content of `PACKAGES` arch dependent instead of this logic that ignores uninstall errors.

scripts/kata-install/Dockerfile
ANJANA-ARK: Sure

scripts/kata-install/osc-kata-install.sh
```diff
@@ -164,6 +164,12 @@ uninstall() {
 	# Set installation status to uninstalling
 	set_status_uninstalling
 
+	# Override package list for s390x
+	ARCH=$(chroot /host uname -m)
+	if [ "$ARCH" = "s390x" ]; then
+		PACKAGES="capstone kata-containers libfdt libpng pixman qemu-img qemu-kvm-common qemu-kvm-core virtiofsd"
+	fi
+
```
gkurz: The package list for s390x should be defined at the same place as the one for x86_64, with a  `case...esac`, so that multi-arch support is explicit and unmissable. Non supported archs should cause the script to fail. Typically a change to do in a preliminary patch :wink:

gkurz: Rest LGTM. I won't ask for more changes when that's done.

ANJANA-ARK: Sure @gkurz I've made these changes 👍

gkurz (approved):
/lgtm

Thanks @ANJANA-ARK !

PR #1339, opened 2025-11-13
Add Intel components's related images to CSV

dbkreling:
These images are expected to be accessible in a disconnected setup so they need to be added to the CSV yaml.

Issue: https://issues.redhat.com/browse/KATA-4300

PR #1343, opened 2025-11-13
Add Intel's related images to manager.yaml

dbkreling:
This is the only place where we should change the RELATED_IMAGES. After we change them here, we should update the rest with "make bundle" and then commit the changes created in the CSV file.

In this commit I'm also update the date when it was generated (rounding up a few hours).

For further information, please see https://github.com/openshift/sandboxed-containers-operator/pull/1339 and commit 58a821a5d.

When this commit is merged, all images will be synced accordingly.


gkurz (approved):
/lgtm

Thanks @dbkreling !

PR #1344, opened 2025-11-13
Add OCP version check for CoCo bare metal support

beraldoleal:
Validate OpenShift version >= 4.20 before enabling CoCo bare metal. Set UnsupportedOCPVersion condition and skip reconciliation on older versions.

Fixes: rhjira#[KATA-3965](https://issues.redhat.com//browse/KATA-3965)

### DIFF
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -125,6 +125,22 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 // It manages kata-cc runtime classes with TEE-specific handlers (Intel TDX, AMD SNP or IBM SE).
 func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state FeatureGateState) error {
 	if state == Enabled {
+		if !r.kataConfig.Spec.EnablePeerPods {
+			isLess, version, err := r.isOCPVersionLessThan("4.20")
+			if err != nil {
+				// Return error to trigger reconcile retry (cluster version not available yet or API error)
+				return err
+			}
+			if isLess {
+				r.Log.Info("WARNING: OpenShift version does not support CoCo bare metal", "version", version, "minVersion", "4.20")
+				cond := r.retrieveInProgressConditionForChange()
+				cond.Status = corev1.ConditionFalse
+				cond.Reason = "UnsupportedOCPVersion"
+				cond.Message = fmt.Sprintf("OpenShift version %s does not support CoCo bare metal (minimum required: 4.20)", version)
+				return nil
+			}
+		}
+
 		r.Log.Info("Creating " + kataCCRuntimeClassName + " runtime class for confidential containers")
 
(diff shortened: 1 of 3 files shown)


controllers/confidential_handler.go
```diff
@@ -125,6 +125,22 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
...
+			if err != nil {
+				// Return error to trigger reconcile retry (cluster version not available yet or API error)
+				return err
+			}
+			if isLess {
+				r.Log.Info("WARNING: OpenShift version does not support CoCo bare metal", "version", version, "minVersion", "4.19")
+				cond := r.retrieveInProgressConditionForChange()
+				cond.Status = corev1.ConditionFalse
+				cond.Reason = "UnsupportedOCPVersion"
+				cond.Message = fmt.Sprintf("OpenShift version %s does not support CoCo bare metal (minimum required: 4.19)", version)
```
gkurz: Just a note that this scenario should be documented if not already.

controllers/openshift_controller.go
```diff
@@ -570,6 +570,44 @@ func (r *KataConfigOpenShiftReconciler) getExtensionName() (string, error) {
...
+		return false, currentVersion, fmt.Errorf("invalid minimum version format: %s", minVersion)
+	}
+
+	if currentMajor < minMajor {
+		return true, currentVersion, nil
+	}
+	if currentMajor == minMajor && currentMinor < minMinor {
+		return true, currentVersion, nil
+	}
+
```
gkurz: What about leveraging https://pkg.go.dev/github.com/hashicorp/go-version ?

controllers/confidential_handler.go
beraldoleal: I will reinforce this with the changelog on this PR.

controllers/openshift_controller.go
beraldoleal: What about Masterminds/semver/v3 ? It seems to be already an indirect dependence of us, but a more popular project for that goal. I updated the code with it, let me know if it was a bad move or not.

controllers/confidential_handler.go
```diff
@@ -125,6 +125,22 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 // It manages kata-cc runtime classes with TEE-specific handlers (Intel TDX, AMD SNP or IBM SE).
 func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state FeatureGateState) error {
 	if state == Enabled {
+		if !r.kataConfig.Spec.EnablePeerPods {
+			isLess, version, err := r.isOCPVersionLessThan("4.19")
```
gkurz: Baremetal CoCo was originally targetting OCP 4.19 indeed but it got re-targeted to OCP 4.20 over the summer. We just forgot to update epics and stories accordingly... sorry for that. Can you upgrade you PR ?

gkurz (approved):
/lgtm

Thanks @beraldoleal !

PR #1349, opened 2025-11-14
Daemonset install without reboot

Pacho20:
The DaemonSet installation mode requires manual node reboots, which complicates both installation and uninstallation. This can confuse users and results in a poor user experience. The introduced changes aim to eliminate the need for reboots and make the process more seamless, although the current solution is not fully complete.

**- What I did**
Removed the configuration script and DaemonSet, as they made the installation process unnecessarily complex. Added `rpm-ostree live-apply` to enable Kata on worker nodes without rebooting. Extended the controller to schedule the installation process so that nodes are updated one at a time.

Initially, I tried to use `systemctl reload` for CRI-O instead of a full restart. This would have been a better solution because it avoids interrupting both CRI-O and kubelet and does not rely on their state restoration (which can fail in some cases). While `reload` works and CRI-O reloads its configuration, it fails to locate the executable for Kata runtime, returning the error:  
`stat: no such file or directory` for `/usr/bin/containerd-shim-kata`.  
The binary exists and works, but CRI-O cannot find it. I investigated multiple possibilities: checking the file in CRI-O’s namespace using `nsenter`, verifying permissions, SELinux flags, mount options, kernel parameters—everything suggests CRI-O should be able to invoke the binary. I still have a few ideas to check the interaction between CRI-O and the kernel during this lookup. If you have any insight into why this happens or how to fix it, that would greatly simplify the installation process.

As a fallback, I verified that CRI-O can invoke the Kata runtime after a full restart, so that is the current approach. However, this is not ideal because restarting CRI-O also triggers a kubelet restart (at least on ROKS). Installation works reliably because `rpm-ostree install` takes time, giving kubelet a chance to recover. Uninstallation, however, fails: although the script waits for the node to be in a "Ready" state, the node never becomes "NotReady" during kubelet restart. This means uninstall runs on other nodes while kubelet is still recovering, and triggers kubelet restarts on those nodes too, leaving the cluster in a broken state where most pods enter `ImagePullBackOff` or `CrashLoopBackOff`. Recovery is possible by restarting pods in the right order, but this should not happen. I could not find a reliable way to detect when kubelet and CRI-O are fully restored, so for now I reintroduced manual reboot for uninstallation.

One reason I pursued this approach is that the community operator uses a similar method successfully. They use the kata-deploy script - see:  
https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/scripts/kata-deploy.sh#L764C10-L778.

Currently, I see three possible paths forward:
- Figure out how to make `systemctl reload` work for CRI-O without breaking Kata runtime.
- Find a reliable way to ensure kubelet and CRI-O are fully recover before proceeding.
- Drain nodes before installation and uninstallation. This approach is already implemented in other operators and would ensure stability, but I wanted to avoid it because it adds significant operational complexity and increases overall duration (draining and uncordoning nodes takes considerable time).

**- How to verify it**
Build the operator using the updated `scripts/kata-install/Dockerfile`. Apply the `KataConfig` CR and wait until all nodes reach the "installed" status.

Change DaemonSet mode to eliminate node reboots (installation only; uninstallation still requires reboot for now).

EDIT: this is expected to fix https://issues.redhat.com/browse/KATA-4233


c3d:
Hey @Pacho20, I may be mistaken, but I think that `live-apply` works by remounting `/usr` with a new overlayfs. So I see it as a possibility that an existing process might still see the old mount and old directory if it kept an old file descriptor for the old directory. A bit in the same way that a process can see a deleted file if it still has a file descriptor to it. Just an hypothesis, but you could patch crio to show the content of the filesystem to validate it.

That would explain why you need to kill and restart the process.

c3d:
The title of the first commit is a big long. And GitHub still does not have a way to comment directly on commit messages.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -183,6 +183,11 @@ main() {
 		exit 1
 	}
 
+	[[ -z "${NODE_LABEL:-}" ]] && {
```
c3d, line 555: What is the point of `:-` here?

scripts/kata-install/Dockerfile
```diff
@@ -6,7 +6,7 @@ ADD 50-kata-remote configuration-remote.toml /files/
 
 RUN mkdir -p /scripts
 
-ADD osc-kata-install.sh osc-configs-script.sh osc-log-level.sh lib.sh /scripts/
+ADD osc-kata-install.sh osc-log-level.sh lib.sh /scripts/
```
c3d, line 9: I don't really understand why merging two scripts into one makes things simpler. Isn't it clearer what each step does when they are separated? Aren't there cases where you would want to reconfigure without reinstalling?

scripts/kata-install/osc-kata-install.sh
```diff
@@ -143,11 +196,14 @@ install() {
 	# Clean up temp dir
 	rm -rf /host/tmp/extensions/
 
+	# Copy configs
+	copy_kata_remote_config_files
```
c3d, line 472: If the goal is to simplify the workflow, the same overall simplification could be achieved by simply invoking the config script, no?

controllers/daemonset_reconcile.go
```diff
@@ -141,6 +142,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequestDaemonSet(
 		for _, node := range nodes {
 			r.Log.Info("node must be rebooted", "node", node.Name)
 		}
+		//r.scheduleInstallation(UninstallKata)
```
c3d: Is that a TODO or a leftover? AFAICT Uninstall is implemented, so looks to me like that code should be uncommented and the code above it removed?

scripts/kata-install/osc-kata-install.sh
```diff
@@ -100,30 +100,117 @@ wait_for_reboot_clear() {
...
+		sleep 1
+	done
+}
+
+restart_crio() {
+	# Restart crio
+	exec_on_host "systemctl daemon-reload"
+	exec_on_host "systemctl restart crio"
+
+	wait_till_node_is_ready
```
c3d, line 324: Can we add a timeout on the various wait and error out if we exceed it?

If you want to use the `timeout` command, you will need to either export the wait functions or put them in some separate script. Or you can add your own custom timeout. But having no node reach the ready state seems like a condition we should be ready to deal with gracefully.

controllers/daemonset_reconcile.go
c3d: Ah, I think that the problem is that uninstall without reboot won't work. Add a comment here then

gcoon151:
> Hey @Pacho20, I may be mistaken, but I think that `live-apply` works by remounting `/usr` with a new overlayfs. So I see it as a possibility that an existing process might still see the old mount and old directory if it kept an old file descriptor for the old directory. A bit in the same way that a process can see a deleted file if it still has a file descriptor to it. Just an hypothesis, but you could patch crio to show the content of the filesystem to validate it.
> 
> That would explain why you need to kill and restart the process.

I offered node debug and lsof might answer this also

controllers/daemonset_reconcile.go
gcoon151: correct. uninstall ran into issues so it's just disabled. As the primary use case I want fixed with this is worker updates, uninstall I'm fine leaving as customer must manually reboot worker to finish uninstall. It was left as a may implement later.

scripts/kata-install/Dockerfile
Pacho20: I removed most of the content from that script, so I thought moving it into the other one wouldn’t be a problem. I don’t think there are any cases that require reconfiguration. The two functions I moved to the other script will be removed anyway, since these config files will be part of the RPM package. Nevertheless, I think you’re right - it makes more sense to keep them in separate scripts.

scripts/kata-install/osc-kata-install.sh
Pacho20: Yup, you're right about that.

scripts/kata-install/osc-kata-install.sh
Pacho20: The `-u` option is set, so if the variable is unset, it results in an error. Therefore, we don’t necessarily need `:-`, but using it allows us to set the variable to an empty string and thereby control the error message.

Pacho20:
> Hey @Pacho20, I may be mistaken, but I think that `live-apply` works by remounting `/usr` with a new overlayfs. So I see it as a possibility that an existing process might still see the old mount and old directory if it kept an old file descriptor for the old directory. A bit in the same way that a process can see a deleted file if it still has a file descriptor to it. Just an hypothesis, but you could patch crio to show the content of the filesystem to validate it.
> 
> That would explain why you need to kill and restart the process.

I had a similar hypothesis. I checked the file descriptors with `lsof` as @gcoon151 suggested, along with other checks. Everything seemed to indicate that CRI-O should see the new binary. So instead of using the script, I performed the reload manually, and it seems that after `live-apply` there needs to be some time for CRI-O to recognize the new mount option. If I wait a few seconds, that part of the reload works - I can see in the logs that the `kata-remote` runtime handler loads. Although it logs an error message:
```
:39:01.361971609Z" level=error msg="Getting /usr/bin/containerd-shim-kata-v2 OCI runtime features failed: io.containerd.kata.v2: shim namespace cannot be empty: exit status 1" file="config/config.go:1379"
```
I checked where [it comes from](https://github.com/cri-o/cri-o/blob/66ff87374160f6d79f769113be2fbbeadcc6e55d/pkg/config/config.go#L1524), and the config validator seems to ignore it and load the new runtime anyway. The relevant part of `crio status config`'s output:

```
[crio.runtime.runtimes.kata-remote]
  runtime_config_path = "/opt/kata/configuration-remote.toml"
  runtime_path = "/usr/bin/containerd-shim-kata-v2"
  runtime_type = "vm"
  runtime_root = "/run/vc"
  privileged_without_host_devices = true
  allowed_annotations = ["io.kubernetes.cri-o.Devices"]
  runtime_pull_image = true
  container_min_memory = "12MiB"
  no_sync_log = false
```

But even with successful reloading, when kubelet tries to invoke the `RunPodSandbox` function, CRI-O returns this error:

```
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: I1121 15:04:16.321546    2570 kuberuntime_sandbox.go:65] "Running pod with runtime handler" pod="openshift-sandboxed-containers-operator/helloworld-887478c6-hjgbc" runtimeHandler="kata-remote"
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: E1121 15:04:16.322549    2570 log.go:32] "RunPodSandbox from runtime service failed" err="rpc error: code = Unknown desc = failed to find runtime handler kata-remote from runtime list map[crun:0xc0004cbb00 runc:0xc0004cb680]"
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: E1121 15:04:16.322616    2570 kuberuntime_sandbox.go:72] "Failed to create sandbox for pod" err="rpc error: code = Unknown desc = failed to find runtime handler kata-remote from runtime list map[crun:0xc0004cbb00 runc:0xc0004cb680]" pod="openshift-sandboxed-containers-operator/helloworld-887478c6-hjgbc"
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: E1121 15:04:16.322637    2570 kuberuntime_manager.go:1237] "CreatePodSandbox for pod failed" err="rpc error: code = Unknown desc = failed to find runtime handler kata-remote from runtime list map[crun:0xc0004cbb00 runc:0xc0004cb680]" pod="openshift-sandboxed-containers-operator/helloworld-887478c6-hjgbc"
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: E1121 15:04:16.322680    2570 pod_workers.go:1301] "Error syncing pod, skipping" err="failed to \"CreatePodSandbox\" for \"helloworld-887478c6-hjgbc_openshift-sandboxed-containers-operator(f9d52ed0-f321-4400-848f-046e943b408a)\" with CreatePodSandboxError: \"Failed to create sandbox for pod \\\"helloworld-887478c6-hjgbc_openshift-sandboxed-containers-operator(f9d52ed0-f321-4400-848f-046e943b408a)\\\": rpc error: code = Unknown desc = failed to find runtime handler kata-remote from runtime list map[crun:0xc0004cbb00 runc:0xc0004cb680]\"" pod="openshift-sandboxed-containers-operator/helloworld-887478c6-hjgbc" podUID="f9d52ed0-f321-4400-848f-046e943b408a"
```

You can find the code related to the error message [here](https://github.com/cri-o/cri-o/blob/66ff87374160f6d79f769113be2fbbeadcc6e55d/internal/oci/oci.go#L119) and the [method using it](https://github.com/cri-o/cri-o/blob/66ff87374160f6d79f769113be2fbbeadcc6e55d/server/sandbox_run.go#L54), which is used in every `RunPodSandbox` implementation. So it seems that for some reason the `Server` does not have the same runtime list. I didn’t have time to figure out the exact reason for that, but for now, reload is still not an option.

The other thing I noticed (which is probably worse) while checking the kubelet logs is that the `openshift-sandboxed-containers-monitor` pods can’t start after the `live-apply`. The error is the following: `Error: container create failed: write to /proc/self/attr/keycreate: Invalid argument`. I couldn’t find much about it, but it seems to be an SELinux-related problem. It goes away after a node reboot, but that’s what we’re trying to avoid. I still need to understand the exact reason. I see a possibility where this could prevent the installation without reboots, but we’ll know more after further investigation.

If anyone has more insight, I’d gladly hear it.

c3d:
@Pacho20 In the current state, this clearly needs quite a bit more work. Would you mind flagging it as do-not-merge until the PR is in a more complete state?

scripts/kata-install/osc-kata-install.sh
c3d: I meant: You are adding a test to check if `NODE_LABEL` is not set, with an error message that essentially is a almost exactly what `bash` itself would give you.

I see no functional difference between
```
ERROR: NODE_LABEL env var must be set
```

and 

```
NODEL_LABEL: unbound variable
```

So that code segment seems pointless to me.

Alternative: make the error message better, and send it to stderr instead of stdout.

Pacho20:
@c3d Yeah I know it's not ready. @gkurz asked me to open the PR so we can figure this out together.
I don’t know how to flag the PR, but I can convert it to a draft instead if that’s fine.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -192,15 +282,21 @@ install_kata() {
...
 	# Clean up temp dir
 	rm -rf /host/tmp/extensions/
 
 	# Copy configs
 	copy_kata_remote_config_files
 
-	# Wait again: rpm-ostree install stages changes, requiring a reboot
-	wait_for_reboot_clear
+	# Install SELinux policy
+	semodule -i /usr/share/kata-containers/defaults/osc_monitor.cil
```
gkurz: Why ? This is supposed to be done by the RPM already...

Pacho20: Maybe it's different on RHEL and RHCOS but it does not load after the installation.
I got this error for the openshift-sandboxed-containers-monitor pods:
`Error: container create failed: write to /proc/self/attr/keycreate: Invalid argument`
And after this modification the error message was gone.

Pacho20:
Small Status Update

Currently, I am facing the following issues with this installation method:

## systemctl reload crio does not work as expected
systemctl reload crio does not behave as expected, so I need to use restart instead (which can lead to issues depending on the crio and kubelet configurations).
As I previously described, after the installation I tried to reload crio. The reload completes without errors, and the configuration appears to load correctly (verified using the crio status command). However, the grpc server still uses the old configuration that does not include the Kata runtime. These are the error messages from the crio grpc server:
```
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: I1121 15:04:16.321546    2570 kuberuntime_sandbox.go:65] "Running pod with runtime handler" pod="openshift-sandboxed-containers-operator/helloworld-887478c6-hjgbc" runtimeHandler="kata-remote"
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: E1121 15:04:16.322549    2570 log.go:32] "RunPodSandbox from runtime service failed" err="rpc error: code = Unknown desc = failed to find runtime handler kata-remote from runtime list map[crun:0xc0004cbb00 runc:0xc0004cb680]"
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: E1121 15:04:16.322616    2570 kuberuntime_sandbox.go:72] "Failed to create sandbox for pod" err="rpc error: code = Unknown desc = failed to find runtime handler kata-remote from runtime list map[crun:0xc0004cbb00 runc:0xc0004cb680]" pod="openshift-sandboxed-containers-operator/helloworld-887478c6-hjgbc"
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: E1121 15:04:16.322637    2570 kuberuntime_manager.go:1237] "CreatePodSandbox for pod failed" err="rpc error: code = Unknown desc = failed to find runtime handler kata-remote from runtime list map[crun:0xc0004cbb00 runc:0xc0004cb680]" pod="openshift-sandboxed-containers-operator/helloworld-887478c6-hjgbc"
Nov 21 15:04:16 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000203 kubenswrapper[2570]: E1121 15:04:16.322680    2570 pod_workers.go:1301] "Error syncing pod, skipping" err="failed to \"CreatePodSandbox\" for \"helloworld-887478c6-hjgbc_openshift-sandboxed-containers-operator(f9d52ed0-f321-4400-848f-046e943b408a)\" with CreatePodSandboxError: \"Failed to create sandbox for pod \\\"helloworld-887478c6-hjgbc_openshift-sandboxed-containers-operator(f9d52ed0-f321-4400-848f-046e943b408a)\\\": rpc error: code = Unknown desc = failed to find runtime handler kata-remote from runtime list map[crun:0xc0004cbb00 runc:0xc0004cb680]\"" pod="openshift-sandboxed-containers-operator/helloworld-887478c6-hjgbc" podUID="f9d52ed0-f321-4400-848f-046e943b408a"
```
More details about this issue can be found in my [previous comment](https://github.com/openshift/sandboxed-containers-operator/pull/1349#issuecomment-3564208840).

## rpm-ostree uninstall with apply-live does not work as expected

Using the following commands:
```
rpm-ostree uninstall $PACKAGES"
rpm-ostree apply-live --allow-replacement
```
results in a different filesystem then expected.
We have some local configuration files in /etc/crio/crio.conf.d in addition to the Kata configurations. Different operators place configs there, so these files are not owned by any packages. When I run rpm-ostree uninstall followed by a reboot, everything works as expected: the Kata configuration is removed while the other files remain. However, when I use apply-live instead of rebooting, the entire /etc/crio/crio.conf.d directory is removed.
rpm-ostree uses a different algorithm to figure out how the filesystem should look like after the modifications when apply-live is used. It should still respect the local changes in /etc but for some reason it does not. If a directory contains both locally created files and files owned by packages that will be removed, the whole directory is removed instead of just the package-owned files. [This file contains the functions](https://github.com/coreos/rpm-ostree/blob/654e6ee0ba5953665b2bc2606bd6ab89c3871454/rust/src/live.rs) responsible for the new filesystem tree and it has a [special function for /etc](https://github.com/coreos/rpm-ostree/blob/654e6ee0ba5953665b2bc2606bd6ab89c3871454/rust/src/live.rs#L189).

As a short term solution I created functions that find all effected files and directories and back them up then restore them after apply-live runs (pushed that as a separate commit so you can check it). I thought this would solve the issues with the uninstallation. But it does not solve all the issues. After the uninstall and apply-live if I restart kubelet or try to reboot the node, kubelet does not start. I always used a ROKS cluster for the whole process so I can't access the worker node without kubelet running. I don't know what exactly happens but I had another running worker node (after apply-live but did not restart anything).  systemctl status shows a degraded state with a failed service called `kata-osbuilder-generate.service`.
```
systemctl status kata-osbuilder-generate.service
× kata-osbuilder-generate.service
     Loaded: not-found (Reason: Unit kata-osbuilder-generate.service not found.)
     Active: failed (Result: exit-code) since Wed 2025-12-10 15:41:03 UTC; 1 week 1 day ago
   Main PID: 67561 (code=exited, status=1/FAILURE)
        CPU: 6.970s

Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 kata-osbuilder.sh[71311]: INFO: Install tmp.mount in ./etc/systemd/system
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 kata-osbuilder.sh[71322]: cp: cannot stat './usr/share/systemd/tmp.mount': No such file or directory
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 kata-osbuilder.sh[71311]: INFO: Create /tmp/kata-dracut-rootfs-s0lpjc/etc
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 kata-osbuilder.sh[71311]: INFO: Configure chrony file /tmp/kata-dracut-rootfs-s0lpjc/etc/chrony.conf
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 kata-osbuilder.sh[71311]: [OK] cp /usr/libexec/kata-containers/agent/usr/bin/kata-agent /tmp/kata-dracut-rootfs-s0lpjc/usr/bin/kata-agent
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 kata-osbuilder.sh[71311]: ERROR: /tmp/kata-dracut-rootfs-s0lpjc/usr/bin/kata-agent is not installed in /tmp/kata-dracut-rootfs-s0lpjc
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 systemd[1]: kata-osbuilder-generate.service: Main process exited, code=exited, status=1/FAILURE
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 systemd[1]: kata-osbuilder-generate.service: Failed with result 'exit-code'.
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 systemd[1]: Failed to start Generate Kata appliance image for host kernel.
Dec 10 15:41:03 kube-d4e4pr6w0o0iqm3niodg-fodoroscdev-default-00000326 systemd[1]: kata-osbuilder-generate.service: Consumed 6.970s CPU time.
```

I tried running systemctl daemon-reload, but that was not enough to resolve the issue.

So, uninstall with apply-live does not work properly and prevents kubelet from starting again. I have not yet identified the root cause.

scripts/kata-install/osc-kata-install.sh
gkurz: This seems to indicate that the kata-containers RPM scriplets did not run. This change is thus basically a band-aid over something that is clearly wrong here and that must be investigated.

gkurz:
> So, uninstall with apply-live does not work properly and prevents kubelet from starting again. I have not yet identified the root cause.

Let's take a step back. You're trying to re-implement concepts that are already implemented in the MCO. Did you consider looking into their code to see how they use `rpm-ostree` and any other action needed on the nodes ?

gkurz:
> > So, uninstall with apply-live does not work properly and prevents kubelet from starting again. I have not yet identified the root cause.
> 
> Let's take a step back. You're trying to re-implement concepts that are already implemented in the MCO. Did you consider looking into their code to see how they use `rpm-ostree` and any other action needed on the nodes ?

Especially, MCO doesn't use `live-apply`. It relies on ad-hoc code to avoid reboot in specific cases instead. There might be a reason since everyone is obviously interested in not rebooting nodes.

Let's go back to something more aligned with what the MCO does : 
- new deployment with `rpm-ostree install/uninstall`
- cordon+drain
- reboot

Pacho20:
> > > So, uninstall with apply-live does not work properly and prevents kubelet from starting again. I have not yet identified the root cause.
> > 
> > 
> > Let's take a step back. You're trying to re-implement concepts that are already implemented in the MCO. Did you consider looking into their code to see how they use `rpm-ostree` and any other action needed on the nodes ?
> 
> Especially, MCO doesn't use `live-apply`. It relies on ad-hoc code to avoid reboot in specific cases instead. There might be a reason since everyone is obviously interested in not rebooting nodes.
> 
> Let's go back to something more aligned with what the MCO does :
> 
> * new deployment with `rpm-ostree install/uninstall`
> * cordon+drain
> * reboot

Yeah, MCO uses a different approach. I hoped it would be easy to do the install/uninstall with live-apply, but clearly it’s not. I agree that install/uninstall, cordon and drain, then reboot would be the way forward.

gkurz:
> > > > So, uninstall with apply-live does not work properly and prevents kubelet from starting again. I have not yet identified the root cause.
> > > 
> > > 
> > > Let's take a step back. You're trying to re-implement concepts that are already implemented in the MCO. Did you consider looking into their code to see how they use `rpm-ostree` and any other action needed on the nodes ?
> > 
> > 
> > Especially, MCO doesn't use `live-apply`. It relies on ad-hoc code to avoid reboot in specific cases instead. There might be a reason since everyone is obviously interested in not rebooting nodes.
> > Let's go back to something more aligned with what the MCO does :
> > 
> > * new deployment with `rpm-ostree install/uninstall`
> > * cordon+drain
> > * reboot
> 
> Yeah, MCO uses a different approach. I hoped it would be easy to do the install/uninstall with live-apply, but clearly it’s not. I agree that install/uninstall, cordon and drain, then reboot would be the way forward.

I suggest we keep this PR for reference of a tentative to use `apply-live` and you create a new one for the other approach.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -192,15 +282,21 @@ install_kata() {
 	# Run install inside chroot
 	echo "Running in chroot: $install_cmd"
 	chroot /host bash -c "$install_cmd"
+	chroot /host bash -c "rpm-ostree apply-live --allow-replacement"
```
gkurz, line 285: After reading a bit more on `apply-live` :
- https://github.com/coreos/rpm-ostree/blob/main/docs/apply-live.md
- https://github.com/coreos/rpm-ostree/blob/main/man/rpm-ostree.xml#L811

This adds a transient `overlayfs` layer that doesn't survive reboots. If the worker gets rebooted for some arbitrary reason later on, is `rpm-ostree apply-live` run again ?

gkurz: `apply-live` also creates a deployment that persists reboots so this should be fine.

c3d:
Regarding the use of `rpmostree live-apply`, not running the scripts is intentional. Here is how GPT5.2-codex explains it:
> Is rpm-ostree live-apply supposed to run %post and other scripts? If not, why not? Can you find the best web references on the web discussing this topic?

No—`rpm-ostree live-apply/apply-live` is not intended to run `%post/%posttrans scriptlets`. Live apply works by computing a filesystem diff between OSTree commits and copying those changes into a transient `overlayfs` for `/usr`, plus special-case handling for `/etc` and `/var`. There’s no RPM transaction phase in this flow, so there is no scriptlet execution stage.

This design keeps live changes minimal, avoids running arbitrary scripts on a live system, and acknowledges that live changes are non-transactional and can “leak” (e.g., `/etc` changes are applied directly). The docs explicitly describe live apply as “copying diffs” and transient overlay state rather than a full compose/transaction with scriptlet execution.

> Architecture of apply-live ... overlayfs ... compute diff ... copy /usr ... update /etc ... update /var ...

By contrast, in a normal `rpm-ostree compose/deploy` path, `%post` and `%posttrans` scripts are run as part of building a new filesystem tree “from scratch.”

> ... regenerate filesystem "from scratch" and all RPM %post scripts rerun ...

### Best web references
* Architecture of `apply-live` — explains the diff/overlay workflow and why it doesn’t resemble a full RPM transaction with scriptlets.
https://coreos.github.io/rpm-ostree/apply-live/

* RPM packages, ostree commits — describes the normal compose path where `%post` and `%posttrans` do run.
https://coreos.github.io/rpm-ostree/architecture-core/

c3d:
Hi @Pacho20 there is another PR that addresses the same issue, #2312. I have done a comparison and testing of both approaches, which is documented in https://redhat.atlassian.net/browse/KATA-5292. At the moment, #2312 looks like the better candidate, but I would like your input.

PR #1367, opened 2025-11-18
Update operator, must-gather and podvm-builder ubi references

dbkreling:
Update ubi-minimal references from to **9.7-1763362218** and ubi to **9.7-1763340522** on the operator, must-gather and podvm-builder.

Also replace the "latest" tag on pccs.yaml.in by an epoch tag in order to improve stability, control and reprodicibility.

Issue: https://issues.redhat.com/browse/KATA-4311

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.24.6-1762373805 as builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.25.3-1764620329 as builder
 
 # Required by the ubi based go-toolset image
 USER root
@@ -29,7 +29,7 @@ RUN . ./controller-tools-ver && mv bin/controller-gen bin/controller-gen-$CONTRO
 RUN . ./controller-tools-ver && make build
 
 # Use OpenShift base image
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.6-1760515502
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1764578379
 WORKDIR /
 COPY --from=builder /workspace/bin/manager .
 COPY --from=builder /workspace/bin/metrics-server .
--- a/must-gather/Dockerfile
+++ b/must-gather/Dockerfile
@@ -1,6 +1,6 @@
 FROM registry.redhat.io/openshift4/ose-must-gather-rhel9:v4.19 as builder
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.6-1760515502
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1764578379
 
 # For gathering data from nodes
 # NOTE for hermetic build: any change to the packages installed here must
--- a/scripts/kata-install/Dockerfile
+++ b/scripts/kata-install/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/skopeo:9.6-1760517870
+FROM registry.access.redhat.com/ubi9/skopeo:9.7-1764607501
 
 RUN mkdir -p /files
 
--- a/config/peerpods/podvm/Dockerfile.podvm-builder
+++ b/config/peerpods/podvm/Dockerfile.podvm-builder
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi:9.6-1760340943
+FROM registry.access.redhat.com/ubi9/ubi:9.7-1764578509
 
 # azure-podvm-image-handler.sh script under /scripts/azure-podvm-image-handler.sh
 # aws-podvm-image-handler.sh script under /scripts/aws-podvm-image-handler.sh
--- a/config/peerpods/podvm/redhat.repo
+++ b/config/peerpods/podvm/redhat.repo
@@ -9,2830 +9,16 @@
 # a "yum repolist" to refresh available repos
 #
 
-[codeready-builder-for-rhel-9-$basearch-source-rpms]
-name = Red Hat CodeReady Linux Builder for RHEL 9 $basearch (Source RPMs)
-baseurl = https://cdn.redhat.com/content/dist/rhel9/$releasever/$basearch/codeready-builder/source/SRPMS
-enabled = 1
-gpgcheck = 1
-gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
-sslverify = 1
-sslcacert = /etc/rhsm/ca/redhat-uep.pem
-sslclientkey = /etc/pki/entitlement/4455887482394223345-key.pem
-sslclientcert = /etc/pki/entitlement/4455887482394223345.pem
-sslverifystatus = 1
-metadata_expire = 86400
-enabled_metadata = 0
-
-[rhelai-3.0-gaudi-for-rhel-9-$basearch-debug-rpms]
-name = Red Hat Enterprise Linux AI (3.0) for RHEL 9 $basearch - Gaudi (Debug RPMs)
-baseurl = https://cdn.redhat.com/content/dist/layered/rhel9/$basearch/rhelai-gaudi/3.0/debug
-enabled = 0
-gpgcheck = 1
-gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
-sslverify = 1
-sslcacert = /etc/rhsm/ca/redhat-uep.pem
-sslclientkey = /etc/pki/entitlement/4455887482394223345-key.pem
-sslclientcert = /etc/pki/entitlement/4455887482394223345.pem
-sslverifystatus = 1
-metadata_expire = 86400
-enabled_metadata = 0
-
-[rhel-9-for-$basearch-sap-netweaver-rhui-debug-rpms]
-name = Red Hat Enterprise Linux 9 for $basearch - SAP NetWeaver (Debug RPMs) from RHUI
-baseurl = https://cdn.redhat.com/content/dist/rhel9/rhui/$releasever/$basearch/sap/debug
-enabled = 0
-gpgcheck = 1
-gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
-sslverify = 1
-sslcacert = /etc/rhsm/ca/redhat-uep.pem
-sslclientkey = /etc/pki/entitlement/4455887482394223345-key.pem
-sslclientcert = /etc/pki/entitlement/4455887482394223345.pem
-sslverifystatus = 1
-metadata_expire = 86400
-enabled_metadata = 0
-
-[lvms-4.17-for-rhel-9-$basearch-source-rpms]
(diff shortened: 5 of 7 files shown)


must-gather/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM registry.redhat.io/openshift4/ose-must-gather-rhel9:v4.19 as builder
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.6-1760515502
+FROM registry.access.redhat.com/ubi10/ubi-minimal:10.1-1762189261
```
littlejawa: This tag is for registry.access.redhat.com/ubi10/ubi
ubi and ubi-minimal don't have the same tags. This is why the build is failing.

Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN . ./controller-tools-ver && mv bin/controller-gen bin/controller-gen-$CONTRO
 RUN . ./controller-tools-ver && make build
 
 # Use OpenShift base image
-FROM registry.access.redhat.com/ubi9/ubi-minimal:9.6-1760515502
+FROM registry.access.redhat.com/ubi10/ubi-minimal:10.1-1762189261
```
gkurz: Since this is bumping the operator as well, maybe update the patch title to ` ... (must-gather, operator)` ?

littlejawa: This tag is for registry.access.redhat.com/ubi10/ubi
ubi and ubi-minimal don't have the same tags. This is why the build is failing.

gkurz (dismissed):
With the suggested commit title fix *and* the right tag for ubi10 minimal :wink:

/lgtm

Thanks @dbkreling !

Dockerfile
gkurz: Yes, just realized that. Tag for ubi10 minimal should be `10.1-1763362715`

littlejawa:
You will need to add a RPM lockfile for the operator's build now that you install packages in it.
Also, do you really need the `redhat.repo` file for must-gather? It seems all the packages referenced in the lock file come from ubi repo... I think we don't need a subscription to build the must-gather actually.

dbkreling:
Last forced push removed the redhat.repo for must-gather and also added "rpms.in.yaml" and "rpms.lock.file".

wainersm:
Hi @dbkreling ! The ci/prow/* jobs are failing because they are still using the rhel-9 buildroot as you suspected.

The current base image is `base-rhel9` (see https://github.com/openshift/release/blob/master/ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml#L5) and we will need to bump to `base-rhel10`. I don't know if `base-rhel10` is available in 4.19, otherwise we will need to bump the line https://github.com/openshift/release/blob/master/ci-operator/config/openshift/sandboxed-containers-operator/openshift-sandboxed-containers-operator-devel.yaml#L3 too.

I think above is enough, but more things to consider:

* It will also run `binary_build_commands` (line 6) in a container with `rhel-9-golang-1.24-openshift-4.20` (line 11). Do we want rhel-10 too?
* Likewise `GOFLAGS="" make test` in line 39.
* In line 20, instructs to replace `registry.redhat.io/openshift-sandboxed-containers/osc-rhel9-operator@sha256:.*` with the just built operator image in the bundle image. We need to check if this should be changed too.

Dockerfile
gkurz: @dbkreling, what's the motivation to bump the operator to ubi10 ? Asking because the jump from ubi9 to ubi10 appears to generate quite some extra work on the prow side. If it isn't strictly required I guess we should do this post release and simply bump to current latest [9.7-1763362218](https://catalog.redhat.com/en/software/containers/ubi9-minimal/61832888c0d15aff4912fe0d) to fix the health index.

must-gather/Dockerfile
gkurz: Same question here. We certainly do want to move away from the B health index of [9.6-1760515502](https://catalog.redhat.com/en/software/containers/ubi9/ubi/615bcf606feffc5384e8452e?image=68ecb4d792068b299f0cf7ff)  but wouldn't it be safer to stick to ubi9 ?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/go-toolset:1.24.6-1762373805 as builder
+FROM registry.access.redhat.com/ubi9/go-toolset:1.25-1763633888 as builder
```
gkurz: Please use the `1.25.3-1763633888` tag, it is more informative.

Dockerfile
```diff
@@ -29,13 +29,12 @@ RUN . ./controller-tools-ver && mv bin/controller-gen bin/controller-gen-$CONTRO
 RUN . ./controller-tools-ver && make build
 
 # Use OpenShift base image
-FROM registry.access.redhat.com/ubi10/ubi-minimal:10.1-1763362715
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.7-1763362218
```
gkurz: The 10.1 bump is de facto not needed since we downgrade it in the same PR. For the sake of simplicity, can you please drop the change that does the 10.1 bump ?

gkurz:
> Downgrade non-critical references of ubi and ubi-minimal to 9.7
> Prow is not yet fully ready to handle ubi10 images and due to time constraints
> we should downgrade to ubi-9.7.
> 
> However, rhel-9 images should be upgraded to use the latest health index A of
> rhel-9.7 ones.
> 
> pccs.yaml.in 'latest' reference was changed due to future reproducibility.
> 
> Signed-off-by: Daniel Kreling <dkreling@redhat.com>
> Reviewed-by: Gregory Kurz <gkurz@redhat.com>
> 

We don't use `Reviewed-by:` tags in this repo. We use approval at the PR level in GH instead.
Also, this tag would only be added when I consider the commit to be ready, which isn't the case yet.
Please discard it.

Dockerfile
gkurz: I think this change should be dropped since we're downgrading it to 9.7 at the end.

must-gather/Dockerfile
gkurz: I think this change should be dropped since we're downgrading it to 9.7 at the end. This should simplify the changes to the RPM lock file.

Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN . ./controller-tools-ver && mv bin/controller-gen bin/controller-gen-$CONTRO
 RUN . ./controller-tools-ver && make build
 
 # Use OpenShift base image
-FROM registry.access.redhat.com/ubi10/ubi-minimal:10.1-1762189261
+FROM registry.access.redhat.com/ubi10/ubi-minimal:10.1-1763362715
```
gkurz: For the same reason, this commit can be dropped since we're sticking to 9.7.

config/peerpods/podvm/ubi.repo
```diff
@@ -1,62 +1,62 @@
-[ubi-9-for-$basearch-baseos-rpms]
+[ubi-9-baseos-rpms]
```
gkurz: I strongly suspect `for-$basearch-` is required for multi-arch support.

config/peerpods/podvm/ubi.repo
```diff
@@ -1,62 +1,62 @@
...
 enabled = 0
 gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
 gpgcheck = 1
 
-[ubi-9-for-$basearch-baseos-source-rpms]
+[ubi-9-baseos-source-rpms]
 name = Red Hat Universal Base Image 9 (Source RPMs) - BaseOS
 baseurl = https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/$basearch/baseos/source/SRPMS
-enabled = 1
+enabled = 0
```
gkurz: This got enabled by 154f467dd6e55. I think we should keep it.

config/peerpods/podvm/ubi.repo
```diff
@@ -1,62 +1,62 @@
...
 enabled = 0
 gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
 gpgcheck = 1
 
-[ubi-9-for-$basearch-appstream-source-rpms]
+[ubi-9-appstream-source-rpms]
 name = Red Hat Universal Base Image 9 (Source RPMs) - AppStream
 baseurl = https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/$basearch/appstream/source/SRPMS
-enabled = 1
+enabled = 0
```
gkurz: ditto

config/peerpods/podvm/ubi.repo
```diff
@@ -1,62 +1,62 @@
...
 enabled = 0
 gpgkey = file:///etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release
 gpgcheck = 1
 
-[codeready-builder-for-ubi-9-$basearch-source-rpms]
+[ubi-9-codeready-builder-source-rpms]
 name = Red Hat Universal Base Image 9 (Source RPMs) - CodeReady Builder
 baseurl = https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/$basearch/codeready-builder/source/SRPMS
-enabled = 1
+enabled = 0
```
gkurz: ditto

must-gather/rpms.lock.yaml
```diff
@@ -5,60 +5,36 @@ arches:
 - arch: s390x
   packages:
   - url: https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/s390x/baseos/os/Packages/r/rsync-3.2.5-3.el9.s390x.rpm
-    repoid: ubi-9-for-s390x-baseos-rpms
+    repoid: ubi-9-baseos-rpms
```
gkurz: It looks like this might break multi-arch support and all the `for-....` should be kept in this PR.

scripts/kata-install/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/skopeo:9.6-1760517870
+FROM registry.access.redhat.com/ubi9/skopeo:9.7-1763368371
```
gkurz: Wrong commit.

gkurz:
Postponing to OSC 1.11.z.

/hold

littlejawa:
I have updated the lock file for podvm-builder.
Running the lock file update for must gather didn't bring any change - the few packages we install apparently didn't change since 9.6.

scripts/kata-install/Dockerfile
littlejawa: Yep, I think we should drop that commit - we don't use this Dockerfile in our builds. If needs an update, it probably belongs to a separate PR.

gkurz: My comment was because it was hijacking a commit for another component. I'm personally fine to bump this in this PR, while here, as soon as it is in its own commit.

gkurz:
> I have updated the lock file for podvm-builder.

Maybe squash this change into commit `Update UBI references to 9.7 (podvm-builder)` ?

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #1378, opened 2025-11-19
konflux: bump the version in the cpe label

littlejawa:


gkurz (approved):
/lgtm

Good catch @littlejawa !

PR #1381, opened 2025-11-19
test-fbc: Handle pagination in quay.io API

gkurz:
Our pipeline is currently failing to create test catalogs. It appears to be looping on :

  jq: error (at <stdin>:1): Cannot iterate over null (null)
  jq: error (at <stdin>:1): Cannot iterate over null (null)
  jq: error (at <stdin>:1): Cannot iterate over null (null)
  jq: error (at <stdin>:1): Cannot iterate over null (null)
  jq: error (at <stdin>:1): Cannot iterate over null (null)

This comes from a ininite loop when trying to generate the changelog. Leverage `.has_additional` from the quay.io to break out the loop.

### DIFF
--- a/fbc/test-fbc/changelog.sh
+++ b/fbc/test-fbc/changelog.sh
@@ -56,6 +56,8 @@ function get_commit_for_image() {
                 fi
             done
         fi
+        HAS_MORE=$(echo "$RESPONSE" | jq -r '.has_additional')
+        [[ "$HAS_MORE" != "true" ]] && break
         PAGE=$((PAGE + 1))
     done
 }


fbc/test-fbc/changelog.sh
littlejawa, line 71: I see another problem in the log for this PR's build :

```suggestion
        "osc-cloud-api-adaptor" | "osc-cloud-api-adaptor-webhook" | "osc-podvm-payload")
```

littlejawa:
For future debugging: the error shows up because of this commit 58a821a
The script compares old image reference to new image reference, assuming only one image change per commit.
58a821a adds 3 new images that didn't exist before, and therefore have nothing to compare to...
Some more work is needed to get the script to work properly in this case.

gkurz:
Superseded by #1470 .

PR #1401, opened 2025-11-21
Update fbcs version v1.11

dbkreling:
Update fbc versions to v1.11

Signed-off-by: Daniel Kreling <dkreling@redhat.com>

### DIFF
--- a/fbc/v4.17/catalog-template.yaml
+++ b/fbc/v4.17/catalog-template.yaml
@@ -56,6 +56,9 @@ entries:
       - name: sandboxed-containers-operator.v1.10.3
         replaces: sandboxed-containers-operator.v1.10.2
         skipRange: '>=1.1.0 <1.10.3'
+      - name: sandboxed-containers-operator.v1.11.0
+        replaces: sandboxed-containers-operator.v1.10.3
+        skipRange: '>=1.1.0 <1.11.0'
     name: stable
     package: sandboxed-containers-operator
     schema: olm.channel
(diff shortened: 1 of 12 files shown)


dbkreling:
TODOs:
- Update the reference for the osc-operator-bundle image with the newest one, which is done after the 1.11 release is completed.
- Confirm the version pattern for this PR ("1.11" vs "1.11.0").

littlejawa:
> Confirm the version pattern for this PR ("1.11" vs "1.11.0").

Nope :) 
My comment about version strings is valid only for the label we put inside the images.
The catalog actually needs to list every single release we make. If we don't specify the 3rd digit here, it will be pretty hard to differentiate 1.11.0 from 1.11.1 (and more) from the customer' and support's point of view. And it doesn't bring anything to us, as we need to rebuild the catalogs anyway.

The version label in the image, on the other hand, is not used by anybody (that we know of) and visible only when you inspect the image. That's why we can afford to keep it vague, as it allows us to avoid rebuilding an image if it is only because of the 3rd digit version update.

littlejawa:
Enterprise Contract failure comes from the FIPS check, complaining about the missing openssl lib in the dm-verity image.
We know about this problem, we need to ignore this until it is properly addressed on the test side.

fbc/v4.17/catalog-template.yaml
```diff
@@ -56,6 +56,9 @@ entries:
       - name: sandboxed-containers-operator.v1.10.3
         replaces: sandboxed-containers-operator.v1.10.2
         skipRange: '>=1.1.0 <1.10.3'
+      - name: sandboxed-containers-operator.v1.11.0
+        replaces: sandboxed-containers-operator.v1.10.3
+        skipRange: '>=1.1.0 <1.11'
```
gkurz: It seems we always put the `.0` before. Any reason to not do it here ? Just asking per curiosity.

littlejawa: Ah, that's a miss :-( Good catch.
I fixed it on the "name:" line, but missed the skipRange.

fbc/v4.17/catalog-template.yaml
```diff
@@ -103,6 +106,8 @@ entries:
     schema: olm.bundle
   - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:4477810692429ac498b6ec1b295e6ead0f8fea25e91f6ddf17255d7e349b82e1
     schema: olm.bundle
+  - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:f63fe210b7c83db08358ae3696e95d27aadd079f3f1cc6c5b0c89c1959eadffc
+    schema: olm.bundle
```
gkurz, line 110: Latest pushed bundle as expected :+1: .

fbc/v4.17/catalog-template.yaml
gkurz: I just stumbled on your previous comment at https://github.com/openshift/sandboxed-containers-operator/pull/1401#issuecomment-3601898497 :smiley: . Always putting the patch level seems better indeed.

gkurz (approved):
/lgtm

Thanks @dbkreling and @littlejawa !

PR #1403, opened 2025-11-24
Fix min version check for CoCo baremetal

bpradipt:
Extract only the core version (major.minor.patch) from both versions
Ignore prerelease and build metadata when comparing

### DIFF
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -105,7 +105,7 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 	} else {
 		peerpodsCMData = map[string]string{
 			"DISABLECVM": "true",
-			"INITDATA": defaultNonCCInitdata,
+			"INITDATA":   defaultNonCCInitdata,
 		}
 		keysToRemove = nil
 	}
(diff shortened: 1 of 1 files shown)


bpradipt:
This is not a must for 1.11 release, but will be good to have it eventually imho

beraldoleal (approved):
Thank you @bpradipt , LGTM.

bpradipt:
Updated the min version to 4.20.6 based on testing.

@gkurz @beraldoleal do you mind taking a relook

controllers/openshift_controller.go
```diff
@@ -591,12 +593,17 @@ func (r *KataConfigOpenShiftReconciler) isOCPVersionLessThan(minVersion string)
 		return false, currentVersion, fmt.Errorf("unable to parse current OCP version %s: %w", currentVersion, err)
 	}
 
-	min, err := semver.NewVersion(minVersion)
+	// Use constraint with -0 suffix to include prerelease versions in the comparison
+	// This allows "4.20.0-0.nightly-*" to satisfy ">= 4.20.0-0"
+	constraint, err := semver.NewConstraint(">= " + minVersion + "-0")
```
gkurz: AFAICT this can only work when patch level in `minVersion` is 0, right ? It seems that the base problem is that nightlies don't bump the patch level... so I'm wondering if we shouldn't rather have some knob to disable the check.

bpradipt:
Based on slack discussion updated the PR to only bump up min version to OCP 4.20.6. No change in detection logic

controllers/confidential_handler.go
```diff
@@ -105,7 +105,7 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 	} else {
 		peerpodsCMData = map[string]string{
 			"DISABLECVM": "true",
-			"INITDATA": defaultNonCCInitdata,
+			"INITDATA":   defaultNonCCInitdata,
```
gkurz, line 108: Heh this one has been a pain in my dev workflow that includes `make fmt`. Thanks ! :wink:

We should have CI to block ill-formated PRs.

gkurz (approved):
/lgtm

Thanks @bpradipt !

controllers/confidential_handler.go
bpradipt: True. I think another option is to have pre-commit hooks.

PR #1415, opened 2025-11-27
Bump the osc-daemonset image to 1.11

gkurz:
The image doesn't exist yet but we do want OSC 1.11 to reference it.

Note that daemonset won't be usable at all as long as the image isn't available in quay.

Cc @ANJANA-A-R-K @ajaypvictor @Pacho20 @gcoon151


ajaypvictor (approved):
LGTM, Thanks..

PR #1429, opened 2025-12-02
Release 1.11: update the catalogs with the latest bundle reference

littlejawa:
Modified the catalog templates manually, then ran render.sh to generate the other files.

We had a build issue and needed to rebuild the osc-monitor, which updated the bundle.


dbkreling (approved):

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #1434, opened 2025-12-03
mirror-set: Remove references to OSC 1.10

gkurz:
Released images should be accessed from the official RH registry.

Intermediate builds are time-bombed (5 days).

We don't have anything to pull from the v1-10 repos at quay.io.


littlejawa (approved):
/lgtm

PR #1468, opened 2025-12-09
ibmcloud: check worker node version

Pacho20:
This fixes issues where worker nodes run a different OCP version than the control plane. This can cause issues with the DaemonSet installation mode.
**- What I did**
Add a function to check the cloud provider in the DaemonSet install script. If the provider is ibmcloud, retrieve the node's version and use it to get the extension image for that specific version instead of using the environment variable.

### DIFF
--- a/scripts/kata-install/lib.sh
+++ b/scripts/kata-install/lib.sh
@@ -55,3 +55,9 @@ client_tools() {
 	chmod +x /usr/bin/oc
 	chmod +x /usr/bin/kubectl
 }
+
+error() {
+	local msg="$*"
+	printf 'ERROR: %s\n' "$msg" >&2
+	return 1
+}
--- a/scripts/kata-install/osc-kata-install.sh
+++ b/scripts/kata-install/osc-kata-install.sh
@@ -187,9 +187,99 @@ uninstall() {
 	wait_for_reboot_clear
 }
 
+get_cloud_provider() {
+	local provider
+
+	# Run kubectl: capture rc and stdout
+	if ! provider="$(kubectl get configmap/peer-pods-cm \
+		-n openshift-sandboxed-containers-operator \
+		-o jsonpath='{.data.CLOUD_PROVIDER}')"; then
+		error "get_cloud_provider: kubectl failed to query peer-pods-cm"
+		return
+	fi
+
+	# Reject empty/whitespace-only results
+	if [[ -z "${provider//[[:space:]]/}" ]]; then
+		error "get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm"
+		return
+	fi
+
+	echo "${provider}"
+}
+
+# Reads worker version label and extracts the OpenShift version (prefix before '_').
+get_worker_node_version_ibmcloud() {
+	local version
+	local raw_version
+
+	if ! raw_version="$(kubectl get "nodes/${NODE_NAME}" \
+		-o jsonpath='{.metadata.labels.ibm-cloud\.kubernetes\.io/worker-version}')"; then
+		error "get_worker_node_version_ibmcloud: kubectl failed to read worker-version label on node ${NODE_NAME}"
+		return
+	fi
+
+	if [[ -z "${raw_version//[[:space:]]/}" ]]; then
+		error "get_worker_node_version_ibmcloud: worker-version label is empty on node ${NODE_NAME}"
+		return
+	fi
+
+	# Extract the part before the first underscore
+	version="${raw_version%%_*}"
+
+	if [[ -z "${version//[[:space:]]/}" ]]; then
+		error "get_worker_node_version_ibmcloud: cannot parse '${raw_version}' as an OpenShift version"
+		return
+	fi
+
+	echo "${version}"
+}
+
+# get_extension_image VERSION
+# Resolves the rhel-coreos-extensions image for a given release VERSION.
+get_extension_image() {
+	[[ -n ${1-} ]] || {
+		error "get_extension_image: usage: get_extension_image VERSION"
+		return
+	}
+
+	local version=$1
+	local image
+
+	if ! image="$(oc adm release info --image-for rhel-coreos-extensions "$version")"; then
+		error "get_extension_image: 'oc adm release info' failed for version: $version"
+		return
+	fi
+
+	if [[ -z "${image//[[:space:]]/}" ]]; then
+		error "get_extension_image: empty image string for version: $version"
+		return
+	fi
+
+	echo "${image}"
+}
+
+# FIXME : This logic should be moved into the Operator
 rpm-extensions() {
+	local provider
+	local extension_image
+
+	provider="$(get_cloud_provider)"
+
+	case "${provider}" in
+	ibmcloud)
+		local version
+		version="$(get_worker_node_version_ibmcloud)"
+		extension_image="$(get_extension_image $version)"
+		;;
+
+	*)
+		extension_image=$EXTENSION_IMAGE
+		;;
+
+	esac
+
 	mkdir -p "/usr/share/rpm-ostree/extensions"
-	extract_container_image "$EXTENSION_IMAGE" "/usr/share/rpm-ostree/extensions" "/usr/share/rpm-ostree" "/tmp/regauth/auth.json"
+	extract_container_image "$extension_image" "/usr/share/rpm-ostree/extensions" "/usr/share/rpm-ostree" "/tmp/regauth/auth.json"
 }
 
 main() {


scripts/kata-install/osc-kata-install.sh
```diff
@@ -187,9 +187,66 @@ uninstall() {
...
 }
 
+get_cloud_provider() {
+	provider="$(kubectl get configmap/peer-pods-cm -n openshift-sandboxed-containers-operator -o jsonpath='{.data.CLOUD_PROVIDER}')"
+	
+	if [[ -z "${provider}" ]]; then
+		echo "Can't find the CLOUD_PROVIDER in the peer-pods-cm"
+		exit 1
+	fi
+
```
gkurz: Please drop the leading tab.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -187,9 +187,66 @@ uninstall() {
 	wait_for_reboot_clear
 }
 
+get_cloud_provider() {
+	provider="$(kubectl get configmap/peer-pods-cm -n openshift-sandboxed-containers-operator -o jsonpath='{.data.CLOUD_PROVIDER}')"
```
gkurz: It is a bit unfortunate to make `peer-pods-cm` a requirement for this script. What about getting the information from a `CLOUD_PROVIDER` environment variable ? This can be easily provided by the controller.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -187,9 +187,66 @@ uninstall() {
...
+	fi
+	
+	echo "${provider}"
+}
+
+get_worker_node_version_ibmcloud() {
+	raw_version="$(kubectl get nodes/${NODE_NAME} -o jsonpath='{.metadata.labels.ibm-cloud\.kubernetes\.io/worker-version}')"
+
+	if [[ -z "${raw_version}" ]]; then
+		echo "Can't find the ibm-cloud.kubernetes.io/worker-version label"
```
gkurz: Errors should go to stderr. Since this script has quite some instances of that, I'd suggest to introduce an `error()` helper that writes to stderr and does `exit 1` (in a preparatory commit).

scripts/kata-install/osc-kata-install.sh
```diff
@@ -187,9 +187,66 @@ uninstall() {
...
+
+	*)
+		extension_image=$EXTENSION_IMAGE
+		;;
+
+	esac
+
 	mkdir -p "/usr/share/rpm-ostree/extensions"
-	extract_container_image "$EXTENSION_IMAGE" "/usr/share/rpm-ostree/extensions" "/usr/share/rpm-ostree" "/tmp/regauth/auth.json"
+	extract_container_image "$extension_image" "/usr/share/rpm-ostree/extensions" "/usr/share/rpm-ostree" "/tmp/regauth/auth.json"
```
gkurz, line 282: This basically means that the OSC controller generates an incorrect `EXTENSION_IMAGE` in the IBM Cloud case and we have to override it here. It is acceptable for a quick fix but for the long term, I'd prefer the controller to implement the appropriate logic directly.

gcoon151: agree this is short term fix to get it working again and we'll longer term move this to better places.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -187,9 +187,63 @@ uninstall() {
...
+	provider="$(kubectl get configmap/peer-pods-cm -n openshift-sandboxed-containers-operator -o jsonpath='{.data.CLOUD_PROVIDER}')"
+	
+	if [[ -z "${provider}" ]]; then
+		error "Can't find the CLOUD_PROVIDER in the peer-pods-cm"
+	fi
+	
+	echo "${provider}"
+}
+
+get_worker_node_version_ibmcloud() {
```
bpradipt, line 211: This appears to be sane logic to detect the version of the worker node for retrieving the correct extension image.
Why this has to be provider specific? Is this due to the non-availability of the annotation in other providers ?

scripts/kata-install/osc-kata-install.sh
Pacho20: True, but it would require a new release of the operator. If we plan to move the entire logic to the operator anyway, can we address this at that point and make this PR a quick fix? If you think I should pass the provider as an environment variable now I can do that.

scripts/kata-install/osc-kata-install.sh
Pacho20: I agree; I marked the function with a FIXME. This PR is intended as a quick fix, and we already have plans to implement this in the operator.

scripts/kata-install/osc-kata-install.sh
Pacho20: Yes, that's the main reason. I don't know if other providers have a similar label or annotation, but I can make the function non-provider-specific if you think that's a better approach. I didn’t really consider this because it’s intended as a quick fix, and the logic will be moved into the operator. See [Greg's comment](https://github.com/openshift/sandboxed-containers-operator/pull/1468#discussion_r2607644857).

scripts/kata-install/osc-kata-install.sh
gkurz: That's fine for the temporary quick fix.

scripts/kata-install/osc-kata-install.sh
gkurz: Works for me.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -187,9 +187,112 @@ uninstall() {
 	wait_for_reboot_clear
 }
 
+# get_cloud_provider OUT_VAR
+# Writes the cloud provider string to OUT_VAR.
```
gkurz: This shell construct is far too unusual and it obfuscates the code for no benefice. Please revert to do the usual way, i.e. return result with `echo`.

Pacho20: done

scripts/kata-install/lib.sh
```diff
@@ -55,3 +55,17 @@ client_tools() {
...
+
+# Non-fatal: log and return
+error() {
+	local msg="$*"
+	printf 'ERROR: %s\n' "$msg" >&2
+	return 1
+}
+
+# Fatal: log and exit
+die() {
```
gkurz: All the user scripts are started with `set -e` => `return 1` in `error()` is already fatal.

`die()` isn't needed in this context and it doesn't have any user in this series. Please drop it for the sake of simplicity. The `Non-fatal/Fatal` comments should be removed as well.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -187,9 +187,102 @@ uninstall() {
...
 
+get_cloud_provider() {
+	local provider
+
+	# Run kubectl: capture rc and stdout
+	if ! provider="$(kubectl get configmap/peer-pods-cm \
+		-n openshift-sandboxed-containers-operator \
+		-o jsonpath='{.data.CLOUD_PROVIDER}')"; then
+		error "get_cloud_provider: kubectl failed to query peer-pods-cm"
+		return
```
gkurz, line 198: `return` is dead code since `set -e` will cause `error()` to fail. Please remove it and other occurences.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -187,9 +187,102 @@ uninstall() {
...
+	fi
+
+	# Reject empty/whitespace-only results
+	if [[ -z "${provider//[[:space:]]/}" ]]; then
+		error "get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm"
+		return
+	fi
+
+	echo "${provider}"
+	return 0
```
gkurz: Not needed. Please drop it and other occurences.

gkurz (commented):
Thanks for the changes @Pacho20 . Some more polishing and it should be good to go.

scripts/kata-install/lib.sh
Pacho20: Done

scripts/kata-install/osc-kata-install.sh
Pacho20: Done

scripts/kata-install/osc-kata-install.sh
Pacho20: As I understand it, return is necessary when using command substitution. Functions executed via command substitution run in a subshell, and the subshell does not inherit set -e.
Without return, the script continues execution even after an error occurs.

Some logs running this without `return`:
```
...
++ get_cloud_provider
++ local provider
+++ kubectl get configmap/peer-pods-cm -n openshift-sandboxed-containers-operator -o 'jsonpath={.data.CLOUD_PROVIDER}'
++ provider=
++ [[ -z '' ]]
++ error 'get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm'
++ local 'msg=get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm'
++ printf 'ERROR: %s\n' 'get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm'
ERROR: get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm
++ return 1
++ echo ''
+ provider=
+ case "${provider}" in
+ extension_image=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:e40ac6498a409397d37c563f377675408a871f9971d354bb8f1a0b1845a0381f
+ mkdir -p /usr/share/rpm-ostree/extensions
+ extract_container_image quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:e40ac6498a409397d37c563f377675408a871f9971d354bb8f1a0b1845a0381f /usr/share/rpm-ostree/extensions /usr/share/rpm-ostree /tmp/regauth/auth.json
...

the script continues "normally"
```

Some logs running with `return`:
```
...
++ get_cloud_provider
++ local provider
+++ kubectl get configmap/peer-pods-cm -n openshift-sandboxed-containers-operator -o 'jsonpath={.data.CLOUD_PROVIDER}'
++ provider=
++ [[ -z '' ]]
++ error 'get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm'
++ local 'msg=get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm'
++ printf 'ERROR: %s\n' 'get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm'
ERROR: get_cloud_provider: CLOUD_PROVIDER not set in configmap/peer-pods-cm
++ return 1
++ return
+ provider=

script exits here, Pod goes into error state
```

The other solutions I found for this problem seemed more complicated than simply using return.

gkurz: Oh yes you're right. Let's keep that then.

gkurz (approved):
/lgtm

Thanks @Pacho20 !

PR #1469, opened 2025-12-09
tests/osc-test-fbc-integration: new parameters and misc updates

wainersm:
The current osc-test-fbc-integration pipeline in Konflux which runs to test the test-fbc catalog have a broken test because it expects the old operator version of 1.11.0. I could have it fixed in Prow side, but I preferred to add a new `EXPECTED_OPERATOR_VERSION` parameter to the pipeline so we can bump the version at same time test-fbc version is raised.

In order to have `EXPECTED_OPERATOR_VERSION` taking effect I had to fix a issue (that's the commit 3c6630582b9fd496997a3a8f2c6c47030af867ca).

While in here, because testing these changes is time-consuming, I added two more changes:

* passing the `FORCE_SUCCESS_EXIT=no` parameter to force the job report fail to any kind of error (CI infra or tests failures)
* bumped the konflux-tasks to latest and greatest to pick up some nice features like job submission retry

### DIFF
--- a/tests/osc-test-fbc-integration.yaml
+++ b/tests/osc-test-fbc-integration.yaml
@@ -38,7 +38,7 @@ spec:
               echo "Snapshot: ${SNAPSHOT}"
               catalogImage=$(jq -r '.components[] | select(.name=="osc-test-fbc") | .containerImage' <<< "${SNAPSHOT}")
               echo "Catalog image: ${catalogImage}"
-              echo "${catalogImage}" > $(results.CATALOG_IMAGE.path)
+              echo -n "${catalogImage}" > $(results.CATALOG_IMAGE.path)
     - name: prow-job
       displayName: "Running prow job $(params.PROWJOB_NAME)"
       timeout: "4h"
--- a/tests/osc-test-fbc-integration.yaml
+++ b/tests/osc-test-fbc-integration.yaml
@@ -59,7 +59,7 @@ spec:
         - name: VARIANT
           value: downstream-candidate
         - name: ENVS
-          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE)"
+          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE),EXPECTED_OPERATOR_VERSION=1.11.1"
       matrix:
         params:
           - name: PROWJOB_NAME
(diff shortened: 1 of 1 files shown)


wainersm:
Managed to trigger konflux to test these changes in https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/ose-osc-tenant/applications/osc-test-catalog/pipelineruns/osc-test-catalog-integration-6dvxq?integrationTestName=osc-test-catalog-integration

littlejawa (approved):
/lgtm
Thanks @wainersm !

wainersm:
/retest

The latest failure on CI seems to be in Prow infra.

snir911 (approved):

tbuskey (approved):

tests/osc-test-fbc-integration.yaml
```diff
@@ -59,7 +59,7 @@ spec:
         - name: VARIANT
           value: downstream-candidate
         - name: ENVS
-          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE)"
+          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE),EXPECTED_OPERATOR_VERSION=1.11.1"
```
gkurz, line 62: Should this bump occur at the same time we bump versions with the `## OSC_VERSION` tag as documented at https://github.com/openshift/sandboxed-containers-operator/blob/devel/docs/DEVELOPMENT.md#updating-versions ? Or should this be only bumped after the bundle was pushed and the test-catalog udpated accordingly as in f70661f620afdd2b93a8b97e0f6f1b0179a38ae3 ?

wainersm: Hi @gkurz !

This pipeline aims to test OSC on updates of the test-catalog, so it should be only bumped alongside the commit https://github.com/openshift/sandboxed-containers-operator/commit/f70661f620afdd2b93a8b97e0f6f1b0179a38ae3

I want to document it somewhere or have the bump done automatically, that's why I asked you and @littlejawa on slack. IIUC test-catalog version is bumped manually so I want to have a small section in fbc/README.md that explains how test-catalog version (and this pipeline) is updated. Makes sense?

wainersm:
Hi @bpradipt ! It seems that it needs your review to have this merged. :)

PR #1470, opened 2025-12-10
Fix changelog generator

littlejawa:
I've been testing the script on various situations, and fixing problems as I find them.
Hopefully, this should be usable again.

### DIFF
--- a/fbc/test-fbc/Dockerfile
+++ b/fbc/test-fbc/Dockerfile
@@ -3,6 +3,8 @@
 FROM brew.registry.redhat.io/rh-osbs/openshift-ose-operator-registry-rhel9:v4.19 as builder
 ARG GIT_URL=""
 ARG GIT_REVISION=""
+ARG OLDCOMMIT=""
+ARG NEWCOMMIT=""
 USER root
 
 WORKDIR /workdir
@@ -19,6 +21,8 @@ RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]
 # Optionally clone the git repo and generate changelog from it
 ENV GIT_URL=${GIT_URL}
 ENV GIT_REVISION=${GIT_REVISION}
+ENV OLDCOMMIT=${OLDCOMMIT}
+ENV NEWCOMMIT=${NEWCOMMIT}
 RUN if [[ -n "${GIT_URL}" && -n "${GIT_REVISION}" ]]; then \
         echo GIT_URL="${GIT_URL}" ; \
         echo GIT_REVISION="${GIT_REVISION}" ; \
@@ -26,7 +30,7 @@ RUN if [[ -n "${GIT_URL}" && -n "${GIT_REVISION}" ]]; then \
         pushd repo && \
         git checkout "${GIT_REVISION}" && \
         cd fbc/test-fbc && \
-        ./changelog.sh 2>&1 | tee CHANGELOG.err || true && \
+        ./changelog.sh ${OLDCOMMIT} ${NEWCOMMIT} 2>&1 | tee CHANGELOG.err || true && \
         mv CHANGELOG CHANGELOG.err /workdir/ && \
         popd ; \
     fi
--- a/fbc/test-fbc/Dockerfile
+++ b/fbc/test-fbc/Dockerfile
@@ -17,22 +17,21 @@ RUN cp -r test-fbc/catalog/ /configs
 RUN ["/bin/opm", "serve", "/configs", "--cache-dir=/tmp/cache", "--cache-only"]
 
 # Optionally clone the git repo and generate changelog from it
-#
-# FIXME : temporarily unplug changelog generation until changelog.sh is fixed
-#         see https://github.com/openshift/sandboxed-containers-operator/pull/1381
-#
-# ENV GIT_URL=${GIT_URL}
-# ENV GIT_REVISION=${GIT_REVISION}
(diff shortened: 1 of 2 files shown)


fbc/test-fbc/changelog.sh
```diff
@@ -46,6 +46,10 @@ function get_commit_for_image() {
             break
         fi
         TAGS=$(echo "$RESPONSE" | jq -r '.tags[] | select(.manifest_digest=="'"$IMAGE_DIGEST"'") | .name')
+        if [ $? -ne 0 ]; then
+            echo "Error: Failed to parse JSON response from quay.io for image $IMAGE_NAME" >&2
+            break
+        fi
```
gkurz, line 62: FWIW I had identified the reason for the infinite loop as a lack of GH pagination handling (see https://github.com/gkurz/sandboxed-containers-operator/commit/993830782603ae0fbbe5aea5b39908e970707d29).

littlejawa: I have added your commit too. It's cleaner to handle the pagination than to wait for an error to happen.
I keep the error check here anyway, just in case something else causes a problem.

fbc/test-fbc/changelog.sh
```diff
@@ -115,22 +115,36 @@ COMMIT_LIST=$(git log --oneline "${OLD_BUNDLE_COMMIT}..${NEW_BUNDLE_COMMIT}" ../
...
+        echo "-----" | tee -a CHANGELOG
+        echo "" | tee -a CHANGELOG
+        continue
+    fi
+
+    # from this point, we assume we have only one image changed per commit
+    # take the first one, if there are more
+    OLD_IMAGE=$(echo "$OLD_IMAGES" | head -n1)
+    NEW_IMAGE=$(echo "$NEW_IMAGES" | head -n1)
+
```
gkurz, line 158: Maybe add a sanity check to ensure NEW_IMAGE is indeed a new version of OLD_IMAGE ? Just in case, to avoid confusion.

fbc/test-fbc/changelog.sh
```diff
@@ -145,6 +145,16 @@ for COMMIT in $COMMIT_LIST; do
     OLD_IMAGE=$(echo "$OLD_IMAGES" | head -n1)
     NEW_IMAGE=$(echo "$NEW_IMAGES" | head -n1)
 
+    # Corner case: when we update the version strings, we get OSC_VERSION as the
+    # image reference. This is because we mark all version strings to update with
+    # the comment "# OSC_VERSION".
```
gkurz: This is clever but it seems a bit hacky to rely on this implementation detail in the parsing of `image:` lines. What about an explicit check upfront to skip `image: .* ## OSC_VERSION` ?

littlejawa: Right... Filtering out the lines with OSC_VERSION from the start is much cleaner actually. 
I just modified this commit (see c18c1b6)

fbc/test-fbc/Dockerfile
```diff
@@ -17,22 +17,21 @@ RUN cp -r test-fbc/catalog/ /configs
...
+ENV GIT_URL=${GIT_URL}
+ENV GIT_REVISION=${GIT_REVISION}
+RUN if [[ -n "${GIT_URL}" && -n "${GIT_REVISION}" ]]; then \
+        echo GIT_URL="${GIT_URL}" ; \
+        echo GIT_REVISION="${GIT_REVISION}" ; \
+        git clone "${GIT_URL}" repo && \
+        pushd repo && \
+        git checkout "${GIT_REVISION}" && \
+        cd fbc/test-fbc && \
+        ./changelog.sh 2>&1 | tee CHANGELOG.err || true && \
```
gkurz: `2>&1` will result in CHANGELOG.err to also have the content of CHANGELOG since `changelog.sh` does `tee CHANGELOG` already. Is that intended ?

fbc/test-fbc/Dockerfile
```diff
@@ -3,6 +3,8 @@
 FROM brew.registry.redhat.io/rh-osbs/openshift-ose-operator-registry-rhel9:v4.19 as builder
 ARG GIT_URL=""
 ARG GIT_REVISION=""
+ARG OLDCOMMIT=""
+ARG NEWCOMMIT=""
```
gkurz, line 7: This is for the case where one doesn't want the script directly in their repo, right ? If so, maybe worth to document this in some README.md ?

gkurz: Not strictly needed.

gkurz (approved):
Great work ! Thanks @littlejawa !

/lgtm

fbc/test-fbc/Dockerfile
gkurz: Yes it is otherwise it would be hard to guess which command failed :smiley:

fbc/test-fbc/changelog.sh
littlejawa: Discussed with Greg, putting our conclusion here:
This is nice to have, but should come on a later update. The reason is that this script is meant to be used in the CI, where there is 0 chance that we get bad input for commit references.
The situation may arise for local execution, and we certainly want to sanitize the inputs there, but it can come later.
First, let's get this script running in test-fbc builds, and get changelogs with our catalog.

fbc/test-fbc/Dockerfile
littlejawa: Yes, that's the intent.
Having only the errors in CHANGELOG.err makes it difficult to troubleshoot, because we don't know what the error messages relates to in the script's execution.

PR #1475, opened 2025-12-11
konflux: centralize pipeline specs into a separate file.

littlejawa:
As documented in build pipeline management patterns in Konflux doc (https://konflux.pages.redhat.com/docs/users/patterns/centralize-pipeline-definitions.html)

The pipeline specs are shared by all components, except the bundle which is not multi-arch. Centralizing the definition makes it easier to maintain.

NOTE: we've already done this for the FBCs, now applying to the other images


gkurz (approved):
Very nice diffstat ! Thanks @littlejawa !

/lgtm

PR #1479, opened 2025-12-15
Update fbcs for release 1.11.1

littlejawa:
Updating the catalogs for release 1.11.1
Using the bundle reference from our successful stage release

### DIFF
--- a/fbc/v4.17/catalog-template.yaml
+++ b/fbc/v4.17/catalog-template.yaml
@@ -59,6 +59,9 @@ entries:
       - name: sandboxed-containers-operator.v1.11.0
         replaces: sandboxed-containers-operator.v1.10.3
         skipRange: '>=1.1.0 <1.11.0'
+      - name: sandboxed-containers-operator.v1.11.1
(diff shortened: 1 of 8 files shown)


littlejawa:
Enterprise contract failures expected: this is the FIPS check with dm-verity

gkurz:
> Enterprise contract failures expected: this is the FIPS check with dm-verity

Would you know why is the check marked as `failing` for 4.20 and `neutral` for the other ones ?

littlejawa:
> > Enterprise contract failures expected: this is the FIPS check with dm-verity
> 
> Would you know why is the check marked as `failing` for 4.20 and `neutral` for the other ones ?

Yes :-)
Konflux marks tests as "failed" if they are mandatory, and "neutral" if they are optional.
The 4.20 Enterprise Contract was marked "mandatory", while the others are optional - probably an overlook.
I modified it to be "optional" like all the others, so it should be consistent in future runs.

dbkreling (approved):

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #1480, opened 2025-12-16
konflux: add a renovate.json file to control Mintmaker PR flow

littlejawa:
Trying to add renovate (Mintmaker) config to reduce the number of PRs we get.
The rationale is
- we usually don't want to merge major versions of dependencies, as they are generally more risky. We should do that in a controlled, manual way.
- we may want to keep minor version updates. But at this point I think we should reduce drastically the number of PRs we get every week.

With this setting, we should get only "patch" version updates, which in theory gets us security fixes, with a reduced risk of breaking things.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,7 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json?raw=true",
+        "github>konflux-ci/mintmaker-presets:disable-minor-updates"
+    ]
+}


renovate.json
```diff
@@ -0,0 +1,7 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "extends": [
+        "https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json?raw=true",
+        "github>konflux-ci/mintmaker-presets:disable-minor-updates"
+    ]
+}
```
gkurz, line 7: After looking at :
- https://konflux-ci.dev/docs/mintmaker/user/#create-your-custom-configuration
- https://konflux-ci.dev/docs/mintmaker/user/#configuration-presets
- https://github.com/konflux-ci/mintmaker-presets/blob/main/README.md

`https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json` is applied to all onboarded repos already and I have the impression that `renovate.json` should just be:

```
{
    "extends": [
        "github>konflux-ci/mintmaker-presets:disable-minor-updates"
    ]
}
```

Can you clarify @littlejawa ?

littlejawa: My understanding is that it is effectively applied to all repositories by default, UNLESS you customize it.
When you customize it, you need to reference a base config.
i.e: adding a "renovate.json" file breaks the default, which needs to be stated in the file.

littlejawa:
Closing in favor of #1519

PR #1485, opened 2025-12-18
Support node-specific extension images in DaemonSet mode

Pacho20:
This fixes issues where worker nodes run a different OCP version than the control plane. This can cause issues with the DaemonSet installation mode.
**- What I did**
Introduced new image component logic. By default, the control plane version is used for all images, but based on the provider, it is now possible to get any component image for any version. Each node now receives a separate image string, and environment variable names include the node name to ensure uniqueness.
**- How to verify it**
Deploy the Operator in DaemonSet mode. Verify that installation pods use the correct image for each node.
Support node-specific extension images in DaemonSet mode

### DIFF
--- a/controllers/image_component.go
+++ b/controllers/image_component.go
@@ -8,11 +8,13 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 	"sync"
 
 	semver "github.com/Masterminds/semver/v3"
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/oc/pkg/cli/admin/release"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/cli-runtime/pkg/genericiooptions"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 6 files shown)


controllers/image_component.go
```diff
@@ -0,0 +1,59 @@
...
+			// we found the short name in ImageStream
+			if tag.From != nil && tag.From.Kind == "DockerImage" {
+				return tag.From.Name, nil
+			}
+		}
+	}
+
+	// Didn't find it
+	return "", fmt.Errorf("empty result for image name")
+}
```
gkurz: Missing terminating newline

controllers/image_component.go
```diff
@@ -0,0 +1,271 @@
...
+		},
+		http.DefaultTransport,
+	)
+	if err != nil {
+		return "", err
+	}
+	httpClient := &http.Client{Transport: transport}
+
+	u.RawQuery = url.Values{
+		"channel": []string{fmt.Sprintf("%s-%d.%d", "stable", semanticVersion.Major(), semanticVersion.Minor())},
```
Pacho20, line 234: Something I’ve been thinking about recently: I essentially restricted the channel to stable, which may not be the best solution. I guess this would prevent testing on new releases that haven’t reached the stable channel yet.

gkurz: Yes, it should at least be `fast` or maybe even `candidate`.

controllers/image_component.go
gkurz, line 59: Missing terminating newline. `make fmt` can fix that. It is good practice to always run it before committing to your branch.

PR #1486, opened 2025-12-19
test-fbc: fix boundaries of the commit range in changelog.sh

littlejawa:
When using "git rev-list", the first element of the range is ignored. This causes issue in the changelog generation if the first item actually references a change in catalog-template.yaml.
This commit makes sure that the first item is included in the search.

### DIFF
--- a/fbc/test-fbc/changelog.sh
+++ b/fbc/test-fbc/changelog.sh
@@ -109,8 +109,9 @@ fi
 
 # Make sure OLD_COMMIT and NEW_COMMIT reference changes to catalog-template.yaml
 # Find the first and last commit that references it in the range OLD_COMMIT..NEW_COMMIT
-OLD_COMMIT=$(git rev-list "${OLD_COMMIT}..${NEW_COMMIT}" -- catalog-template.yaml | tail -n1)
-NEW_COMMIT=$(git rev-list "${OLD_COMMIT}..${NEW_COMMIT}" -- catalog-template.yaml | head -n1)
+# NOTE: we use ~1 to include the OLD_COMMIT itself in the search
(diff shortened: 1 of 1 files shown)


fbc/test-fbc/changelog.sh
```diff
@@ -109,8 +109,9 @@ fi
 
 # Make sure OLD_COMMIT and NEW_COMMIT reference changes to catalog-template.yaml
 # Find the first and last commit that references it in the range OLD_COMMIT..NEW_COMMIT
-OLD_COMMIT=$(git rev-list "${OLD_COMMIT}..${NEW_COMMIT}" -- catalog-template.yaml | tail -n1)
-NEW_COMMIT=$(git rev-list "${OLD_COMMIT}..${NEW_COMMIT}" -- catalog-template.yaml | head -n1)
+# NOTE: we use ~1 to include the OLD_COMMIT itself in the search
```
gkurz, line 112: FWIW `^` can be used instead of `~1` but this is nit picking :wink:

gkurz (approved):
/lgtm

Thanks @littlejawa !

fbc/test-fbc/changelog.sh
littlejawa: Isn't it the opposite?
From the `git rev-list --help` command :

```
Thus, the following command:
$ git rev-list foo bar ^baz
means "list all the commits which are reachable from foo or bar, but not from baz".

A special notation "<commit1>..<commit2>" can be used as a short-hand for "^<commit1> <commit2>".
For example, either of the following may be used interchangeably:
$ git rev-list origin..HEAD
$ git rev-list HEAD ^origin
```

I.e: my understanding is that the `^` notation will exclude a commit. And adding `~1` brings it back.

PR #1516, opened 2026-01-09
azure: enhance storage account security in podvm-image-handler

bpradipt:
Add security hardening to Azure storage account creation in the VHD upload process:
- Enforce minimum TLS version 1.2
- Disable cross-tenant object replication
- Disable public network access
- Enforce HTTPS-only connections
- Disable public blob access

These changes ensure the temporary storage accounts used for VHD uploads comply with Azure security best practices and organizational security policies.

**- What I did**

**- How to verify it**

### DIFF
--- a/config/peerpods/podvm/azure-podvm-image-handler.sh
+++ b/config/peerpods/podvm/azure-podvm-image-handler.sh
@@ -510,6 +510,10 @@ function create_azure_image_from_prebuilt_artifact() {
 
     # Clean up
     rm "${podvm_image_path}"
+
+    # Delete private endpoint before deleting storage account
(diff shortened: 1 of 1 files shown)


bpradipt:
Need to test this in a regular connected Azure cluster.

bpradipt:
Tested on self-managed Azure

snir911 (approved):

bpradipt:
Tested successfully on both ARO and Azure connected clusters

bpradipt:
For complete disconnected support the podvm-builder image must have the deps preinstalled. This will be addressed in a follow-on PR.

PR #1553, opened 2026-01-14
konflux: update the tekton files for building 1.11 from new branch

littlejawa:
For the pipeline to trigger on the new branch, we need the tekton files to be updated so that they reference the new Konflux object names, branch name, and quay.io image location.


littlejawa:
/hold
I need to control when this gets merged, so that I can verify nudges are working as expected for the (new) 1.11 pipelines

PR #1557, opened 2026-01-15
Prepare 1.11.2

littlejawa:
Updating the FBCs with latest builds for 1.11.2

### DIFF
--- a/fbc/v4.17/catalog-template.yaml
+++ b/fbc/v4.17/catalog-template.yaml
@@ -116,7 +116,7 @@ entries:
     schema: olm.bundle
   - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:7b2035904ec464fe32c55ace525f573a491b046b5ef7cd5721d7a67cfc8dae25
     schema: olm.bundle
-  - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:1cfc8fed959fb609765a75e0490649b858c17eb23f88b4065596d8920e71b513
+  - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:33ed98ca0a02d5a99a09b1d232c0ba31ec5ec0f7990828c1896c876902e213f3
     schema: olm.bundle
   - schema: olm.deprecations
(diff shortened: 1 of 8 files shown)


fbc/v4.17/catalog-template.yaml
```diff
@@ -116,7 +116,7 @@ entries:
     schema: olm.bundle
   - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:7b2035904ec464fe32c55ace525f573a491b046b5ef7cd5721d7a67cfc8dae25
     schema: olm.bundle
-  - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:1cfc8fed959fb609765a75e0490649b858c17eb23f88b4065596d8920e71b513
+  - image: registry.redhat.io/openshift-sandboxed-containers/osc-operator-bundle@sha256:33ed98ca0a02d5a99a09b1d232c0ba31ec5ec0f7990828c1896c876902e213f3
```
gkurz, line 119: This is tag `osc-operator-bundle-on-push-rv5g2-build-container` which is the latest build indeed... I'm not sure about testing coverage though since we had some azure quota limits yesterday. Any idea @littlejawa ?

littlejawa: @tbuskey made some tests in https://github.com/openshift/release/pull/73514
The azure tests failed because of the quota, but I see the AWS+peerpods test suceeded.
Given that this new build is just about updating the base images - no code change on our side - I'm asking for a simple regression test, and I think that this is what we just had. I don't think we *need* to test across all cloud providers.

gkurz (approved):
/lgtm

Thanks @littlejawa !

And now this should be cherry-picked on the osc-release-v1.11 branch, right ?

littlejawa:
> And now this should be cherry-picked on the osc-release-v1.11 branch, right ?

Nope :-)
The build of FBC is independent from the images that they reference. FBCs are built from the devel branch, and are not duplicated.
Actually, we *could* have a totally separate repository for FBCs. That's what we do for Trustee for instance. But we thought we already had too many repositories to maintain, so we've put the FBC files here.

PR #1578, opened 2026-01-19
fix: Removing backup kata-se/config file

chathuryaadapa:
For clean rpm uninstall to happen. i.e, removal of kata-se folder while uninstalling
Fixes: [KATA-4605](https://issues.redhat.com/browse/KATA-4605)

``` Once the cleanup job `osc-rpm-uninstall-xxxx` is triggered, we see that in worker node `kata-se` directory is not getting removed automatically ```

**- What I did**

``` In the demonset flow there is a block of code which do the backup of `kata-se/configuration.toml` as `kata-se/configuration.toml.backup-1767604472` because of this while unistalling this directory is not getting cleanup. Hence I removed backing up of `kata-se/configuration.toml` since `kata-se/configuration.toml` can be directly used just like other runtimeclass configuration.toml.```

**- How to verify it**

### DIFF
--- a/scripts/kata-install/osc-kata-addons-install.sh
+++ b/scripts/kata-install/osc-kata-addons-install.sh
@@ -28,11 +28,6 @@ update_config() {
         return 1
     fi
     
-    echo "Updating configuration: $config_file"
-    
-    # Backup config
-    chroot /host cp "$config_file" "${config_file}.backup-$(date +%s)"
-    
     # Update kernel if provided
 	chroot /host sed -i "s|^\(kernel[[:space:]]*=[[:space:]]*\)\".*\"|\1\"$kernel_path\"|g" "$config_file"
 	echo "  Updated kernel: $kernel_path"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/kata-install/osc-kata-addons-install.sh [hunk 1]
```diff
@@ -28,11 +28,6 @@ update_config() {
         return 1
     fi
     
-    echo "Updating configuration: $config_file"
-    
-    # Backup config
-    chroot /host cp "$config_file" "${config_file}.backup-$(date +%s)"
```
c3d, line 34: What file is being copied here in practice? If the cp succeeds that means there is a file there, and uninstall should restore it, no?

Shouldn't we keep the configuration if it predated the install, and remove the file and directory only if nothing existed before?

c3d (approved):
With a question about where the configuration file being copied as a backup is coming from, this otherwise looks good to me.

scripts/kata-install/osc-kata-addons-install.sh
chathuryaadapa: we were taking the back up configuration.toml file under kata-se directory initially but now we are not using this backup configuration.toml file.
which resulting in two files under /etc/kata-containers/kata-se directory
```
ls /etc/kata-containers/kata-se
configuration.toml configuration.toml.backup-1767604472
```

Hence while uninstalling since the rpm structure doesn't include this backup file `configuration.toml.backup-1767604472`
```
rpm -ql kata-containers.s390x  
/etc/crio/crio.conf.d/50-kata
/etc/crio/crio.conf.d/50-kata-nvidia-gpu
.....
/etc/kata-containers/kata-se
/etc/kata-containers/kata-se/config.d
/etc/kata-containers/kata-se/configuration.toml
/etc/kata-containers/kata-snp
```
the directory `/etc/kata-containers/kata-se` is not removed during uninstall. As a result, the directory is left behind even though it serves no purpose after uninstall.

Hence I am removed the code which does backup of the configuration.toml file as we are not relying on this backup configuration.toml now.

gkurz (approved):
The kata RPM does provide the appropriate configuration for IBM SE already. No need for OSC to do anything in this area indeed.

/lgtm

Thanks @chathuryaadapa !

PR #1488, opened 2026-01-02
Introduce the Prowjob analyzer tool

wainersm:
**- Description of the problem which is fixed/What is the use case**

We are almost done with the QE testing migration from Jenkins so that most of our downstream CI job are now running on Prow. Also, there is already integration with Konflux to automatic trigger CI when new builds of the OSC (test) catalog happens.  The developers can also trigger jobs either via clusterbot or opening pull request to openshift/release repository. Thus, respite some needed improvements here and there (increase coverage included), triggering Prow jobs is no longer a big issue.

Understanding the result of a job is still a concern because our team still lacks of knowledge on Prow as well as QE's automated tests.

Let's have https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-sandboxed-containers-operator-devel-downstream-candidate-azure-ipi-kata/1998028866194509824 as an example. That job is marked as "passed" but in reality two tests failed...there are some nuances in our job implementation that explain that behavior, but a first look it seems confusing, huh? One of the tests that failed mention the version of the operator that doesn't match the expect, but what's exactly the build tested? Moreover, what kind of workload (kata, peerpods, coco) is tested? What's the platform? In summary, there are many question that could be only answered by an experienced developer on the Prow logs, workflow, and internals.

This PR introduces a new tool to help on the analysis of Prow jobs, aiming to provide detailed but high-level and use-friendly information as well as an initial assessment (and debug tips) on any fails that occurred.

**- What I did**

The tool was implemented as a slash-command for Claude (`/prowjob-analyze`) where the user passes the URL for the Prow job she/he wants analyzed.

It was created scripts and auxiliary libraries in `scripts/prowjob-analyzer`. These scripts are used to grab Prow log files and parse them, to extract the relevant information for analysis. Notice that they can be executed as stand-alone tools, if one cannot rely on Claude's analysis or be used by a different tool (reuse). Check out `scripts/prowjob-analyzer/README.md` for details about that part of the tool.

Then it was created a custom Claude command in `.claude/commands/prowjob-analyze.md` which in reality is just a prompt. That contains the algorithm and contexts to carry out the analysis. The scripts in `scripts/prowjob-analyzer` are used just for fetching information in a semi-structured way. 

**- How to verify it**

1. Open Claude in the repository's root directory
2. Send the message `/prowjob-analyze <Job URL>` where `<Job URL>` is the URL for the job's results

[...]

PR #1607, opened 2026-01-22
GPT-recommended fixes to prowjob analyzer

c3d:

I ran a review of the prowjob analyzer using GPT5.1 and GPT5.2 in Cursor. It came up with a number of issues, see end of this description.

**- What I did**

Asked GPT5.2-codex to review the GPT5.1 findings and confirm them, then asked GPT5.2-codex to apply the changes one by one, and quickly reviewed each step, which looked sane t me.

**- How to verify it**

Check that this does not disturb the existing workflow

Apply recommended improvements from GPT5.1-codex-mini and GPT5.2-codex reviews.

**- Depends on**

This PR depends on #1488 

**- Review details follows**
# GPT Code Review Report

## Overview
- Reviewed the new `.claude` slash command along with the Python-based `scripts/prowjob-analyzer` tooling added relative to `devel`.
- Focused on dependency coverage, artifact handling resilience, and failure-path behavior across the analyzer plus the supplemental `failed_tests_report.py`.

## Findings
1. **Undocumented dependency**: `scripts/prowjob-analyzer` and `failed_tests_report.py` both depend on `pyyaml`, yet there is no mention of that requirement in any README, docs, or dependency manifest. Users running the CLI without installing it will hit silent parse failures.
2. **URL parsing is brittle**: `lib/fetcher.parse_prow_url()` assumes the raw URL matches the regex exactly, so legitimate prow URLs carrying query strings (`?foo=bar`) fail validation. Normalizing (`url.split('?')[0]`) or allowing optional query parts would avoid misreported “Invalid Prow URL” errors.
3. **Missing-artifact handling is inconsistent**: When `test-results.yaml` isn’t available, `analyze_prowjob()` only logs a warning but keeps going; downstream helpers (`get_job_status`, `analyze_failure`, report generation) assume parsed test data exists. This can surface as confusing failures later rather than the actual artifact absence.
4. **Repeated artifact listings**: `failure_analyzer.identify_failure_location()` invokes `get_failed_steps()` each time it runs, which fetches the same step directory listing repeatedly during a single command. Caching that listing per job/variant would avoid redundant HTTP requests and speed up failure analysis.
5. **Failure analyzer reporting hides missing dependencies**: When `pyyaml` is absent, `parser.parse_test_results()` already logs “pyyaml not available,” but `failed_tests_report.py` then reports “Could not parse test-results.yaml” before exiting. A clearer exit message (or enforcing the dependency up front) would spare hours of debugging for a missing package.

## Questions
- Should the analyzer be accompanied by automated tests (e.g., sample job data + expected report) to guard future refactors?
- Is this tooling expected to run only in environments where the dependency stack is pre-installed, or should the repo ship a `requirements.txt` / doc updates that pin `pyyaml` (and any other third-party libs)?

## Recommended Next Steps
1. Document (and optionally bundle) the Python dependencies (`pyyaml`, etc.) so the analyzer fails fast if they’re missing.
2. Normalize prow job URLs before running the regex matcher and guard downstream logic from `None` test-result objects.
3. Cache artifact listings when multiple failure detection helpers run to reduce redundant network calls and keep analysis snappy.
 
# GPT-5.2 Code Review Report

## Overview
- Reviewed the new `scripts/prowjob-analyzer` tooling and the `.claude` slash command additions relative to `devel`.
- Focused on job status determination, failure classification, and how test-results formats are handled in reporting.

## Findings
1. **High — failingScenarios/errors can be reported as success**
   - `get_job_status()` only checks `failures`, so jobs using `failingScenarios` or `errors` can be marked as `success` and exit with code 0 even when tests failed.
   - The report also gates the failed test list on `failures > 0`, so real failures can be hidden in the human output.
   - Fix by deriving failure count from `errors + failures` and/or `extract_failing_tests()` so alternate formats are treated as failures.
   - Evidence:
```
108:143:scripts/prowjob-analyzer/lib/parser.py
    if state == 'success':
        # Double-check with test results if available
        if test_results and test_results.get('failures', 0) > 0:
            return 'failure'
        return 'success'
...
        if test_results:
            if test_results.get('failures', 0) == 0:
                return 'success'
            else:
                return 'failure'
```
```
31:78:scripts/prowjob-analyzer/lib/report_generator.py
    failures = test_results.get('failures', 0)
    passed = total - failures
    skipped = test_results.get('skipped', 0)
...
    if failures > 0:
        failing_tests_by_cat = failure_analysis.get('failing_tests_by_category', {})
```

2. **Medium — aborted jobs are labeled as timeouts**
   - `identify_failure_location()` returns `timeout` when state is `aborted`, which conflates distinct outcomes and


scripts/prowjob-analyzer/lib/parser.py
```diff
@@ -0,0 +1,368 @@
...
+                    # This is the nested structure, unwrap it
+                    return first_value
+
+        return data
+    except (yaml.YAMLError, UnicodeDecodeError) as e:
+        logger.error(f"Failed to parse test-results.yaml: {e}")
+        return None
+
+
+def get_test_failures_count(test_results: Optional[Dict]) -> int:
```
tbuskey, line 115: We should try to keep the prow step failures separate from test failures during the prow test step.

If openshift-extended-test has `failures: 0`, the test is a SUCCESS.  The pre steps ran correctly to build a cluster & configure it for the test step.  The test step finished because the file was created.  Having the test step finish successfully does not mean it passed any tests.

From a test perspective, the post steps don't matter.

In another PR (soon) I'd like to label test failures as SUCCESS/ERROR and steps as pass/fail so its obvious they are different.

scripts/prowjob-analyzer/lib/report_generator.py
```diff
@@ -0,0 +1,341 @@
...
+    }
+    return status_emojis.get(status.lower(), '❓')
+
+
+def generate_test_results_section(test_results: Dict, failure_analysis: Dict) -> str:
+    """Generate test results section for report."""
+    if not test_results:
+        return "\n## Test Results\n\nNo test results available.\n"
+
+    # Handle different test results formats
```
tbuskey, line 36: We want to keep step failures separate from failures during the test.

tbuskey (changes requested):
We want to keep step errors separate from failures in the tests.

steps failures can result from infrastructure or setup issues.

Right now, installing the operator/kataconfig is only in the test step.  We will be moving that  to a step eventually.

Test failures do not cause the test step to fail.  If openshift-extended-test doesn't timeout or get aborted, the step finishes and records the results.  That has nothing to do w/ test success/error.  They could all fail, but the step finishes.

PR #1608, opened 2026-01-22
tests/osc-test-fbc-integration: optimize the execution of prowjobs

wainersm:
Currently we are running all the prowjobs in a single phase.

In order to save resources as well as avoid the azure quota limit problem, I refactor the pipeline to run the jobs in phases and in groups.

### DIFF
--- a/tests/osc-test-fbc-integration.yaml
+++ b/tests/osc-test-fbc-integration.yaml
@@ -14,6 +14,13 @@ spec:
       name: SNAPSHOT
       default: '{"components": [{"name":"osc-test-fbc", "containerImage": "quay.io/redhat-user-workloads/ose-osc-tenant/osc-test-fbc:latest"}]}'
       type: string
+    - name: RUN_PROWJOBS
+      type: string
+      description: >-
+        Which prowjobs to run. Options: "sanity" (default, runs only sanity checks),
+        "all" (runs all prowjobs), or specific job names like "ocp419 ocp420"
+        (runs only those OCP version jobs, space-separated). Available job names: ocp418, ocp419, ocp420
+      default: "sanity"
   tasks:
     - name: get-catalog-image
       params:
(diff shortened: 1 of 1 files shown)


littlejawa (approved):
/lgtm

Thanks @wainersm !

ldoktor (approved):
lgtm, one thing to consider is to reorder the tests so they are ordered as executed. Also you might consider executing only part of the configurations and leave the thorough testing for weekly/monthly/prerelease testing (I mean run one with kata/peer/coco and the other ocp versions just one each)

wainersm:
Hi @ldoktor !

> lgtm, one thing to consider is to reorder the tests so they are ordered as executed. Also you might consider executing only part of the configurations and leave the thorough testing for weekly/monthly/prerelease testing (I mean run one with kata/peer/coco and the other ocp versions just one each)

Based in your comment above I added a new commit that introduces a parameter called `RUN_PROWJOBS` where we will be able to better select groups of jobs. This is how I see us using it:

- early in development cycle we run the sanity checks only, that include one job for kata and peerpods each in the latest supported OCP version
- if we need to test on specific ocp versions, we just enable it
- as we approach the end of development cycle, we should enable to trigger all job. IMPORTANT: the trigger "all" will still run the jobs in group.

ps: there is one situation that I need to test (actually these changes aren't tested yet...) which is when user selects only given ocp versions. For example, user select ocp417 but that version "run-after" ocp418...I don't know if tekton will assume the ocp418 ran even though it actually just skipped.

wainersm:
Yesterday I made a couple of tests on that PR and had to fix some stuffs. By default, it will run only the sanity jobs (two jobs) like below:

image

Now I will be testing if the jobs filter works. For example, I want to specify that only the OCP 4.18 jobs will run. I will post the result here.

wainersm:
There is one thing that I'm not totally happy with konflux UI that might even be a bug.

When the job is running like in the screenshot below, you gonna see two entries for the prowjob-sanity and that's expected because it actually run a matrix of tasks. Now, compare with the screenshot on the previous comment, where you see only one prowjob-sanity entry. After the pipeline finished it collapsed all the entries into a single prowjob-sanity, and I don't know how to access individual tasks anymore.

image

wainersm:
Here is an execution where I specify only a subset of jobs I want to executed, in this case OCP 4.18 ones. I set the parameter on konflux's configuration as:

image

Then I only got OCP 4.18 ones running:

image

It's working!

PR #1609, opened 2026-01-23
daemonset deployment: fix baremetal installation

Pacho20:

The newly introduced rpm-extensions function, added to resolve installation issues on IBM Cloud, inadvertently broke installation on baremetal. The get_cloud_provider function expected the peer-pods ConfigMap, which is not used on baremetal.

**- What I did**

This update changes the behavior so that get_cloud_provider returns an empty value instead of an error when the peer-pods ConfigMap is not present, restoring correct behavior for baremetal deployments.

**- How to verify it**

Build a new version of the daemonset image and use that on baremetal with DaemonSet mode. 

daemonset deployment: fix baremetal installation

### DIFF
--- a/scripts/kata-install/osc-kata-install.sh
+++ b/scripts/kata-install/osc-kata-install.sh
@@ -194,7 +194,6 @@ get_cloud_provider() {
 	if ! provider="$(kubectl get configmap/peer-pods-cm \
 		-n openshift-sandboxed-containers-operator \
 		-o jsonpath='{.data.CLOUD_PROVIDER}')"; then
-		error "get_cloud_provider: kubectl failed to query peer-pods-cm"
 		return
 	fi
 


scripts/kata-install/osc-kata-install.sh
```diff
@@ -194,7 +194,7 @@ get_cloud_provider() {
 	if ! provider="$(kubectl get configmap/peer-pods-cm \
 		-n openshift-sandboxed-containers-operator \
 		-o jsonpath='{.data.CLOUD_PROVIDER}')"; then
-		error "get_cloud_provider: kubectl failed to query peer-pods-cm"
+		echo "get_cloud_provider: kubectl failed to query peer-pods-cm, assuming bare-metal" >&2
```
gkurz: I would drop that line. No `peer-pods-cm` simply indicates that peer pods aren't enabled.

ajaypvictor (approved):
Tested this fix on baremetal flow and looks to be all good.
Thanks..

scripts/kata-install/osc-kata-install.sh
Pacho20: Alright.

gkurz (approved):
/lgtm

Thanks for the quick fix @Pacho20 !

PR #1648, opened 2026-02-04
mintmaker: change schedule config

littlejawa:
We don't need to have Mintmaker run every 4 hours.
This commit changes the schedule to run it only once per week, on weekends.

See:
- https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#scheduling
- https://docs.renovatebot.com/key-concepts/scheduling/

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,5 +1,6 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "schedule": ["* * * * 0,6"],
     "extends": [
         "https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json?raw=true"
     ],


renovate.json
```diff
@@ -1,5 +1,6 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "schedule": ["* * * * 0,6"],
```
gkurz, line 3: If this were a cron job, this syntax would cause it to run :
- every minute (maybe not for renovate that doesn't support minute granularity)
- every hours
- every day of month
- every month
- on Sundays and Saturdays

Not sure what renovate does with this syntax but it looks like something will run all over the WE. Is this what we want ?

littlejawa: Based on the links I provided, I think Mintmaker actually doesn't apply it as is.
From: https://konflux.pages.redhat.com/docs/users/mintmaker/user.html#scheduling
> MintMaker has a base schedule which runs every 4 hours, starting at 00:00 UTC (i.e. 04:00, 08:00, 12:00, …​). While you can configure some aspects of your renovate schedule (see below), you cannot influence MintMaker to run any more frequently than this base schedule.

So I suspect it still runs every 4 hours, and uses our "schedule" parameter to decide whether or not it should proceed with our repo.

The syntax `[ * * * * 0,6]` is actually used as example in the above link, and also found here as "every weekend" schedule:
https://docs.renovatebot.com/key-concepts/scheduling/#recommended-cron-syntax

gkurz: Thanks for the clarification !

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #1654, opened 2026-02-04
local builds: add the files needed to build a hermeto-based image

littlejawa:
In order to troubleshoot build issues in our CI, we need to have a way to build our images in a process that is as close as possible to how Konflux builds them. This requires building hermetically during local builds. Now Hermeto doesn't support subscription by default, it needs to be added to its workflow. Also, we don't want to have the same steps repeated in the Makefiles of all our repositories.

This commit brings a very simple customization to a base Hermeto image. The resulting image will be published to our quay.io repository, and used in the Makefiles of our container images, to allow easier local hermetic builds.

NOTE: this PR extracts the hermeto-specific part from https://github.com/openshift/cloud-api-adaptor/pull/317

### DIFF
--- a/scripts/hermeto-image/README.md
+++ b/scripts/hermeto-image/README.md
@@ -1,0 +1,9 @@
+This folder contains the files necessary to build a custom version of the hermeto
+container.
+The goal is to enable subscription as part of the hermeto run, so that RPMs
+coming from RHEL can be fetched.
+See entrypoint.sh for further details on how to use the resulting container.
+
+The container is published to quay.io/repository/redhat-user-workloads/ose-osc-tenant/osc-hermeto
+and is used by Makefiles in our different repositories to make local hermetic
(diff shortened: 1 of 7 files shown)


littlejawa:
/cc @gkurz @c3d 

FYI: I can't push the resulting image to quay.io/openshift_sandboxed_containers, but I was able to test it locally anyway.

littlejawa:
Cherry-picked the Konflux pipeline creation from #1655

scripts/hermeto-image/README.md
```diff
@@ -0,0 +1,9 @@
+This folder contains the files necessary to build a custom version of the hermetp
```
gkurz: s/hermetp/hermeto

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #1663, opened 2026-02-05
changelog.sh: simplify the process, trying to get some logs in more situations

littlejawa:
Trying to filter out changes that did not impact the bundle proved counter productive: we filtered too much, to the point of getting nothing. Typically: if we get a catalog from a change in the CI test script, this will generate no CHANGELOG, because one of the boundaries of our commit list does not impact the bundle.

This commit removes that filter, looking for changes to the CSV file, without looking for bundle image's commits first.
Since both FBC and bundle are in the same repo, the result *should* be similar, and we will avoid issues. It simplifies the process too, but we will need to monitor the resulting CHANGELOG, to make sure we don't get too many/repetitive entries.

### DIFF
--- a/fbc/test-fbc/changelog.sh
+++ b/fbc/test-fbc/changelog.sh
@@ -18,7 +18,7 @@
 # be included in the built image.
 #
 
-#set -x
+set -x
 
 # Takes an image reference with digest, and looks at the quay.io repository
 # for all the tags that point to the same image digest.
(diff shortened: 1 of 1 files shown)


dbkreling (approved):

gkurz (approved):
Definitely simpler !

/lgtm

Thanks @littlejawa !

PR #1665, opened 2026-02-05
Add --csv output to prowjob analyzer script

tbuskey:
[KATA-4656](https://issues.redhat.com/browse/KATA-4656) Add option to prowjob analyzer script save results to a spreadsheet

Add --csv output to script for import into a spreadsheet

- Consolidate complete output data into json
  - Human and csv are subsets
- Change csv output to be closer to record-metadata in prow
- A sha256 catalog will now get its build time from quay to match  record-metadata
- Adding an initial RCA

To add a Konflux run, download all task logs to `task.log`
```
 for i in $(grep https://prow task.log)
do
  scripts/prowjob-analyzer/analyze.py --csv --no-header $i
done > import.csv
```

In your spreadsheet, go to the line where you want the data and import.
Replace the data at that line and choose comma deliminated

### DIFF
--- a/scripts/prowjob-analyzer/analyze.py
+++ b/scripts/prowjob-analyzer/analyze.py
@@ -28,7 +28,7 @@ from lib.parser import (
 )
 from lib.metadata_extractor import extract_metadata
 from lib.failure_analyzer import analyze_failure
-from lib.report_generator import generate_human_report, generate_json_report
+from lib.report_generator import generate_human_report, generate_json_report, generate_csv_report
 
 # Configure logging
 logger = logging.getLogger(__name__)
@@ -138,6 +138,9 @@ Examples:
 
   %(prog)s --json <URL> > report.json
 
+  %(prog)s --csv <URL> >> jobs.csv
+  %(prog)s --csv --no-header <URL>
+
   %(prog)s --verbose --no-wait <URL>
         '''
     )
@@ -151,6 +154,18 @@ Examples:
         '--json',
         action='store_true',
         help='Output machine-readable JSON format'
+    )
+
+    parser.add_argument(
+        '--csv',
+        action='store_true',
+        help='Output CSV format (one row: start_time, catalog_full_tag, catalog_build_date, provider, ocp_version, prowjob url, workload_type, kata_rpm_version, status, trigger, root_cause)'
+    )
+
+    parser.add_argument(
+        '--no-header',
+        action='store_true',
+        help='Omit CSV header row (only applies with --csv)'
     )
 
     parser.add_argument(
@@ -199,6 +214,17 @@ Examples:
             results['failure_analysis'],
             results['base_url']
         )
+    elif args.csv:
+        logger.info("Generating CSV report...")
+        report = generate_csv_report(
+            results['prowjob_data'],
+            results['metadata'],
+            results['status'],
+            results['test_results'],
+            results['failure_analysis'],
+            results['base_url'],
+            header=not args.no_header,
+        )
     else:
         logger.info("Generating human-readable report...")
         report = generate_human_report(
@@ -211,7 +237,8 @@ Examples:
         )
 
     # Output report
-    print("\n" + "="*80 + "\n")
+    if not args.csv:
+        print("\n" + "="*80 + "\n")
     print(report)
 
     # Exit code based on job status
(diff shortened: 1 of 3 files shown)


wainersm:
Hi @tbuskey , the change looks good to me. I ran the script --csv and it seemed to work fine :)

tbuskey:
Removed 'long description' boilerplate from commit msg

wainersm:
/lgtm
/approve

Thanks @tbuskey !

ldoktor:
Hello @tbuskey I started the review, thought the `_get_root_cause_for_csv` is unnecessarily complex, wanted to find it in sources and it was gone. Then I noticed a second commit which completely changes the PR. Could you please always squash changes that doesn't add to the process (sometimes the process is necessary but in this case it is not). Also brief look at the second commit uncovered multiple changed things not really tied to the overall change (fetch_build_date_from_quay) which should be a separate commit.

Please format your commits before sending as reviewing mixed commits is hard, time-consuming and can lead to overlooked issues (not talking about broken reverts).

tbuskey:
@ldoktor The 1st stab at the _get_root_cause was too complex also.  I should have squashed it.

The 2nd commit adds things that are needed for the csv.

The json, human readable and csv had different processes and data.  There should be 1 source.  It will always be in the json and the other formats can leave out what is not needed.  The csv output needed other items added to the json for it to be useful.

The csv must have the build date so we can correlate the SHA to when it was built.  The X.Y.Z format has the date in the tag and is close to the build date.  Looking that tag up to precisely match the time isn't worth the effort.  Knowing the date, hour and within 10 minutes is enough to correlate.

A preliminary RCA, if possible, is very useful in the csv.  A test run will be output into a csv.  Then imported into a spreadsheet.  From there, it can be reviewed and edit.  An RCA can eliminate the need to run it through the AI.

ldoktor:
> @ldoktor The 1st stab at the _get_root_cause was too complex also. I should have squashed it.
> 

I understand that and if it was already merged I'd have nothing against removing it. As this process happened while working on this PR it should be squashed or removed. PR is a process and can evolve, the usual way is to force-push or new PR that supersedes the old one.

> The 2nd commit adds things that are needed for the csv.

Great, then each new addition should be a separate commit (sometimes multiple ones can be bundled if they are connected, eg. adding multiple variables at the same time...

> 
> The json, human readable and csv had different processes and data. There should be 1 source. It will always be in the json and the other formats can leave out what is not needed. The csv output needed other items added to the json for it to be useful.

And that was something I wanted to comment on the first commit. This is definitely the way to go.

> 
> The csv must have the build date so we can correlate the SHA to when it was built. The X.Y.Z format has the date in the tag and is close to the build date. Looking that tag up to precisely match the time isn't worth the effort. Knowing the date, hour and within 10 minutes is enough to correlate.

Understandable, please extract it into a separate commit.

> 
> A preliminary RCA, if possible, is very useful in the csv. A test run will be output into a csv. Then imported into a spreadsheet. From there, it can be reviewed and edit. An RCA can eliminate the need to run it through the AI.

Also great addition which deserves to be a standalone commit.


What I'm trying to say is I like the addition, I like the individual enhancements, but to be able to review those changes it's important to separate them and remove in-process changes. Remember what Pavel mentioned, he spends more time writing the commit message and polishing the patches then actually developing the feature. I'm often in the same situation, developing something, adding things that are required or just fixing little issues I notice while working on it, but before committing these I always need to stop and say: "OK, what am I changing now" and leave everything unrelated to elsewhere. In such situations I usually use `git cola` (`dnf install git-gui git-cola`), I go through the changes, select the style/comment/... fixes and commit them individually. Then I select the pre-requisites (like the build-date) and again include them one by one along with a proper comment so the reviewer reads what are you changing, then reviews the patches and can say whether you are really changing what you are changing. In the end there should only be the remaining big chunk that implements the feature I started working on in the first place. And even then it might happen that I need several commits to implement it and splitting it is often tricky. But it always pays-off, because:

1. I might (and often do) notice corner-cases or little issues as going through the changes
2. I prepare the patches so the reviewer can thoroughly review (and not just blindly accept or spend ages trying to understand why are you doing this change when it doesn't makes much sense or trying to improve something that you rework in the next commit). Remember there is one to submit patches and several to review the patches. Time of all is important, but there are multiple reviewers that multiplies the necessity to have good commits. Also note that the developer is not responsible for the code that goes in, it's the reviewer!
3. changes are atomic, testable and revertable (which is crucial as there might be hidden problems)

So please take a look at your changes, use `git rebase` or even some gui to help you separate the patches and force-push a polished change set.

tbuskey:
Replacing with [PR 1701]( https://github.com/openshift/sandboxed-containers-operator/pull/1701)

PR #1666, opened 2026-02-06
prowjob-analyzer: enhance the analysis when most/all tests fail

wainersm:
The commits are self-explanatory but let me try to summarize the changes:

```
    This helps the AI avoid being misled by error frequency and correctly
    identify the root cause in the first test's logs rather than focusing
    on cascading errors that appear more frequently in subsequent tests.
```

We had recently jobs for OCP 4.17 and 4.18 failing because of a bug on the Kata 3.25 rpm, for example https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-sandboxed-containers-operator-devel-downstream-candidate417-azure-ipi-peerpods/2019354177586597888. The prowjob-analyzer was able detect the job failed to install the RPM but could not derive the correct problem cause. This happened because of a characteristic of our test framework (every and each test will try to setup the test environment, unless it was already setup'ed by a previous test) misled the analysis of AI.

In this particular case of rpm install failure, the first test failed to install the rpm due dependency (qemu) not found. Subsequent tests tried to install the rpm likewise, however, they failed to actually make the filesystem read-write because the first did that already! With so many errors, all over the tests, related with filesystem errors, AI thought that was the cause of the problem (and not the lack of a given version of qemu).

Here I'm trying to teach AI how to make the analysis in a similar situation.

### DIFF
--- a/.claude/commands/prowjob-analyze.md
+++ b/.claude/commands/prowjob-analyze.md
@@ -29,9 +29,12 @@ python3 scripts/prowjob-analyzer/analyze.py --no-wait "$@"
 **Case A: Tests failed** (Failed Step is `openshift-extended-test` AND Failed Tests are listed)
 - This means the job ran through infrastructure setup and failed during test execution
 - **IMPORTANT**: Check the failure count:
-  - If **>90% of tests failed** (e.g., 10+ tests failed): Run analysis on **ONLY the FIRST test**
+  - If **>90% of tests failed** (e.g., 10+ tests failed):
+    - Each test in the "Failed Tests" section shows metadata including "Execution Order: N"
+    - Find the test with "Execution Order: 1" - this is the FIRST test that executed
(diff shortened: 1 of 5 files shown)


tbuskey:
/lgtm

If the 1st test is skipped (or successful), there will be no log output unless `--include-success` is used
We should look into changing the prow test section to have that option

wainersm:
Hi @tbuskey ,

> /lgtm
> 
> If the 1st test is skipped (or successful), there will be no log output unless `--include-success` is used We should look into changing the prow test section to have that option

You are absolutely right, and I overlooked that. Note that the prompt is instruct to look at the first test when > 90% of tests fail and in this scenario the first test will have a high probability of fail (i.e. unless it skips). The analysis is still not 100% accurate, yes, but if you don't mind I would like to tackle that later.

PR #1672, opened 2026-02-09
fbc: remove 1.11.2 entries

littlejawa:
OSC 1.11.2 was never released, we need to remove that from our catalog entries.

This reverts the following commits:
95138fa04e16ea2665c7b7ab481d7a1d0c0a0514
29dee912534e5dfd15cabe924a958bcb8b8d5d64
ae108e68351f4727017c66cd720f3341fe8aa4d2
454e56ba99b64a1db6054b40fe0498b35265ca83


gkurz (approved):
And yes... we need to revert that indeed for the public FBCs.

/lgtm

Thanks @littlejawa !

PR #1701, opened 2026-02-16
Add option to prowjob analyzer script save results to a spreadsheet

tbuskey:
[KATA-4656](https://issues.redhat.com/browse/KATA-4656) Add option to prowjob analyzer script save results to a spreadsheet

This replaces [PR 1665](https://github.com/openshift/sandboxed-containers-operator/pull/1665).  It seperates the commits better.

- Consolidate complete output data into json
  - Human and csv are subsets
- A sha256 catalog will now get its build time from quay
- Adding an initial RCA
  - Include quota issue
  - Infrastructure issues
- Add --csv output to script for import into a spreadsheet

To add a Konflux run, download all task logs to task.log

```
for i in $(grep https://prow task.log)
do
  scripts/prowjob-analyzer/analyze.py --csv --no-header $i
done > import.csv
```

In your spreadsheet, go to the line where you want the data
_File -> import_
Upload import.csv
Import location: _Replace the data at selected cell_
_Import data_


scripts/prowjob-analyzer/lib/report_generator.py
```diff
@@ -244,28 +331,17 @@ def generate_human_report(
 
     if metadata.get('catalog_source_image'):
         catalog = metadata['catalog_source_image']
-
-        # For digest format (@sha256:...), show only the image name without digest
```
ldoktor, line 248: These comments seem quite useful, any reason to remove them?

scripts/prowjob-analyzer/lib/metadata_extractor.py
```diff
@@ -147,6 +150,63 @@ def extract_ocp_version(prowjob_data: Dict, base_url: str = None) -> str:
...
+        with urllib.request.urlopen(req, timeout=15) as resp:
+            data = json.loads(resp.read().decode('utf-8'))
+    except (urllib.error.URLError, urllib.error.HTTPError, OSError, json.JSONDecodeError) as e:
+        logger.debug(f"Quay API request failed for {repo}: {e}")
+        return 'quay-unavailable'
+
+    tags = data.get('tags') or []
+    for t in tags:
+        manifest_digest = t.get('manifest_digest') or ''
+        if manifest_digest == digest or manifest_digest == digest.replace('sha256:', '', 1):
```
ldoktor: replace is slow. As we are checking it starts with the `sha256:` and replacing it only once it should work well, but better solution would IMO be `digest[7:]`

scripts/prowjob-analyzer/lib/metadata_extractor.py
```diff
@@ -147,6 +150,63 @@ def extract_ocp_version(prowjob_data: Dict, base_url: str = None) -> str:
...
+
+    Returns:
+        ISO-style build date string, or one of:
+        "quay-unavailable" (network/API error),
+        "quay-not-found" (digest not in tag list),
+        "unknown" (not a quay digest image or generic failure).
+    """
+    if not catalog_image or '@' not in catalog_image:
+        return 'unknown'
+    tag_part = catalog_image.split('@')[-1]
```
ldoktor, line 166: why not to use `digest` here in the first place? (you are re-naming it to it on line 177)

scripts/prowjob-analyzer/lib/metadata_extractor.py
```diff
@@ -157,16 +217,17 @@ def parse_catalog_tag(catalog_image: str) -> Dict:
     - plain version: 1.11.1
     - latest: latest
     - git SHA: abc123def456
-    - sha256 digest: sha256:41ca9598b816ddc784de8f345e7e586c2630747f2d10c1a45ad5e082ff228a1f
+    - sha256 digest: sha256:... (build_date from Quay API when quay.io)
+
+    Problem values for build_date: "not set", "invalid-timestamp", "quay-unavailable",
```
ldoktor, line 221: The `invalid-timestamp` is never set anywhere (this code sets `unknown` instead, IMO it'd be better to print the received value to get at least something even if not in the expected format).

How about using `error: XXX` notation for all errors, then we could use `startswith("error")` to get the failure and pass additional details in the error.

scripts/prowjob-analyzer/lib/metadata_extractor.py
```diff
@@ -175,22 +236,20 @@ def parse_catalog_tag(catalog_image: str) -> Dict:
...
-
-        # Extract hash without algorithm prefix (sha256:)
         if ':' in tag:
             algorithm, hash_value = tag.split(':', 1)
             catalog_info['base_version'] = hash_value
         else:
             catalog_info['base_version'] = tag
-
-        catalog_info['build_date'] = 'unknown'
+        catalog_info['build_date'] = fetch_build_date_from_quay(catalog_image)
```
ldoktor: As we are then merging this with the `build_info`, shouldn't we use more specific name then `build_date`? (eg. `catalog_build_date`)

scripts/prowjob-analyzer/lib/failure_analyzer.py
```diff
@@ -340,6 +358,16 @@ def find_pattern_source(pattern_name: str) -> Optional[str]:
...
             root_cause['note'] = 'Multiple "Deployment is already in unlocked state" errors are cascading failures from the initial RPM dependency error'
+    else:
+        # Unknown pattern type: use first detected pattern
+        if detected_patterns:
+            primary = detected_patterns[0]
+            pattern_name = primary.get('pattern', 'unknown') if isinstance(primary, dict) else 'unknown'
+            root_cause['primary_pattern'] = pattern_name
+            root_cause['pattern_source'] = primary.get('source') if isinstance(primary, dict) else None
+            root_cause['likely_cause'] = f"Failure pattern: {pattern_name}"
+            root_cause['confidence'] = 'medium'
```
ldoktor, line 369: This should only happen when the pattern is in the `FAILURE_PATTERNS` but not here and generally should mean a bug in this code. I don't think we should use `medum` confidence here.

TBH I don't much like this whole function with reverse mapping, I believe it can be performed in more organized way (without reverse mapping and 2 places where we define attributes) but that wouldn't be related to this commit so take it just for your consideration and are for improvement.

tests/osc-test-fbc-integration.yaml
```diff
@@ -59,7 +59,7 @@ spec:
         - name: VARIANT
           value: downstream-candidate417
         - name: ENVS
-          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE),EXPECTED_OPERATOR_VERSION=1.11.2,FORCE_SUCCESS_EXIT=no,INSTALL_KATA_RPM=true,KATA_RPM_VERSION=3.25.0-1.rhaos4.17.el9"
```
wainersm: Hi @tbuskey , could you send this change to `tests/osc-test-fbc-integration.yaml` in a separated PR?

tbuskey: I did it before you commented 
[PR 1759](https://github.com/openshift/sandboxed-containers-operator/pull/1759 )

scripts/prowjob-analyzer/README.md
```diff
@@ -72,9 +84,44 @@ The analyzer supports both job URL patterns:
...
+for URL in $URLS
+do
+    echo $URL
+    ./dig.py $URL --csv --no-header --no-wait >> <output.csv>
+done
+dos2unix <output.csv> # to conver to Unix EOL
+grep -v <output.csv> > import.csv
+```
+
+We've been using [OSC Testing spreadsheet](https://docs.google.com/spreadsheets/d/13k-rIFap5BTLqDcYfbgMH37hAWJI6hf9Af9LHcA3haw/edit?gid=1173858770#gid=1173858770) to gather results.
```
ldoktor: This document is not publicly available so it doesn't belong to upstream documentation

ldoktor (commented):
Thanks @tbuskey, looks almost good to go. The biggest problem IMO is the google docs link and the "PR issue cleanup" commit should be extracted and put into the individual commits rather than adding as a separate commit. Last but not least are the commit messages which are quite cryptic to me. Not sure if it's just me, but I rarely understand what, why and how that commit changes and I only guess that based on the commit content. It should be the other way around, the commit message is the first thing reviewer reads and a good commit message very much simplifies review process as one knows what to focus on. With those commits I usually need to read it multiple times, then go through the change, once I understand what and why it changed I need to go through it again to focus on things related to that kind of commit.

tbuskey:
@ldoktor I edited the commit message but I did not separate them

wainersm:
Hi @ldoktor , I'm going to merge this PR as @tbuskey has used it daily, and keep working on branches is hard for him. While the commits are still a little mess, he addressed all your other suggestions I think. On next PR we will do better! :)

PR #1707, opened 2026-02-19
make-tests: Set CPU and memory resources

gkurz:
Konflux workers have 32 CPU, the task has no CPU limit and requests 100m. The go compiler ends up thinking it has a lot more CPU than what it gets in practice. It spawns 32 compile threads that quickly saturate both CPU and memory available to the task. This causes `go vet` to run for more than 2 hours and the task times out.

Set limits and requests in the task spec and tell go about the maximum number of CPU it can use.


gkurz:
I'm going to merge this right away to unblock pending PRs.

PR #1708, opened 2026-02-19
Add 'openshift-kube-proxy' to the list

gcoon151:
It's a long story how we found this. IBM Cloud uses calico CNI. Even after configuring calico exclusions the OSC peer pod webhook was causing problems because openshift-kube-proxy wasn't excluded.

The peer pod webhook can cause havoc in alternate CNIs causing them to not start correctly if the underlying app isn't responding. Even after adding exclusions to the webhook for the obvious namespaces our CNI used, this one eluded us. 

**- What I did**
added an exclusion for an additional openshift namespace commonly used on startup resources

**- How to verify it**
Create a CNI that depends on notifying the peer pod webhook to start up but there is no peer pod app to respond to it thereby hanging the CNI from starting. 

Add openshift-kube-proxy to exclude peer pod webhook namespaces

### DIFF
--- a/controllers/ppwebhook.go
+++ b/controllers/ppwebhook.go
@@ -309,6 +309,7 @@ func (r *KataConfigOpenShiftReconciler) createMutatingWebhookConfig() error {
 		"openshift-kube-apiserver",
 		"openshift-kube-apiserver-operator",
 		"openshift-kube-controller-manager",
+		"openshift-kube-proxy",
 		"openshift-kube-scheduler",
 		"openshift-kube-scheduler-operator",
 		"openshift-kube-storage-version-migrator",


bradbehle:
I can confirm, this should definitely be excluding the openshift-kube-proxy namespace since those are very necessary pods when anything other than OVN is used as the CNI for the cluster.  It would also be good to exclude `tigera-operator`, `calico-system`, and `calico-apiserver` namespaces as well (either always, or when Calico is detected as the CNI), but this PR is a necessary first step.

gkurz:
Hi @gcoon151 ! Thanks for this fix. Please add you Signed-off-by to the commit BTW :wink:

gcoon151:
commit message sign-off fixed.

gkurz (approved):
/lgtm

Thanks @gcoon151 !

PR #1729, opened 2026-02-23
Bump osc 1.12

gkurz:
Let's bump OSC to 1.12 now for clarity.

OSC version bump script is improved a bit on the way.


dbkreling:
We also have versions entries to bump in the `fbc` directory, but those are for later in the release.
Also, the `scripts/prowjob-analyzer/lib/metadata_extractor.py` contains comments stating `1.11` that do not interfere with the code, they're only examples about how to use the version level in the code. If we want, we can change those in this commit as well, but not necessary.

As I don't think those changes are necessary for this PR, this one LGTM.

dbkreling (approved):

gkurz:
Ignoring the EC check failure for 4.21 at this stage.

PR #1738, opened 2026-02-25
feat(runtimeclass): add NVIDIA GPU support for kata runtime classes

thejasn:
Closes: [KATA-4613](https://issues.redhat.com/browse/KATA-4613) and [KATA-4614](https://issues.redhat.com/browse/KATA-4614)

TBD

**- What I did**
Introduce runtime class support for NVIDIA GPUs in both confidential and non-confidential baremetal deployments
 - Add `kata-nvidia-gpu` runtime class for standard GPU workloads
  - Add `kata-cc-nvidia-gpu` runtime class for confidential GPU workloads (Intel TDX and AMD SNP)
  - Refactor TEE handler detection to only guess TEE platform
  - Update createRuntimeClass to accept label maps for node scheduling
  - Update `createRuntimeClass()` for conditional runtime class creation

**- How to verify it**
TBD

Introduce runtime class support for NVIDIA GPUs in both confidential and non-confidential baremetal deployment

### DIFF
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -164,7 +164,7 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
 		}
 
 		// Create kata-cc runtime class restricted to the detected TEE subset
-		err = r.createRuntimeClass(kataCCRuntimeClassName, kataCCRuntimeClassCpuOverhead, kataCCRuntimeClassMemOverhead, kataCCRuntimeClassExtResOverhead, handler, nodeLabel)
+		err = r.createRuntimeClass(kataCCRuntimeClassName, kataCCRuntimeClassCpuOverhead, kataCCRuntimeClassMemOverhead, kataCCRuntimeClassExtResOverhead, handler, map[string]string{nodeLabel: "true"})
 		if err != nil {
 			r.Log.Info("Error creating "+kataCCRuntimeClassName+" runtime class", "err", err)
 			return fmt.Errorf("Error creating "+kataCCRuntimeClassName+" runtime class: %w", err)
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"maps"
 
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -10,6 +11,9 @@ import (
 )
 
 const (
+
+	// CPU
+
 	// kata-cc runtime class for CoCo BM
 	kataCCRuntimeClassName        = "kata-cc"
 	kataCCRuntimeClassCpuOverhead = "0.25"
@@ -29,10 +33,40 @@ const (
 	intelTDXExtendedResource = "tdx.intel.com/keys"
 	amdSNPExtendedResource   = "sev-snp.amd.com/esids"
 
+	// GPU
+
+	// kata-cc-nvidia-gpu runtime class for CoCo BM GPU
+	kataNvidiaGPUCCRuntimeClassName        = "kata-cc-nvidia-gpu"
+	kataNvidiaGPUCCRuntimeClassCpuOverhead = "0.25"
+	kataNvidiaGPUCCRuntimeClassMemOverhead = "350Mi"
+
+	// RuntimeClass handlers for TEE
+	kataNvidiaGPUCCIntelHandler = "kata-tdx-nvidia-gpu"
+	kataNvidiaGPUCCAmdHandler   = "kata-snp-nvidia-gpu"
+
 	// INITDATA value for non-confidential peer pods, this is required in order
 	// to override the default restrictive CoCo agent policy
 	// created from sourced plaintxt: cat config/peerpods/default-non-cc-initdata.toml | gzip | base64 -w0
 	defaultNonCCInitdata = "H4sIAAAAAAAAA42UwW7bMAyG734KwT3ktKDDehgG9NAl2VZgWQw7bQ7DMDAWYxOVRU+i03pPP7lGhx0muwddyE8U+ZPUhdrX5BVZEg0C6kQG1SN4VToEQa2OvZIaFfvyDbfoQNglYCp2JHWjrlXqa3j3/ipNzug8sR1Ml8u3y8s0Sb4PIX8kKcBSuDFp8C0Wi2Q4SdqyobJfOqz4xdFC+QAVqnCs/ByBJNF4gs6IutH6Js++IVX1kZ3P8VeHXtSHayWuw3+x4hHamHtl2GMhmmyU4Lb/FGSI+p+VWbEVIItuGivA6iM/xaB1MDruZ6jNE5aZ4xJ9tOrPKFsUR+UUsdttN+cgbRQZrGsMdZlomK/k5dYKuhOEfKaonDuJE1tsvrC0pqs+9qG2Y1TunTVB5lV2F27EmAw6P9+RrDPmtgnDFQNyBF1I6Fv0oRwbPs+/NGKFgMF7ckJ88kUNDrfcWYlKkqNH1HmYBW7WeJ7AumY+hwJl7GeYwj010aIDlz1vWhSgyoKZmb9Qq5P5nAZq76AkW00w4l8RiduZQHvpD8OWe/odLf6u1a/Z5RHbtDU24Qs0020c4b87Mo1NL8kBSGaEP4SPGP8/tMOX+geD8J3Z4AUAAA=="
+)
+
+var (
+	// TEE node labels for NVIDIA GPUs
+	nvidiaGPUCCNodeLabels = map[string]string{
+		"nvidia.com/cc.mode.state":  "on",
+		"nvidia.com/cc.ready.state": "true",
+	}
+
+	// TEE node labels for CPUs
+	intelTDXNodeLabels = map[string]string{
+		intelTDXNodeLabel: "true",
+	}
+	amdSNPNodeLabels = map[string]string{
+		amdSNPNodeLabel: "true",
+	}
+	imbSENodeLabels = map[string]string{
+		ibmSENodeLabel: "true",
+	}
 )
 
 // When the feature is enabled, handleFeatureConfidential configures confidential computing support.
@@ -124,6 +158,7 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 // handleConfidentialBaremetal configures confidential computing for baremetal deployments.
 // It manages kata-cc runtime classes with TEE-specific handlers (Intel TDX, AMD SNP or IBM SE).
 func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state FeatureGateState) error {
+	// if confidential feature gate enabled
 	if state == Enabled {
 		if !r.kataConfig.Spec.EnablePeerPods {
 			isLess, version, err := r.isOCPVersionLessThan("4.20.6")
@@ -143,7 +178,8 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
 
 		r.Log.Info("Creating " + kataCCRuntimeClassName + " runtime class for confidential containers")
 
-		handler, nodeLabel, err := r.computeTEEHandlerAndLabel()
+		// compute TEE
+		hasIntelTDX, hasAmdSNP, hasIbmSE, err := r.computeTEE()
 		if err != nil {
 			// If peer pods is enabled, just warn and skip baremetal coco
 			if r.kataConfig.Spec.EnablePeerPods {
@@ -155,19 +191,64 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
 			return err
 		}
 
-		// Determine extended resource based on TEE type
-		var kataCCRuntimeClassExtResOverhead string
-		if handler == kataCCIntelHandler {
+		// Determine handler and extended resource based on TEE type
+		var handler, kataCCRuntimeClassExtResOverhead string
+		additionalLabels := map[string]string{}
+		if hasIntelTDX {
 			kataCCRuntimeClassExtResOverhead = intelTDXExtendedResource
-		} else if handler == kataCCAmdHandler {
+			handler = kataCCIntelHandler
+			maps.Copy(additionalLabels, intelTDXNodeLabels)
+		} else if hasAmdSNP {
 			kataCCRuntimeClassExtResOverhead = amdSNPExtendedResource
+			handler = kataCCAmdHandler
+			maps.Copy(additionalLabels, amdSNPNodeLabels)
+		} else if hasIbmSE {
+			handler = kataCCIbmHandler
+			maps.Copy(additionalLabels, imbSENodeLabels)
 		}
 
 		// Create kata-cc runtime class restricted to the detected TEE subset
-		err = r.createRuntimeClass(kataCCRuntimeClassName, kataCCRuntimeClassCpuOverhead, kataCCRuntimeClassMemOverhead, kataCCRuntimeClassExtResOverhead, handler, map[string]string{nodeLabel: "true"})
+		err = r.createRuntimeClass(
+			kataCCRuntimeClassName,
+			kataCCRuntimeClassCpuOverhead,
+			kataCCRuntimeClassMemOverhead,
+			kataCCRuntimeClassExtResOverhead,
+			handler,
+			additionalLabels)
 		if err != nil {
-			r.Log.Info("Error creating "+kataCCRuntimeClassName+" runtime class", "err", err)
-			return fmt.Errorf("Error creating "+kataCCRuntimeClassName+" runtime class: %w", err)
+			r.Log.Error(err, "aborting, failed to create runtimeclass")
+			return err
+		}
+
+		additionalLabels = map[string]string{}
+		// for kata-cc-nvidia-gpu
+		if hasIntelTDX {
+			kataCCRuntimeClassExtResOverhead = intelTDXExtendedResource
+			handler = kataNvidiaGPUCCIntelHandler
+			maps.Copy(additionalLabels, nvidiaGPUNodeLabels)
+			// nvidiaGPUCCNodeLabels overrides values of base gpu labels
+			maps.Copy(additionalLabels, nvidiaGPUCCNodeLabels)
+			maps.Copy(additionalLabels, intelTDXNodeLabels)
+		} else if hasAmdSNP {
+			kataCCRuntimeClassExtResOverhead = amdSNPExtendedResource
+			handler = kataNvidiaGPUCCAmdHandler
+			maps.Copy(additionalLabels, nvidiaGPUNodeLabels)
+			// nvidiaGPUCCNodeLabels overrides values of base gpu labels
+			maps.Copy(additionalLabels, nvidiaGPUCCNodeLabels)
+			maps.Copy(additionalLabels, amdSNPNodeLabels)
+		}
+
+		// Create kata-cc-nvidia-gpu runtime class restricted to the detected GPU TEE subset
+		err = r.createRuntimeClass(
+			kataNvidiaGPUCCRuntimeClassName,
+			kataNvidiaGPUCCRuntimeClassCpuOverhead,
+			kataNvidiaGPUCCRuntimeClassMemOverhead,
+			kataCCRuntimeClassExtResOverhead,
+			handler,
+			additionalLabels)
+		if err != nil {
+			r.Log.Error(err, "aborting, failed to create runtimeclass")
+			return err
 		}
 
 	} else {
@@ -179,26 +260,35 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
 			r.Log.Info("Error deleting "+kataCCRuntimeClassName+" runtime class", "err", err)
 			return fmt.Errorf("Error deleting "+kataCCRuntimeClassName+" runtime class: %w", err)
 		}
+
+		r.Log.Info("Deleting " + kataNvidiaGPUCCRuntimeClassName + " runtime class for confidential containers")
+
+		// Delete kata-cc runtime class
+		err = r.deleteRuntimeClass(kataNvidiaGPUCCRuntimeClassName)
+		if err != nil {
+			r.Log.Info("Error deleting "+kataNvidiaGPUCCRuntimeClassName+" runtime class", "err", err)
+			return fmt.Errorf("Error deleting "+kataNvidiaGPUCCRuntimeClassName+" runtime class: %w", err)
+		}
 	}
 
 	return nil
 }
 
-func (r *KataConfigOpenShiftReconciler) computeTEEHandlerAndLabel() (string, string, error) {
+func (r *KataConfigOpenShiftReconciler) computeTEE() (hasIntelTDX, hasAmdSNP, hasIbmSE bool, err error) {
 	selector, err := r.getKataConfigNodeSelectorAsSelector()
 	if err != nil {
-		return "", "", fmt.Errorf("failed to build node selector: %w", err)
+		err = fmt.Errorf("failed to build node selector: %w", err)
+		return
 	}
 
 	nodes := &corev1.NodeList{}
 	listOpts := []client.ListOption{
 		client.MatchingLabelsSelector{Selector: selector},
 	}
-	if err := r.Client.List(context.TODO(), nodes, listOpts...); err != nil {
-		return "", "", fmt.Errorf("failed to list nodes: %w", err)
+	if err = r.Client.List(context.TODO(), nodes, listOpts...); err != nil {
+		err = fmt.Errorf("failed to list nodes: %w", err)
+		return
 	}
-
-	var hasIntelTDX, hasAmdSNP, hasIbmSE bool
 
 	for _, n := range nodes.Items {
 		if v, ok := n.Labels[intelTDXNodeLabel]; ok && v == "true" {
@@ -223,19 +313,14 @@ func (r *KataConfigOpenShiftReconciler) computeTEEHandlerAndLabel() (string, str
 		count++
 	}
 
+	if count == 0 {
+		err = fmt.Errorf("no TEE platform labels found (expected %s, %s or %s)", intelTDXNodeLabel, amdSNPNodeLabel, ibmSENodeLabel)
+		return
+	}
 	if count >= 2 {
-		return "", "", fmt.Errorf("multiple TEE platforms detected; only one per cluster supported")
+		err = fmt.Errorf("multiple TEE platforms detected; only one per cluster supported")
+		return
 	}
 
-	if hasIntelTDX {
-		return kataCCIntelHandler, intelTDXNodeLabel, nil
-	}
-	if hasAmdSNP {
-		return kataCCAmdHandler, amdSNPNodeLabel, nil
-	}
-	if hasIbmSE {
-		return kataCCIbmHandler, ibmSENodeLabel, nil
-	}
-
-	return "", "", fmt.Errorf("no TEE platform labels found (expected %s, %s or %s)", intelTDXNodeLabel, amdSNPNodeLabel, ibmSENodeLabel)
+	return
 }
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -978,7 +978,7 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(
 			NodeSelector: nodeSelector,
 		}
 
-		r.Log.Info("RuntimeClass NodeSelector:", "nodeSelector", nodeSelector)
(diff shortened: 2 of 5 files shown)


gkurz:
Hi @thejasn !

A general remark on expectations regarding PRs :
- changes should be broken down into a series of small and focused commits
- all commits should have a clear and descriptive message that helps understanding the change
- all commits should have your Signed-off-by:

This is required to ease maintenance, bug fixing and most importantly, this populates `git log` with valuable data that future contributors will be glad to find easily.

We do a regular merge with a merge commit. All the commits of the PR will thus appear in `git log`. The `tide/merge-method-squash` label isn't an option.

controllers/openshift_controller.go
```diff
@@ -827,7 +831,13 @@ func (r *KataConfigOpenShiftReconciler) createDaemonsetForMonitor() error {
...
 }
 
-func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName string, cpuOverhead string, memoryOverhead string, extResOverhead string, handler string, additionalNodeLabel string) error {
+func (r *KataConfigOpenShiftReconciler) createRuntimeClass(
+	runtimeClassName string,
+	cpuOverhead string,
+	memoryOverhead string,
+	extResOverhead string,
+	handler string,
+	additionalNodeLabel map[string]string) error {
```
gkurz: Please move this change to a preparatory commit that should simply introduce the new implementation and fix **all** existing callers across the repo without behavior change. Description should explain that we're going to add new runtime classes that require more labels.

wainersm:
Hi @thejasn ,

Initially, I've one comment about the runtimeclass name. It think we should treat it as a namespace. Currently it has `kata` (bare-metal, non-CC), `kata-cc` (baremetal, CC) and `kata-remote` (cloud). I would rename `kata-nvidia-gpu-cc` to `kata-cc-nvidia-gpu` so it end up with `kata` (bare-metal, non-CC), `kata-nvidia-gpu` (bare-metal, non-CC, nvidia gpu), `kata-cc` (baremetal, CC), `kata-cc-nvidia-gpu` (baremetal, CC, nvidia gpu) and `kata-remote` (cloud).

I know that we configure the kata-cc handler for cri-o via kata rpm specfile and perhaps a rename on the runtimeclass here will require a change on kata rpm side too.

controllers/openshift_controller.go
```diff
@@ -1149,7 +1159,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		if err != nil {
 			// If no nodes are found, requeue to check again for eligible nodes
 			r.Log.Error(err, "Failed to check Node eligibility for running Kata containers")
-			return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 20}, err
+			return ctrl.Result{}, err
```
gkurz: Why ?

controllers/openshift_controller.go
```diff
@@ -1270,7 +1280,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequest() (ctrl.
 		}
 		// FIXME : dead code, revisit postKataInstallation() return paths
 		if err != nil {
-			return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+			return ctrl.Result{}, err
```
gkurz: Why ?

controllers/openshift_controller.go
```diff
@@ -2182,6 +2192,7 @@ func (r *KataConfigOpenShiftReconciler) isUpdating() bool {
 	return cond.Status == corev1.ConditionTrue && cond.Reason == "Updating"
 }
 
+// FIX: unused function
```
gkurz: I see you have a commit to remove the function. Simply make this commit the first one in the PR.

controllers/openshift_controller.go
thejasn: So currently when both are non-nil, we get a warning printed in the logs that says the following.
```
Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes reqeueuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler
```
ref: https://github.com/kubernetes-sigs/controller-runtime/pull/2451

gkurz: Ok but please move that to the relevant commit.

controllers/openshift_controller.go
```diff
@@ -2274,13 +2285,65 @@ func (r *KataConfigOpenShiftReconciler) deleteScc() error {
...
+		nodes, err := r.getNodesWithLabels(requiredNodeLabels)
+		if err != nil {
+			r.Log.Error(err, "error in getting list of nodes", "labels", requiredNodeLabels)
+			return err
+		}
+		if len(nodes.Items) == 0 {
+			r.Log.Info("skipping creating runtimeclass due to missing labels", "runtimeclass", runtimeClassName)
+			return nil
+		}
+	}
```
gkurz: The current additional labels are for `kata-cc` and they are actually required. This means that this helper isn't required and you could just embed this code in `createRuntimeClass()`.

controllers/openshift_controller.go
```diff
@@ -2274,13 +2285,65 @@ func (r *KataConfigOpenShiftReconciler) deleteScc() error {
...
+		},
+		kataRuntimeClassName,
+		kataRuntimeClassCpuOverhead,
+		kataRuntimeClassMemOverhead,
+		"",                   /* nil extended resource overhead */
+		kataRuntimeClassName, /* reused for handler */
+		nil /* empty additional node labels */)
 	if err != nil {
-		return &ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+		return nil, err
```
gkurz: Please move this to the other commit.

controllers/openshift_controller.go
thejasn: This helper function isn't really applicable in CoCo case since TEE label computation based on the handler would handle verifying node labels.

controllers/openshift_controller.go
thejasn: Done

controllers/confidential_handler.go
```diff
@@ -184,18 +184,18 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
 	return nil
 }
 
-func (r *KataConfigOpenShiftReconciler) computeTEEHandlerAndLabel() (string, string, error) {
+func (r *KataConfigOpenShiftReconciler) computeTEEHandlerAndLabel() (string, map[string]string, error) {
```
gkurz: It is quite confusing to propagate that up to here : there's no clear link with the signature change of `createRuntimeClass()` .

I understand this is a temporary change since the code is heavily changed later in the PR, but all you need at this point is to update the user of `createRuntimeClass()` in this file to ``createRuntimeClass(kataCCRuntimeClassName, kataCCRuntimeClassCpuOverhead, kataCCRuntimeClassMemOverhead, kataCCRuntimeClassExtResOverhead, handler, map[string]string{nodeLabel})``.

controllers/openshift_controller.go
```diff
@@ -2281,13 +2285,75 @@ func (r *KataConfigOpenShiftReconciler) deleteScc() error {
...
+		"",                   /* nil extended resource overhead */
+		kataRuntimeClassName, /* reused for handler */
+		nil /* empty additional node labels */)
 	if err != nil {
 		return &ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
 	}
+
+	// creating kata-nvidia-gpu runtime class if node labels exist
+	err = r.nodeBasedCreateRuntimeClass(
+		nvidiaGPUNodeLabels,
```
gkurz:
```
$ make vet
go vet ./...
# github.com/openshift/sandboxed-containers-operator/controllers
controllers/openshift_controller.go:2346:3: undefined: nvidiaGPUNodeLabels
controllers/openshift_controller.go:2352:3: undefined: nvidiaGPUNodeLabels
# github.com/openshift/sandboxed-containers-operator/controllers
# [github.com/openshift/sandboxed-containers-operator/controllers]
vet: controllers/openshift_controller.go:2346:3: undefined: nvidiaGPUNodeLabels
make: *** [Makefile:119: vet] Error 1
```

It is important that each individual commit is buildable for the sake of bisectability.

controllers/openshift_controller.go
```diff
@@ -827,7 +828,13 @@ func (r *KataConfigOpenShiftReconciler) createDaemonsetForMonitor() error {
...
 }
 
-func (r *KataConfigOpenShiftReconciler) createRuntimeClass(runtimeClassName string, cpuOverhead string, memoryOverhead string, extResOverhead string, handler string, additionalNodeLabel string) error {
+func (r *KataConfigOpenShiftReconciler) createRuntimeClass(
+	runtimeClassName string,
+	cpuOverhead string,
+	memoryOverhead string,
+	extResOverhead string,
+	handler string,
+	additionalNodeLabel map[string]string) error {
```
gkurz:
```suggestion
	additionalNodeLabels map[string]string) error {
```

controllers/openshift_controller.go
gkurz: Detection/Validation of node features is messy. We'll need to revisit it at some point... probably out of scope for the current effort.

Anyway, we only have two situations to consider here :
- node eligibility is true : we want to enforce all required node labels
- node eligibility is false : we don't care for node labels at all

There is no point in having two separate label maps arguments for this helper : just one that is checked and passed to `createRuntimeClass` or ignored.

controllers/confidential_handler.go
```diff
@@ -29,12 +28,73 @@ const (
...
 	// INITDATA value for non-confidential peer pods, this is required in order
 	// to override the default restrictive CoCo agent policy
 	// created from sourced plaintxt: cat config/peerpods/default-non-cc-initdata.toml | gzip | base64 -w0
 	defaultNonCCInitdata = "H4sIAAAAAAAAA42UwW7bMAyG734KwT3ktKDDehgG9NAl2VZgWQw7bQ7DMDAWYxOVRU+i03pPP7lGhx0muwddyE8U+ZPUhdrX5BVZEg0C6kQG1SN4VToEQa2OvZIaFfvyDbfoQNglYCp2JHWjrlXqa3j3/ipNzug8sR1Ml8u3y8s0Sb4PIX8kKcBSuDFp8C0Wi2Q4SdqyobJfOqz4xdFC+QAVqnCs/ByBJNF4gs6IutH6Js++IVX1kZ3P8VeHXtSHayWuw3+x4hHamHtl2GMhmmyU4Lb/FGSI+p+VWbEVIItuGivA6iM/xaB1MDruZ6jNE5aZ4xJ9tOrPKFsUR+UUsdttN+cgbRQZrGsMdZlomK/k5dYKuhOEfKaonDuJE1tsvrC0pqs+9qG2Y1TunTVB5lV2F27EmAw6P9+RrDPmtgnDFQNyBF1I6Fv0oRwbPs+/NGKFgMF7ckJ88kUNDrfcWYlKkqNH1HmYBW7WeJ7AumY+hwJl7GeYwj010aIDlz1vWhSgyoKZmb9Qq5P5nAZq76AkW00w4l8RiduZQHvpD8OWe/odLf6u1a/Z5RHbtDU24Qs0020c4b87Mo1NL8kBSGaEP4SPGP8/tMOX+geD8J3Z4AUAAA=="
 )
 
+var (
+
+	// TEE node labels for NVIDIA GPU
+	nvidiaGPUNodeLabels = map[string]string{
```
gkurz: This isn't a TEE setup and it is required by a previous commit actually. Please move it there.

controllers/openshift_controller.go
```diff
@@ -643,6 +643,12 @@ func (r *KataConfigOpenShiftReconciler) newMCForCR(machinePool string) (*mcfgv1.
 			Namespace: OperatorNamespace,
 		},
 		Spec: mcfgv1.MachineConfigSpec{
+			// added for GPU on bare metals
+			// TODO: test if ignored on non-GPU clusters
```
gkurz: `amd_iommu` and `intel_iommu` won't be ignored. This tells the kernel to plug the IOMMU HW from the host board into the I/O flow. This can have an impact depending on the setups. If we want to stay on the safe side, I would say we should only add these kernel args when we know we're going to do VFIO.

controllers/openshift_controller.go
```diff
@@ -877,7 +883,7 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(
 			NodeSelector: nodeSelector,
 		}
 
-		r.Log.Info("RuntimeClass NodeSelector:", "nodeSelector", nodeSelector)
+		r.Log.Info("RuntimeClass", "name", runtimeClassName, "nodeSelector", nodeSelector)
```
gkurz, line 981: Change is nice but please do that in some other commit.

controllers/openshift_controller.go
```diff
@@ -897,7 +903,7 @@ func (r *KataConfigOpenShiftReconciler) createRuntimeClass(
 		r.Log.Info("Creating a new RuntimeClass", "rc.Name", rc.Name)
 		err = r.Client.Create(context.TODO(), rc)
 		if err != nil {
-			return err
+			return fmt.Errorf("error creating %s runtime class: %w", rc.Name, err)
```
gkurz, line 1001: Same here.

gkurz (commented):
Hi @thejasn ! I'm struggling with commit 4, so I'm posting a partial review for now.

controllers/openshift_controller.go
thejasn: Upon reading up a bit more, looks like according to nvidia [docs](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/confidential-containers-deploy.html#prerequisites) these kernel parameters are a prerequisite to installing their operator.  

Chronologically since the GPU operator is installed prior to OSC we might not need this. Adding this via OSC essentially acts as a override on previous rendered MC which I think we shouldn't do. Maybe this could just be a check behind `checkNodeEligibility`. 

/cc @bpradipt since this addition was suggest by him.

controllers/openshift_controller.go
thejasn: Done, moved to https://github.com/openshift/sandboxed-containers-operator/pull/1738/commits/b5b75a45429e769524bdf03038cc941089fa9330

controllers/openshift_controller.go
thejasn: Done, moved to https://github.com/openshift/sandboxed-containers-operator/commit/b5b75a45429e769524bdf03038cc941089fa9330

thejasn:
rebased upstream

controllers/openshift_controller.go
bpradipt: In that case, we'll need to add this MachineConfig as part of docs only.

controllers/confidential_handler.go
```diff
@@ -29,12 +33,66 @@ const (
 	intelTDXExtendedResource = "tdx.intel.com/keys"
 	amdSNPExtendedResource   = "sev-snp.amd.com/esids"
 
+	// GPU
+
+	// kata-nvidia-cc runtime class for CoCo BM GPU
+	kataNvidiaGPUCCRuntimeClassName        = "kata-nvidia-gpu-cc"
```
bpradipt: s/kata-nvidia-gpu-cc/kata-cc-nvidia-gpu

controllers/confidential_handler.go
```diff
@@ -29,12 +33,66 @@ const (
...
 	// to override the default restrictive CoCo agent policy
 	// created from sourced plaintxt: cat config/peerpods/default-non-cc-initdata.toml | gzip | base64 -w0
 	defaultNonCCInitdata = "H4sIAAAAAAAAA42UwW7bMAyG734KwT3ktKDDehgG9NAl2VZgWQw7bQ7DMDAWYxOVRU+i03pPP7lGhx0muwddyE8U+ZPUhdrX5BVZEg0C6kQG1SN4VToEQa2OvZIaFfvyDbfoQNglYCp2JHWjrlXqa3j3/ipNzug8sR1Ml8u3y8s0Sb4PIX8kKcBSuDFp8C0Wi2Q4SdqyobJfOqz4xdFC+QAVqnCs/ByBJNF4gs6IutH6Js++IVX1kZ3P8VeHXtSHayWuw3+x4hHamHtl2GMhmmyU4Lb/FGSI+p+VWbEVIItuGivA6iM/xaB1MDruZ6jNE5aZ4xJ9tOrPKFsUR+UUsdttN+cgbRQZrGsMdZlomK/k5dYKuhOEfKaonDuJE1tsvrC0pqs+9qG2Y1TunTVB5lV2F27EmAw6P9+RrDPmtgnDFQNyBF1I6Fv0oRwbPs+/NGKFgMF7ckJ88kUNDrfcWYlKkqNH1HmYBW7WeJ7AumY+hwJl7GeYwj010aIDlz1vWhSgyoKZmb9Qq5P5nAZq76AkW00w4l8RiduZQHvpD8OWe/odLf6u1a/Z5RHbtDU24Qs0020c4b87Mo1NL8kBSGaEP4SPGP8/tMOX+geD8J3Z4AUAAA=="
 )
 
+var (
+	// TEE node labels for NVIDIA GPUs
+	nvidiaGPUCCNodeLabels = map[string]string{
+		"nvidia.com/cc.mode.state":         "on",
+		"nvidia.com/cc.ready.state":        "true",
```
bpradipt: I tested with a single `ClusterPolicy` for both CC and regular GPU and found that these two are important labels to distinguish between CC and regular GPU. For regular GPU these are set to false/off
`nvidia.com/cc.ready.state: false`
and `nvidia.com/cc.mode.state: off`

controllers/confidential_handler.go
```diff
@@ -29,12 +33,66 @@ const (
...
+		"nvidia.com/gpu.deploy.cc-manager": "true",
```
bpradipt: With the new ClusterPolicy I'm defaulting to , this will be true for both regular and CC GPU. You can remove this from consideration

controllers/openshift_controller.go
gkurz: > Maybe this could just be a check behind checkNodeEligibility.

FWIW I like the idea of double checking that the host kernel is configured properly and it shouldn't be that hard to come up with appropriate NFD rules. Future work.

> In that case, we'll need to add this MachineConfig as part of docs only.

Or maybe this is already covered in some way by the GPU operator ?

bpradipt: I have not seen it in the GPU operator instructions. Also this MachineConfig is specifically required for Kata and GPU operator instructions for OpenShift focuses on non-Kata use case

controllers/confidential_handler.go
```diff
@@ -29,12 +33,66 @@ const (
...
 
+	// GPU
+
+	// kata-nvidia-cc runtime class for CoCo BM GPU
+	kataNvidiaGPUCCRuntimeClassName        = "kata-nvidia-gpu-cc"
+	kataNvidiaGPUCCRuntimeClassCpuOverhead = "0.25"
+	kataNvidiaGPUCCRuntimeClassMemOverhead = "350Mi"
+
+	// RuntimeClass handlers for TEE
+	kataNvidiaGPUCCIntelHandler = "kata-nvidia-gpu-tdx"
```
gkurz: For consistency : s/kata-nvidia-gpu-tdx/kata-tdx-nvidia-gpu

controllers/openshift_controller.go
gkurz: Please create a dedicated commit for these logging improvements.

controllers/openshift_controller.go
gkurz: Please move that to the logging improvement commit.

controllers/openshift_controller.go
```diff
@@ -2359,13 +2374,72 @@ func (r *KataConfigOpenShiftReconciler) deleteScc() error {
 	return nil
 }
 
+// nodeBasedCreateRuntimeClass checks the node for required labels before
+// creating the required runtime class.
+func (r *KataConfigOpenShiftReconciler) nodeBasedCreateRuntimeClass(
+	requiredNodeLabels map[string]string,
```
gkurz: As mentioned earlier, I strongly believe that this wrapper will eventually be merged into `createRuntimeClass()`. In order to facilitate that, please re-order the arguments so that `nodeBasedCreateRuntimeClass()` has the same signature as `createRuntimeClass()`.

controllers/openshift_controller.go
```diff
@@ -88,6 +88,21 @@ const (
 	kataRuntimeClassCpuOverhead = "0.25"
 	// We need a higher value than upstream (see https://github.com/openshift/sandboxed-containers-operator/pull/84)
 	kataRuntimeClassMemOverhead = "350Mi"
+
+	kataNvidiaGPURuntimeClassName        = "kata-nvidia-gpu"
+	kataNvidiaGPURuntimeClassCpuOverhead = "0.25"
+	kataNvidiaGPURuntimeClassMemOverhead = "350Mi"
```
gkurz: FYI the upstream reference for runtime class overheads is https://github.com/kata-containers/kata-containers/blob/main/tools/packaging/kata-deploy/helm-chart/kata-deploy/templates/runtimeclasses.yaml#L84

```
"qemu-nvidia-gpu" (dict "memory" "4096Mi" "cpu" "1.0")
```

controllers/migrate.go
```diff
@@ -88,9 +88,11 @@ func (r *KataConfigOpenShiftReconciler) migratePeerPodsLimit() error {
 // ensureRuntimeClassFinalizers adds finalizers to existing runtime classes that don't have them
 func (r *KataConfigOpenShiftReconciler) ensureRuntimeClassFinalizers() error {
 	runtimeClasses := []string{
+		kataRuntimeClassName,
+		kataNvidiaGPURuntimeClassName,
```
gkurz: This function was introduced by 47e37f7d019eb8482d90b73850a41b5a3c1668c0 to migrate existing setups. I don't think it is needed for the new GPU runtime classes. WDYT @beraldoleal ?

controllers/confidential_handler.go
```diff
@@ -179,63 +276,56 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
...
-			hasIntelTDX = true
-		}
-		if v, ok := n.Labels[amdSNPNodeLabel]; ok && v == "true" {
-			hasAmdSNP = true
-		}
-		if v, ok := n.Labels[ibmSENodeLabel]; ok && v == "true" {
-			hasIbmSE = true
+		for k, v := range teeMapping {
+			// check if node labels contains all the labels
+			// a particular handler requires
```
gkurz: As discussed on slack, I don't think we should check all labels here... but rather leave that to `nodeBasedCreateRuntimeClass`. The only thing we want to guess here is the TEE.

controllers/daemonset_reconcile.go
```diff
@@ -205,7 +205,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequestDaemonSet
 		if err != nil {
 			// If no nodes are found, requeue to check again for eligible nodes
 			r.Log.Error(err, "Failed to check Node eligibility for running Kata containers")
-			return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 20}, err
```
snir911, line 208: nit: s/adn/and in commit msg

controllers/daemonset_reconcile.go
```diff
@@ -205,7 +205,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequestDaemonSet
...
+			return ctrl.Result{}, err
```
snir911: Do we want exponential backoff here? aren't we prefer a fixed backoff in this case as we expect eligible nodes to appear at some point. cc @Pacho20

controllers/migrate.go
beraldoleal: Indeed @gkurz . this was for backward compatibility, new classes are not needed.

controllers/openshift_controller.go
thejasn: Will update the ref values. Also looks like other runtime class overheads are not set by the operator at all.

controllers/openshift_controller.go
```diff
@@ -2356,10 +2404,34 @@ func (r *KataConfigOpenShiftReconciler) deleteScc() error {
...
+
+	// creating kata runtime class if node labels exist
+	err := r.createRuntimeClass(
+		kataRuntimeClassName,
+		kataRuntimeClassCpuOverhead,
+		kataRuntimeClassMemOverhead,
+		"",                   /* nil extended resource overhead */
+		kataRuntimeClassName, /* reused for handler */
+		map[string]string{
+			"feature.node.kubernetes.io/runtime.kata": "true",
```
gkurz, line 2369: This label is needed by all baremetal runtime classes actually. Please move it to a global `kataNodeLabels` map variable and have all other label maps to use it.

thejasn: As a FYI, this kind of goes into cleaning up the rest of the usages of `checkNodeEligibility` as currently this label is by default checked in ` getKataConfigNodeSelectorAsLabelSelector()` that is called as part of the list nodes prior to creating runtime class. 

For now moving this out to `kataNodeLabels` makes it explicit and doesn't bury this check (but does duplicate the label). But as you previously mentioned at some point we should revisit usages of `checkNodeEligibility` to de-duplicate stuff.

controllers/daemonset_reconcile.go
thejasn: I can update this to `ctrl.Result{Requeue: true, RequeueAfter: time.Second * 20}, nil`, but essentially this diff doesn't really change anything since in the old version `ctrl.Result` was anyways being ignored. 

I'm thinking we re-visit all non-nil ctrl.Results later in a separate PR as there a lot more of these misconfigured returns. WDYT?

controllers/openshift_controller.go
gkurz: Fair enough.

controllers/daemonset_reconcile.go
gkurz: This commit doesn't even address all the call sites in `controllers/openshift_controller.go`... so yes, I think all of this should be revisited in its own PR later, including this commit. Please just drop it.

controllers/confidential_handler.go
```diff
@@ -155,19 +191,63 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
...
 			kataCCRuntimeClassExtResOverhead = intelTDXExtendedResource
-		} else if handler == kataCCAmdHandler {
+			handler = kataCCIntelHandler
+			maps.Copy(additionalLabels, intelTDXNodeLabels)
+		} else if hasAmdSNP {
 			kataCCRuntimeClassExtResOverhead = amdSNPExtendedResource
+			handler = kataCCAmdHandler
+			maps.Copy(additionalLabels, amdSNPNodeLabels)
+		} else if hasIbmSE {
+			handler = kataCCIbmHandler
```
gkurz, line 206: What about the IBM SE labels ?

```suggestion
			handler = kataCCIbmHandler
			maps.Copy(additionalLabels, ibmSENodeLabels)
```

controllers/confidential_handler.go
```diff
@@ -179,27 +259,34 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
...
 
 	return nil
 }
 
-func (r *KataConfigOpenShiftReconciler) computeTEEHandlerAndLabel() (string, string, error) {
+func (r *KataConfigOpenShiftReconciler) computeTEE() (hasIntelTDX, hasAmdSNP, hasIbmSE bool, err error) {
 	selector, err := r.getKataConfigNodeSelectorAsSelector()
 	if err != nil {
-		return "", "", fmt.Errorf("failed to build node selector: %w", err)
+		return hasIntelTDX, hasAmdSNP, hasIbmSE, fmt.Errorf("failed to build node selector: %w", err)
```
gkurz: It is a bit misleading to return the named arguments in this context. What about a naked return like already present in this function ?

```suggestion
		err = fmt.Errorf("failed to build node selector: %w", err)
		return
```

controllers/confidential_handler.go
```diff
@@ -179,27 +259,34 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
...
 	}
 
 	nodes := &corev1.NodeList{}
 	listOpts := []client.ListOption{
 		client.MatchingLabelsSelector{Selector: selector},
 	}
 	if err := r.Client.List(context.TODO(), nodes, listOpts...); err != nil {
-		return "", "", fmt.Errorf("failed to list nodes: %w", err)
+		return hasIntelTDX, hasAmdSNP, hasIbmSE, fmt.Errorf("failed to list nodes: %w", err)
```
gkurz:
```suggestion
		err = fmt.Errorf("failed to list nodes: %w", err)
		return
```

gkurz (commented):
Last comments hopefully and we should be good.

controllers/confidential_handler.go
thejasn: Good catch! Updated

gkurz (approved):
/lgtm

Great thanks @thejasn for the hard work !

PR #1740, opened 2026-02-25
daemonset deployment: Remove deprecated steps

gkurz:
Following the bump of OSC to 1.12 across the repo, I had to fix the Dockerfile in order to create a new osc-daemonset image.


Pacho20:
Thanks for fixing this @gkurz!
LGTM

gkurz:
This file isn't used in CI. No need to wait.

PR #1753, opened 2026-02-25
tests/osc-test-fbc-integration: do not run the pipeline on push

wainersm:
Avoid triggering the prowjobs when the pipeline definition (tests/osc-test-fbc-integration.yaml) only changed and was pushed because it is waste of resources.

The .tekton/osc-test-fbc-pull-request.yaml is still configured to trigger osc-test-fbc-integration pipeline when its own definition changed, but rather on pull_request. This way we still can test changes on the osc-test-fbc-integration pipeline at pull_request time.

To make it clear: any changes on the catalog (test-fbc) that are pushed will still be triggering the osc-test-fbc-integration pipeline.


gkurz (approved):
/lgtm

Thanks @wainersm !

PR #1775, opened 2026-03-03
cache: Only cache required objects to prevent OOM

gkurz:
The operator accesses a variety of object types. The controller runtime
caches all objects of all types in RAM by default. Some types may have
too many instances, causing the cache to inflate until OOM. The way to
go is to fine tune the cache so that it only keeps objects that the
operator actually needs.

Good first candidates for that are secrets and config maps. Let's
scope the cache to namespaces where we access any of these.

A full review of the operator code is needed if we want to find
other candidates.

First commit is just because I spotted the literal string when looking for secrets in the code.

OOM Reproducer :

- run the following script

```
#!/bin/bash

#
# Create 1000 secrets with a 100KB payload. Controller will need ~400MB to
# 600MB of heap to cache them all.
#
SECRET_COUNT=1000
PREFIX="dummy-secret"

# Create a 100KB payload string
PAYLOAD=$(head -c 102400 /dev/urandom | base64 | head -c 102400)

echo "Deploying $SECRET_COUNT secrets (100KB each)..."

for i in $(seq 1 $SECRET_COUNT); do
  kubectl create secret generic "$PREFIX-$i" \
    --from-literal=payload="$PAYLOAD" \
    --namespace=default \
    --dry-run=client -o yaml | kubectl apply -f -

  if [ $((i % 100)) -eq 0 ]; then
    echo "Sent $i/1000 secrets..."
  fi
done
```

- install latest OSC

Assisted by gemini to find the root cause and to come up with a reproducer script.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -36,6 +36,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -105,6 +106,24 @@ func main() {
 		},
 		LeaderElection:   enableLeaderElection,
 		LeaderElectionID: "290f4947.kataconfiguration.openshift.io",
+		// The controller runtime can end up caching all objects in the cluster and cause OOM.
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&corev1.Secret{}: {
+					// Restrict caching to namespaces where we access secrets
+					Namespaces: map[string]cache.Config{
+						controllers.OperatorNamespace: {},
+						"openshift-config":            {}, // for global pull-secrets
+					},
+				},
+				&corev1.ConfigMap{}: {
+					// Same for config maps
+					Namespaces: map[string]cache.Config{
+						controllers.OperatorNamespace: {},
+					},
+				},
+			},
+		},
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
--- a/controllers/ppwebhook.go
+++ b/controllers/ppwebhook.go
@@ -270,7 +270,7 @@ func (r *KataConfigOpenShiftReconciler) createMutatingWebhookConfig() error {
 	// Add list of namespaces to exclude
 	namespacesToExclude := []string{
 		"peer-pods-webhook-system",
-		"openshift-sandboxed-containers-operator",
+		OperatorNamespace,
 		"openshift",
 		"openshift-apiserver",
 		"openshift-apiserver-operator",
--- a/controllers/scc.go
+++ b/controllers/scc.go
@@ -35,6 +35,6 @@ func GetScc() *secv1.SecurityContextConstraints {
 			Type: secv1.SELinuxStrategyRunAsAny,
 		},
 		Volumes: []secv1.FSType{secv1.FSTypeAll},
-		Users:   []string{"system:serviceaccount:openshift-sandboxed-containers-operator:monitor"},
+		Users:   []string{"system:serviceaccount:" + OperatorNamespace + ":monitor"},
 	}
 }
--- a/controllers/utils.go
+++ b/controllers/utils.go
@@ -156,7 +156,7 @@ func getPeerPodsSecret(c client.Client) (*corev1.Secret, error) {
 
 	err := c.Get(context.TODO(), types.NamespacedName{
 		Name:      peerPodsSecretName,
-		Namespace: "openshift-sandboxed-containers-operator",
+		Namespace: OperatorNamespace,
 	}, peerPodsSecret)
 
 	if err != nil {


controllers/credentials_controller.go
```diff
@@ -49,7 +49,7 @@ type SecretReconciler struct {
 }
 
 const (
-	credentialsRequestSecretRefName         = "cco-secret"
+	CredentialsRequestSecretRefName         = "cco-secret"
```
snir911: isn't lower case first letter is the convention?

cmd/manager/main.go
```diff
@@ -105,6 +107,13 @@ func main() {
 		},
 		LeaderElection:   enableLeaderElection,
 		LeaderElectionID: "290f4947.kataconfiguration.openshift.io",
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&corev1.Secret{}: {
+					Field: fields.OneTermEqualSelector("metadata.name", controllers.CredentialsRequestSecretRefName),
```
snir911: IIUC it says "enabling the cache only for this specific secret"
however, i think it will disable the cache also for other secret we fetch (but not watch) like:
1. peer-pods-secret
2. peer-pods-webhook-svc-secret
3. pull-secret
4. auth-json-secret

IIUC 3 & 4 will be deprecated soon though, is it possible to filter based on ns maybe?

controllers/credentials_controller.go
gkurz: This tells go about the scope. Lowercase is local to the package and uppercase is public. Required to be accessible from `main` package.

cmd/manager/main.go
gkurz: Good point @snir911 ... I'll look into an alternative. Thanks !

controllers/credentials_controller.go
gkurz: Dropped this change. Now considering namespaces.

cmd/manager/main.go
gkurz: Now considering namespaces.

cmd/manager/main.go
```diff
@@ -105,6 +106,19 @@ func main() {
 		},
 		LeaderElection:   enableLeaderElection,
 		LeaderElectionID: "290f4947.kataconfiguration.openshift.io",
+		// A production cluster can have thousands of secrets that will end up in the cache and cause OOM.
+		// Only keep the ones we care for, i.e. the ones from the OSC operator namespace and the global pull
+		// secrets from the cluster config.
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
```
c3d, line 111: I think this is the best we can do given that the infrastructure does not let you either size-limit the cache or refill it on demand. So /lgtm.

c3d (approved):
I'm not super familiar with that code, but FWIW, looks good to me.

(I would prefer if there was a way to size-limit or demand-load the cache, but according to Copilot that's not possible)

controllers/credentials_controller.go
wainersm: Hi @gkurz , you said you dropped that change but ... forgot to push?

wainersm:
@gkurz @snir911 , at this point you should already know that it's caching more than secrets and the correct solution seems to enable cache of only two namespaces. Even though, let me share an interesting analysis made by Claude (** don't think it's hallucination but I am not expert):

```
What Resources Are Cached by Default?
                                                                                                                                                  
  By default, controller-runtime caches ALL instances of any resource type that ANY controller watches, regardless of predicates/filters.
                                                                                                                                                  
  Predicates only control which events trigger reconciliation, NOT what gets cached!

  Current Cached Resources (Before PR #1775):

  ┌───────────────────┬────────────────┬───────────────────────────────────────┬─────────────────────────────────────────┐
  │     Resource      │     Scope      │                Source                 │              Memory Impact              │
  ├───────────────────┼────────────────┼───────────────────────────────────────┼─────────────────────────────────────────┤
  │ Secret            │ ALL namespaces │ SecretReconciler watches              │ 🔴 HIGH (1000s in large clusters)       │
  ├───────────────────┼────────────────┼───────────────────────────────────────┼─────────────────────────────────────────┤
  │ ConfigMap         │ ALL namespaces │ KataConfigOpenShiftReconciler watches │ 🔴 HIGH (1000s in large clusters)       │
  ├───────────────────┼────────────────┼───────────────────────────────────────┼─────────────────────────────────────────┤
  │ Pod               │ ALL namespaces │ RuntimeClassReconciler lists          │ 🔴 CRITICAL (10,000s in large clusters) │
  ├───────────────────┼────────────────┼───────────────────────────────────────┼─────────────────────────────────────────┤
  │ Node              │ ALL            │ KataConfigOpenShiftReconciler watches │ 🟡 Medium (100s-1000s)                  │
  ├───────────────────┼────────────────┼───────────────────────────────────────┼─────────────────────────────────────────┤
  │ MachineConfigPool │ ALL            │ KataConfigOpenShiftReconciler watches │ 🟢 Low (< 10)                           │
  ├───────────────────┼────────────────┼───────────────────────────────────────┼─────────────────────────────────────────┤
  │ RuntimeClass      │ ALL            │ RuntimeClassReconciler watches        │ 🟢 Low (< 10)                           │
  ├───────────────────┼────────────────┼───────────────────────────────────────┼─────────────────────────────────────────┤
  │ KataConfig        │ ALL            │ Multiple reconcilers                  │ 🟢 Low (< 5)                            │
  └───────────────────┴────────────────┴───────────────────────────────────────┴─────────────────────────────────────────┘

  The Problem with PR #1775:

  The PR only fixes Secrets! It doesn't address:

  1. ConfigMaps - Still caching ALL cluster ConfigMaps (watched by KataConfigOpenShiftReconciler)
  2. Pods - Still caching ALL cluster Pods (listed by RuntimeClassReconciler line 105)

  The ConfigMapEventHandler filters by namespace (line 26, 45), but that's only for event processing, not caching!

  Why This Matters:

  In a large cluster with:
  - 10,000 Pods
  - 5,000 ConfigMaps
  - 3,000 Secrets

  The operator is caching ALL OF THEM, but only needs:
  - ~5 Secrets (from 2 namespaces)
  - ~2 ConfigMaps (from operator namespace)
  - 0 Pods initially (only needs to list on-demand during RuntimeClass deletion)
```

It's really weird that the controller manager has cache enabled by default.

snir911:
@wainersm interesting , however, AFAICT pods are only listed, hence, they are not cached (i think, otherwise we would have hit OOM much earlier i suppose, so it make some sense).
For CMs maybe it worth to limit the cache also, @gkurz WDYT?

controllers/credentials_controller.go
gkurz: Forgot to `stg refresh` ;-)

gkurz:
> @wainersm interesting , however, AFAICT pods are only listed, hence, they are not cached (i think, otherwise we would have hit OOM much earlier i suppose, so it make some sense). For CMs maybe it worth to limit the cache also, @gkurz WDYT?

IIUC the problem is triggered by `Watches()` and we have :
```
$ git grep -A1 Watch
controllers/credentials_controller.go:          Watches(
controllers/credentials_controller.go-                  &corev1.Secret{},
--
controllers/credentials_controller.go:          Watches(&kataconfigurationv1.KataConfig{},
controllers/credentials_controller.go-                  &KataConfigHandler{r}).
--
controllers/openshift_controller.go:            Watches(
controllers/openshift_controller.go-                    &mcfgv1.MachineConfigPool{},
--
controllers/openshift_controller.go:            Watches(
controllers/openshift_controller.go-                    &corev1.Node{},
--
controllers/openshift_controller.go:            Watches(
controllers/openshift_controller.go-                    &corev1.ConfigMap{},
```

KataConfig is obviously fine since we can only have one single instance.
I guess MachineConfigPool and Node are fine because it is unlikely to have thousands of these in a cluster.
ConfigMap could potentially be a problem.

gkurz:
> IIUC the problem is triggered by Watches() and we have :

Yikes, claude is nacking this assumption... all object types that we `Get` or `List` will also go to the cache by default.

I suspect cache induced OOMs aren't frequent because OSC isn't deployed on huge clusters.

In order to be 100% OOM proof, we should carefully review all the code and configure the cache accordingly. Depending on each case, solution can be namespace scoping, label scoping for objects managed by the operator or even fully by-passing the cache for objects that aren't accessed that often.

I suggest we plan that for a future release, and we stick to handle secrets and CMs for now. WDYT ?

wainersm:
> > IIUC the problem is triggered by Watches() and we have :
> 
> Yikes, claude is nacking this assumption... all object types that we `Get` or `List` will also go to the cache by default.
> 
> I suspect cache induced OOMs aren't frequent because OSC isn't deployed on huge clusters.
> 
> In order to be 100% OOM proof, we should carefully review all the code and configure the cache accordingly. Depending on each case, solution can be namespace scoping, label scoping for objects managed by the operator or even fully by-passing the cache for objects that aren't accessed that often.
>

I agree with your assessment @gkurz .
 
> I suggest we plan that for a future release, and we stick to handle secrets and CMs for now. WDYT ?

Sound as a good plan.

wainersm (approved):
Thanks @gkurz !

snir911 (approved):

PR #1798, opened 2026-03-05
claude: add osc-user-guide skill for OSC documentation access

wainersm:
**- Description of the problem which is fixed/What is the use case**

While developing or reviewing a PR, sometimes I need to get some information from the user guide of OSC.

For example, recently I was reviewing a PR about adding automation to tests. I wanted to check if the code was implementing the setup that was describe in our user guide; then I used the Claude skill I am proposing here to automatically and easily fetch the information needed to compare.

Claude would be capable of getting the OSC guide without this still, but maybe it would get an older version. Or maybe it would spend some cycles figuring out how to get and parse the content. This skill automates the process so that you don't need to teach claude over and over how to get the information.

**- What I did**

Adds a comprehensive Claude Code skill that provides access to all three OpenShift Sandboxed Containers 1.11 documentation guides:
- Deploying Red Hat OpenShift Sandboxed Containers
- Deploying Confidential Containers
- Deploying Red Hat Build of Trustee

Features:
- Automatic environment setup with Python venv
- Auto-downloads documentation PDFs on first run (~2 MB total)
- Efficient PDF search with keyword, topic, and section search modes
- Relevance scoring and ranking for search results
- Page and page range retrieval
- Team-friendly: no manual setup required

The skill automatically triggers when users ask questions about OSC installation, configuration, deployment, troubleshooting, PeerPods, confidential containers, TEE, or Trustee.

Files:
- SKILL.md: Skill definition with detailed instructions for Claude
- skill.py: Main entry point (handles environment and PDF downloads)
- search_pdfs.py: PDF search CLI tool with multiple search modes
- pdf_utils.py: PDF utility library for text extraction and searching
- README.md: Documentation for team members
- requirements.txt: Python dependencies (PyPDF2)
- .gitignore: Excludes venv and downloaded PDFs from git

Assisted-by: Claude

**- How to verify it**

You can test that PR by:

*  pull & checkout it
*  launch claude
* Ask something like "What is the runtimeclass that I should use for Confidential Containers? Use the user guide" (you don't necessarily need to mention the user guide as it can infer -- look at the skill prompt that there are some conditions when the skill is triggered)

Example of output:

```
❯ What is the runtimeclass that I should use for Confidential Containers? Use the user guide.                                                     
                                                                                              
● I'll use the OSC user guide to find information about the runtimeclass for Confidential Containers.                                             
                                                                                                                                                  
● Skill(osc-user-guide)                                                                                                                           
  ⎿  Successfully loaded skill
[...]

PR #1813, opened 2026-03-06
claude: add osc-user-guide skill for OSC documentation access

wainersm:
Supersedes: #1798 

While developing or reviewing a PR, sometimes I need to get some information from the user guide of OSC.

For example, recently I was reviewing a PR about adding automation to tests. I wanted to check if the code was implementing the setup that was describe in our user guide; then I used the Claude skill I am proposing here to automatically and easily fetch the information needed to compare.

Claude would be capable of getting the OSC guide without this still, but maybe it would get an older version. Or maybe it would spend some cycles figuring out how to get and parse the content. This skill automates the process so that you don't need to teach claude over and over how to get the information.

**- What I did**

Adds a comprehensive Claude Code skill that provides access to all three OpenShift Sandboxed Containers documentation guides:
- Deploying Red Hat OpenShift Sandboxed Containers
- Deploying Confidential Containers
- Deploying Red Hat Build of Trustee

How it works:
1. Downloads PDF guides using a simple bash script (curl/wget)
2. Reads entire PDFs using built-in PDF reading capability
3. Provides a comprehensive answer with examples and citations

The skill automatically triggers when users ask questions about OSC installation, configuration, deployment, troubleshooting, PeerPods, confidential containers, TEE, or Trustee.

Assisted-by: Claude

**- How to verify it**

You can test that PR by:

*  pull & checkout it
*  launch claude
* Ask something like "What is the runtimeclass that I should use for Confidential Containers? Use the user guide" (you don't necessarily need to mention the user guide as it can infer -- look at the skill prompt that there are some conditions when the skill is triggered)

Example of output:

```
❯ What is the runtimeclass that I should use for Confidential Containers? Use the user guide. 

● I'll use the OSC user guide to find information about the runtimeclass for Confidential Containers.

❯ Tool loaded. 

● Skill(osc-user-guide)
  ⎿  Successfully loaded skill

● I'll help you find the runtimeclass for Confidential Containers by consulting the official OSC documentation.

❯ Tool loaded. 

● Fetch(https://docs.redhat.com/en/documentation/openshift_sandboxed_containers)
  ⎿  Received 289.8KB (200 OK)
                     
❯ Tool loaded.                 
                                                                                                                                                  
● Bash(bash .claude/skills/osc-user-guide/download-pdfs.sh 1.11)                                                                                  
  ⎿  OSC 1.11 User Guide PDFs                                                                                                                     
     ========================                                                                                                                     
     … +11 lines (ctrl+o to expand)                                                                                                               
                                                                                                                                                  
❯ Tool loaded.                                                                                                                                    
                                                           
● Read 1 file (ctrl+o to expand)                                                                                                                  
                                                                                                                  
● Based on the OSC 1.11 Confidential Containers deployment guide, the runtimeclass depends on your deployment type:                               
                                                                                                                                                  
  RuntimeClass for Confidential Containers                                                                                                        
                                                                                                                                                  
  For Bare Metal deployments:                                                                                                                     
  - kata-cc                                                                                                                                       
                                                                                                                                                  
  For Peer Pods (cloud) deployments:                                                                                                              
  - kata-remote

  Key Details

  From the documentation (pages 17-18, 35):

  1. kata-cc is installed as a runtime class when you create the KataConfig CR for bare metal confidential containers deployments (Intel TDX, AMD
  SEV-SNP, IBM SE)
  2. kata-remote is installed as a runtime class when you create the KataConfig CR for peer pods deployments on cloud platforms (Azure, IBM Z, IBM
   LinuxONE)

  Both are installed as secondary, optional runtimes on the cluster, not as the primary runtime.

  Example Pod Manifest

  To use confidential containers, specify the runtimeClassName in your pod spec:

  apiVersion: v1
  kind: Pod
  metadata:
    name: ocp-cc-pod
  spec:
    runtimeClassName: kata-cc  # or kata-remote for peer pods
    containers:
      - name: <container_name>
        image: registry.access.redhat.com/ubi9/ubi:latest

  Source: Deploying Confidential Containers guide, pages 17-18 (bare metal), page 35 (Azure peer pods)

✻ Cogitated for 3m 39s
```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added an automated Claude skill that fetches and caches official OpenShift Sandboxed Containers guides to provide step-by-step answers with examples, YAML snippets, prerequisites, and page-cited references.

* **Documentation**
  * Added skill documentation describing selection rules, caching behavior, runtime requirements, and response expectations.

* **Chores**
  * Added tooling to download/cache PDF guides and updated .gitignore to avoid tracking downloaded PDFs.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/.claude/skills/osc-user-guide/SKILL.md
+++ b/.claude/skills/osc-user-guide/SKILL.md
@@ -1,0 +1,130 @@
+---
+name: osc-user-guide
+description: Provides comprehensive guidance on OpenShift Sandboxed Containers (OSC) by reading official Red Hat documentation. Use when questions involve OSC installation, KataConfig configuration, node setup, bare-metal or cloud deployments, PeerPods, confidential containers, TEE, Trustee, attestation, or troubleshooting.
+---
+
+# OSC User Guide Reader
+
+When this skill is triggered, follow these steps to answer the user's question using the official OSC documentation:
+
+## Instructions for Claude
+
+1. **Determine the OSC version** to use:
+
+   - If the user mentions a specific version (e.g., "with OSC 1.11"), use that version.
+   - Otherwise, determine the latest version using the WebFetch tool:
+   ```text
+   WebFetch(url='https://docs.redhat.com/en/documentation/openshift_sandboxed_containers',
+            prompt='What is the latest version of OpenShift Sandboxed Containers documentation? Return only the version number (e.g., 1.12)')
+   ```
+
+2. **Download PDFs** using the Bash tool with the version:
+   ```bash
+   bash .claude/skills/osc-user-guide/download-pdfs.sh <version>
+   ```
+
+   For example, if the version is 1.12:
+   ```bash
+   bash .claude/skills/osc-user-guide/download-pdfs.sh 1.12
+   ```
+
+   This script will:
+   - Dynamically discover all available guides from the Red Hat docs site
+   - Download PDFs into a per-version subdirectory (e.g., `.claude/skills/osc-user-guide/1.12/`)
+   - Skip already-cached files
+   - Works with any version — no hardcoded guide lists
+   - No dependencies required (just curl or wget)
+
+3. **Read the relevant PDF(s)** using Claude's built-in Read tool:
+
+   The download script output will show the directory where PDFs are stored.
+   Use the Read tool to access the documentation.
+
+   **How to read PDFs:**
+   - Use the Read tool with the `pages` parameter
+   - Maximum 20 pages per Read call
+   - Example: `Read(file_path='.claude/skills/osc-user-guide/1.12/<filename>.pdf', pages='1-20')`
+
+   **Reading strategy:**
+   - First, read the table of contents (typically pages 1-5) to understand the PDF structure
+   - Then read only the specific pages relevant to the user's question
+   - This is more efficient than reading the entire PDF
+   - If needed, read additional PDFs using the same TOC-first approach
+
+   **Guide selection — choose based on the user's question:**
+
+   The download script discovers all available guides automatically. PDF filenames
+   indicate their topic, e.g. `...-Deploying_confidential_containers_on_bare-metal_servers-en-US.pdf`.
+
+   Guides fall into three categories:
+   - **Deploying OpenShift Sandboxed Containers** — Operator install, KataConfig, kata runtime, peer pods
+   - **Deploying Confidential Containers** — CoCo, TEE, kata-cc/kata-remote, initdata, attestation
+   - **Deploying Red Hat Build of Trustee** — Trustee/KBS, attestation tokens, secrets
+
+   Each category may have per-platform variants (bare-metal, Azure, AWS, IBM Z, etc.).
+   Pick the guide matching the user's platform. If the user doesn't specify a platform, default to bare-metal.
+   If unsure which guide, read the most relevant one first — it's usually enough.
+
+4. **Answer the question comprehensively:**
+   - Provide a direct answer to the user's question
+   - Include step-by-step instructions when applicable
+   - Show code examples and YAML configurations
+   - List prerequisites and requirements
+   - Cite which guide(s) and page numbers the information comes from
+   - Mention related topics the user should know about
+
+## How This Skill Works
+
+This skill is **automatically triggered** by Claude when you ask questions about OpenShift Sandboxed Containers. You don't need to explicitly invoke it - just ask your question naturally.
+
+When triggered, the skill:
+1. Determines the OSC version (from user's question or latest available)
+2. Dynamically discovers and downloads PDFs for that version (first run only, cached afterward)
+3. Claude reads the table of contents first, then the relevant pages
+4. Claude provides a comprehensive answer with examples and citations
+
+Multiple versions can coexist — each version's PDFs are stored in their own subdirectory.
+
+## Example Questions That Trigger This Skill
+
+Ask questions naturally - the skill will activate automatically:
+
+**General deployment:**
+- "How do I install the OSC operator?"
+- "Show me KataConfig configuration options"
+- "How do I deploy OSC on a subset of worker nodes?"
+- "What are the node eligibility requirements?"
+
+**Confidential containers:**
+- "How do I deploy confidential containers?"
+- "What hardware is required for TEE?"
+- "How does attestation work in confidential containers?"
+
+**Version-specific questions:**
+- "Why is my KataConfig failing with OSC 1.11?"
+- "What changed in OSC 1.12 for peer pods?"
+
+**Trustee:**
+- "How do I configure Red Hat build of Trustee?"
+- "How do I create attestation tokens?"
+- "What is the key broker service?"
+
+**Troubleshooting:**
+- "Why is my KataConfig failing?"
+- "How do I verify Kata runtime installation?"
+
+## Requirements
+
+- `curl` or `wget` for downloading PDFs
+- Internet connection (for initial download only)
+- Read access to .claude/skills directory
+
+## Setup for Team Members
+
+**No manual setup required!**
+
+The first time you run this skill:
+1. Detects the OSC documentation version (from user's question or latest available)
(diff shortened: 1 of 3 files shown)


ldoktor (commented):
Looks interesting (quite verbose though, but that's AI...). My biggest problem is the fragility of the pdf urls so I had claude try some approaches and and it found those PDFs are in site-maps, which might be safer way to get all related documentations:

```
  version="$1"
  for i in {0..22}; do
    curl -sL "https://docs.redhat.com/sitemaps/docs/pages/sitemap-$i.xml" | \
      grep -oP "https://[^<]*openshift_sandboxed_containers/${version}/pdf/[^<]*\.pdf"
  done
```

I tried a few questions and it seemed to respond better than without the documentation. Sometimes the costs were higher but most of the time it was actually cheaper to use the docs as it didn't have to scan this projects sources/docs.

Btw ideal solution would IMO be to put docs in this project, but I guess that's not the option :-( (I like upstream, AI can just read the related documentation in-place in native format)

ldoktor:
Oh and just for the record, it has extra deps, one needs to have poppler-utils installed in order for that pdf scraper to work ;-)

wainersm:
Updated the skill to handle PDFs for OSC 1.12 as the guide is now split in more parts. I kept backward compatibilities so you can still do things like compare instructions between 1.11 and 1.12.

wainersm:
Updated the skill to use @ldoktor 's suggestion to use sitemap and avoid hard-coding the links to the pdfs

wainersm:
@beraldoleal mind to review as well?

.claude/skills/osc-user-guide/download-pdfs.sh
```diff
@@ -0,0 +1,110 @@
...
+
+download_pdf() {
+    local pdf_url="$1"
+    local filename
+    filename=$(basename "$pdf_url")
+    local pdf_file="${skill_dir}/${filename}"
+
+    # Check cache
+    if [ -f "$pdf_file" ]; then
+        local size_kb=$(($(stat -f%z "$pdf_file" 2>/dev/null || stat -c%s "$pdf_file") / 1024))
```
ldoktor: This check is used multiple times, how about extracting it into a function?

.claude/skills/osc-user-guide/SKILL.md
```diff
@@ -0,0 +1,123 @@
...
+   The download script output will show the paths to the available PDFs.
+   Use those paths to read the documentation.
+
+   **How to read PDFs:**
+   - Use the Read tool with the `pages` parameter
+   - Maximum 20 pages per Read call
+   - Example: `Read(file_path='<path_from_download_script_output>', pages='1-20')`
+
+   **Reading strategy:**
+   - Determine which PDF is most relevant to the user's question
```
ldoktor: Our guidelines start with header, leagel notice and table of content. Wouldn't it be useful to tell claude to first look at the table of content (~5-10 pages should do) and based on the table of content ask for more?

.claude/skills/osc-user-guide/download-pdfs.sh
```diff
@@ -0,0 +1,110 @@
...
+        fi
+        rm -f "$pdf_file"
+    fi
+
+    echo "  [downloading] ${filename}..."
+
+    if command -v curl &> /dev/null; then
+        curl --fail -sL "$pdf_url" -o "$pdf_file" || { echo "  [FAILED] ${filename} - download error"; rm -f "$pdf_file"; return 1; }
+    elif command -v wget &> /dev/null; then
+        wget -q "$pdf_url" -O "$pdf_file" || { echo "  [FAILED] ${filename} - download error"; rm -f "$pdf_file"; return 1; }
```
ldoktor, line 64: it might be useful to use the version as subdir (`.claude/skills/osc-user-guide/1.11/*`) to be able to respond questions about various versions (and also to fetch new one in case it's available)

.claude/skills/osc-user-guide/SKILL.md
```diff
@@ -0,0 +1,123 @@
...
+- `curl` or `wget` for downloading PDFs
+- Internet connection (for initial download only)
+- Read access to .claude/skills directory
+
+## Setup for Team Members
+
+**No manual setup required!**
+
+The first time you run this skill:
+1. Automatically detects the latest OSC documentation version
```
ldoktor: Do we have to emphasize the latest version or wouldn't it be possible to mention it but also suggest the user might mention their version manually? Eg. questions like `Why is my KataConfig failing with 1.11 osc?` (which actually works well and it executes: `Bash(bash .claude/skills/osc-user-guide/download-pdfs.sh 1.11)`)

ldoktor: and it's even clever enough to delete the previous (different version) pdfs... Still I think we should use subfolders hierarchy to allow queries for different versions.

bpradipt:
An orthogonal thought - should we have an agent friendly doc derived from the public OSC user guide, workshop doc, blogs etc and have it in this repo, and let the agent skill refer to that?

ldoktor:
If it's fine sharing the documentation sources than having them here would be the best solution (and one wouldn't even need skill for that as "docs" is pretty known way of sharing docs and claude and others would simply use it, even local llm...).

.claude/skills/osc-user-guide/SKILL.md
wainersm: Hi @ldoktor ,

my first implementation [1] of this skill was a lot more sophisticated and optimized, however, I had its own PDF reader in python. But @bpradipt suggested to use claude's builtin PDF, and I implemented that new version. It turns out that the builtin reader is very simple: it just read entire PDFs (in chunks of 20 pages if I remember). The builtin consumes more tokens (see analysis in [1]) and fill the context more easily but, on the other hand, it is one less piece of code that we would need to maintain (i.e. the custom pdf reader). At this point I don't care about tokens (as the company doesn't either) and even with a filled context it seems to produce correct answers, so I decided to stick with the simplest implementation.

[1] https://github.com/openshift/sandboxed-containers-operator/pull/1798

.claude/skills/osc-user-guide/SKILL.md
```diff
@@ -0,0 +1,123 @@
...
+2. Discovers and downloads all available PDF guides for that version from the Red Hat docs site
+
+On subsequent runs, it uses the cached PDFs (unless a new version is detected).
```
apinnick: How does the skill detect new PDFs if they have the same version number?

.claude/skills/osc-user-guide/download-pdfs.sh
wainersm: Hi @ldoktor ,

I've added this feature. If user asks for a given version it will look first at the local cache, otherwise download it. The pdfs are now saved in directories per-version.

@apinnick ^ that answers your question below. BTW, thanks for you review too.

.claude/skills/osc-user-guide/SKILL.md
wainersm: @ldoktor I was wrong, the builtin PDF reader is able to read random pages. So I implemented the algorithm you describe above: read the TOC then only the relevant pages. I don't know if it will degrade the accuracy of the answers (because less page in the context) but let's give it a try.

wainersm:
> An orthogonal thought - should we have an agent friendly doc derived from the public OSC user guide, workshop doc, blogs etc and have it in this repo, and let the agent skill refer to that?

It seems a good idea to me. I was thinking in indexing the internal guides, but having all the other docs is even better. That reminds me the knowledge trove doc that we had years ago, now we can have it in form of an agent/skill.

ldoktor (approved):
lgtm, thanks (one thing I noticed is that in some cases it prefers the documentation rather than the osc sources now, for example question like: `what kata version does osc 1.11 use?` failed, but follow-up with `isn't there an rpm version mentioned somewhere?` made it look at osc sources and detected `OSC 1.11 uses Kata Containers 3.25.0`. But I guess claude will improve by usage :-)

ldoktor (approved):

PR #1818, opened 2026-03-09
peerpods: Use downstream CAA

gkurz:
Use the CAA submodule. This ensures that the operator is consistent with the rest of the downstream peer pods code.


littlejawa (approved):
/lgtm
Thanks @gkurz !

snir911 (approved):

PR #1828, opened 2026-03-11
tools: add script to update rpms.lock.yaml file

dbkreling:
Whenever we update the RPMs we install in the images are updated, we (as developers) need to update the rpms.lock.yaml file.

The update is a multi-step process that requires installation of multiple packages, deployment of a disposable container and files updates.

The automation of the process will be a great benefit when that task is necessary

Issue: [KATA-4748](https://issues.redhat.com/browse/KATA-4748)

### DIFF
--- a/scripts/update-rpms-lock-yaml/update-rpms-lock.sh
+++ b/scripts/update-rpms-lock-yaml/update-rpms-lock.sh
@@ -1,0 +1,172 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# This script automates generating rpms.lock.yaml (notes provided by @jrope during Q4 2025).
+# It must be run from the directory containing rpms.in.yaml.
+# It also updates the redhat.repo and ubi.repo files in the same directory.
+# Note: if all "url" lines in rpms.lock.yaml reference "cdn.redhat.com" (and none
+# reference "cdn-ubi.redhat.com"), then ubi.repo is not needed and can be removed.
+#
+# Requirements on host:
+# - podman
+# - activation key
+# - organization id
+# - registry username
+#
+# Note: An activation key can be created following this documentation:
+# https://konflux.pages.redhat.com/docs/users/building/activation-keys-subscription.html#create-the-activation-key
+#
+# Inputs (flags or env vars):
+# - -k ACTIVATION_KEY (or env ACTIVATION_KEY)
+# - -o ORG_ID (or env ORG_ID)
+# - -u REGISTRY_USERNAME (or env REGISTRY_USERNAME)
+# - -p REGISTRY_PASSWORD (or env REGISTRY_PASSWORD)
+# - -i IMAGE (defaults to registry.access.redhat.com/ubi9/ubi:latest)
+#
+# Example:
+#   ./update-rpms-lock.sh \
+#     -k "$ACTIVATION_KEY" -o "$ORG_ID" \
+#     -u "$REGISTRY_USERNAME" -p "$REGISTRY_PASSWORD"
+#
+# Author: Julien Rope <jrope@redhat.com>
+# Author: Daniel Kreling <dkreling@redhat.com>
+# Made with the help of Cursor and Claude Opus 4.6
+
+
+IMAGE_DEFAULT="registry.access.redhat.com/ubi9/ubi:latest"
+
+usage() {
+  cat <<EOF
+Usage: $(basename "$0") [-k ACTIVATION_KEY] [-o ORG_ID] [-u REGISTRY_USERNAME] [-p REGISTRY_PASSWORD] [-i IMAGE]
+
+Environment variables may be used instead of flags: ACTIVATION_KEY, ORG_ID, REGISTRY_USERNAME, REGISTRY_PASSWORD, IMAGE.
+Defaults: IMAGE=${IMAGE_DEFAULT}
+
+Note: run from the folder containing rpms.in.yaml.
+EOF
+}
+
+ACTIVATION_KEY="${ACTIVATION_KEY:-}"
+ORG_ID="${ORG_ID:-}"
+REGISTRY_USERNAME="${REGISTRY_USERNAME:-}"
+REGISTRY_PASSWORD="${REGISTRY_PASSWORD:-}"
+IMAGE="${IMAGE:-${IMAGE_DEFAULT}}"
+
+while getopts ":k:o:u:p:i:h" opt; do
+  case "${opt}" in
+    k) ACTIVATION_KEY="$OPTARG" ;;
+    o) ORG_ID="$OPTARG" ;;
+    u) REGISTRY_USERNAME="$OPTARG" ;;
+    p) REGISTRY_PASSWORD="$OPTARG" ;;
+    i) IMAGE="$OPTARG" ;;
+    h) usage; exit 0 ;;
+    :) echo "Missing argument for -$OPTARG" >&2; usage; exit 2 ;;
+    \?) echo "Unknown option -$OPTARG" >&2; usage; exit 2 ;;
+  esac
+done
+shift $((OPTIND-1)) || true
+
+# Extract UBI version from the image name (e.g. "ubi10" -> "10", "ubi9" -> "9")
+UBI_VERSION=$(echo "${IMAGE}" | grep -oP 'ubi\K[0-9]+')
+if [[ -z "${UBI_VERSION}" ]]; then
+  echo "Error: could not determine UBI version from IMAGE: ${IMAGE}" >&2
+  exit 1
+fi
+
+# Basic validations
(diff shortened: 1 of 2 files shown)


scripts/update-rpms-lock-yaml/update-rpms-lock.sh
```diff
@@ -0,0 +1,150 @@
...
+echo "Copying yum repo files (redhat.repo, ubi.repo) to /source..."
+cp -f /etc/yum.repos.d/redhat.repo /source/redhat.repo
+cp -f /etc/yum.repos.d/ubi.repo /source/ubi.repo
+
+cd /source
+
+# Adjust architecture placeholders for multi-arch
+echo "Replacing architecture-specific values with \$basearch placeholders..."
+sed -i "s/$(uname -m)/\$basearch/g" redhat.repo
+sed -i "s/ubi-9-codeready-builder/codeready-builder-for-ubi-9-\$basearch/" ubi.repo
```
gkurz: This would only work for a ubi-9 base image but we have ubi-10 ones as well. You should come up with a smarter regex I guess.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
```diff
@@ -63,6 +66,13 @@ while getopts ":k:o:u:p:i:h" opt; do
 done
 shift $((OPTIND-1)) || true
 
+# Extract UBI version from the image name (e.g. "ubi10" -> "10", "ubi9" -> "9")
+UBI_VERSION=$(echo "${IMAGE}" | grep -oP 'ubi\K[0-9]+')
```
gkurz, line 70: Basing on the image name is fragile. I was more thinking to a more powerful regex in `sed`.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
```diff
@@ -126,15 +137,20 @@ cd /source
 # Adjust architecture placeholders for multi-arch
 echo "Replacing architecture-specific values with \$basearch placeholders..."
 sed -i "s/$(uname -m)/\$basearch/g" redhat.repo
-sed -i "s/ubi-9-codeready-builder/codeready-builder-for-ubi-9-\$basearch/" ubi.repo
-sed -i "s/\[ubi-9/[ubi-9-for-\$basearch/" ubi.repo
+sed -i "s/ubi-${UBI_VERSION}-codeready-builder/codeready-builder-for-ubi-${UBI_VERSION}-\$basearch/" ubi.repo
```
gkurz, line 152: With something like below, you don't need to extract `UBI_VERSION`.

```
sed -i -E 's/ubi-([0-9]+)-codeready-builder/codeready-builder-for-ubi-\1-\$basearch/g' ubi.repo
```

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
```diff
@@ -0,0 +1,150 @@
...
+
+Environment variables may be used instead of flags: ACTIVATION_KEY, ORG_ID, REGISTRY_USERNAME, REGISTRY_PASSWORD, IMAGE.
+Defaults: IMAGE=${IMAGE_DEFAULT}
+
+Note: run from the folder containing rpms.in.yaml.
+EOF
+}
+
+ACTIVATION_KEY="${ACTIVATION_KEY:-}"
+ORG_ID="${ORG_ID:-${ORG_ID_DEFAULT}}"
```
gkurz:
```
$ ./scripts/update-rpms-lock-yaml/update-rpms-lock.sh -h
./scripts/update-rpms-lock-yaml/update-rpms-lock.sh: line 50: ORG_ID_DEFAULT: unbound variable
```

What's the story around `ORG_ID_DEFAULT` ?

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
```diff
@@ -0,0 +1,166 @@
...
+# Adjust architecture placeholders for multi-arch
+echo "Replacing architecture-specific values with \$basearch placeholders..."
+sed -i "s/$(uname -m)/\$basearch/g" redhat.repo
+sed -i "s/ubi-${UBI_VERSION}-codeready-builder/codeready-builder-for-ubi-${UBI_VERSION}-\$basearch/" ubi.repo
+sed -i "s/\[ubi-${UBI_VERSION}/[ubi-${UBI_VERSION}-for-\$basearch/" ubi.repo
+
+# Enable source repositories in the ubi.repo and redhat.repo files
+# This finds lines containing "-source]" and the first "enabled = 0" following it
+echo "Enabling source RPM repositories in redhat.repo and ubi.repo..."
+sed -i "/-source-rpms]/,/enabled = 0/ s/enabled = 0/enabled = 1/" redhat.repo
```
littlejawa: You should not enable all source RPMs for all repositories in redhat.repo. A lot of them are unneeded, and if they're not enabled in your activation key, it will generate a lot of warning.
I think that's the cause of the problem you have when running the script and getting a huge list of warnings.
From my experience, the source repos we need are already enabled in redhat.repo by default (because you enabled them in your activation key), so you don't need to edit it.
Only "ubi.repo" needs to be edited here.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
```diff
@@ -0,0 +1,166 @@
...
+sed -i "/-source-rpms]/,/enabled = 0/ s/enabled = 0/enabled = 1/" ubi.repo
+
+# Allow repos that do not exist (e.g. EUS, some source repos) to be skipped without error
+echo "Setting skip_if_unavailable=1 for all enabled repositories..."
+sed -i "/^enabled = 1$/a skip_if_unavailable = 1" redhat.repo
+sed -i "/^enabled = 1$/a skip_if_unavailable = 1" ubi.repo
```
littlejawa: I would not skip them. If I understand correctly, you added that because of the above warnings?
You probably should try without the source repos enabled, and if it works, remove this "skip" section entirely.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
gkurz: I had a lot of warnings. They go away when I remove this line.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
gkurz: I confirm that the script ran smoothly with these lines removed as well.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
gkurz: Oops... ubi-10 already has multi-arch support so this is really only needed for ubi-9. Sorry for the misguidance.

Maybe then just add a comment to make it clear.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
gkurz: Not needed as per https://github.com/openshift/sandboxed-containers-operator/pull/1828#discussion_r2958841599.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
gkurz: Not needed as per https://github.com/openshift/sandboxed-containers-operator/pull/1828#discussion_r2958841599.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
dbkreling: Sorry, some leftover of previous implementation. Fixed.

scripts/update-rpms-lock-yaml/update-rpms-lock.sh
```diff
@@ -0,0 +1,160 @@
...
+  -e REGISTRY_USERNAME="${REGISTRY_USERNAME}" \
+  -e REGISTRY_PASSWORD="${REGISTRY_PASSWORD}" \
+  -e UBI_VERSION="${UBI_VERSION}" \
+  "${IMAGE}" bash -lc '
+set -euo pipefail
+
+# Ensure required tools (subscription-manager, skopeo, python3-pip) are available
+# Notes used "pip"; on UBI9 the package is typically "python3-pip".
+echo "Installing required packages: subscription-manager, skopeo, python3-pip..."
+dnf install -y subscription-manager skopeo python3-pip >/dev/null
```
gkurz: This assumes that `dnf` is available in the base image but UBI9 minimal has `microdnf` instead...

dbkreling:
This script lives now in its own repository inside OSC Gitlab space.

PR #1829, opened 2026-03-11
Fix OCP version check for baremetal CoCo

gkurz:
Baremetal CoCo was TP in OSC 1.11 with a very limited _support_ matrix : OCP 4.20 only, starting with OCP 4.20.6.

Baremetal CoCo will be GA with OSC 1.12 and the support matrix is bigger : OCP 4.19, 4.20 and 4.21, starting with specific z-stream versions that will ship the appropriate kata-containers RPM.

Adapt the validation logic to handle these new OCP versions. The respective z-streams are hardcoded and not configurable for the sake of simplicity. They can be adjusted just before the release of OSC 1.12 if needed.

A _hidden_ knob is provided to override the OCP version used in the validation check. This is to allow testing of
baremetal CoCo on with existing OCP versions that would fail obviously validation.

### DIFF
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"slices"
+	"strings"
 
 	semver "github.com/Masterminds/semver/v3"
 	configv1 "github.com/openshift/api/config/v1"
@@ -159,12 +161,16 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 }
 
 // validateOCPVersion checks if the current OpenShift cluster is valid based on
-// minimal version in minOCPVersion.
-// Returns false if OCP version is < to minimal version
-// Returns true otherwise
+// minimal versions in minOCPVersions.
+// Returns false if OCP version is < to all minimal versions
+// Returns false if OCP version is < to the minimal version matching major.minor
+// Returns true otherwise (i.e. assume that any higher OCP version has support)
 // Returns error if cluster version is not available or cannot be retrieved.
 func (r *KataConfigOpenShiftReconciler) validateOCPVersion() (bool, error) {
-	minOCPVersion := "4.20.6"
+	// Sorted slice of minimal OCP z-stream releases.
+	minOCPVersions := slices.Sorted(slices.Values([]string{
+		"4.20.6",
+	}))
 
 	clusterVersion := &configv1.ClusterVersion{}
 	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "version"}, clusterVersion)
@@ -182,22 +188,50 @@ func (r *KataConfigOpenShiftReconciler) validateOCPVersion() (bool, error) {
 		return false, fmt.Errorf("unable to parse current OCP version %s: %w", currentVersion, err)
 	}
 
-	minVersion := minOCPVersion
-	min, err := semver.NewVersion(minVersion)
-	if err != nil {
-		return false, fmt.Errorf("invalid minimum version format %s: %w", minVersion, err)
+	supported := false
+	for i, minVersion := range minOCPVersions {
+		min, err := semver.NewVersion(minVersion)
+		if err != nil {
+			return false, fmt.Errorf("invalid minimum version format %s: %w", minVersion, err)
+		}
+
+		if current.Major() < min.Major() {
+			// major is too low to be even considered.
+			break
+		}
+
+		if current.Major() == min.Major() {
+			if current.Minor() < min.Minor() {
+				// minor is too low to be even considered.
+				break
+			}
+
+			if current.Minor() == min.Minor() {
+				// Same major/minor, just compare the patch.
+				supported = !current.LessThan(min)
+				break
+			}
+		}
+
+		if i == len(minOCPVersions)-1 {
+			// Higher major/minor than the last version is assumed to have proper support.
+			supported = true
+		}
 	}
 
-	if current.LessThan(min) {
-		r.Log.Info("WARNING: OpenShift version does not support CoCo bare metal", "version", currentVersion, "minVersion", minVersion)
+	if !supported {
+		minVersions := strings.Join(minOCPVersions, ", ")
+
+		r.Log.Info("WARNING: OpenShift version does not support CoCo bare metal", "version", currentVersion, "minVersions", minVersions)
 		cond := r.retrieveInProgressConditionForChange()
 		cond.Status = corev1.ConditionFalse
 		cond.Reason = "UnsupportedOCPVersion"
-		cond.Message = fmt.Sprintf("OpenShift version %s does not support CoCo bare metal (minimum required: %s or higher)", currentVersion, minVersion)
+		cond.Message = fmt.Sprintf("OpenShift version %s does not support CoCo bare metal (minimum required: %s or higher)", currentVersion, minVersions)
 
 		return false, nil
 	}
 
+	r.Log.Info("OpenShift version supports CoCo bare metal", "version", currentVersion)
 	return true, nil
 }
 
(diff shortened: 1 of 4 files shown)


controllers/confidential_handler.go
```diff
@@ -121,6 +124,34 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 	return nil
 }
 
+// isOCPVersionLessThan checks if the current OpenShift cluster version is less than the specified minimum version.
```
bpradipt: A better place imho is utils.go.

controllers/confidential_handler.go
```diff
@@ -124,50 +124,60 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
...
-// isOCPVersionLessThan checks if the current OpenShift cluster version is less than the specified minimum version.
-// Returns (true, version, nil) if current version < minVersion, (false, version, nil) if current >= minVersion.
+// validateOCPVersion checks if the current OpenShift cluster is valid based on
+// minimal version in minOCPVersion.
+// Returns false if OCP version is < to minimal version
+// Returns true otherwise
 // Returns error if cluster version is not available or cannot be retrieved.
-func (r *KataConfigOpenShiftReconciler) isOCPVersionLessThan(minVersion string) (bool, string, error) {
+func (r *KataConfigOpenShiftReconciler) validateOCPVersion() (bool, error) {
+	minOCPVersion := "4.20.6"
```
bpradipt: preferably this should be a global constant

controllers/confidential_handler.go
```diff
@@ -135,7 +135,9 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialPeerPods(state Feature
 func (r *KataConfigOpenShiftReconciler) validateOCPVersion() (bool, error) {
 	// Sorted slice of minimal OCP z-stream releases.
 	minOCPVersions := slices.Sorted(slices.Values([]string{
-		"4.20.6",
+		"4.19.27",
+		"4.20.17",
+		"4.21.8",
```
bpradipt, line 175: nit: It would be better to have this in a separate method or declared at package scope for easy visibility. 

```
var minOCPVersions = []string{
	"4.19.27",
	"4.20.17",
	"4.21.8",
}

or 

var minOCPVersions = slices.Sorted([]string{
	"4.19.27",
	"4.20.17",
	"4.21.8",
})
```

controllers/confidential_handler.go
```diff
@@ -140,15 +141,18 @@ func (r *KataConfigOpenShiftReconciler) validateOCPVersion() (bool, error) {
...
-	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: "version"}, clusterVersion)
-	if err != nil {
-		return false, err
-	}
-
-	currentVersion := clusterVersion.Status.Desired.Version
+	currentVersion := os.Getenv("BM_COCO_OVERRIDE_OCP_VERSION")
 	if currentVersion == "" {
-		return false, fmt.Errorf("cluster version not available yet")
+		clusterVersion := &configv1.ClusterVersion{}
```
bpradipt, line 181: We have clusterVersion retrieval code in other places as well :-( 
Probably add a FIXME to refactor this into a common util method

controllers/confidential_handler.go
gkurz: The code is moved there on purpose as a preparatory step, in order to have pretty diff in the second commit that doesn't cross files.

controllers/confidential_handler.go
gkurz: I don't quite want this to be exposed and abusively used for unrelated purposes :wink:

controllers/confidential_handler.go
gkurz: Same remark. Restricted scope.

controllers/confidential_handler.go
```diff
@@ -148,18 +154,38 @@ func (r *KataConfigOpenShiftReconciler) validateOCPVersion() (bool, error) {
...
-	if err != nil {
-		return false, fmt.Errorf("invalid minimum version format %s: %w", minVersion, err)
+	supported := false
+	for i, minVersion := range minOCPVersions {
+		min, err := semver.NewVersion(minVersion)
+		if err != nil {
+			return false, fmt.Errorf("invalid minimum version format %s: %w", minVersion, err)
+		}
+
+		if current.Major() < min.Major() || current.Minor() < min.Minor() {
```
littlejawa: OCP 5.0 will be considered unsupported, because `current.Minor() < min.Minor()`, for all elements in the list.

Maybe you need:
```suggestion
		if current.Major() < min.Major() || (current.Major() == min.Major() && current.Minor() < min.Minor()) {
```

controllers/confidential_handler.go
gkurz: Honestly, I hope we will get rid of this function as soon as we're able to support baremetal coco anywhere.
Also, other users of ClusterVersion have separate needs so I don't quite see what refactoring you have in mind. Can you please elaborate ?

controllers/confidential_handler.go
gkurz: Good catch ! Will fix.

controllers/confidential_handler.go
bpradipt: My thinking was having a single util function to return the ClusterVersion

gkurz: I don't quite want to go sideways at this point so I'll just add a FIXME :wink:

controllers/confidential_handler.go
gkurz: I've broken down the check for clarity and claude helped to add tests for `make test` :wink:

littlejawa (approved):
/lgtm

bpradipt (approved):
/lgtm
thanks @gkurz

gkurz:
In order to test Baremetal CoCo with _unsupported_ OCP versions, it is required to manually override the version to a _supported_ value (e.g. 4.22.0) using the `BM_COCO_OVERRIDE_OCP_VERSION` environment variable.

```
# Add BM_COCO_OVERRIDE_OCP_VERSION=4.22.0 to the CSV of an already installed OSC operator 1.12
oc get csv sandboxed-containers-operator.v1.12.0 -n openshift-sandboxed-containers-operator -o json \
| jq '.spec.install.spec.deployments[0].spec.template.spec.containers |= map(if .name == "manager" then if (.env | any(.name == "BM_COCO_OVERRIDE_OCP_VERSION")) then (.env[] | select(.name == "BM_COCO_OVERRIDE_OCP_VERSION").value) = "4.22.0" else .env += [{"name": "BM_COCO_OVERRIDE_OCP_VERSION", "value": "4.22.0"}] end else . end)' \
| oc apply -f -
```

Cc @xiangchunfu @pezhang

gkurz:
> In order to test Baremetal CoCo with _unsupported_ OCP versions, it is required to manually override the version to a _supported_ value (e.g. 4.22.0) using the `BM_COCO_OVERRIDE_OCP_VERSION` environment variable.
> 
> ```
> # Add BM_COCO_OVERRIDE_OCP_VERSION=4.22.0 to the CSV of an already installed OSC operator 1.12
> oc get csv sandboxed-containers-operator.v1.12.0 -n openshift-sandboxed-containers-operator -o json \
> | jq '.spec.install.spec.deployments[0].spec.template.spec.containers |= map(if .name == "manager" then if (.env | any(.name == "BM_COCO_OVERRIDE_OCP_VERSION")) then (.env[] | select(.name == "BM_COCO_OVERRIDE_OCP_VERSION").value) = "4.22.0" else .env += [{"name": "BM_COCO_OVERRIDE_OCP_VERSION", "value": "4.22.0"}] end else . end)' \
> | oc apply -f -
> ```
> 
> Cc @xiangchunfu @pezhang

Got positive feedback from @pezhang on slack that this is appropriate for baremetal coco testing.

PR #1842, opened 2026-03-13
osc-podvm-payload: update the cloud-api-adaptor submodule to latest

littlejawa:

PR #1844, opened 2026-03-14
Fix `make-test` failures in CI

gkurz:
Fix `make test` failures when `git clone --recurse-submodules` doesn't clone submodules.


gkurz:
Fix already merged with #1842 .

PR #1847, opened 2026-03-16
runtimeclass: fix nvidia sandbox device plugin label

thejasn:
Update sandbox device plugin label to match latest nvidia gpu-operator label for kata sandbox device plugins

Fixes: https://redhat.atlassian.net/browse/KATA-4756

Label verification fails since old label is not used anymore.

**- What I did**
Update label

**- How to verify it**
TBD

Fix incorrect label for kata sandbox device plugin

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -96,11 +96,11 @@ const (
 var (
 	// node labels for NVIDIA GPU
 	nvidiaGPUNodeLabels = map[string]string{
-		"nvidia.com/gpu.present":                      "true",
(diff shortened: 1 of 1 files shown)


bpradipt (approved):
/lgtm
Thanks @thejasn

gkurz:
> /hold until ref jira ticket

Given the extremely tight schedule, I think we should not hold this PR because of a jira ticket reference... especially in this case, the code was never released so we don't need release notes. This is basically a follow-up to the work you did previously.

PR #1909, opened 2026-03-24
chore(deps): Manually update osc-must-gather to 284e6a028

dbkreling:
Manually changing it because the latest on-push pipeline failed on Konflux. 

Image fetched from the latest image in quay.io [1]

[1] https://quay.io/repository/redhat-user-workloads/ose-osc-tenant/osc-must-gather?tab=tags&tag=latest

Reference: https://github.com/openshift/sandboxed-containers-operator/pull/1899


gkurz (approved):
This is indeed the latest build.

/lgtm

I'd still like to have some feedback from @littlejawa if we manage to contact him.

littlejawa (approved):
/lgtm
I see that the build succeeded, and then some other task failed (infra issue) and prevented the nudge PR to be sent.

PR #1917, opened 2026-03-25
peerpods: fix CAA daemonset recycling on every reconciliation

wainersm:
## Problem

  The CAA (cloud-api-adaptor) DaemonSet gets recycled on every reconciliation loop, causing peer-pod VMs to lose their ttrpc connections and
  become unmanageable.

  **Root cause**: The operator's `enablePeerPodsMiscConfigs()` used a pattern where it built a fresh DaemonSet object and called `Update()` on
  every reconciliation. Because the freshly-built object didn't include API server defaults (like `HostPath.Type`, `FieldRef.APIVersion`,
  `MaxSurge`, `RevisionHistoryLimit`), the API server treated each Update as a spec change, incrementing the DaemonSet generation and triggering a
   rolling update of CAA pods.

  **Impact**: When a CAA pod is recycled, the new instance has no knowledge of existing peer-pod VMs. The kata-remote runtime shims lose their
  ttrpc connection to the old CAA, resulting in `ttrpc: closed` errors. Affected pods cannot be terminated gracefully — they hang in `Terminating`
   state because the delete request goes to a CAA that doesn't know about them.

  **Evidence**: This was uncovered while analyzing a [failing Azure peer-pods Prow job](https://prow.ci.openshift.org/view/gs/test-platform-result
  s/logs/periodic-ci-openshift-sandboxed-containers-operator-devel-downstream-candidate421-azure-ipi-peerpods/2036283647832625152). The
  must-gather showed the CAA DaemonSet at `generation: 4` with 4 rounds of pod create/delete events in its lifetime. The build log confirmed test
  C00317 (delete KataConfig with running workloads) failed because a peer-pod could not be terminated after its CAA pod was recycled mid-test.

  ## Test coverage

  The new envtest suite (`controllers/peerpods_test.go`) directly validates the fix by creating a CAA DaemonSet via `CreateOrUpdate`, then calling
   it again on the API server's version and asserting `OperationResultNone` (no update). This test caught two additional server-default fields
  (`RevisionHistoryLimit` and `MaxSurge`) that were missing from the initial fix.

### DIFF
--- a/controllers/peerpods.go
+++ b/controllers/peerpods.go
@@ -43,7 +43,10 @@ func MountProgagationRef(mode corev1.MountPropagationMode) *corev1.MountPropagat
 	return &mode
 }
 
-func (r *KataConfigOpenShiftReconciler) processDaemonsetForCAA() *appsv1.DaemonSet {
+// mutateCAADaemonSet sets the desired state for the cloud-api-adaptor DaemonSet.
+// It builds the complete spec including provider-specific volumes so that it is
+// safe to call on both new and existing objects (idempotent, no appending).
+func (r *KataConfigOpenShiftReconciler) mutateCAADaemonSet(ds *appsv1.DaemonSet) error {
 	var (
 		runPrivileged                = true
 		runAsUser              int64 = 0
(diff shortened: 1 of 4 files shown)


gkurz:
> When CAA pods are recycled, the new CAA instance has no knowledge of existing peer-pod VMs. 

This is a terrible flaw in a k8s environment where pods can be recycled for a variety of reasons. Switching to `CreateOrUpdate` is certainly an improvement for OSC but it is merely a band aid with respect to the CAA limitation.

controllers/peerpods.go
```diff
@@ -257,53 +364,97 @@ func (r *KataConfigOpenShiftReconciler) processProviderConfigCAA(ds *appsv1.Daem
...
+			})
 		}
+	}
 
-		return nil
-	default:
-		return nil
+	// Set pod spec fields individually to preserve server-set defaults
+	// (e.g. terminationGracePeriodSeconds, dnsPolicy) on existing objects
+	ds.Spec.Template.Spec.ServiceAccountName = "default"
```
thejasn, line 227: This approach duplicates a bunch of values from `processDaemonsetForCAA()`, and if we want to maintain this I recommend maybe using common builder funcs.

thejasn:
> an improvement for OSC but it is merely a band aid with respect to the CAA limitation 

I would agree this is a band aid fix and moving to `CreateOrUpdate()` definitely is in the right direction for a long term fix (separate peerpods controller) and this is not the only instance where we are missing `CreateOrUpdate()` (eg: runtimeclasses as well)

I think we could handle such limitations in a much better manner on a separate controller i.e. 
a separate controller that,
- Watches
  - kataconfig
  - caa daemonset
  - job, cm, secrets, etc resources that are required/managed for peerpods
- Reconciles
  - if there is a diff (create/update) 
  
IMO, 
- If we are going to properly fix this then I think we can fix the above issue in a simpler but hacky way similar to how other resources are created for peerpods (webhooks, etc) (reduces uncertainty for 1.12 release)
```patch
diff --git a/controllers/peerpods.go b/controllers/peerpods.go
index bd3b1719..3fa0223a 100644
--- a/controllers/peerpods.go
+++ b/controllers/peerpods.go
@@ -295,18 +295,16 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPodsMiscConfigs() error {
 		return err
 	}
 
-	err := r.Client.Update(context.TODO(), ds)
-	if err != nil && k8serrors.IsNotFound(err) {
-		r.Log.Error(err, "cloud-api-adaptor daemonset doesn't exist. Creating")
-		err = r.Client.Create(context.TODO(), ds)
-		if err != nil {
+	if err := r.Client.Create(context.Background(), ds); err != nil {
+		// Check if the CAA ds already exists
+		if !k8serrors.IsAlreadyExists(err) {
 			r.Log.Error(err, "failed to create cloud-api-adaptor daemonset")
 			return err
 		}
 	}
 
 	// Create the mutating webhook deployment
-	err = r.createMutatingWebhookDeployment()
+	err := r.createMutatingWebhookDeployment()
 	if err != nil {
 		r.Log.Info("Error in creating mutating webhook deployment for peerpods", "err", err)
 		return err

```

- If we are not fixing this anytime soon, the above direction definitely makes sense

controllers/peerpods.go
wainersm: Yeah, claude generated a bad code (as I would expect) that we can improve

wainersm:
> > an improvement for OSC but it is merely a band aid with respect to the CAA limitation
> 
> I would agree this is a band aid fix and moving to `CreateOrUpdate()` definitely is in the right direction for a long term fix (separate peerpods controller) and this is not the only instance where we are missing `CreateOrUpdate()` (eg: runtimeclasses as well)
> 
> I think we could handle such limitations in a much better manner on a separate controller i.e. a separate controller that,
> 
>     * Watches
>       
>       * kataconfig
>       * caa daemonset
>       * job, cm, secrets, etc resources that are required/managed for peerpods
> 
>     * Reconciles
>       
>       * if there is a diff (create/update)
> 
>

I'm all in for improving the code but...
 
> IMO,
> 
>     * If we are going to properly fix this then I think we can fix the above issue in a simpler but hacky way similar to how other resources are created for peerpods (webhooks, etc) (reduces uncertainty for 1.12 release)
>

... for 1.12 release I think we just need the simplest fix as possible. As I am not expert here, I would like help to assess if what claude proposes makes sense. It does make sense to my eyes, but again I'm not knowledgeable of that area. 
 
> 
> ```diff
> diff --git a/controllers/peerpods.go b/controllers/peerpods.go
> index bd3b1719..3fa0223a 100644
> --- a/controllers/peerpods.go
> +++ b/controllers/peerpods.go
> @@ -295,18 +295,16 @@ func (r *KataConfigOpenShiftReconciler) enablePeerPodsMiscConfigs() error {
>  		return err
>  	}
>  
> -	err := r.Client.Update(context.TODO(), ds)
> -	if err != nil && k8serrors.IsNotFound(err) {
> -		r.Log.Error(err, "cloud-api-adaptor daemonset doesn't exist. Creating")
> -		err = r.Client.Create(context.TODO(), ds)
> -		if err != nil {
> +	if err := r.Client.Create(context.Background(), ds); err != nil {
> +		// Check if the CAA ds already exists
> +		if !k8serrors.IsAlreadyExists(err) {
>  			r.Log.Error(err, "failed to create cloud-api-adaptor daemonset")
>  			return err
>  		}
>  	}
>

I didn't understand how it improves the situation. So instead of update, it is going to create a new daemonset. If it exists then it's no-op. What happens if it really needs an update (say, the caa image changed)?
  
>  	// Create the mutating webhook deployment
> -	err = r.createMutatingWebhookDeployment()
> +	err := r.createMutatingWebhookDeployment()
>  	if err != nil {
>  		r.Log.Info("Error in creating mutating webhook deployment for peerpods", "err", err)
>  		return err
> ```
> 
>     * If we are not fixing this anytime soon, the above direction definitely makes sense

thejasn:
> I didn't understand how it improves the situation. So instead of update, it is going to create a new daemonset. If it exists then it's no-op. What happens if it really needs an update (say, the caa image changed)?

You are right here, this would not work for updates. Since I was also looking for the simplest change possible I kind of borrowed the pattern with webhook deployments setup that is done just after CAA DS creation (they have this bug as well).

> for 1.12 release I think we just need the simplest fix as possible. 

We could always explicitly check fields that the OSC operator cares i.e.
- get desired DS
- get actual DS
  - create a new DS if doesn't exist
- Check if desired != actual 
  - it does look like only CAA image and node selectors are configurable and prone to changing by the end user (I'm not familiar with peerpods/CAA flow so might need another set of eyes on this)
  - update only new changes only

The above approach is one of simpler ways to implement reconcile loops. The `createOrUpdate()` proposed by claude also does do something similar internally,
```
- get object
- if does not exist, run mutate and create with object after mutate
- if exists, run mutate, check if there was a change using `semantic.deepEqual()`, update with object after mutate
```
We can just go with your changes, although I'm wondering if we need the entire daemonset builder changes. AFAICT `mutateCAADaemonSet()` just combines `processDaemonsetForCAA()` and `processProviderConfigCAA()`. I believe then we could just do update `processDaemonsetForCAA` to be parameterized and pass it as is in the mutate fn in `CreateOrUpdate()`.

```patch
diff --git a/controllers/peerpods.go b/controllers/peerpods.go
index bd3b1719..e43a8397 100644
--- a/controllers/peerpods.go
+++ b/controllers/peerpods.go
@@ -282,28 +282,32 @@ func (r *KataConfigOpenShiftReconciler) processProviderConfigCAA(ds *appsv1.Daem
 
 // Create the PeerPodConfig CRDs and misc configs required for peer-pods
 func (r *KataConfigOpenShiftReconciler) enablePeerPodsMiscConfigs() error {
-	// Create the CAA daemonset
-	ds := r.processDaemonsetForCAA()
-	if err := r.processProviderConfigCAA(ds); err != nil {
-		r.Log.Error(err, "Failed setting cloud provider specific configuration for cloud-api-adaptor DS")
-		return err
-	}
-	r.Log.Info("Got CAA ds manifest", "ds", ds)
-
-	if err := controllerutil.SetControllerReference(r.kataConfig, ds, r.Scheme); err != nil {
-		r.Log.Error(err, "Failed setting ControllerReference for cloud-api-adaptor DS")
-		return err
+	ds := &appsv1.DaemonSet{
+			Name:      caaDsName,
+			Namespace: os.Getenv("PEERPODS_NAMESPACE"),
+		},
 	}
+	op, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, ds, func() error {
+		// Create the CAA daemonset
+		r.processDaemonsetForCAA(ds) // update this to be parameterized to accept an empty ds
+		if err := r.processProviderConfigCAA(ds); err != nil {
+			r.Log.Error(err, "Failed setting cloud provider specific configuration for cloud-api-adaptor DS")
+			return err
+		}
+		r.Log.Info("Got CAA ds manifest", "ds", ds)
 
-	err := r.Client.Update(context.TODO(), ds)
-	if err != nil && k8serrors.IsNotFound(err) {
-		r.Log.Error(err, "cloud-api-adaptor daemonset doesn't exist. Creating")
-		err = r.Client.Create(context.TODO(), ds)
-		if err != nil {
-			r.Log.Error(err, "failed to create cloud-api-adaptor daemonset")
+		if err := controllerutil.SetControllerReference(r.kataConfig, ds, r.Scheme); err != nil {
+			r.Log.Error(err, "Failed setting ControllerReference for cloud-api-adaptor DS")
 			return err
 		}
+		return nil
+	})
+	if err != nil {
+		r.Log.Error(err, "failed to reconcile cloud-api-adaptor daemonset")
+		return err
 	}
+	r.Log.Info("CAA daemonset reconciled", "result", op)
 
 	// Create the mutating webhook deployment
 	err = r.createMutatingWebhookDeployment()

```

snir911:
As you guys mentioned i agree the CreateOrUpdate is the more decent option ATM, although I don't think we expect anything to be changed on the fly for the DS .
The only thing is that i'm afraid we are missing some flow which is currently utilizes this CAA restart in-explicitly (e.g. cm update)

wainersm:
Hi @thejasn , update this PR to address your suggestion to remove the duplicates. Claude is stuborn and as such still refactored slightly different from what you suggested, but the duplication is gone. I will manage to run some tests. what do you think? Ah, let me know if you prefer yourself to provide a fix then I can just drop this.

wainersm:
> As you guys mentioned i agree the CreateOrUpdate is the more decent option ATM, although I don't think we expect anything to be changed on the fly for the DS .

This is one interesting point. We've seen two different tests that fails on pod deletion:

1. In one it deletes csv/subscription and reinstall
2. No deletion of csv/subscription happen but deletion still fails
 
I don't know what happens caa DS to restart on the (2) honestly.

>The only thing is that i'm afraid we are missing some flow which is currently utilizes this CAA restart in-explicitly (e.g. cm update)

I'm afraid of introducing breaks in another part. But I checked here and the reconcile seems to not run on peer-pods-cm updates. I don't know what else could be affected.

thejasn:
> Hi @thejasn , update this PR to address your suggestion to remove the duplicates. Claude is stuborn and as such still refactored slightly different from what you suggested, but the duplication is gone. I will manage to run some tests. what do you think? Ah, let me know if you prefer yourself to provide a fix then I can just drop this.

As I'm still understand the peer pods flow, I'm using #1950 to understand the bug and the potential fix (what I was hoping would fix it). Still running some tests to see if it does work.

> This is one interesting point. We've seen two different tests that fails on pod deletion:
I was also wondering if there is a way I can trigger the automated test on the PR?

gkurz:
> 
> As I'm still understand the peer pods flow, I'm using #1950 to understand the bug and the potential fix (what I was hoping would fix it). Still running some tests to see if it does work.
> 

FWIW I prefer #1950 over the huge commit in this PR :wink:

wainersm:
## Konflux Build Triage

| Field | Details |
|---|---|
| **Pipeline** | `osc-operator-on-pull-request` |
| **PipelineRun** | `osc-operator-on-pull-request-7jftb` |
| **Failed task** | `clone-repository` |
| **Status** | `Failed` |
| **Classification** | Transient/Infrastructure — Git Clone Failure |
| **Log source** | GitHub check run summary only (PipelineRun not on cluster, tkn-results not configured) |

### Failure Details

The `clone-repository` task failed after exhausting all retries, likely due to a transient GitHub/network issue:

```
error: Command failed after 10 tries with status 1
Command exited with non-zero status 1
```

### Auto-retest

`/retest` posted at 13:50 UTC | 1/2 retests used

---
<sub>🤖 Generated by konflux-triage</sub>

wainersm:
@thejasn has worked on a fix to that problem, so this no longer makes sense.

PR #1919, opened 2026-03-27
Adding more output and RCA to dig.py

tbuskey:
dig.py does a static pass over the prowjob URL to determine if a HUMAN needs to analyze it further

If the openshift-extended-test step is SUCCESS, no analysis of the test results needs to be done.
If a prow step after the test step fails, the output show flag it as a prow error, but not necessarily a test FAILURE

If a prow step before the test step fails, the test step is a FAILURE

Some of the pre steps failure can be detected and output so further analysis is not need.  ex: azure quota, rpm failed to upload

Usually dig.py is used to create a csv to import into  a spreadsheet.  Many runs can be added at one time and the spreadsheet can be looked at to decide analysis.  A human will correct the spreadsheet after analysis.

This PR is to add to the root cause and other output.  This should reduce analysis time.  Most of the code changes were done with claude and cursor

-  Test success is primarily determined by the openshift-extended-test results in prow
   - dig.py was waiting for the post steps to finish
   - dig.py was not always using the test-results.yaml of finished.json artifacts as the primary way to find success
   - A human had to keep correcting output
- If a post test fails, don't change the test result.  Flag it in the output
- Output the failed test case numbers, especially in the csv
- When the test fails, show # failed/passed/skipped.  If the test was interrupted find the last test that was started which will have (# failed/#started/# total tests) and show the test was timed out
- Add the duration of the test step to the output
  -  add a column to the csv output for this
  - this will be in the test step or in the time of the steps
- remove blank lines in the csv output
  - if URLS is a list of the prowjob URLs this will make import easier:
 ```bash
echo '' > /tmp/x
for URL in $URLS
do
  echo $URL; 
  scripts/prowjob-analyzer/dig.py $URL --csv --no-header --no-wait >> /tmp/x
done
grep -v '^$' /tmp/x > ~/Downloads/x.csv
```
- Add rpm install failure to the RCA so analysis is not needed

### DIFF
--- a/.claude/commands/prowjob-analyze.md
+++ b/.claude/commands/prowjob-analyze.md
@@ -190,7 +190,7 @@ Tests use hardcoded defaults that can be overridden via `osc-config` ConfigMap:
 - Runtime class name (kata, kata-remote)
 - Enable peer pods (true/false)
 - Workload type to test (kata, peer-pods, coco)
-****- Install Kata RPM (true/false)
+- Install Kata RPM (true/false)
 - Enable GPU to include Nvidia GPU drivers in podvm image and run GPU tests
 - Set initdata for the OSC operator
 - Must-gather image
--- a/.claude/commands/prowjob-analyze.md
+++ b/.claude/commands/prowjob-analyze.md
@@ -190,7 +190,7 @@ Tests use hardcoded defaults that can be overridden via `osc-config` ConfigMap:
 - Runtime class name (kata, kata-remote)
 - Enable peer pods (true/false)
 - Workload type to test (kata, peer-pods, coco)
-- Install Kata RPM (true/false)
+****- Install Kata RPM (true/false)
 - Enable GPU to include Nvidia GPU drivers in podvm image and run GPU tests
 - Set initdata for the OSC operator
 - Must-gather image
--- a/scripts/prowjob-analyzer/lib/fetcher.py
+++ b/scripts/prowjob-analyzer/lib/fetcher.py
@@ -2,17 +2,32 @@
 Artifact Fetcher Module
 
 Handles URL parsing and fetching artifacts from Prow/GCS.
+Supports in-memory caching, a local directory tree, or a .tar.gz of the same layout.
 """
 
-import re
-import time
 import json
 import logging
+import os
+import re
+import shutil
+import subprocess
+import sys
+import tarfile
+import tempfile
+import time
+from pathlib import Path
 from typing import Dict, Optional, Tuple, List
 from urllib.request import urlopen, Request
 from urllib.error import HTTPError, URLError
 
 logger = logging.getLogger(__name__)
+
+# Bytes fetched over HTTP or read from disk (keyed by root + relative path).
+_artifact_cache: Dict[str, Optional[bytes]] = {}
+# Resolved directory containing prowjob.json (and artifacts/, …); None = fetch over HTTP only.
+_local_artifacts_root: Optional[str] = None
+# Temp directory created when configuring from a .tar.gz (removed on next configure or process exit).
+_tar_extract_temp_dir: Optional[str] = None
 
 
 def parse_prow_url(url: str) -> Tuple[str, str, str]:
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: .claude/commands/prowjob-analyze.md [hunk 1]
```diff
@@ -190,7 +190,7 @@ Tests use hardcoded defaults that can be overridden via `osc-config` ConfigMap:
 - Runtime class name (kata, kata-remote)
 - Enable peer pods (true/false)
 - Workload type to test (kata, peer-pods, coco)
-- Install Kata RPM (true/false)
+****- Install Kata RPM (true/false)
```
wainersm: Hi @tbuskey !

I didn't get why that change.

scripts/prowjob-analyzer/lib/parser.py
```diff
@@ -105,23 +105,48 @@ def parse_test_results(test_results_content: bytes) -> Optional[Dict]:
         return None
 
 
-def get_job_status(prowjob_data: Dict, test_results: Optional[Dict] = None) -> str:
+def get_job_status(
```
wainersm, line 169: It seems that `get_job_status` is already returning the right job status. Do you have an job example where it's not getting it correctly?

The current logic is simple:

1. Look at job status.
2. is it success? Then check the test_results because failing tests might not fail the entire job
3. else just return the job status.

.claude/commands/prowjob-analyze.md
tbuskey: Removing

scripts/prowjob-analyzer/lib/parser.py
tbuskey: This looks at test-results.yaml if it exists, then finishes.json if it exists
It handles if the test doesn't get started or is interrupted

Then if looks at if the pre prowjobs succeeded or failed

wainersm: my point it that the current function is already doing that. The changes that claude made seems to simply had the code more complex without a good reason. 

Do you have a job where it was not handling the prow status correctly?

wainersm:
@tbuskey here is review clade made to this change:

```
Review                                                                                                                                          
                                                                                                                                                  
  Overall this PR tackles a real problem — the analyzer was conflating test failures with Prow pipeline failures, and the new priority logic in   
  get_job_status (test-results.yaml → finished.json → prowjob state) is the right approach. The RPM install detection, test case ID extraction,   
  and elapsed time tracking are all useful additions.
                                                                                                                                                  
  A few issues I spotted:                                                                                                                         
   
  Bugs                                                                                                                                            
                  
  1. get_job_status ignores errors field (lib/parser.py:134-135)                                                                                  
                  
  if test_results is not None and 'failures' in test_results:                                                                                     
      return 'success' if test_results['failures'] == 0 else 'failure'                                                                            
                                                                                                                                                  
  A run with failures=0, errors=3 would be reported as success. Should check both:                                                                
                                                                                                                                                  
  if test_results is not None and ('failures' in test_results or 'errors' in test_results):                                                       
      failures = int(test_results.get('failures', 0) or 0)                                                                                        
      errors = int(test_results.get('errors', 0) or 0)                                                                                            
      return 'success' if failures == 0 and errors == 0 else 'failure'                                                                            
                                                                                                                                                  
  2. prefer_next flag has no effect (lib/fetcher.py:982-989)                                                                                      
                                                                                                                                                  
  if sec_tf is not None and sec_tf > 0:                                                                                                           
      sec_wall = sec_tf                                                                                                                           
  elif prefer_next and sec_next is not None and sec_next > 0:                                                                                     
      sec_wall = sec_next                                                                                                                         
  elif sec_next is not None and sec_next > 0:                                                                                                     
      sec_wall = sec_next                                                                                                                         
                                                                                                                                                  
  The second and third branches do the same thing. When sec_tf is None, sec_next is used regardless of prefer_next. Looks like a logic error — the
   third branch should probably be sec_wall = None or removed.                                                                                    
                                                                                                                                                  
  3. Duplicate regex pattern (lib/failure_analyzer.py:72 and 75)                                                                                  
   
  r'error:\s*Failed dependencies',   # line 72, new                                                                                               
  r'error:.*Failed dependencies',    # line 75, existing                                                                                          
                                                                                                                                                  
  These match the same strings. Remove one.                                                                                                       
                                                                                                                                                  
  Nits                                                                                                                                            
                  
  4. lib/report_generator.py:399 — typo: "canonical repor dict" → "canonical report dict"                                                         
                  
  Suggestions                                                                                                                                     
                  
  5. The new parsing and duration functions (go_duration_to_seconds, extract_failed_case_ids_from_extended_build_log,                             
  summarize_openshift_extended_test_build_log, etc.) are great unit test candidates. Given these are regex-heavy and AI-generated, tests would go
  a long way toward building confidence.                                                                                                          
                  
  6. compute_openshift_extended_test_elapsed_display fetches build-log.txt independently, and analyze_failure fetches the same file again. Not    
  blocking, but passing the already-fetched content through would avoid redundant HTTP requests.
```

tbuskey:
Fixed 1-4
Added unit tests (5). 

For 6, I'd like to be able to download all the artifacts to a directory or tar.gz file and use that in addition to using the URL.  That could be the caching mechanism.  It would also allow us to keep a directory of interesting test runs to verify dig&sift changes

wainersm:
Hi @tbuskey ,

Asked Claude for another round of review:

```
Here's my review of PR #1919 "Adding more output and RCA to dig.py":                                                                            
                                                                                                                                                  
  ---                                                                                                                                             
  Overview                                                                                                                                        
                                                                                                                                                  
  This is a substantial PR (1247+/57-) to scripts/prowjob-analyzer/ that improves how dig.py determines test success/failure and enriches the     
  output with more root cause details. The changes were largely AI-assisted (Claude/Cursor per the description).                                  
                                                                                                                                                  
  What it does well                                                                                                                               
                                                                                                                                                
  - Better status determination — get_job_status now has a clear priority chain: test-results.yaml > finished.json > prowjob.json, fixing cases   
  where post-test Prow step failures were incorrectly reported as test failures.                                                                
  - Post-test failure distinction — analyze_post_test_prow_failure properly separates "tests passed but Prow pipeline failed later" from actual   
  test failures. This directly reduces false analysis.                                                                                            
  - Failed test case IDs — Extracting C##### IDs from build-log.txt is useful for CSV/spreadsheet workflows.
  - Test elapsed time — Adding test_elapsed_time to output is a practical addition.                                                               
  - Logging to stderr — Good fix for --csv piping workflows.                                                                                      
  - CSV blank line fix — lineterminator='\n' and print(report, end="") properly handles appending.                                                
  - Unit tests — Good coverage for the new pure functions.                                                                                        
                                                                                                                                                  
  Issues                                                                                                                                          
                                                                                                                                                  
  Bugs / Correctness                                                                                                                              
                  
  1. get_job_status short-circuits too aggressively (lib/parser.py new lines ~129-141): When test_results has failures/errors keys, it returns    
  immediately without ever checking extended_test_finished. But test_results could have failures: 0, errors: 0 AND extended_test_finished could
  indicate failure. The early return for success is correct, but the function now has redundant logic — the same test_results check appears again 
  in the state == 'success' branch below (~line 164). Clean up the duplication.
  2. get_failed_steps side-effect: fetching build-log inside a loop (lib/fetcher.py ~lines 462-479): For every failed openshift-extended-test
  step, it fetches build-log.txt to generate a label. This is the same file fetched again in analyze_failure. That's two full downloads of        
  potentially large logs. Consider passing the already-fetched content or caching.
  3. _kata_rpm_install_log_snippet test helper (tests/test_failure_analyzer.py ~line 100): The string concatenation + "x" * 20 + "\n" followed by 
  another string literal without + relies on implicit string concatenation, but the \n at the end of the first string means the second string     
  starts a new implicit line. This works but is fragile — a stray space would break it.
                                                                                                                                                  
  Design concerns                                                                                                                                 
   
  4. get_failed_steps returns mixed types — Step names are strings like "gather-must-gather", but for openshift-extended-test it now returns "3   
  fail, 10 pass, 2 skip". This mixes identifiers with display labels. Downstream code needs is_openshift_extended_test_failure_label() and
  artifact_dir_for_failed_step_label() to reverse-map. A struct/tuple (step_name, label) would be cleaner.                                        
  5. Duplicate build-log.txt fetches — compute_openshift_extended_test_elapsed_display fetches build-log.txt, get_failed_steps fetches it, and
  analyze_failure fetches it. That's up to 3 downloads of the same (large) file. Consider a simple cache or passing the content through.          
  6. compute_openshift_extended_test_elapsed_display complexity — This function and its ~10 helper functions (_oet_step_ended,
  _oet_step_has_finished, _merge_wall_and_log_seconds, etc.) span ~200 lines for computing elapsed time. The layered fallback logic (finished.json
   timestamps → next step start → build-log durations → ginkgo failed durations → generic parenthesized durations) is thorough but hard to follow
  and debug.                                                                                                                                      
                  
  Style / Minor                                                                                                                                   
   
  7. dig.py line 69-75: The if variant and variant != 'unknown' check is duplicated twice in sequence (lines 100 and 110 in the new code). Merge  
  them.           
  8. failure_analyzer.py: extended_build_log_first_test_prefix uses max_chars: int = 1_500_000 — consider documenting why 1.5M chars is the       
  cutoff.                                                                                                                                         
  9. Several functions return Optional[str] but callers don't consistently check — e.g., extract_kata_rpm_dependency_detail returns None but
  determine_root_cause only uses it when truthy (fine), though the pattern is inconsistent across the file.                                       
  10. _line_is_prow_entrypoint_noise — String containment checks like '"component"' in s are fragile; this works for known log formats but could
  false-positive on test output containing those strings.                                                                                         
                  
  Verdict                                                                                                                                         
                  
  The core improvements (status priority chain, post-test failure separation, test case IDs) are valuable and will reduce manual analysis. The    
  main concerns are:
  - Duplicate large file fetches (performance)                                                                                                    
  - Mixed step name/label semantics                                                                                                               
  - Complexity of the elapsed time computation
                                                                                                                                                  
  I'd suggest addressing #2/#5 (duplicate fetches) before merge as it's a performance issue with large logs, and #4 (mixed return types in        
  get_failed_steps) as it'll cause maintenance headaches. The rest can be follow-ups.
```

scripts/prowjob-analyzer/lib/fetcher.py
```diff
@@ -406,8 +460,333 @@ def get_failed_steps(base_url: str, variant: str) -> List[str]:
...
+
+    ended = _oet_step_ended(ts, tf, fin_json, build_log_text, sec_next)
+    finished_prow = _oet_step_has_finished(ts, tf, fin_json, build_log_text)
+    not_started = ts is None and not finished_prow
+    running = ts is not None and not ended
+
+    sec_log = _parse_build_log_duration_seconds(build_log_text)
+
+    if not_started or running:
+        return '00:00'
```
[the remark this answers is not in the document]
tbuskey: Its ok to have 0:00

PR #1949, opened 2026-03-30
tests/osc-test-fbc-integration: continue pipeline on error

wainersm:
The konflux pipelines are fail-fast (e.g. abort on first error) by default. However, in our integration tests pipeline we want to run all the prow jobs. Well, expect for the sanity tests jobs, because these are meant to prevent prow jobs to run unnecessarily in case of infraestructure problems.

So on this PR:

* run pipeline tasks with `onError: continue` parameter to keep the pipeline alive on any error. In practice, tekton mark the task failed but ignore it and continue on.
* the last task run on `finally` and should fail the entire pipeline status if any task failed
  *  I tried to make at last task to consume and report an aggregate of the results of the previous tasks, however, if any of these tasks are skipped then the finally task is simple skipped too. In other words, it is not guaranteed the finally task to run if it it depends on the results of other tasks

### DIFF
--- a/tests/osc-test-fbc-integration.yaml
+++ b/tests/osc-test-fbc-integration.yaml
@@ -127,7 +127,7 @@ spec:
       displayName: "Running prow job $(params.PROWJOB_NAME)"
       # A fail on sanity should stop and fail the pipeline
       onError: stopAndFail
-      timeout: "4h"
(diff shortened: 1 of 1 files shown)


littlejawa (approved):
/lgtm

Thanks @wainersm

PR #1961, opened 2026-03-31
runtimeclass: remove GPU runtime classes on s390x

ajaypvictor:
Skip creation of kata-nvidia-gpu runtime class on s390x (IBM Z) systems where NVIDIA GPU support is not available. Uses runtime.GOARCH for architecture detection, consistent with existing checks in the codebase.

The kata-cc-nvidia-gpu runtime class is already filtered on s390x through existing TEE type logic (only created for Intel TDX or AMD SNP, never for IBM SE).

Fixes: https://redhat.atlassian.net/browse/KATA-4801

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -23,6 +23,7 @@ import (
 	"maps"
 	"os"
 	"reflect"
+	goruntime "runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


gkurz (commented):
/ok-to-test

gkurz (approved):
/lgtm

Thanks @ajaypvictor !

PR #1963, opened 2026-04-01
coderabbit: add basic coderabbit config

thejasn:
Add a basic coderabbit config yaml
**- What I did**

**- How to verify it**

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,66 @@
+# CodeRabbit Configuration
+# Reference: https://docs.coderabbit.ai/reference/configuration
+
+# Language for review comments (ISO code). Default: "en-US"
+language: en-US
+
+# Inherit settings from parent/organization config.
+inheritance: true
+
+reviews:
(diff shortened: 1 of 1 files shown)


gkurz:
Hi @thejasn ! People likely need more information in order to understand what this PR is actually doing, e.g. [Using CodeRabbit at Red Hat](https://docs.google.com/document/d/1LdyFK6nqH2ZovXc4Q9Hvf7v4B5ujzhi8DgzeOFcX_QE/edit?tab=t.0.) and maybe some reference from the openshift org in GH. Please add everything that can help review in the commit message.

wainersm:
Hi @thejasn , I read about usage of coderabbit within red hat while ago but I forgot most of the details. One thing that I remember is that users need to do something to enable it on their accounts. Could you also add some words on the docs explaining the tool, how to use, interpreting results...etc.etc? Maybe the CONTRIBUTING.md is a good candidate.

wainersm:
Hi @thejasn , I was wondering if it's possible to configure it to not review the PR opened by Mintmaker. I don't think its reviews to those PRs add any value, besides being a waste of money :)

PR #1964, opened 2026-04-01
claude: add rehearse-jobs skill and command

wainersm:
Added a claude skill to automate the workflow to open a PR on openshift/release to rehearse OSC Prow CI jobs with a custom catalog image and optionally a specific kata RPM version.

This skill can be used, for example, to test a fix to the operator in the CI environment.

### DIFF
--- a/.claude/commands/rehearse-jobs.md
+++ b/.claude/commands/rehearse-jobs.md
@@ -1,0 +1,12 @@
+---
+description: Open a PR on openshift/release to rehearse OSC Prow CI jobs with a custom catalog and optional kata RPM
+argument-hint: "<catalog-image> <ocp-versions> <job-name> [kata-rpm: yes|no]"
+---
+
+Rehearse OSC Prow CI jobs via a PR on openshift/release using the pj-rehearse workflow.
+
+Follow the instructions in the **Rehearse Jobs** skill (`skills/rehearse-jobs/SKILL.md`) with:
+- **Catalog image**: `$1`
+- **OCP versions**: `$2` (comma-separated, e.g. `4.18,4.20,4.21`)
+- **Job name**: `$3` (e.g. `azure-ipi-peerpods`)
+- **Kata RPM**: `$4` (optional, `yes` or `no`, default: ask the user)
--- a/.claude/skills/rehearse-jobs/SKILL.md
+++ b/.claude/skills/rehearse-jobs/SKILL.md
@@ -1,0 +1,217 @@
+---
+name: Rehearse Jobs
+description: |
+  Open a PR on openshift/release to rehearse OSC Prow CI jobs with a custom catalog image
+  and optionally a specific kata RPM version. Uses the pj-rehearse workflow.
+
+  Triggers: "rehearse jobs", "rehearse prow jobs", "run prow jobs", "test catalog image",
+  "pj-rehearse", "trigger ci jobs", "run ci jobs"
+---
+
+# Rehearse Jobs Skill
+
+Open a PR on openshift/release to trigger pj-rehearse for a subset of OSC Prow CI jobs
+with a custom catalog image and optionally a specific kata RPM version.
+
+## Prerequisites
+
+- **gh** (GitHub CLI): Must be installed and authenticated (`gh auth status`).
+  If not available, ask the user to install it (https://cli.github.com/) and run `gh auth login`.
+
+## Required Input
+
+Ask the user for the following if not provided:
+
+- **Catalog image**: The catalog image to test (e.g. from `/build-catalog`)
+- **OCP version(s)**: Which versions to target (4.18, 4.19, 4.20, 4.21)
+- **Job(s) to run**: Which jobs from the candidate config to modify. Available jobs
+  vary by version but typically include:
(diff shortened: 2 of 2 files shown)


wainersm:
Example of job triggered with that claude skill: https://github.com/openshift/release/pull/77202

tbuskey:
/assign @dbkreling 
Let @dbkreling know we're using his script!

ldoktor:
I'm wondering who are the users here. I can not imagine running this without `claude --permission-mode bypassPermissions` which is forbidden to us as far as I know.

Even then this requires so many deps that running this as a one-off action seems quite long an problematic, so I assume it's for someone, who does that regularly. And in such case I'd expect them having scripts with existing repositories.

I think for one-off thing people could use clusterbot, which seems more straight forward. As for the regular thing we have the gangway (or at least the generate script).

wainersm:
Hi @ldoktor ,
thanks for your comments! and here goes some replies:

> I'm wondering who are the users here. I can not imagine running this without `claude --permission-mode bypassPermissions` which is forbidden to us as far as I know.
>

I made a custom catalog image then used claude to submit it for testing via pj-rehearse. As it worked so well I asked claude to generate a skill out of the session so I could used it in the future. Recently I saw a dev of our team asking @tbuskey to test a fix, then I thought: "well, if he had that same skill maybe we could test it himself easily". Hence, I decided to share the skill here.

As for running with `--permission-mode bypassPermissions`, I never do that. Essentially I become a "yes" and "no" operator. 
 
> Even then this requires so many deps that running this as a one-off action seems quite long an problematic, so I assume it's for someone, who does that regularly. And in such case I'd expect them having scripts with existing repositories.
>

I don't think it's much deps. You will need `gh` and a couple of python libraries in your system and only if you need to run the script to figure out the kata rpms. Or my system has already all the necessary tools and I didn't notice it is demanding a lot of deps?

I thought in generate a script instead of a skill. The script is more efficient (computer-ability  speaking) and deterministic, and that's good. On the other hand, I would need to maintain an stable interface between the script and some outside systems (quay, brew, prow..etc), whereas, an LLM can detect when something goes wrong and it will try to fix. Another advantage of the LLM is that I note necessarily need to implement all the cases, for example, filtering tests is not part of the default inputs and is not describe in the skill, but an user could ask to filter for specific tests, as the LLM can adapt for new requests. I weighted all those things and preferred the skill over a script.
 
> I think for one-off thing people could use clusterbot, which seems more straight forward. As for the regular thing we have the gangway (or at least the generate script).

I considered using clusterbot but then I assumed Claude couldn't easily interact with Slack. That's actually something I can give a try and report here. I would prefer the workflow via clusterbot better than github.

And the gangway has a problem that user requires a token and the token expires very quickly. The skill would need to fetch a fresh token but I don't know how to do that. I looked to me too complicated to implement :)

tbuskey:
I like that others can use it, but it is a bit complex.
Run this for my runs
```
V=4.19  INSTALL_KATA_RPM=true  KATA_RPM_VERSION="3.25.0-5.rhaos${V}.el9"  OCP_VERSION=$V SLEEP_DURATION=4h  MUST_GATHER_ON_FAILURE_ONLY=false INITDATA=$INITDATA TRUSTEE_URL=$TRUSTEE_URL   TEST_TIMEOUT=120 TEST_RELEASE_TYPE=Pre-GA ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed-containers-operator-create-prowjob-commands.sh  create
```
Maybe the skill can use the script with the parameters and we keep the script updated?

wainersm:
Hi @tbuskey , thanks for the review!

> I like that others can use it, but it is a bit complex. Run this for my runs
> 

Yeah, indeed it's a bit complex.

> ```
> V=4.19  INSTALL_KATA_RPM=true  KATA_RPM_VERSION="3.25.0-5.rhaos${V}.el9"  OCP_VERSION=$V SLEEP_DURATION=4h  MUST_GATHER_ON_FAILURE_ONLY=false INITDATA=$INITDATA TRUSTEE_URL=$TRUSTEE_URL   TEST_TIMEOUT=120 TEST_RELEASE_TYPE=Pre-GA ci-operator/step-registry/sandboxed-containers-operator/create-prowjob/sandboxed-containers-operator-create-prowjob-commands.sh  create
> ```
> 
> Maybe the skill can use the script with the parameters and we keep the script updated?

It simplify the process a lot. My only concern is about the GANGWAY API token....I was playing with claude today and it seems that I can stream-line the process of getting a API token.

I will make this open a couple of days more just in case anyone has more feedbacks, but ultimately I will close it.

/hold

ldoktor:
Testing ought to be simple so you are right. As prow is complex and hard to tackle, perhaps a skill like this might help. Still I'd focus on reproducibility and reliability. I like the idea of using clusterbot (generate command to be passed to clusterbot) or gangway, should that be feasible. Ideally abstracting as much as not-needed and decreasing the dependencies, while caching repos that are being used (rather tham mktemp use .claude/ dir?). What do you think?

wainersm:
@tbuskey built a very similar skill so this one is no longer needed.

PR #1965, opened 2026-04-01
ci: add GitHub workflow for commit message validation

wainersm:
Run hack/commit-msg-check.sh on pull requests to enforce commit message conventions (subsystem prefix, body, Signed-off-by, line lengths).

Assisted-by: Claude

### DIFF
--- a/.github/workflows/commit-msg-check.yaml
+++ b/.github/workflows/commit-msg-check.yaml
@@ -1,0 +1,26 @@
+# Validate commit messages against project conventions.
+---
+name: commit-msg-check
+
+on:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+permissions: {}
+
+jobs:
+  commit-msg-check:
+    name: commit-msg-check
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      - name: Validate commit messages
+        run: ./hack/commit-msg-check.sh origin/${{ github.event.pull_request.base.ref }} ${{ github.event.pull_request.head.sha }}


c3d (approved):
That seems like a good idea, but I wonder if this will not get in the way of the many chore PRs.

For example I believe that this check would fail with a `chore(deps)` prefix:
```
    # Check subsystem prefix
    if ! echo "$subject" | grep -qE '^[a-zA-Z_/-]+:'; then
        echo "ERROR: Missing subsystem prefix (e.g., 'feat:', 'fix:'): $sha"
        echo "  $subject"
        FAILED=1
    fi
```

and the body check might fail on the chores as well, and does not for the wrong reason:

```
% git log --format=%b -n 1 HEAD
Signed-off-by: red-hat-konflux <126015336+red-hat-konflux[bot]@users.noreply.github.com>
```

(Unless we consider that a body that only contains a signed-off-by is sufficient?)

**Update**: Tested it, and it works because we specifically skip bot commits. So ack, @wainersm…

wainersm:
@littlejawa @gkurz please help here too? :)

snir911 (approved):
LGTM thanks

.github/workflows/commit-msg-check.yaml
```diff
@@ -0,0 +1,26 @@
+# Validate commit messages against project conventions.
+---
+name: commit-msg-check
+
+on:
+  pull_request:
```
littlejawa, line 6: I'm not familiar with GHA, so I'm not sure how often this will trigger. Does it run for any action done on the pull request?
If so, should we add some specific activities, like:
```
    types: [opened, reopened, edited]
```

I.e: I don't think we need to run it when the PR is closed, assigned, labeled, etc...

(based on https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)

littlejawa: Ah, I actually missed it:

```
By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. 
```

And : 
> [littlejawa](https://github.com/littlejawa) assigned [wainersm](https://github.com/wainersm) [2 minutes ago](https://github.com/openshift/sandboxed-containers-operator/pull/1965#event-24665283199)

... didn't trigger the action :-)

littlejawa (approved):
/lgtm

Thanks @wainersm !

PR #2012, opened 2026-04-07
confidential: Bump OCP z-stream for 4.19 and 4.21

gkurz:
This matches the following :

4.19.28 Dev cut-off : Apr 8, Fast channel : Apr 15
4.21.9  Dev cut-off : Apr 1, Fast channel : Apr 8

Kata 3.25.0-5 is already available in 4.21.9 and it is present in 4.19.0 nightlies.

Assisted by Claude AI to refresh the test.

### DIFF
--- a/controllers/confidential_handler_test.go
+++ b/controllers/confidential_handler_test.go
@@ -64,11 +64,11 @@ var _ = Describe("validateOCPVersion", func() {
 			},
 			Entry("version 4.18.0 (major.minor too low)", "4.18.0"),
 			Entry("version 4.19.0 (patch too low)", "4.19.0"),
-			Entry("version 4.19.26 (patch one below minimum)", "4.19.26"),
+			Entry("version 4.19.27 (patch one below minimum)", "4.19.27"),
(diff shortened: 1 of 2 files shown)


controllers/confidential_handler_test.go
```diff
@@ -175,7 +175,7 @@ var _ = Describe("validateOCPVersion integration", func() {
 		})
 
 		It("should use environment variable when set", func() {
-			os.Setenv("BM_COCO_OVERRIDE_OCP_VERSION", "4.21.8")
+			os.Setenv("BM_COCO_OVERRIDE_OCP_VERSION", "4.21.9")
```
dbkreling, line 178: Wouldn't it be more interesting if we parameterized the versions to reduce the number of identical updates of the OCP all across the file?
Other than that, the versions LGTM to me.

gkurz: Yes it might be possible to factor out things a bit. I'd rather do that in a subsequent PR as I really want this be merged ASAP.

dbkreling (approved):

PR #2021, opened 2026-04-10
runtimeclass: skip kata-cc-nvidia-gpu creation for IBM SE

ajaypvictor:
Skip creation of kata-cc-nvidia-gpu runtime class for IBM SE (s390x) systems where NVIDIA GPU confidential computing is not supported.

This change adds an explicit conditional check (hasIntelTDX || hasAmdSNP) around the GPU runtime class creation to ensure it's never created on IBM SE deployments, even if the logic changes in the future.

Fixes: https://redhat.atlassian.net/browse/KATA-4801

Follow up of https://github.com/openshift/sandboxed-containers-operator/pull/1961

### DIFF
--- a/controllers/confidential_handler.go
+++ b/controllers/confidential_handler.go
@@ -320,17 +320,19 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
 			maps.Copy(additionalLabels, amdSNPNodeLabels)
 		}
 
-		// Create kata-cc-nvidia-gpu runtime class restricted to the detected GPU TEE subset
-		err = r.createRuntimeClass(
-			kataNvidiaGPUCCRuntimeClassName,
-			kataNvidiaGPUCCRuntimeClassCpuOverhead,
-			kataNvidiaGPUCCRuntimeClassMemOverhead,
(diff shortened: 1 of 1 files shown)


controllers/confidential_handler.go
```diff
@@ -320,17 +320,19 @@ func (r *KataConfigOpenShiftReconciler) handleConfidentialBaremetal(state Featur
...
-			kataNvidiaGPUCCRuntimeClassCpuOverhead,
-			kataNvidiaGPUCCRuntimeClassMemOverhead,
-			kataCCRuntimeClassExtResOverhead,
-			handler,
-			additionalLabels)
-		if err != nil {
-			r.Log.Error(err, "aborting, failed to create runtimeclass")
-			return err
+		// Create kata-cc-nvidia-gpu runtime class only for Intel TDX and AMD SNP
+		if hasIntelTDX || hasAmdSNP {
```
c3d, line 324: I remember seeing a PR not that long ago, #1961, that was similar but disabled S390x.

So I'm curious why the logic is different here (and inconsistent with the label that states "disable SE", not "enable only on TDX or SNP").

PR #2040, opened 2026-04-15
Deprecate packer based podvm builds

snir911:
## Summary by CodeRabbit

* **Chores**
  * Simplified Pod VM image creation to support only pre-built artifacts specified via image URI; removed operator-built image capability and associated build tooling.
  * Removed build-time configuration options (distro selection, NVIDIA drivers, FIPS support, cloud-config, source downloads) from all cloud provider configurations.
  * Cleaned up scripts and documentation reflecting the streamlined image handling approach.


gkurz:
IBM explicitly asked we keep this around as they don't have a working alternative yet.

Add `hold` label to make this explicit

/hold

PR #2062, opened 2026-04-23
airgap: add ImageSetConfiguration

balintTobik:
To install OSC in disconnected environments, users need an ImageSetConfiguration for oc-mirror and all operand images listed in the bundle. This PR adds both.

**- What I did**
Add imageset-config files for OCP 4.17-4.21 and a README under docs/disconnected/.

**- How to verify it**
Follow the README.md under docs/disconnected on a air-gapped environment.

Add imageset-config files for OCP 4.17-4.21 and a README under docs/disconnected/. This is an initial attempt to provide a supported way to mirror the OSC operator in air-gapped environments using oc-mirror.

### DIFF
--- a/docs/disconnected/README.md
+++ b/docs/disconnected/README.md
@@ -1,0 +1,35 @@
+# Disconnected Installation
+
+This directory contains `ImageSetConfiguration` files for mirroring the
+Sandbox Containers operator in disconnected (air-gapped) environments using `oc-mirror`.
+
+## What is covered
+
+Each `imageset-config-<ocp-version>.yaml` file mirrors:
+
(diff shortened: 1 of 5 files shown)


beraldoleal (commented):
@balintTobik thanks for this. Apart from some minor suggestions from coderabiit, this lgtm.

@gkurz, we need this file for disconnected envs and we used the same approach with the trustee operator. Do you see any problem with keeping it here?

snir911 (approved):

beraldoleal (approved):
I just tested this on Azure on 4.20.18 and its working as expected.

The operator is still not fully functional, but this out of the scope of this PR.

PR #2127, opened 2026-05-12
osc-podvm-builder: update the submodule reference to cloud-api-adaptor

littlejawa:


gkurz (approved):
Matches current HEAD of downstream CAA.

/lgtm

Thanks @littlejawa !

PR #2130, opened 2026-05-12
controller: fix non-targeted nodes appearing in waitingToUninstall status

ranjithrajaram:
Addressing issue reported in : #[KATA-4768](https://redhat.atlassian.net/browse/KATA-4768)

The updateStatus() function processes all worker nodes and incorrectly marks regular workers as "waitingToUninstall" when they have any MachineConfig drift, even though Kata was never installed on them.

The root cause is in putNodeOnStatusList() which evaluates all workers. When a non-kata node has nodeCurrMc != nodeTargetMc (common during cluster updates), the isNodeWaitingToUninstall() predicate returns true because:
- nodeMcoState == NodeDone (MCO finished processing)
- nodeCurrMc != nodeTargetMc (any MC drift)
- !isKataEnabledOnNode (true for non-kata labeled nodes)

**- What I did**

This fix adds an early-return check to skip nodes that don't match the configured kataConfigPoolSelector on non-converged clusters. This ensures only nodes that are actually targeted by the KataConfig (regardless of the specific selector used) are tracked in status lists, supporting both default and custom node selectors.

**- How to verify it**

Reproducer steps in https://redhat.atlassian.net/browse/KATA-4768

This fix adds an early-return check to skip nodes that don't match the configured kataConfigPoolSelector on non-converged clusters

### DIFF
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -1903,8 +1903,30 @@ func (r *KataConfigOpenShiftReconciler) updateStatus() error {
 		return len(nodes.Items)
 	}()
 
+	// Pre-compute cluster-level attributes outside the node loop for performance.
+	// isConvergedCluster is a cluster attribute, not a node-level attribute.
+	isConvergedCluster, err := r.checkConvergedCluster()
+	if err != nil {
+		return fmt.Errorf("failed to check if cluster is converged: %w", err)
+	}
+
+	// Build the effective KataConfig node selector once for the entire reconciliation.
+	kataNodeSelector, err := r.getKataConfigNodeSelectorAsSelector()
+	if err != nil {
+		return fmt.Errorf("failed to build KataConfig node selector: %w", err)
+	}
+
 	for _, node := range nodeList.Items {
-		e := r.putNodeOnStatusList(&node)
+		// On non-converged clusters, skip nodes that don't match the selector AND
+		// never had kata-oc label (i.e., were never targeted for Kata installation).
+		// Include nodes with kata-oc label even if they no longer match the selector,
+		// as they may be uninstalling Kata and should appear in status lists.
+		_, hasKataLabel := node.Labels["node-role.kubernetes.io/kata-oc"]
+		if !isConvergedCluster && !kataNodeSelector.Matches(labels.Set(node.Labels)) && !hasKataLabel {
+			continue
+		}
+
+		e := r.putNodeOnStatusList(&node, isConvergedCluster)
 		if e != nil {
 			err = e
 		}
@@ -1953,12 +1975,7 @@ func isNodeFailedToUninstall(nodeMcoState string, nodeCurrMc string, nodeTargetM
 	return nodeMcoState == NodeDegraded && !isKataEnabledOnNode
 }
 
-func (r *KataConfigOpenShiftReconciler) putNodeOnStatusList(node *corev1.Node) error {
-
-	isConvergedCluster, err := r.checkConvergedCluster()
-	if err != nil {
-		return err
-	}
+func (r *KataConfigOpenShiftReconciler) putNodeOnStatusList(node *corev1.Node, isConvergedCluster bool) error {
 
 	targetMcpName := func() string {
 		if isConvergedCluster {


ranjithrajaram:
@gkurz when time permits, can you please take a look

ranjithrajaram:
@snir911 can you review this PR ?

controllers/openshift_controller.go
```diff
@@ -1903,8 +1903,12 @@ func (r *KataConfigOpenShiftReconciler) updateStatus() error {
 		return len(nodes.Items)
 	}()
 
+	kataSelector, selErr := r.getKataConfigNodeSelectorAsSelector()
+	if selErr != nil {
+		return selErr
+	}
```
thejasn:
```suggestion
	if err != nil {
		return err
	}
```

controllers/openshift_controller.go
```diff
@@ -1903,8 +1903,12 @@ func (r *KataConfigOpenShiftReconciler) updateStatus() error {
 		return len(nodes.Items)
 	}()
 
+	kataSelector, selErr := r.getKataConfigNodeSelectorAsSelector()
```
thejasn:
```suggestion
	kataNodeSelector, selErr := r.getKataConfigNodeSelectorAsSelector()
```

controllers/openshift_controller.go
```diff
@@ -1953,13 +1957,23 @@ func isNodeFailedToUninstall(nodeMcoState string, nodeCurrMc string, nodeTargetM
...
+	// Skip nodes that don't match the effective KataConfig node selector on non-converged clusters.
+	// These are regular workers that were never targeted for Kata installation
+	// and should not appear in any KataConfig status lists (including waitingToUninstall).
+	if !isConvergedCluster {
+		// Check if node matches the effective KataConfig node selector.
+		if !kataNodeSelector.Matches(labels.Set(node.Labels)) {
+			return nil
+		}
+	}
+
```
thejasn: I would move this logic out to `updateStatus()` so that we retain the decision making in the caller function. 

```suggestion
```

controllers/openshift_controller.go
```diff
@@ -1903,8 +1903,12 @@ func (r *KataConfigOpenShiftReconciler) updateStatus() error {
 		return len(nodes.Items)
 	}()
 
+	kataSelector, selErr := r.getKataConfigNodeSelectorAsSelector()
+	if selErr != nil {
+		return fmt.Errorf("failed to build KataConfig node selector: %w", selErr)
+	}
 	for _, node := range nodeList.Items {
-		e := r.putNodeOnStatusList(&node)
+		e := r.putNodeOnStatusList(&node, kataSelector)
```
thejasn:
```suggestion
        if !isConvergedCluster && !kataNodeSelector.Matches(labels.Set(node.Labels)) {
            continue
        }
        e := r.putNodeOnStatusList(&node, isConvergedCluster)
```

This also pulls out the `checkConvergedCluster()` outside of `putNodeOnStatusList()` since this is a cluster attribute and not a node level attribute (we don't need recomputing for every node).

thejasn (changes requested):
Hi @ranjithrajaram! 
Thanks for the PR. A few suggestions, PTAL. 

Also the RCA seems to be a bit different. Nodes in custom MCPs (e.g. `border`) have `currentConfig` = rendered-border-* but `putNodeOnStatusList()` hardcodes `targetMcpName = "worker"` for any node without the kata-oc label, so `nodeTargetMc` is always rendered-worker-* i.e. a permanent mismatch that satisfies all three conditions of `isNodeWaitingToUninstall` (`NodeDone` && `currMc` != `targetMc` && `!isKataEnabled`), incorrectly placing those nodes in the `waitingToUninstall` status list despite Kata never having been installed on them.

Please do update the commit message accordingly.

ranjithrajaram:
@thejasn thanks for the feedback. I have added one more condition on top of your suggested changes. 

For this scenario: 

Node has node-role.kubernetes.io/kata-oc label and Kata installed (selector: disk=ssd)
KataConfig selector changed to disk=nvme (node no longer matches)
updateNodeLabels() removes kata-oc label; node moves to worker MCP
Node begins uninstalling Kata via MCO
updateStatus() runs: node doesn't match new selector → skipped
Node doesn't appear in WaitingToUninstall or Uninstalling status lists
User loses visibility into Kata uninstallation progress
The skip condition should only exclude nodes that both don't match the selector and never had the kata-oc label. Nodes transitioning out of Kata must remain tracked.

Plz check last commit

controllers/openshift_controller.go
ranjithrajaram: Added error wrapping

PR #2158, opened 2026-05-19
pre-install all binaries and prefetch azure CLI for podvm

balintTobik:
install all binaries and pre-fetch azure CLI in docker build phase to remove dependencies in case of disconnected setup

During podvm creation required binaries and CLI are downloaded and installed. 
In case of disconnected setups the podvm creation fails because these binaries and CLIs 
cannot download because lack of external connection.

**- What I did**
Install all required binaries and pre-fetch azure CLI rpm in docker build time.

**- How to verify it**
Run osc-podvm-create-job.yaml with new image and check podvm image is created and uploaded to the cloud provider on disconnected azure environment.

install all binaries and pre-fetch azure CLI rpm in docker build phase to remove dependencies in case of disconnected setup


config/peerpods/podvm/azure-podvm-image-handler.sh
```diff
@@ -119,7 +119,18 @@ function install_azure_cli() {
         return
     fi
 
-    # Add azure cli repo
+    # Check if Azure CLI RPMs were prefetched by Konflux (x86_64 only)
```
littlejawa: If things work well, the prefetcher is transparent.
You don't need to check whether you install with or without it - you just do "dnf install azure-cli", and it will find it if it is prefetched correctly.

i.e: you should not have to edit your script because of our build framework.

.tekton/osc-podvm-builder-push.yaml
```diff
@@ -39,7 +39,7 @@ spec:
   - name: build-args-file
     value: .tekton/build-args.env
   - name: prefetch-input
-    value: '{"type": "rpm", "path": "./config/peerpods/podvm/"}'
+    value: '[{"type": "rpm", "path": "./config/peerpods/podvm/"}, {"type": "rpm", "path": "./config/peerpods/podvm/cloud-cli/"}, {"type": "generic", "path": "./config/peerpods/podvm/", "lockfile": "artifacts.lock.yaml"}]'
```
littlejawa: Why do you use a second "rpm" prefetcher ? Why not edit the existing rpms.in.yaml with your new repo + packages?

config/peerpods/podvm/cloud-cli/rpms.lock.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
littlejawa: Was this file generated with "rpms-lockfile-prototype"?

config/peerpods/podvm/artifacts.lock.yaml
```diff
@@ -0,0 +1,15 @@
+metadata:
```
littlejawa: We should not download our binaries from third parties. We need to install them from RPM repositories, or from sources. You may not have it yet (hidden by other errors), but I suspect you'll get a Conforma failure because of this.

I know that "yq" is used in other projects that build in konflux, and I think they use a "known image" of yq build. Maybe you could use that too?
I don't know for the rest.

In general, I've tried to avoid this (see https://github.com/openshift/kata-containers/blob/72c1c3eb329f6e839def49faf4cf8e851bbf592c/Dockerfile.monitor#L8-L11)

And when I could not avoid it, I installed from sources using this kind of hack:
https://github.com/openshift/sandboxed-containers-operator/blob/devel/controller-gen
=> use "go" prefetcher to get the sources, then use "go tool" to run the tool from the prefetched sources.

I'm afraid you'll need to find your own solution, for each of those tools...

config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -37,6 +38,26 @@ ARG CAA_SRC=https://github.com/confidential-containers/cloud-api-adaptor
 ARG CAA_REF=main
 ARG CERT_RPM
 
+# Extract and install the binaries
+RUN cd /cachi2/output/deps/generic/ && \
```
littlejawa: As stated in another comment: I don't think this way of installing build tools is conforma-compliant...

littlejawa (commented):
Doing a quick review, I'm afraid some of your changes will raise other problems down the road...
Have a look, and double-check the errors you get from Conforma.

beraldoleal:
I'm trying to trigger konflux to build the img again, since it expired. I would like to test the img generated (even tho is not finished) in a disconnected cluster.

/test all

beraldoleal:
@littlejawa not sure if the prow jobs I just triggered will rebuild the img, I guess only with a new push from Balint?

littlejawa:
@beraldoleal - you need to use '/retest' to trigger the Konflux pipelines again.

config/peerpods/podvm/cloud-cli/rpms.lock.yaml
balintTobik: yes

config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -37,6 +37,35 @@ ARG CAA_SRC=https://github.com/confidential-containers/cloud-api-adaptor
...
 ARG CERT_RPM
 
+RUN dnf install -y golang python3 python3-pip python3-wheel krb5-devel gpgme-devel libassuan-devel gcc libffi-devel python3-devel rust cargo openssl-devel && dnf clean all
+
+# Copy go.mod, go.sum and tools.go to use our dependency versions
+COPY go.mod go.sum /tmp/tools/
+WORKDIR /tmp/tools
+
+# Build and install tools using our module's dependency versions
+RUN . /cachi2/cachi2.env && \
```
littlejawa: you don't need to add the ". /cachi2/cachi2.env &&" line. It is added automatically by Konflux during an hermetic build

balintTobik: Thanks! I fixed it!

config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -37,6 +37,36 @@ ARG CAA_SRC=https://github.com/confidential-containers/cloud-api-adaptor
...
 ARG CERT_RPM
 
+RUN dnf install -y golang python3 python3-pip python3-wheel krb5-devel gpgme-devel libassuan-devel gcc libffi-devel python3-devel rust cargo openssl-devel && dnf clean all
+
+# Copy go.mod, go.sum and tools.go to use our dependency versions
+COPY go.mod go.sum azure-cli-pip.txt /tmp/tools/
+WORKDIR /tmp/tools
+
+# Build and install tools using our module's dependency versions
+RUN GOSUMDB=off GOBIN=/usr/local/bin go install github.com/hashicorp/packer && \
```
snir911: if we install this by default, shouldn't we remove these from installtion at runtime?

config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -37,6 +37,36 @@ ARG CAA_SRC=https://github.com/confidential-containers/cloud-api-adaptor
...
+    GOSUMDB=off GOBIN=/usr/local/bin go install github.com/mikefarah/yq/v4 && \
+    GOSUMDB=off GOBIN=/usr/local/bin go install github.com/opencontainers/umoci/cmd/umoci && \
+    GOSUMDB=off GOBIN=/usr/local/bin go install github.com/openshift/oc/cmd/oc && \
+    ln -s /usr/local/bin/oc /usr/local/bin/kubectl
+
+# Install Azure CLI from prefetched pip packages
+RUN if [ -d "/cachi2/output/deps/pip" ]; then \
```
snir911, line 53: shouldn't we do the else case here and let it install not from cache for the non-konflux case?

config/peerpods/podvm/Dockerfile.podvm-builder
balintTobik: In runtime the script check if the binary is exists in /usr/local/bin path.

config/peerpods/podvm/Dockerfile.podvm-builder
balintTobik: done

config/peerpods/podvm/Dockerfile.podvm-builder
```diff
@@ -37,6 +37,34 @@ ARG CAA_SRC=https://github.com/confidential-containers/cloud-api-adaptor
 ARG CAA_REF=main
 ARG CERT_RPM
 
+RUN dnf install -y golang python3 python3-pip python3-wheel krb5-devel gpgme-devel libassuan-devel gcc libffi-devel python3-devel rust cargo openssl-devel && dnf clean all
+
+# Copy go.mod, go.sum and tools.go to use our dependency versions
```
snir911: tools.go?

config/peerpods/podvm/redhat.repo
```diff
@@ -9,12560 +9,13277 @@
 # a "yum repolist" to refresh available repos
 #
 
-[cnv-4.16-for-rhel-9-$basearch-rpms]
-name = Red Hat Container Native Virtualization 4.16 for RHEL 9 $basearch (RPMs)
-baseurl = https://cdn.redhat.com/content/dist/layered/rhel9/$basearch/cnv/4.16/os
+[rhceph-5-tools-for-rhel-9-$basearch-debug-rpms]
```
snir911, line 12: the changes in redhat.repo are auto generated?

.tekton/osc-podvm-builder-pull-request.yaml
```diff
@@ -42,7 +42,12 @@ spec:
   - name: build-args-file
     value: .tekton/build-args.env
   - name: prefetch-input
-    value: '{"type": "rpm", "path": "./config/peerpods/podvm/"}'
```
snir911, line 45: pleas mention in the commit message all the pkgs which are now pre-installed, also mention this is intended to help in disconnected setup on Azure (and x86) as all the other cli tools aren't installed ( or install them :) but i'm afraid it will increase the image too much)

snir911 (commented):
Do you know how much it increases the image size?
The builder is becoming a Swiss knife of public cloud tools, no ideal, however, the only alternative i can think of is to maintain multiple images, if someone has better idea pls suggest..
cc @littlejawa for approving the hermetic build view
cc @ajayvic to ensure it doesn't affect the s390 things

config/peerpods/podvm/go.mod
```diff
@@ -0,0 +1,388 @@
...
+	github.com/containerd/typeurl/v2 v2.2.3 // indirect
+	github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01 // indirect
+	github.com/containers/ocicrypt v1.3.0 // indirect
+	github.com/coreos/go-oidc/v3 v3.18.0 // indirect
+	github.com/cpuguy83/go-md2man/v2 v2.0.7 // indirect
+	github.com/cyberphone/json-canonicalization v0.0.0-20241213102144-19d51d7fe467 // indirect
+	github.com/cyphar/filepath-securejoin v0.6.1 // indirect
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
+	github.com/dimchansky/utfbom v1.1.1 // indirect
+	github.com/distribution/distribution/v3 v3.0.0-20230519140516-983358f8e250 // indirect
```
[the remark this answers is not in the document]
beraldoleal: @balintTobik please try to add a replace directives to go.mod to fix the high-severity CVEs flagged by CodeRabbit:

```
  replace github.com/distribution/distribution/v3 => github.com/distribution/distribution/v3 v3.1.1
  replace github.com/moby/spdystream => github.com/moby/spdystream v0.5.1
  replace github.com/go-git/go-git/v5 => github.com/go-git/go-git/v5 v5.16.5
```

beraldoleal (commented):
From what I could see, the image built successfully here and all binaries are present and functional (packer, yq, oc, umoci, az). However, I noticed the PR pipeline built without cachi2 (Go modules and Azure CLI were downloaded from the internet using proxy.golang.org and PyPI, rather than using prefetched dependencies.

@littlejawa, the Tekton pipeline configs include prefetch-input for `gomod/pip/rpm`, but I'm not sure how the build system works here. Can the PR pipeline run with hermetic builds enabled, or does that only happen post-merge? Want to make sure the cachi2 conditional paths are tested since this is an important img.

config/peerpods/podvm/redhat.repo
balintTobik: yes, it was generated by https://gitlab.cee.redhat.com/osc/osc-internal-utility/-/blob/main/update-rpms-lock/update-rpms-lock.sh

littlejawa:
@beraldoleal - Yes, the "on-pr" build is hermetic too.
I can't access the logs anymore (time-bombed), but it should have run in network isolation.
If you've seen it download stuff, then there's something very wrong going on.

Now from the build log's perspective, maybe you can't tell the difference - hermeto just configure stuff to look at its local cache, where they are "downloaded" from...

What's important is [the prefetch phase](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/ose-osc-tenant/applications/openshift-sandboxed-containers/pipelineruns/osc-podvm-builder-on-pull-request-ln78w/logs?task=ln78w-prefetch-dependencies), where we can see it has retrieved go, pip and rpm dependencies.

Anyway I will start another build and look how it goes just make sure.

littlejawa:
@beraldoleal -
Build is still ongoing, but I can see the prefetch phase ran successfully, and I see this at the beginning of the build phase:

```
Build will be executed with network isolation
[2026-06-18T16:06:20,758746613+00:00] Setup prefetched
Prefetched content will be made available
````

I can also see that all the dockerfile "RUN" steps have the prepended ". /cachi2/cachi2.env &&" string, which means hermeto did its job.

For me, the build is hermetic.

**Comment by littlejawa** [comment]
> Do you know how much it increases the image size? The builder is becoming a Swiss knife of public cloud tools, no ideal, however, the only alternative i can think of is to maintain multiple images, if someone has better idea pls suggest.. 

For the record:
From ~180MiB to ~500MiB

And also for the build time:
From ~12min to ~45/50 min

That makes podvm-builder as long to build as podvm-payload and dm-verity :-( 

Unfortunately, I don't have an alternative to offer.
This image needs to run those tools, and if we can't download them at run time, we need to ship them...

File: config/peerpods/podvm/Dockerfile.podvm-builder (hunk 9 above)
Comment (balintTobik) [reply config/peerpods/podvm/Dockerfile.podvm-builder]: it's just a leftover from a previous try. I've removed it.

File: .tekton/osc-podvm-builder-pull-request.yaml (hunk 11 above)
Comment (balintTobik) [reply .tekton/osc-podvm-builder-pull-request.yaml:45]: done

File: config/peerpods/podvm/go.mod [hunk 13]
```diff
@@ -0,0 +1,250 @@
...
+	github.com/distribution/reference v0.6.0 // indirect
+	github.com/docker/docker v28.5.2+incompatible // indirect
```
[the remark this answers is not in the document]
balintTobik: github.com/openshift/oc hardcodes docker/docker v28.5.2+incompatible and imports packages (pkg/archive, pkg/idtools, pkg/system, api/types/*) using old import paths

config/peerpods/podvm/go.mod
```diff
@@ -0,0 +1,250 @@
...
+	github.com/containerd/typeurl/v2 v2.2.3 // indirect
+	github.com/containers/libtrust v0.0.0-20230121012942-c1716e8a8d01 // indirect
+	github.com/containers/ocicrypt v1.3.0 // indirect
+	github.com/coreos/go-oidc/v3 v3.18.0 // indirect
+	github.com/cpuguy83/go-md2man/v2 v2.0.7 // indirect
+	github.com/cyberphone/json-canonicalization v0.0.0-20241213102144-19d51d7fe467 // indirect
+	github.com/cyphar/filepath-securejoin v0.6.1 // indirect
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
+	github.com/dimchansky/utfbom v1.1.1 // indirect
+	github.com/distribution/distribution/v3 v3.0.0-20230519140516-983358f8e250 // indirect
```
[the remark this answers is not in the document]
balintTobik: github.com/openshift/oc hardcodes v3.0.0-20230519140516-983358f8e250

config/peerpods/podvm/go.mod
```diff
@@ -0,0 +1,250 @@
...
+	github.com/klauspost/pgzip v1.2.6 // indirect
+	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
+	github.com/lithammer/dedent v1.1.0 // indirect
+	github.com/magiconair/properties v1.8.10 // indirect
+	github.com/mailru/easyjson v0.9.0 // indirect
+	github.com/mattn/go-colorable v0.1.14 // indirect
+	github.com/mattn/go-isatty v0.0.20 // indirect
+	github.com/mikefarah/yq/v4 v4.49.2 // indirect
+	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
+	github.com/moby/buildkit v0.12.5 // indirect
```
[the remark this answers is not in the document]
balintTobik: BuildKit v0.28.1+ requires moby/go-archive v0.2.0, which is incompatible with docker/docker v28.5.2

config/peerpods/podvm/go.mod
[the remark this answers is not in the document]
beraldoleal: @balintTobik ... well.. from your logs, the replace won't work, v3.1.1 doesn't have the registry/client, reference, and manifest/schema1 packages that oc still uses.

But we might not need it anyway. CVE-2026-33540 only affects pull-through cache proxy mode. From what I can see, oc uses distribution for manifest handling and image mirroring, not as a proxy. Probably no real exposure. Not sure if @coderabbitai could investigate further.

So if not a real exposure, I would suggest leaving as-is, maybe add a comment in go.mod explaining the CVE doesn't apply to our usage. And maybe we could wait for openshift/oc to bump their version and catch the fix for free.

beraldoleal:
@littlejawa, checked the pipeline and image to confirm hermetic build is working:

```
  # Image built (latest commit)
  $ gh pr view 2158 --json commits | jq -r '.commits[-1].oid'
  3951d1aa9b5af3587ae731c66e8d86a89843cc7b

  # Verify hermetic paths used
  $ podman history --no-trunc quay.io/.../osc-podvm-builder:on-pr-3951d1aa | grep "pip3 install"
  /bin/sh -c . /cachi2/cachi2.env && if [ -d "/cachi2/output/deps/pip" ]; then \
      echo "Installing Azure CLI from prefetched packages..." && \
      pip3 install --no-index --find-links /cachi2/output/deps/pip --require-hashes -r /tmp/tools/azure-cli-pip.txt

  # Go modules also using cachi2
  $ podman history --no-trunc quay.io/.../osc-podvm-builder:on-pr-3951d1aa | grep "go install"
  /bin/sh -c . /cachi2/cachi2.env && GOSUMDB=off GOBIN=/usr/local/bin go install github.com/mikefarah/yq/v4 && ...

  # Verify binaries installed
  $ podman run --rm --entrypoint=/bin/bash quay.io/.../osc-podvm-builder:on-pr-3951d1aa -c 'az --version | head -1'
  azure-cli                         2.79.0 *
```
So looks like hermetic paths are being triggered this time. Not sure if something changed in the pipeline config, but anyway, good to see it working. Thanks.

beraldoleal:
> > Do you know how much it increases the image size? The builder is becoming a Swiss knife of public cloud tools, no ideal, however, the only alternative i can think of is to maintain multiple images, if someone has better idea pls suggest..
> 
> For the record: From ~180MiB to ~500MiB
> 
> And also for the build time: From ~12min to ~45/50 min
> 
> That makes podvm-builder as long to build as podvm-payload and dm-verity :-(
> 
> Unfortunately, I don't have an alternative to offer. This image needs to run those tools, and if we can't download them at run time, we need to ship them...

Yes, I guess this is expected. :(

config/peerpods/podvm/lib.sh
```diff
@@ -85,7 +86,6 @@ install_binary_packages() {
     else
         # Define the required binary packages
         REQUIRED_BINARY_PACKAGES=(
-            "packer=https://releases.hashicorp.com/packer/1.9.4/packer_1.9.4_linux_amd64.zip"
```
snir911, line 88: LGTM overall thanks!
qq, keeping it without pre-installation will break anything?

balintTobik: Thanks!
Yes! The script test if the binary is exists:
https://github.com/openshift/sandboxed-containers-operator/blob/devel/config/peerpods/podvm/lib.sh#L103-L105

If one is not exists  try to download that and it will fail on disconnected setup.

In case of libvirt on s390x the the packer will be fetched anyway and installed from source:
https://github.com/openshift/sandboxed-containers-operator/blob/382781f0b5e00defecc098ccb38b758ff2b2811e/config/peerpods/podvm/libvirt-podvm-image-handler.sh#L385-L396

beraldoleal (approved):

snir911 (approved):
LGTM thanks

littlejawa (approved):
/lgtm
Thanks @balintTobik !

PR #2189, opened 2026-05-27
peerpods: add AWS IRSA support for workload identity

snir911:
Add support for AWS IAM Roles for Service Accounts (IRSA) to enable workload identity authentication for peer-pods, similar to the existing Azure workload identity support.

- credentials_controller: detect ROLEARN env var and create peer-pods-secret with AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE
- peerpods: mount bound-sa-token volume to CAA DaemonSet when IRSA detected
- image_generator: validate IRSA credentials as alternative to static keys
- Refactor addBoundSATokenVolume into shared helper for Azure and AWS

**- What I did**

**- How to verify it**

### DIFF
--- a/config/peerpods/podvm/ami-helper.sh
+++ b/config/peerpods/podvm/ami-helper.sh
@@ -169,6 +169,26 @@ delete_bucket() {
 	aws s3api delete-bucket --bucket ${BUCKET_NAME} --region ${REGION}
 }
 
+delete_vmimport_role() {
+	echo "Delete vmimport role"
+	if ! aws iam get-role --role-name vmimport --region ${REGION} &>/dev/null; then
+		echo "vmimport role does not exist, skipping deletion"
+		return 0
+	fi
+
+	echo "Deleting inline policy from vmimport role"
+	# Ignore NoSuchEntity error to make cleanup idempotent
+	aws iam delete-role-policy --role-name vmimport --policy-name vmimport --region ${REGION} 2>/dev/null || true
+
+	echo "Deleting vmimport role"
+	if ! aws iam delete-role --role-name vmimport --region ${REGION}; then
+		echo "Error: Failed to delete vmimport role"
+		return 1
+	fi
+
+	echo "Successfully deleted vmimport role"
+}
+
 create_bucket() {
 	echo "Create s3 Bucket named ${BUCKET_NAME} at ${REGION}"
 	if [[ ${REGION} == us-east-1 ]]; then
@@ -303,7 +323,7 @@ prepare
 
 [[ $in_place ]] && extended_credentials
 
-[[ $clean_credentials ]] && clean_credentials; [[ $delete_bucket ]] && delete_bucket; [[ $clean_credentials || $delete_bucket ]] && exit 0
+[[ $delete_bucket ]] && delete_bucket && delete_vmimport_role; [[ $clean_credentials ]] && clean_credentials; [[ $clean_credentials || $delete_bucket ]] && exit 0
 
 create_bucket
 
--- a/config/peerpods/podvm/ami-helper.sh
+++ b/config/peerpods/podvm/ami-helper.sh
@@ -13,9 +13,9 @@ Usage: $(basename $0) [options]
    -c               Clean credentials and exit
    -d               Delete the bucket and exit
    -h               Print this help message
-   -i               In cluster
+   -i               Request and use extended credentials in-place for this execution
    -r <region>      Set the region (otherwise it will be fetched from the cluster)
-   -s               Skip credentials request
+   -s               Skip requesting extended credentials for operator image operations
 EOF
 }
 
@@ -25,13 +25,16 @@ while getopts ":b:cdhir:s" opt; do
 		c ) clean_credentials=true;;
 		d ) delete_bucket=true;;
 		h ) usage && exit 0;;
-		i ) in_cluster=true;;
+		i ) in_place=true;;
 		r ) REGION=$OPTARG;;
 		s ) skip_cr=true;;
 		\? ) echo "Invalid option: -$OPTARG" >&2 && usage && exit 1;;
 	esac
 done
 
+# Request extended credentials from CCO and export them to the current environment
+# for immediate use during this script execution. Used when running with -i flag
+# (typically inside cluster jobs that need elevated permissions).
 extended_credentials() {
 	echo "Creating extended credentials for IAM role management, bucket creation and operation credentials"
 
@@ -224,9 +227,12 @@ clean_credentials() {
 	oc delete credentialsrequest openshift-sandboxed-containers-aws-image -n openshift-cloud-credential-operator
 }
 
+# Request extended credentials from CCO for future operator image creation operations.
+# Creates peer-pods-image-creation-secret that will be used by automated image jobs.
+# Does NOT modify current environment - credentials are saved in secret for later use only.
 get_operation_credentials() {
 	[[ $skip_cr ]] && return
-	echo "Ask for addtional credentials"
+	echo "Ask for additional credentials"
 	oc get ns openshift-sandboxed-containers-operator >/dev/null 2>&1 || (echo "OSC namespace is missing, re-run after installting the operator" && exit 1)
 
 	cat <<EOF > "${CREDENTIALS_REQUEST_YAML_FILE}"
@@ -295,7 +301,7 @@ init
 
 prepare
 
-[[ $in_cluster ]] && extended_credentials
+[[ $in_place ]] && extended_credentials
 
 [[ $clean_credentials ]] && clean_credentials; [[ $delete_bucket ]] && delete_bucket; [[ $clean_credentials || $delete_bucket ]] && exit 0
 
@@ -303,4 +309,4 @@ create_bucket
 
 set_service_role
 
-[[ $in_cluster ]] || get_operation_credentials
+[[ $in_place ]] || get_operation_credentials
--- a/config/peerpods/podvm/ami-helper.sh
+++ b/config/peerpods/podvm/ami-helper.sh
@@ -176,9 +176,25 @@ delete_vmimport_role() {
 		return 0
 	fi
 
-	echo "Deleting inline policy from vmimport role"
-	# Ignore NoSuchEntity error to make cleanup idempotent
-	aws iam delete-role-policy --role-name vmimport --policy-name vmimport --region ${REGION} 2>/dev/null || true
+	echo "Detaching managed policies from vmimport role"
+	# List and detach any managed policies
+	managed_policies=$(aws iam list-attached-role-policies --role-name vmimport --region ${REGION} --query 'AttachedPolicies[*].PolicyArn' --output text 2>/dev/null || true)
+	if [[ -n "$managed_policies" ]]; then
+		for policy_arn in $managed_policies; do
+			echo "Detaching managed policy: $policy_arn"
+			aws iam detach-role-policy --role-name vmimport --policy-arn "$policy_arn" --region ${REGION} || true
+		done
+	fi
+
+	echo "Deleting inline policies from vmimport role"
+	# List and delete all inline policies
+	inline_policies=$(aws iam list-role-policies --role-name vmimport --region ${REGION} --query 'PolicyNames[*]' --output text 2>/dev/null || true)
+	if [[ -n "$inline_policies" ]]; then
+		for policy_name in $inline_policies; do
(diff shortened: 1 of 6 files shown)


config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -215,13 +246,26 @@ function prepare_for_prebuilt_artifact() {
...
+        echo "CCO flow detected, requesting credential extension"
+        /scripts/ami-helper.sh -i -b ${BUCKET_NAME} || error_exit "Failed to extend credentials using the helper script"
+        export AWS_ACCESS_KEY_ID=$(kubectl get secret ${extended_secret_name} -n openshift-sandboxed-containers-operator -o jsonpath='{.data.aws_access_key_id}' | base64 -d)
+        export AWS_SECRET_ACCESS_KEY=$(kubectl get secret ${extended_secret_name} -n openshift-sandboxed-containers-operator -o jsonpath='{.data.aws_secret_access_key}' | base64 -d)
+    else
+        echo "Using manual credentials (expecting s3 ${BUCKET_NAME} bucket to exist and vmimport service role to be set)"
+
+        # Validate bucket exists
+        aws s3api head-bucket --bucket ${BUCKET_NAME} --region ${AWS_REGION} 2>/dev/null || \
+            error_exit "Bucket ${BUCKET_NAME} not found or not accessible. Manual credentials require pre-created infrastructure (bucket and vmimport role)."
```
bpradipt: For my understanding, when using key and secret, who creates the bucket and role ?

snir911: Currently user has to create it manually (the ami-helper.sh would do this), it is mentioned in the bootc doc, however, i should mention it in the credentials doc also, maybe even in the error msg..
The other option is to consider doing it automatically also, downside is - it will require also bucket and role permissions permanently..

bpradipt: Mentioning in the doc should be fine for now. I think its safer to have automation for the workload identity flow only.

bpradipt (approved):
/lgtm

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -200,13 +206,38 @@ function bootc_to_ami() {
     aws_ami_name="${4}"
     aws_bucket="${5}" # bucket must exist
 
-    [[ -n "$AWS_ACCESS_KEY_ID" ]] && [[ -n "$AWS_SECRET_ACCESS_KEY" ]] && [[ -n "$AWS_REGION" ]] || error_exit "bootc_to_ami failed: AWS_* keys or AWS_REGION are missing"
-    # TODO: check permissions
-    run_args="--env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY"
+    # Validate credentials - support both static and IRSA
+    if [[ -n "$AWS_ACCESS_KEY_ID" && -n "$AWS_SECRET_ACCESS_KEY" && -n "$AWS_REGION" ]]; then
```
wainersm: `verify_vars()` seems to ensure that AWS_REGION is always set, so this extra `-n $AWS_REGION` is not necessary?

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -200,13 +206,38 @@ function bootc_to_ami() {
...
+        echo "bootc_to_ami: Using static AWS credentials"
+	# TODO: pass as env file
+        run_args="--env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY"
+    elif [[ -n "$AWS_ROLE_ARN" && -n "$AWS_WEB_IDENTITY_TOKEN_FILE" && -n "$AWS_REGION" ]]; then
```
wainersm: Same here (regarding AWS_REGION)

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -200,13 +206,38 @@ function bootc_to_ami() {
     aws_ami_name="${4}"
     aws_bucket="${5}" # bucket must exist
 
-    [[ -n "$AWS_ACCESS_KEY_ID" ]] && [[ -n "$AWS_SECRET_ACCESS_KEY" ]] && [[ -n "$AWS_REGION" ]] || error_exit "bootc_to_ami failed: AWS_* keys or AWS_REGION are missing"
-    # TODO: check permissions
-    run_args="--env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY"
+    # Validate credentials - support both static and IRSA
+    if [[ -n "$AWS_ACCESS_KEY_ID" && -n "$AWS_SECRET_ACCESS_KEY" && -n "$AWS_REGION" ]]; then
+        echo "bootc_to_ami: Using static AWS credentials"
+	# TODO: pass as env file
```
wainersm, line 214: is there a reason to not address this TODO now?

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -200,13 +206,38 @@ function bootc_to_ami() {
...
+        run_args="--env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY"
+    elif [[ -n "$AWS_ROLE_ARN" && -n "$AWS_WEB_IDENTITY_TOKEN_FILE" && -n "$AWS_REGION" ]]; then
+        error_exit "bootc_to_ami failed: IRSA are not yet supported by the Bootc Image Builder"
+	# see https://github.com/osbuild/bootc-image-builder#aws-credentials-file
```
wainersm, line 218: https://github.com/osbuild/bootc-image-builder#aws-credentials-file doesn't explicit say IRSA is not supported yet. Did you try it?

If this is true then I think it should be documented in the user guide because it will be an important limitation.

What scenario (CoCo vs non-CoCo, pre-built vs built..etc) this bootc is used to publish the podvm ami?

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -200,13 +206,38 @@ function bootc_to_ami() {
...
 
+# Check if peer-pods-secret is based on CCO-created secret (needs permissions extension) or otherwise (should have full permissions).
+# Returns 0 if CCO flow, 1 otherwise (should have full permissions).
+function is_cco_flow_secret() {
+    local cco_label=$(kubectl get secret peer-pods-secret \
+        -n openshift-sandboxed-containers-operator \
+        -o jsonpath='{.metadata.labels.kataconfiguration\.openshift\.io/credentials-request-based}' 2>/dev/null)
+
+    if [[ "$cco_label" == "true" ]]; then
+        echo "CCO flow secret detected (credential extension needed)"
```
wainersm: I find this "echo" (and the next) unneeded. The caller is already logging a message saying the credential will be extended.

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -200,13 +206,38 @@ function bootc_to_ami() {
...
+    fi
+
     bib_args="--type ami --aws-ami-name ${aws_ami_name} --aws-bucket ${aws_bucket} --aws-region ${AWS_REGION}"
     bootc_image_builder_conversion "${container_image_repo_url}" "${image_tag}" "${auth_json_file}" "${run_args}" "${bib_args}"
 }
 
+# Check if peer-pods-secret is based on CCO-created secret (needs permissions extension) or otherwise (should have full permissions).
+# Returns 0 if CCO flow, 1 otherwise (should have full permissions).
+function is_cco_flow_secret() {
+    local cco_label=$(kubectl get secret peer-pods-secret \
```
wainersm, line 230: It seems a little fragile. If there is an error then it is assumed it's not CCO which might not be true. I think we should capture-and-exit-error if it for some reason fails to detect the correct flow.

Also what happens if peer-pods-secret doesn't exist at the time the function is called? is it possible?

config/peerpods/podvm/aws-podvm-image-handler.sh
```diff
@@ -200,13 +206,38 @@ function bootc_to_ami() {
...
+        -n openshift-sandboxed-containers-operator \
+        -o jsonpath='{.metadata.labels.kataconfiguration\.openshift\.io/credentials-request-based}' 2>/dev/null)
+
+    if [[ "$cco_label" == "true" ]]; then
+        echo "CCO flow secret detected (credential extension needed)"
+        return 0
+    else
+        echo "Manually-created or STS secret detected (assuming full permissions)"
+        return 1
+    fi
```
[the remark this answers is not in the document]
wainersm: Related with what I said above

controllers/peerpods.go
```diff
@@ -245,36 +245,16 @@ func (r *KataConfigOpenShiftReconciler) configureCAAProvider(ds *appsv1.DaemonSe
...
-				container := &ds.Spec.Template.Spec.Containers[i]
-				if container.Name == "caa-pod" {
-					container.VolumeMounts = append(container.VolumeMounts, boundSATokenVolumeMount)
-				}
-			}
+		return ds, nil
+	case AWSProvider:
+		// Only add bound-sa-token volume for AWS IRSA if using STS flow
+		// Check for STS environment variable (set during OLM installation)
+		hasAWSSTSCreds := os.Getenv("ROLEARN") != ""
```
wainersm, line 255: Assigning to hasAWSSTSCreds is not needed here as that variable is used once. I code better than Claude! :D

docs/credentials-handling.md
```diff
@@ -38,8 +38,29 @@ data:
   AZURE_CLIENT_SECRET: <base64-encoded-client-secret>
   AZURE_TENANT_ID: <base64-encoded-tenant-id>
   AZURE_SUBSCRIPTION_ID: <base64-encoded-subscription-id>
+  # AWS example
```
wainersm: TBH, I would prefer to break this yaml in two: one for azure and another for aws.

config/peerpods/podvm/ami-helper.sh
```diff
@@ -303,7 +316,7 @@ prepare
 
 [[ $in_place ]] && extended_credentials
 
-[[ $clean_credentials ]] && clean_credentials; [[ $delete_bucket ]] && delete_bucket; [[ $clean_credentials || $delete_bucket ]] && exit 0
+[[ $clean_credentials ]] && clean_credentials; [[ $delete_bucket ]] && delete_bucket && delete_vmimport_role; [[ $clean_credentials || $delete_bucket ]] && exit 0
```
wainersm: Those commands concated in logical operations usually is confusing. I suggest to use good and old `if`blocks.

wainersm:
Hi @snir911 ,

I'v finished my review. Mostly nits, except for the comment about bootc ami publishing not being supported. If the case and this is a user visible limitation then it should be documented on user guide.

config/peerpods/podvm/aws-podvm-image-handler.sh
snir911: just out of scope, i plan to work on the bootc part later in any case

config/peerpods/podvm/aws-podvm-image-handler.sh
snir911: i didn't try it, I'll do it together with the bootc work (it isn't used currently, very convenient for custom images, maybe it should removed? IDK yet)

config/peerpods/podvm/aws-podvm-image-handler.sh
snir911: If it is cco and there's an error it will faile on bucket check and job will be executed again..
If it's manual and it fails, than, it will work as it should

if peer-pods doesn't exist it should fail earlier when validating the vars

controllers/peerpods.go
snir911: >  I code better than Claude!

indeed!

i kept it only for consistency with the Azure part

config/peerpods/podvm/ami-helper.sh
snir911: I've modified it a bit, however, i prefer to keep it oneliner only for logic & consistency with lines before/after

config/peerpods/podvm/aws-podvm-image-handler.sh
wainersm: ok, I trust in your assessment :D

wainersm (approved):

PR #2240, opened 2026-06-02
tests: update KATA_RPM_VERSION to version 3.25.0-7

dbkreling:
Kata-containers version 3.25.0-6 RPMs have been built. Note that the version for 4.18 was not updated because it was not affected by the regression. Staying at *-6.

Signed-off-by: Daniel Kreling <dkreling@redhat.com>


wainersm (approved):
Thanks for the update @dbkreling !

PR #2291, opened 2026-06-05
controller: skip MCP use when MachineConfigPool API is absent on HCP

paulczar:
## Description of the problem
On ROSA HCP guest clusters, the MachineConfigPool API may be absent while other
MCO types exist. OSC 1.12 unconditionally watches MCP at manager startup and
lists MCPs during reconcile, causing crash loops or repeated ERROR logs before
DaemonSet-based Kata install can complete ([KATA-4840](https://redhat.atlassian.net/browse/KATA-4840)).

## What I did
- Probe for MachineConfigPool API availability (same pattern as MachineConfig)
- Register MCP watch only when the API is present
- Short-circuit `checkConvergedCluster()` when MCP is unavailable
- Unit tests for both paths

## How to verify
1. ROSA HCP cluster without MachineConfigPool API
2. Set `osc-feature-gates` `deploymentMode: DaemonSet`
3. Install operator and apply KataConfig
4. Confirm `controller-manager` stays up (no MCP cache sync failure)
5. Confirm reconcile does not ERROR-spam on MCP list each loop
6. Confirm DaemonSet install path progresses (RuntimeClass `kata` when install completes)

## Lab validation (author)
Validated on ROSA HCP with `c5n.metal` workers: DaemonSet mode, compute pool
`auto_repair: false` (ROSA ops — not part of this PR), manual reboot after
`osc-rpm-install`, then `RuntimeClass/kata` and a pod with `runtimeClassName:
kata` running in a KVM sandbox.
Fixes: rhjira#[KATA-4840](https://redhat.atlassian.net/browse/KATA-4840)
Related: [KATA-4233](https://redhat.atlassian.net/browse/KATA-4233) (DaemonSet reboot), [KATA-5177](https://redhat.atlassian.net/browse/KATA-5177) (live-apply), [KATA-4597](https://redhat.atlassian.net/browse/KATA-4597) (DS image)

## Changelog
HCP: skip MachineConfigPool integration when MCP API is absent so DaemonSet
deployment mode can run on guest clusters without full MCO.

### DIFF
--- a/controllers/deployment_mode_handler.go
+++ b/controllers/deployment_mode_handler.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/api/meta"
@@ -29,9 +30,10 @@ const (
 )
 
 const (
-	machineConfigGroup   = "machineconfiguration.openshift.io"
-	machineConfigVersion = "v1"
-	machineConfigKind    = "MachineConfig"
+	machineConfigGroup    = "machineconfiguration.openshift.io"
+	machineConfigVersion  = "v1"
+	machineConfigKind     = "MachineConfig"
+	machineConfigPoolKind = "MachineConfigPool"
 )
 
 func ParseDeploymentModeOption(s string) (DeploymentModeOption, error) {
@@ -106,7 +108,9 @@ func (r *KataConfigOpenShiftReconciler) isMachineConfigAvailable() (bool, error)
 
 	// Attempt to GET any MachineConfig to verify if the GVK is known.
 	// If the CRD isn’t installed, it will return a NoMatchError.
-	err := r.Client.Get(context.Background(), client.ObjectKey{Name: "dummy-machine-config"}, machineConfig)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	err := r.Client.Get(ctx, client.ObjectKey{Name: "dummy-machine-config"}, machineConfig)
 	if err == nil || k8serrors.IsNotFound(err) {
 		r.Log.Info("MachineConfig CRD is present")
 		return true, nil
@@ -114,6 +118,33 @@ func (r *KataConfigOpenShiftReconciler) isMachineConfigAvailable() (bool, error)
 
 	if meta.IsNoMatchError(err) {
 		r.Log.Info("MachineConfig CRD not found")
+		return false, nil
+	}
+
+	return false, err
+}
+
+// isMachineConfigPoolAvailable reports whether the MachineConfigPool API is registered.
+// Hosted control plane guest clusters often omit MCP while still exposing other MCO types;
+// an unconditional MCP watch prevents the operator manager from starting (see KATA-4840).
+func (r *KataConfigOpenShiftReconciler) isMachineConfigPoolAvailable() (bool, error) {
+	mcp := &unstructured.Unstructured{}
+	mcp.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   machineConfigGroup,
+		Version: machineConfigVersion,
+		Kind:    machineConfigPoolKind,
+	})
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	err := r.Client.Get(ctx, client.ObjectKey{Name: "dummy-machine-config-pool"}, mcp)
+	if err == nil || k8serrors.IsNotFound(err) {
+		r.Log.Info("MachineConfigPool CRD is present")
+		return true, nil
+	}
+
+	if meta.IsNoMatchError(err) {
+		r.Log.Info("MachineConfigPool CRD not found")
 		return false, nil
 	}
 
--- a/controllers/openshift_controller.go
+++ b/controllers/openshift_controller.go
@@ -42,6 +42,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	nodeapi "k8s.io/api/node/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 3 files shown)


snir911 (commented):
Thank you! looks good to me, pls fix only the commit msg check failure.

IIRC @c3d  touched hcp, could you pls have a look

controllers/openshift_controller.go
```diff
@@ -1736,18 +1749,29 @@ func (eh *NodeEventHandler) Generic(ctx context.Context, event event.GenericEven
...
 		Watches(
 			&corev1.Node{},
 			&NodeEventHandler{r}).
 		Watches(
 			&corev1.ConfigMap{},
 			&ConfigMapEventHandler{r},
-		).Complete(r)
+		)
+
+	mcpAvailable, err := r.isMachineConfigPoolAvailable()
```
c3d, line 1762: Isn't it possible for an actual MCP CRD to be installed or become available _after_ the manager starts?

controllers/openshift_controller.go
```diff
@@ -761,12 +762,24 @@ func (r *KataConfigOpenShiftReconciler) kataOcExists() (bool, error) {
...
+		return false, nil
+	}
+
+	// Check if only master and worker MCP exists. Worker machinecount should be 0.
 	listOpts := []client.ListOption{}
 	mcpList := &mcfgv1.MachineConfigPoolList{}
-	err := r.Client.List(context.TODO(), mcpList, listOpts...)
+	err = r.Client.List(context.TODO(), mcpList, listOpts...)
 	if err != nil {
+		if meta.IsNoMatchError(err) {
```
c3d, line 779: Is there a case where this happens, i.e. `isMachineConfigPoolAvailable` let us get there and then the MachineConfigPool API is seen as unavailble here?

controllers/deployment_mode_handler.go
```diff
@@ -119,3 +123,30 @@ func (r *KataConfigOpenShiftReconciler) isMachineConfigAvailable() (bool, error)
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	err := r.Client.Get(ctx, client.ObjectKey{Name: "dummy-machine-config-pool"}, mcp)
+	if err == nil || k8serrors.IsNotFound(err) {
+		r.Log.Info("MachineConfigPool CRD is present")
+		return true, nil
+	}
+
+	if meta.IsNoMatchError(err) {
+		r.Log.Info("MachineConfigPool CRD not found")
```
c3d, line 147: What is the difference in meaning, if any, between `MachineConfigPool CRD not found` (here) and `MachineConfigPool API unavailable` (in `openshift_controller.go`)?

If it means the same thing, I suggest we convert all the logging message to state both, something like "CRD not found, API not available" at both places.

controllers/deployment_mode_handler.go
```diff
@@ -106,7 +108,9 @@ func (r *KataConfigOpenShiftReconciler) isMachineConfigAvailable() (bool, error)
 
 	// Attempt to GET any MachineConfig to verify if the GVK is known.
 	// If the CRD isn’t installed, it will return a NoMatchError.
-	err := r.Client.Get(context.Background(), client.ObjectKey{Name: "dummy-machine-config"}, machineConfig)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
c3d, line 111: This is not really related to this PR, but there `context.Background()` here means we may not get a cancellation sent to the reconcile context. So we'd end up with a timeout and not good traceability for the timeout / cancellation. But this is already the case at many other places, so let's not fix it here.

c3d (approved):
Thanks a lot, @paulczar. This is super helpful! It looks good to me, and significant progress for KATA-4840. I did not know you were working on it, so that's great news.

A few minor nits, but nothing really blocking.

controllers/openshift_controller.go
paulczar: I can't think of a case where this would happen,   in my experience on a standard OCP cluster the MCP CRD is installed long before an admin has the chance to deploy operators, and a HCP cluster never has the MCP CRD.

controllers/openshift_controller.go
paulczar: Yes, there is a narrow race: isMachineConfigPoolAvailable could return true and then the CRD is removed before the List call completes. The IsNoMatchError guard at line 779 is there precisely as a defensive safety net for that case (and for any other transient API server state). The dual-check pattern (probe + guard on List) is intentional.

paulczar:
Thanks @c3d I was asked about rosa hcp support for a customer, and decided to investigate and found myself throwing together a PR with what I found ... as one does :)

c3d:
@paulczar I can confirm that this works on my test cluster:

First stage was installation:

> :tada: OSC Installation Complete - Custom Operator Success!
> 
> Installation completed successfully:
> 
> :white_check_mark: Custom OSC operator deployed and running
> - Image: [quay.io/c3d/openshift-sandboxed-containers-operator:hcp-skip-mcp-watch](http://quay.io/c3d/openshift-sandboxed-containers-operator:hcp-skip-mcp-watch)
> - Successfully bypassed MCO dependency for Azure HCP
> - First successful OSC deployment on MCO-less cluster!
> 
> :white_check_mark: Kata packages installed on all worker nodes
> - Both nodes: c3d-hcp-hosted-01-52rp2-8pfx7 and c3d-hcp-hosted-01-52rp2-m6q8x
> - Status: waiting_for_reboot
> - Packages: kata-runtime, qemu-kvm-core, seabios, virtiofsd
> 
> :white_check_mark: RBAC and permissions configured
> - Privileged SCC granted for kata installation
> - Node labeling permissions
> - ConfigMap read permissions
> 
> What this proves:
> - Your custom operator successfully removes the hard MCO dependency
> - OSC DaemonSet mode is now functional on Azure HCP
> - The operator correctly detects MCO absence and adapts

I still need a manual reboot, but after manual reboot everything works fine:

> 🎉 OSC Node Reboot SUCCESS!
> 
> Final Status:
> 
> ✅ Both nodes Ready and operational
> - c3d-hcp-hosted-01-52rp2-8pfx7: Ready
> - c3d-hcp-hosted-01-52rp2-m6q8x: Ready
> 
> ✅ Custom OSC operator running
> - Status: Running (recovered from post-reboot API discovery issue)
> - Restarts: 6 (normal during stabilization)
> 
> ✅ RuntimeClass created!
> - kata: Available (4 minutes ago)
> - kata-nvidia-gpu: Available (4 minutes ago)
> - This means the operator successfully reconciled after reboot
> 
> ✅ kata-install DaemonSet healthy
> - Both pods running on their respective nodes
> - Kata packages loaded and ready
> 
> What this proves:
> - Node reboot successfully loaded kata runtime kernel modules
> - Operator recovered and reconciled KataConfig
> - RuntimeClass resources created automatically
> - Cluster is ready for kata workloads!

controllers/openshift_controller.go
gkurz: FWIW this intentionality would have been best recorded in a comment.

gkurz (commented):
> Fixes: rhjira#KATA-4840

This jira is an epic that covers the full support for HPC clusters.

This PR is fixing a bug. For correctness you should have created an bug in jira instead of linking directly to the epic.

PR #2312, opened 2026-06-16
fix: daemonset: use rpm-ostree install --apply-live to avoid node reboots

paulczar:
Fixes: https://issues.redhat.com/browse/KATA-4233

DaemonSet-based kata installation (used on HCP clusters and when `DaemonSetOption` / `DaemonSetFallbackOption` is selected) previously used plain `rpm-ostree install`, which only stages changes requiring a node reboot. This was disruptive and unnecessary — it forced a full drain/reboot cycle for every kata install and uninstall.

**- What I did**

Switch `rpm-ostree install` to `rpm-ostree install --apply-live` so RPM file contents are applied to the running filesystem immediately, without a node reboot.

Three post-install steps are required that the old reboot path handled automatically:

1. **CRI-O drop-in copy**: `--apply-live` updates `/usr` immediately but OSTree's three-way merge only propagates `/etc` changes at next boot. The kata CRI-O runtime drop-ins land in `/usr/etc` after install; we copy them to `/etc/crio/crio.conf.d/` so CRI-O can see them without a reboot.

2. **Kata VM image symlinks**: `--apply-live` skips RPM `%post` scripts on the live system. The `kata-containers` `%post` normally runs `kata-osbuilder` to build a host-kernel-derived `kata.kernel`/`initrd`. We skip that: the kata guest kernel runs inside a VM and is independent of the host kernel. The RPM ships tested pre-built CC images in `/usr/share/`; we symlink them into the path `configuration.toml` expects.

3. **CRI-O restart**: `systemctl reload` (SIGHUP) only re-reads the config struct; it does not rebuild the internal OCI handler map. Kata only appears in the available runtime list after a full CRI-O restart. Existing pod processes survive the restart — only new CRI operations are briefly unavailable.

For uninstall, `rpm-ostree uninstall` does not support `--apply-live` (only install does). RPM removal is staged for the next reboot, but the CRI-O drop-ins and symlinks are removed immediately and CRI-O is restarted, so kata becomes unavailable without waiting for that reboot.

The `KataWaitingForReboot` node state, `isNodeRebootRequired()` helper, and all associated reboot diagnostic log blocks are removed from the controller — none are needed with `--apply-live`.

**Relationship to #1349**

PR #1349 addresses the same problem but takes a different approach: it moves installation logic into the controller and removes the DaemonSet script. That approach cannot work on HCP clusters, where the controller runs in the management cluster and has no direct access to worker node filesystems — the DaemonSet pod is precisely what enables worker-side `rpm-ostree` execution on HCP. This PR intentionally keeps the existing DaemonSet architecture and makes the minimal change needed to eliminate reboots, ensuring HCP compatibility is preserved.

PR #1349 also found that `systemctl restart crio` triggers a kubelet restart on ROKS, breaking uninstall. We did not observe this on ROSA HCP (OCP 4.20); it may be ROKS-specific behaviour. Testing on bare-metal and standard (non-HCP) ROSA would help confirm.

**- How to verify it**

Deploy on a cluster using DaemonSet mode (e.g. HCP with `DaemonSetFallbackOption`):

1. Create a KataConfig
2. Watch node labels: `kubectl get nodes -L kataconfiguration.openshift.io/kata-ds-rpm-install`
3. Nodes should transition `waiting_to_install` → `installing` → `installed` with no reboot (node uptime unchanged throughout)
4. Run a kata test pod (`runtimeClassName: kata`) — should reach `Running`
5. Confirm QEMU KVM process on node: `oc debug node/<node> -- chroot /host ps -C qemu-kvm`
6. Delete KataConfig and confirm clean uninstall (kata immediately unavailable, node uptime unchanged)

Tested on ROSA HCP (ap-southeast-2, OCP 4.20) — all steps above confirmed passing.

DaemonSet-based kata install/uninstall no longer requires a node reboot; changes take effect immediately using `rpm-ostree --apply-live`.

### DIFF
--- a/scripts/kata-install/osc-kata-install.sh
+++ b/scripts/kata-install/osc-kata-install.sh
@@ -62,8 +62,8 @@ install_kata() {
 		s=\$(rpm-ostree status --json)
 		n_staged=\$(echo \"\$s\" | jq '[.deployments[] | select(.staged)] | length')
 		[[ \$n_staged -eq 0 ]] && exit 1
-		n_kata_staged=\$(echo \"\$s\" | jq '[.deployments[] | select(.staged) | (.\"requested-packages\" // [])[] | select(startswith(\"kata-containers\"))] | length')
-		n_kata_booted=\$(echo \"\$s\" | jq '[.deployments[] | select(.booted) | (.\"requested-packages\" // [])[] | select(startswith(\"kata-containers\"))] | length')
+		n_kata_staged=\$(echo \"\$s\" | jq '[.deployments[] | select(.staged) | ((.\"requested-packages\" // []) + (.\"requested-local-packages\" // []))[] | select(startswith(\"kata-containers\"))] | length')
+		n_kata_booted=\$(echo \"\$s\" | jq '[.deployments[] | select(.booted) | ((.\"requested-packages\" // []) + (.\"requested-local-packages\" // []))[] | select(startswith(\"kata-containers\"))] | length')
 		[[ \$n_kata_booted -gt 0 && \$n_kata_staged -eq 0 ]]
 	"; then
 		echo "Staged removal of kata-containers detected; cancelling pending deployment"
@@ -156,7 +156,7 @@ install_kata() {
 		# Clean up temp dir
 		rm -rf /host/tmp/extensions/
 
-		# rpm-ostree --apply-live updates /usr immediately but /etc changes from
+		# rpm-ostree install --apply-live updates /usr immediately but /etc changes from
 		# RPMs only land after reboot (OSTree three-way merge). Copy the kata
 		# CRI-O drop-ins from the RPM's factory-defaults location now so CRI-O
 		# can see the runtime handlers without a reboot.
@@ -224,7 +224,7 @@ uninstall_kata() {
 	' || true)
 	local uninstall_args=()
 	for pkg in $PACKAGES; do
-		echo "$ostree_pkg_list" | grep -qF "$pkg" && uninstall_args+=("$pkg")
+		echo "$ostree_pkg_list" | grep -xqF "$pkg" && uninstall_args+=("$pkg")
 	done
 	if [[ ${#uninstall_args[@]} -gt 0 ]]; then
 		chroot /host /bin/bash -c "rpm-ostree uninstall ${uninstall_args[*]}"
--- a/scripts/kata-install/osc-kata-install.sh
+++ b/scripts/kata-install/osc-kata-install.sh
@@ -32,47 +32,12 @@ label_node() {
 	kubectl label node "${NODE_NAME}" "${NODE_LABEL}=${state}" --overwrite
 }
 
-# Check whether a package has staged updates
-check_package_updated() {
-	local pkg="$1"
-	local booted staged
-
-	booted=$(chroot /host /bin/bash -c "rpm-ostree status --json | jq -r '.deployments[] | select(.booted==true) | .checksum'")
-	staged=$(chroot /host /bin/bash -c "rpm-ostree status --json | jq -r '.deployments[] | select(.staged==true) | .checksum'")
-
-	if [ -n "$staged" ]; then
-		if chroot /host rpm-ostree db diff "$booted" "$staged" | grep -q "$pkg"; then
-			return 0 # Package was updated
-		fi
-	fi
-
-	return 1 # Package not updated or no staged deployment
-}
-
-# Function to check if a reboot is required by looking for "Staged: yes"
-is_reboot_required() {
-	check_package_updated "kata-containers"
-}
-
-# Loop until reboot is no longer required
-wait_for_reboot_clear() {
-	while is_reboot_required; do
-		echo "Reboot required"
-		set_status_waiting_for_reboot
-		sleep 60
-	done
-}
-
 set_status_installed() {
 	label_node "installed"
 }
 
 set_status_installing() {
 	label_node "installing"
-}
-
-set_status_waiting_for_reboot() {
-	label_node "waiting_for_reboot"
 }
 
 set_status_uninstalling() {
@@ -84,13 +49,26 @@ set_status_uninstalled() {
 }
 
 install_kata() {
-	# Initial wait: avoid doing anything if a previous staged update is pending
-	wait_for_reboot_clear
+	# Compares installed and available package versions. If any packages need
+	# installing or upgrading, runs rpm-ostree install --apply-live so changes
+	# take effect immediately without a node reboot. rpm-ostree cannot upgrade
+	# local layered packages in-place, so outdated versions are uninstalled first.
 
-	# This compares installed and available versions of packages.
-	# If updates or installations are needed, it prepares and runs an rpm-ostree install command with uninstall flags.
-	# rpm-ostree can't update local packages directly, so old versions must be explicitly removed.
-	# If no action is needed, it sleeps indefinitely to prevent pod restarts in a DaemonSet.
+	# If a prior uninstall_kata staged the removal of kata-containers, rpm -q
+	# below still sees kata as installed (the booted deployment still has it)
+	# and would skip reinstalling — leaving the staged removal to silently drop
+	# kata at the next reboot. Detect and cancel that staged removal now.
+	if chroot /host bash -c "
+		s=\$(rpm-ostree status --json)
+		n_staged=\$(echo \"\$s\" | jq '[.deployments[] | select(.staged)] | length')
+		[[ \$n_staged -eq 0 ]] && exit 1
+		n_kata_staged=\$(echo \"\$s\" | jq '[.deployments[] | select(.staged) | (.\"requested-packages\" // [])[] | select(startswith(\"kata-containers\"))] | length')
+		n_kata_booted=\$(echo \"\$s\" | jq '[.deployments[] | select(.booted) | (.\"requested-packages\" // [])[] | select(startswith(\"kata-containers\"))] | length')
+		[[ \$n_kata_booted -gt 0 && \$n_kata_staged -eq 0 ]]
+	"; then
+		echo "Staged removal of kata-containers detected; cancelling pending deployment"
+		chroot /host rpm-ostree cleanup -p
+	fi
 
 	local uninstall_list=()
 	local install_rpms=()
@@ -125,7 +103,28 @@ install_kata() {
 		fi
 	done
 
-	# If nothing to install, exit early
+	# If all packages are live and up-to-date, check they're also staged in
+	# rpm-ostree. After an uninstall cycle without a reboot, packages remain
+	# live via --apply-live but are absent from the staged deployment. Without
+	# re-staging, a subsequent uninstall would fail with "not currently
+	# requested". Force a re-stage if kata is not tracked in any deployment.
+	if [[ ${#install_rpms[@]} -eq 0 ]]; then
+		local n_kata_tracked
+		n_kata_tracked=$(chroot /host bash -c '
+			rpm-ostree status --json | jq "[.deployments[] |
+			  ((.\"requested-packages\" // []) + (.\"requested-local-packages\" // []))[]]
+			| map(select(startswith(\"kata-containers\"))) | length" 2>/dev/null || echo 0
+		')
+		if [[ "${n_kata_tracked:-0}" -eq 0 ]]; then
+			echo "Packages live but absent from rpm-ostree deployments; re-staging all packages"
+			for package in $PACKAGES; do
+				rpm_path=$(find /usr/share/rpm-ostree/extensions/ -maxdepth 1 -type f -name "${package}-*.rpm" 2>/dev/null | head -n1)
+				[[ -n "$rpm_path" ]] && install_rpms+=("$rpm_path")
+			done
+		fi
+	fi
+
+	# If nothing to install or re-stage, exit early
 	if [[ ${#install_rpms[@]} -eq 0 ]]; then
 		set_status_installed
 	else
@@ -140,8 +139,8 @@ install_kata() {
 			cp "$rpm_path" /host/tmp/extensions/
 		done
 
-		# Build install command
-		install_cmd="rpm-ostree install"
+		# Build install command with --apply-live to avoid a node reboot
+		install_cmd="rpm-ostree install --apply-live"
 		for pkg in "${uninstall_list[@]}"; do
 			install_cmd+=" --uninstall=$pkg"
 		done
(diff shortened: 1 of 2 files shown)


c3d:
Hi @paulczar, thanks for the PR. Were you aware of https://github.com/openshift/sandboxed-containers-operator/pull/1349/commits. Can you take a look and contrast the two approaches? I have to choose one 😄

paulczar:
thanks @c3d !  I did see that other PR and I updated this PR description to address the differences.   I think the main thing is the explicit support of HCP in my method.

c3d:
Thanks @paulczar. Based on my own review and testing, which is detailed in https://redhat.atlassian.net/browse/KATA-5292, I am planning to pick up your PR.

Any reason you kept it as "draft"?

c3d:
@gkurz I would appreciate your review on this one if you can find the time.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -157,19 +119,51 @@ install_kata() {
...
+		# system. The kata-containers %post script normally runs kata-osbuilder
+		# to build a host-kernel-derived kata.kernel/kata.initrd at
+		# /var/cache/kata-containers/osbuilder-images/. We skip that step
+		# intentionally: the kata guest kernel runs inside a VM and is fully
+		# independent of the host kernel, so a host-kernel-specific build is
+		# unnecessary. The RPM already ships pre-built CC images in /usr/share/
+		# that are tested against this kata-containers release. We symlink them
+		# into the path that configuration.toml expects. The symlinks are updated
+		# automatically whenever kata-containers is re-installed at a new version.
+		local kata_cache=/host/var/cache/kata-containers/osbuilder-images
```
c3d: @gkurz what do you think about this approach?

On one hand, it makes sense to me to use a pre-built and validated kernel, notably for the CC case. 

On the other hand, this breaks the "contracts" we had so far to run the same kernel as the host.

c3d:
@paulczar In the case of HCP on Azure, I see permission problems creating node labels.

1. The operator creates a DaemonSet but uses the wrong ServiceAccount (`default`, should probably be `kata-install`)
2. It also doesn't create RBAC for `kata-install`. The `kata-install` container needs to label nodes but has no permissions
3. The operator doesn't ensure `SCC` is usable. Even though we created a `kata-install-scc`, it's not bound to the ServiceAccount, so it's not usable as is.

So in my opinion, we need the following 
1. DaemonSet creation should specify serviceAccountName: kata-install
2. Operator should create:
  - ServiceAccount kata-install
  - ClusterRole with node get/list/patch/update permissions
  - ClusterRoleBinding linking them
  - SCC with privileged access
  - SCC binding to the ServiceAccount

I'm going to submit a commit for that.

scripts/kata-install/osc-kata-install.sh
gkurz: I understand the "contract" as a clever way to get security updates and bug fixes from RHCOS for free :wink:

For example, we don't have anything in place to bump the RHEL10 kernel we're shipping in the RPM on a regular basis.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -157,19 +119,51 @@ install_kata() {
...
+
+		# rpm-ostree --apply-live does not run RPM %post scripts on the live
+		# system. The kata-containers %post script normally runs kata-osbuilder
+		# to build a host-kernel-derived kata.kernel/kata.initrd at
+		# /var/cache/kata-containers/osbuilder-images/. We skip that step
+		# intentionally: the kata guest kernel runs inside a VM and is fully
+		# independent of the host kernel, so a host-kernel-specific build is
+		# unnecessary. The RPM already ships pre-built CC images in /usr/share/
+		# that are tested against this kata-containers release. We symlink them
+		# into the path that configuration.toml expects. The symlinks are updated
```
gkurz: Yikes, please don't do that. `kata-cc.initrd` is really tailored for coco. If we're to go with a fixed kernel, I'd rather ship a `kata.initrd` in the RPM.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -157,19 +119,51 @@ install_kata() {
...
-		wait_for_reboot_clear
+		# rpm-ostree --apply-live updates /usr immediately but /etc changes from
+		# RPMs only land after reboot (OSTree three-way merge). Copy the kata
+		# CRI-O drop-ins from the RPM's factory-defaults location now so CRI-O
+		# can see the runtime handlers without a reboot.
+		for conf in /host/usr/etc/crio/crio.conf.d/50-kata*; do
+			[ -f "$conf" ] && cp "$conf" /host/etc/crio/crio.conf.d/
+		done
+
+		# rpm-ostree --apply-live does not run RPM %post scripts on the live
```
gkurz: Does `rpm-ostree install --apply-live` honor `%posttrans` ?

```
%posttrans
# Needs to be here to ensure it happens last during updates
semodule -i %{udica_templates_dir}/base_container.cil \
            %{udica_templates_dir}/net_container.cil \
            %{katadefaults}/osc_monitor.cil
# Grant access to /dev/sev (and other devices) in guest
setsebool -P container_use_devices 1
```

This is mandatory.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -157,19 +119,51 @@ install_kata() {
...
-	wait_for_reboot_clear
-
-	# Check if kata-containers is installed
-	# If kata-containers is not installed we are done
-	# Create uninstalled file to signal readiness
-	# Sleep infinity to prevent pod restart (DaemonSets always restart exited pods)
+	# If kata-containers is already uninstalled, mark the node and sleep to
+	# prevent DaemonSet pod restarts. Otherwise, stage removal via rpm-ostree
+	# uninstall (takes effect at next reboot) and immediately clean up CRI-O
+	# config so kata becomes unavailable without waiting for that reboot.
```
gkurz, line 199: And now the $1M question : what happens if the user finally decides to redeploy kata ? Will `rpm-ostree install --apply-live` be happy with the pending staged removal ?

gkurz (commented):
Hi Paul ! Thanks for reviving this. I've dropped some comments.

scripts/kata-install/osc-kata-install.sh
c3d: So that means we would need to at least install the osbuilder service. Not necessarily run the osbuilder script before the next reboot, but get the update at the next reboot.

paulczar:
@c3d no specific reason I've kept it draft, just assumed there would be a bunch of back and forth over it and it might take some time.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -157,19 +119,51 @@ install_kata() {
 		# Clean up temp dir
 		rm -rf /host/tmp/extensions/
 
-		# Wait again: rpm-ostree install stages changes, requiring a reboot
-		wait_for_reboot_clear
+		# rpm-ostree --apply-live updates /usr immediately but /etc changes from
```
gkurz:
```suggestion
		# rpm-ostree install --apply-live updates /usr immediately but /etc changes from
```

c3d:
Hi @paulczar I added a few commits on top of your PR in #2325. Please take a look and tell me what you think. @gkurz, your review would be appreciated as well.

paulczar:
Updated the PR to address gkurz's review feedback. Here's what changed and the test results:

### Changes made

**kata-cc.initrd symlink (gkurz's comment 2)** — removed entirely. The old approach of symlinking \`kata-cc.*\` CC images was wrong for non-CC kata. Replaced with:
- `chroot /host systemctl start kata-osbuilder-generate.service` — the correct mechanism for building host-kernel-derived VM images on RHCOS; the \`%post\` script installs this service specifically because \`/var/cache/kata-containers\` isn't writable during \`%post\` on RHCOS, so the service defers the build. Starting it directly gives us the right images immediately.

**%posttrans SELinux steps (gkurz's comment 3)** — now run manually after \`--apply-live\`:
```bash
chroot /host semodule -i \
    /usr/share/udica/templates/base_container.cil \
    /usr/share/udica/templates/net_container.cil \
    /usr/share/kata-containers/defaults/osc_monitor.cil
chroot /host setsebool -P container_use_devices 1
```
And on uninstall (\`%preun\` equivalent):
```bash
chroot /host semodule -r osc_monitor
chroot /host setsebool -P container_use_devices 0
```
Paths verified by inspecting the RPM scriptlets directly on a UBI9 pod in the cluster.

**Staged removal detection (gkurz's comment 4)** — added detection and cancellation of pending staged removals at the top of \`install_kata()\`. If a prior uninstall staged kata removal (rpm-ostree uninstall doesn't support --apply-live), the booted deployment still shows kata as installed — so \`rpm -q\` would skip reinstalling and the staged removal would silently drop kata at the next reboot. We now cancel the pending deployment with \`rpm-ostree cleanup -p\` before proceeding.

### Test results (bare-metal cluster, c5n.metal nodes, HCP)

**Install:**
- Both nodes reached \`installed\` state; no node reboots
- \`kata-osbuilder-generate.service\` completed successfully; host-kernel images present in \`/var/cache/kata-containers/osbuilder-images/\`
- SELinux: \`osc_monitor\` module loaded, \`container_use_devices --> on\`
- Kata pod ran successfully: \`1/1 Running\`, QEMU confirmed \`-machine q35,accel=kvm\`, guest kernel matched host (\`5.14.0-570.80.1.el9_6.x86_64\`)

**Uninstall:**
- KataConfig deleted → both nodes transitioned \`installed\` → \`waiting_to_uninstall\` → \`uninstalling\` → \`uninstalled\` in ~33 seconds total, no reboots
- KataConfig deletion completed cleanly; all kata labels removed from nodes

scripts/kata-install/osc-kata-install.sh
paulczar: Good question! There are two defences against this:

1. The staged-removal detection at lines 61–71 is meant to catch exactly this case and cancel the pending deployment with `rpm-ostree cleanup -p` before reinstalling. CodeRabbit correctly flagged (line 71 thread) that the detection was only checking `requested-packages` and missing `requested-local-packages` — local RPM installs land in the latter — so `n_kata_booted` was always 0 and the branch never fired. Fixed in 20dc1e87.

2. Even with that bug present, the re-stage check (lines 111–125) detects kata absent from all deployments and forces a full `rpm-ostree install --apply-live`. That call creates a new staged deployment which supersedes the pending removal, so the reinstall works correctly regardless.

With 20dc1e87 both paths are now correct: the staged-removal detection fires and cancels the old deployment first, and the re-stage check handles any remaining edge cases.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -157,19 +135,47 @@ install_kata() {
...
+		chroot /host setsebool -P container_use_devices 1
+
+		# Restart CRI-O to pick up the new runtime configuration. A reload
+		# (SIGHUP) only re-reads the config struct; it does not rebuild the
+		# internal OCI handler map, so kata would not appear as a valid runtime
+		# until the next full CRI-O start. Existing pod processes survive the
+		# restart — only new CRI operations are briefly unavailable.
+		chroot /host systemctl restart crio
+
+		set_status_installed
```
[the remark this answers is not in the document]
paulczar: The script runs under `set -xeuo pipefail` (line 9), so any non-zero exit from a `chroot` command causes the script to exit immediately with a failure status. The pod enters CrashLoopBackOff and the controller retries — this is the standard DaemonSet failure model and already gives the "node stays unlabeled while broken" behaviour you are looking for. Adding per-command `|| { log; exit 1; }` wrappers would be redundant given `set -e` is already in effect.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -84,13 +49,26 @@ set_status_uninstalled() {
...
+		s=\$(rpm-ostree status --json)
+		n_staged=\$(echo \"\$s\" | jq '[.deployments[] | select(.staged)] | length')
+		[[ \$n_staged -eq 0 ]] && exit 1
+		n_kata_staged=\$(echo \"\$s\" | jq '[.deployments[] | select(.staged) | (.\"requested-packages\" // [])[] | select(startswith(\"kata-containers\"))] | length')
+		n_kata_booted=\$(echo \"\$s\" | jq '[.deployments[] | select(.booted) | (.\"requested-packages\" // [])[] | select(startswith(\"kata-containers\"))] | length')
+		[[ \$n_kata_booted -gt 0 && \$n_kata_staged -eq 0 ]]
+	"; then
+		echo "Staged removal of kata-containers detected; cancelling pending deployment"
+		chroot /host rpm-ostree cleanup -p
+	fi
```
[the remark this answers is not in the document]
paulczar: Fixed in 20dc1e87. Both `n_kata_staged` and `n_kata_booted` now merge `requested-packages` and `requested-local-packages` before filtering, matching the pattern used in the re-stage check at lines 111–125.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -180,11 +207,47 @@ uninstall_kata() {
...
+	local ostree_pkg_list
+	ostree_pkg_list=$(chroot /host bash -c '
+		rpm-ostree status --json | jq -r "[.deployments[] |
+		  ((.\"requested-packages\" // []) + (.\"requested-local-packages\" // []))[]]
+		| .[]" 2>/dev/null
+	' || true)
+	local uninstall_args=()
+	for pkg in $PACKAGES; do
+		echo "$ostree_pkg_list" | grep -qF "$pkg" && uninstall_args+=("$pkg")
+	done
```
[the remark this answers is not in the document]
paulczar: Fixed in 20dc1e87: changed to `grep -xqF` for exact whole-line matching.

scripts/kata-install/osc-kata-install.sh
paulczar: The symlink/pre-built-kernel approach was abandoned in favour of starting `kata-osbuilder-generate.service` directly after install. This is what the kata-containers RPM installs to build host-kernel-derived images on RHCOS — the `%post` scriptlet detects RHCOS, skips the in-place image build, and relies on the service to run it after boot. We start the service explicitly so the images are ready immediately without a reboot. Agreed that having the operator pre-build images would be cleaner; that's a good candidate for a follow-up once the no-reboot baseline lands.

scripts/kata-install/osc-kata-install.sh
paulczar: The `kata-cc.initrd` symlink approach was dropped entirely. The current code uses `chroot /host systemctl start kata-osbuilder-generate.service`, which is the correct mechanism on RHCOS — it builds host-kernel-derived VM images the same way the RPM `%post` scriptlet would have (the scriptlet skips this on RHCOS and installs the service to do it post-boot instead). Starting the service directly means the images are ready immediately without waiting for a reboot.

scripts/kata-install/osc-kata-install.sh
paulczar: Correct — `--apply-live` does **not** run RPM scriptlets on the live system (only on the staged deployment, which takes effect at next reboot). This is exactly why the script now manually replicates the relevant `%post`/`%posttrans` steps after every install:

- `systemctl start kata-osbuilder-generate.service` — builds host-kernel-derived VM images
- `semodule -i` — loads the kata SELinux policy modules (`%posttrans`)
- `setsebool -P container_use_devices 1` — grants container device access (`%posttrans`)
- `systemctl restart crio` — picks up the new runtime configuration

And equivalently on uninstall, `semodule -r osc_monitor` and `setsebool -P container_use_devices 0` replicate what `%preun` would have done. The comments in the script now call this out explicitly.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -180,11 +186,27 @@ uninstall_kata() {
...
-	# Wait again: rpm-ostree uninstall stages changes, requiring a reboot
-	wait_for_reboot_clear
+	# Remove the CRI-O drop-ins we copied during install
+	rm -f /host/etc/crio/crio.conf.d/50-kata*
+
+	# Run the steps from kata-containers %preun manually (rpm-ostree uninstall
+	# does not run scriptlets on the live system): remove the kata SELinux policy
+	# module and revoke device access granted during install.
+	chroot /host semodule -r osc_monitor
+	chroot /host setsebool -P container_use_devices 0
```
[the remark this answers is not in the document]
paulczar: This behaviour mirrors what the kata-containers RPM scriptlets did before `--apply-live` was adopted — `%posttrans` sets `container_use_devices on` and `%preun` reverts it. Since `--apply-live` skips scriptlets on the live system, the script replicates those steps manually to keep parity. The concern about unconditional toggling is valid in principle, but it is pre-existing RPM behaviour, not something introduced by this PR. A follow-up to make it conditional could be tracked separately if desired.

scripts/kata-install/osc-kata-install.sh
paulczar: Suggestion accepted and applied in `009107cb`. The comment now reads "updates /usr immediately but /etc changes from RPMs only land after reboot".

c3d:
Closing as superceded by #2410

PR #2351, opened 2026-06-23
Drop OCP version check

gkurz:
This was introduced to mitigate the lack of support of BM CoCo in older OCP version. It is suboptimal to hardcode OCP versions. It generates extra work, especially painful at release time.

Since we now support BM CoCo in all OCP versions, just drop this code.


littlejawa (approved):
/lgtm

PR #2366, opened 2026-06-23
tests: update to kata-containers 3.31.0-1

dbkreling:
Kata-containers version 3.31.0-1 RPMs have been built.

OSC 1.13 and kata-containers 3.31 will not support OSC 4.18 and will support
OSC 4.22.

Notes:
- The RPM packages for 4.19, 4.20 and 4.21 are "3.31.0-1.rhaos4.19"
- The RPM package for 4.22 is "3.31.0-1.rhaos4.22"

### DIFF
--- a/tests/osc-test-fbc-integration.yaml
+++ b/tests/osc-test-fbc-integration.yaml
@@ -152,7 +152,7 @@ spec:
         - name: VARIANT
           value: downstream-candidate420
         - name: ENVS
-          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE),FORCE_SUCCESS_EXIT=no,INSTALL_KATA_RPM=true,KATA_RPM_VERSION=3.25.0-7.rhaos4.20.el9"
+          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE),FORCE_SUCCESS_EXIT=no,INSTALL_KATA_RPM=true,KATA_RPM_VERSION=3.31.0-1.rhaos4.19.el9"
       matrix:
         params:
           - name: PROWJOB_NAME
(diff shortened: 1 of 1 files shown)


wainersm:
Hi @dbkreling ,

Thanks for the notes on the description, otherwise I would certainly ask you why of that change :)

The change is not wrong but there is one detail that you likely to have missed. We used to run all the jobs in parallel in single shot, however, because we are sharing the Azure profile with the entire openshift org and the amount of clusters created in azure is limited. In some occasions we hit the limit because we were running too many azure jobs in parallel. As a result, we changed the pipeline to run in steps to consume the minimal amount of azure each and at same time minimizing the time to run all jobs.

Currently we got:

sanity -> 4.21, 4.20 (2 azure) -> 4.19 (2 azure) -> 4.18 (2 azure)

With your change:

sanity -> 4.22, 4.21, 4.20 (5 azure) -> 4.19 (2 azure)

That said, we need to re-balance the jobs :D I can send a follow up PR for that, or if you prefer you can try it in this one. What do you say?

dbkreling:
> That said, we need to re-balance the jobs :D I can send a follow up PR for that, or if you prefer you can try it in this one. What do you say?

I'd be happy to change that, but if you do, chances are we will have more efficiency, because I'm not familiar with that. I did not know about the balancing of the jobs, so feel free to change them accordingly.

Thanks @wainersm

wainersm:
> > That said, we need to re-balance the jobs :D I can send a follow up PR for that, or if you prefer you can try it in this one. What do you say?
> 
> I'd be happy to change that, but if you do, chances are we will have more efficiency, because I'm not familiar with that. I did not know about the balancing of the jobs, so feel free to change them accordingly.
> 
> Thanks @wainersm

That works for me @dbkreling . Let's merge this one then I send a follow up.

wainersm (approved):

tests/osc-test-fbc-integration.yaml
```diff
@@ -184,7 +184,7 @@ spec:
         - name: VARIANT
           value: downstream-candidate418
         - name: ENVS
-          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE),FORCE_SUCCESS_EXIT=no,INSTALL_KATA_RPM=true,KATA_RPM_VERSION=3.25.0-6.rhaos4.18.el9"
+          value: "CATALOG_SOURCE_IMAGE=$(tasks.get-catalog-image.results.CATALOG_IMAGE),FORCE_SUCCESS_EXIT=no,INSTALL_KATA_RPM=true,KATA_RPM_VERSION=3.31.0-1.rhaos4.22.el9"
```
gkurz, line 187: Ouch this is very wrong... this RPM relies on a feature that comes with CRI-O 1.35. Fortunately it is fixed by the next commit that removes 4.18.

wainersm: Hi @gkurz , that change could be in the next commit. If you look at the final result, only the 4.22 pipelines will have the 4.22 rpm installed.

PR #2378, opened 2026-06-23
tests: rebalance the jobs execution in osc-test-fbc-integration.yaml

wainersm:
It is required to rebalance the execution of the jobs to avoid hitting the limit to use Azure. Before this change:
sanity -> 4.22,4.41,4.20 (5 azure) -> 4.19 (2 azure)

After:
sanity -> 4.22 (2 zure) -> 4.21,4.20 (3 azure) -> 4.19 (2 azure)


dbkreling (approved):
LGTM

PR #2410, opened 2026-07-01
Implement DaemonSet installation without reboot

c3d:
This PR builds on the latest iteration of #2312 and integrates some of the changes in #2325 + comments there.

Supersedes #1349, #2312 and #2325

### DIFF
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -408,6 +408,8 @@ spec:
                 env:
                 - name: PEERPODS_NAMESPACE
                   value: openshift-sandboxed-containers-operator
+                - name: RELATED_IMAGE_KATA_DAEMONSET
+                  value: quay.io/openshift_sandboxed_containers/osc-daemonset:1.13.1-apply-live
                 - name: RELATED_IMAGE_KATA_MONITOR
                   value: registry.redhat.io/openshift-sandboxed-containers/osc-monitor-rhel9@sha256:2a69854ef50aec3cb1e7ee1e7de9bc698c6c2f52fd87c57e57c5a4c3198904a5
                 - name: RELATED_IMAGE_CAA
@@ -579,6 +581,8 @@ spec:
   provider:
     name: Red Hat
   relatedImages:
+  - image: quay.io/openshift_sandboxed_containers/osc-daemonset:1.13.1-apply-live
+    name: kata-daemonset
   - image: registry.redhat.io/openshift-sandboxed-containers/osc-monitor-rhel9@sha256:2a69854ef50aec3cb1e7ee1e7de9bc698c6c2f52fd87c57e57c5a4c3198904a5
     name: kata-monitor
   - image: registry.redhat.io/openshift-sandboxed-containers/osc-cloud-api-adaptor-rhel9@sha256:07c42ec5b059d3af7face3bd7306e7a71322d6f00240be82285bb303f30c13b1
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -195,8 +195,12 @@ spec:
           - serviceaccounts
           verbs:
           - create
+          - delete
           - get
+          - list
+          - patch
           - update
+          - watch
         - apiGroups:
           - rbac.authorization.k8s.io
           resources:
@@ -204,8 +208,12 @@ spec:
           - clusterrolebindings
           verbs:
           - create
+          - delete
           - get
+          - list
+          - patch
           - update
+          - watch
         - apiGroups:
           - ""
           - machineconfiguration.openshift.io
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -167,6 +167,7 @@ spec:
           - ""
           resources:
           - secrets
+          - serviceaccounts
           verbs:
           - create
           - delete
@@ -189,31 +190,6 @@ spec:
           - get
           - patch
           - update
-        - apiGroups:
-          - ""
-          resources:
-          - serviceaccounts
-          verbs:
-          - create
-          - delete
-          - get
-          - list
-          - patch
-          - update
-          - watch
-        - apiGroups:
-          - rbac.authorization.k8s.io
-          resources:
-          - clusterroles
-          - clusterrolebindings
-          verbs:
-          - create
-          - delete
-          - get
-          - list
-          - patch
-          - update
-          - watch
         - apiGroups:
           - ""
           - machineconfiguration.openshift.io
@@ -354,6 +330,19 @@ spec:
           - node.k8s.io
           resources:
           - runtimeclasses
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
+        - apiGroups:
+          - rbac.authorization.k8s.io
+          resources:
+          - clusterrolebindings
+          - clusterroles
           verbs:
           - create
           - delete
--- a/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
@@ -194,8 +194,18 @@ spec:
           resources:
           - serviceaccounts
           verbs:
+          - create
           - get
-          - list
+          - update
+        - apiGroups:
+          - rbac.authorization.k8s.io
+          resources:
+          - clusterroles
+          - clusterrolebindings
+          verbs:
+          - create
+          - get
+          - update
         - apiGroups:
           - ""
           - machineconfiguration.openshift.io
--- a/config/peerpods/podvm/ami-helper.sh
+++ b/config/peerpods/podvm/ami-helper.sh
@@ -207,11 +207,34 @@ delete_vmimport_role() {
 
 create_bucket() {
 	echo "Create s3 Bucket named ${BUCKET_NAME} at ${REGION}"
-	if [[ ${REGION} == us-east-1 ]]; then
-		aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION}
-	else
-		aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION} --create-bucket-configuration LocationConstraint=${REGION}
+
+	# Check if bucket already exists
+	if aws s3api head-bucket --bucket ${BUCKET_NAME} --region ${REGION} 2>/dev/null; then
+		echo "Bucket ${BUCKET_NAME} already exists and is owned by this account, adopting it"
+		return 0
 	fi
+
+	# Bucket doesn't exist, create it
+	if [[ ${REGION} == us-east-1 ]]; then
+		if ! aws s3api create-bucket --bucket ${BUCKET_NAME} --region ${REGION} 2>&1 | tee /tmp/bucket_create.log; then
+			# Check if error is "BucketAlreadyOwnedByYou"
+			if grep -q "BucketAlreadyOwnedByYou" /tmp/bucket_create.log; then
+				echo "Bucket ${BUCKET_NAME} already exists and is owned by this account"
+				return 0
+			fi
+			return 1
+		fi
+	else
+		if ! aws s3api create-bucket --bucket ${BUCKET_NAME} --region ${REGION} --create-bucket-configuration LocationConstraint=${REGION} 2>&1 | tee /tmp/bucket_create.log; then
+			# Check if error is "BucketAlreadyOwnedByYou"
+			if grep -q "BucketAlreadyOwnedByYou" /tmp/bucket_create.log; then
+				echo "Bucket ${BUCKET_NAME} already exists and is owned by this account"
+				return 0
+			fi
+			return 1
+		fi
+	fi
+	echo "Successfully created bucket ${BUCKET_NAME}"
 }
 
 set_service_role() {
--- a/config/peerpods/podvm/ami-helper.sh
+++ b/config/peerpods/podvm/ami-helper.sh
@@ -215,22 +215,25 @@ create_bucket() {
 	fi
 
 	# Bucket doesn't exist, create it
+	local output
 	if [[ ${REGION} == us-east-1 ]]; then
-		if ! aws s3api create-bucket --bucket ${BUCKET_NAME} --region ${REGION} 2>&1 | tee /tmp/bucket_create.log; then
+		if ! output=$(aws s3api create-bucket --bucket "${BUCKET_NAME}" --region "${REGION}" 2>&1); then
 			# Check if error is "BucketAlreadyOwnedByYou"
-			if grep -q "BucketAlreadyOwnedByYou" /tmp/bucket_create.log; then
+			if echo "$output" | grep -q "BucketAlreadyOwnedByYou"; then
 				echo "Bucket ${BUCKET_NAME} already exists and is owned by this account"
 				return 0
 			fi
+			echo "Failed to create bucket: $output" >&2
 			return 1
 		fi
 	else
-		if ! aws s3api create-bucket --bucket ${BUCKET_NAME} --region ${REGION} --create-bucket-configuration LocationConstraint=${REGION} 2>&1 | tee /tmp/bucket_create.log; then
+		if ! output=$(aws s3api create-bucket --bucket "${BUCKET_NAME}" --region "${REGION}" --create-bucket-configuration LocationConstraint="${REGION}" 2>&1); then
 			# Check if error is "BucketAlreadyOwnedByYou"
-			if grep -q "BucketAlreadyOwnedByYou" /tmp/bucket_create.log; then
+			if echo "$output" | grep -q "BucketAlreadyOwnedByYou"; then
 				echo "Bucket ${BUCKET_NAME} already exists and is owned by this account"
 				return 0
 			fi
+			echo "Failed to create bucket: $output" >&2
 			return 1
 		fi
 	fi
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -57,8 +57,12 @@ rules:
   - serviceaccounts
   verbs:
   - create
+  - delete
   - get
+  - list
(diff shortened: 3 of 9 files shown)


gkurz:
Hi Christophe ! I have to // the review of this PR with other activities. It might give a drive-by impression :wink:

First, I'll just comment that you're reposting someone else's change (commit 1). Please add your S-o-b and mention any change you might have made in this commit (or nothing if taken as is).

scripts/kata-install/osc-kata-install.sh
```diff
@@ -157,19 +156,47 @@ install_kata() {
...
+		#    boot. Start that service now so the images are ready immediately.
+		chroot /host systemctl start kata-osbuilder-generate.service
+
+		# 2. Run the %posttrans SELinux steps: load the kata SELinux policy
+		#    modules and grant containers access to /dev/sev and similar devices.
+		chroot /host semodule -i \
+			/usr/share/udica/templates/base_container.cil \
+			/usr/share/udica/templates/net_container.cil \
+			/usr/share/kata-containers/defaults/osc_monitor.cil
+		chroot /host setsebool -P container_use_devices 1
```
gkurz: It is a bit unfortunate to duplicate the scriplets. What about extracting them from the kata-containers RPM directly ? Something like :
```
[core@greg2-worker-0 ~]$ rpm -q --scripts kata-containers  | awk '/^postinstall scriptlet/{found=1;next} /^[a-z]+ scriptlet/{found=0} found'

 
if [ $1 -eq 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then 
    # Initial installation 
    /usr/lib/systemd/systemd-update-helper install-system-units kata-osbuilder-generate.service || : 
fi 

# Skip running this on Fedora CoreOS / Red Hat CoreOS
if test -w /var/cache/kata-containers; then
    TMPOUT="$(mktemp -t kata-rpm-post-XXXXXX.log)"
    echo "Creating kata appliance initrd..."
    /usr/libexec/kata-containers/osbuilder/kata-osbuilder.sh > ${TMPOUT} 2>&1
    if test "$?" != "0" ; then
        echo "Building failed. Here is the log details:"
        cat ${TMPOUT}
        exit 1
    fi
fi
[core@greg2-worker-0 ~]$ rpm -q --scripts kata-containers | awk '/^posttrans scriptlet/{found=1;next} /^[a-z]+ scriptlet/{found=0} found'
# Needs to be here to ensure it happens last during updates
semodule -i /usr/share/udica/templates/base_container.cil \
            /usr/share/udica/templates/net_container.cil \
            /usr/share/kata-containers/defaults/osc_monitor.cil
# Grant access to /dev/sev (and other devices) in guest
setsebool -P container_use_devices 1
```

scripts/kata-install/osc-kata-install.sh
```diff
@@ -157,19 +156,47 @@ install_kata() {
 		# Clean up temp dir
 		rm -rf /host/tmp/extensions/
 
-		# Wait again: rpm-ostree install stages changes, requiring a reboot
-		wait_for_reboot_clear
+		# rpm-ostree --apply-live updates /usr immediately but /etc changes from
+		# RPMs only land after reboot (OSTree three-way merge). Copy the kata
+		# CRI-O drop-ins from the RPM's factory-defaults location now so CRI-O
+		# can see the runtime handlers without a reboot.
+		for conf in /host/usr/etc/crio/crio.conf.d/50-kata*; do
```
gkurz: For precision, I'd prefer to get the list of  CRI-O drop-ins from the RPM itself :
```
[core@greg2-worker-0 ~]$ rpm -ql kata-containers | grep /etc/crio/crio.conf.d
/etc/crio/crio.conf.d/50-kata
/etc/crio/crio.conf.d/50-kata-nvidia-gpu
/etc/crio/crio.conf.d/50-kata-remote
/etc/crio/crio.conf.d/50-kata-se
/etc/crio/crio.conf.d/50-kata-snp
/etc/crio/crio.conf.d/50-kata-snp-nvidia-gpu
/etc/crio/crio.conf.d/50-kata-tdx
/etc/crio/crio.conf.d/50-kata-tdx-nvidia-gpu
```

scripts/kata-install/osc-kata-install.sh
```diff
@@ -180,11 +207,47 @@ uninstall_kata() {
...
+		echo "$ostree_pkg_list" | grep -qF "$pkg" && uninstall_args+=("$pkg")
+	done
+	if [[ ${#uninstall_args[@]} -gt 0 ]]; then
+		chroot /host /bin/bash -c "rpm-ostree uninstall ${uninstall_args[*]}"
+	else
+		echo "No kata packages tracked in rpm-ostree deployments; skipping rpm-ostree uninstall (packages will be absent after next reboot)"
+	fi
+
+	# Remove the CRI-O drop-ins we copied during install
+	rm -f /host/etc/crio/crio.conf.d/50-kata*
```
gkurz: Same as https://github.com/openshift/sandboxed-containers-operator/pull/2410/changes/7ad49cdfdd5dbfca52ce65d2f05baedf385804bd#r3507080935

controllers/daemonset_reconcile.go
```diff
@@ -230,6 +328,12 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequestDaemonSet
 		}
 	}
 
+	// Ensure RBAC resources exist for kata-install DaemonSet
+	if err := r.ensureKataInstallRBAC(); err != nil {
+		r.Log.Error(err, "Failed to ensure kata-install RBAC resources")
+		return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+	}
+
```
gkurz, line 355: I guess we need the same on the uninstall path for the sake of idempotency.

controllers/daemonset_reconcile.go
```diff
@@ -435,6 +435,16 @@ func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDaemo
...
 	}
 
+	// Check feature gates for custom DaemonSet image override
+	dsImage := daemonSetImage
+	fgStatus, err := r.NewFeatureGateStatus()
+	if err != nil {
+		r.Log.Error(err, "couldn't get feature gate status, using default DaemonSet image")
+	} else if fgStatus.DaemonSetImage != "" {
+		dsImage = fgStatus.DaemonSetImage
+		r.Log.Info("Using custom DaemonSet image from feature gates", "image", dsImage)
```
gkurz: I guess this should help.

gkurz (approved):
I've dropped some comments and suggestions. They don't call for a behavior change so I'll approve this PR. Feel free to process them or to keep that for the future (maybe add some TODO comments so that we don't forget ?)

Just make sure to add you S-o-b tags to each commit and make sure all commit titles follow a `foo: bla bla bla` format to silent `commit-msg-check` in CI.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -154,22 +153,48 @@ install_kata() {
...
+		# rpm-ostree --apply-live does not run RPM scriptlets
+		# Run the scriptlets manually in host context
+		# Normally this includes:
+		# 1. postinstall: Building host-kernel derived kata VM image
+		# 2. posttrans:	  Running the SELinux steps
+		for phase in postinstall posttrans; do
+		    for rpm_path in "${install_rpms[@]}"; do
+			scriptlet="$(rpm -qp --scripts $rpm_path | awk '/^'$phase' scriptlet/{found=1;next} /^[a-z]+ scriptlet/{found=0} found')"
+			chroot /host bash -c "$scriptlet"
+		    done
```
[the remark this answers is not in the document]
gkurz: `$1` meaning :
```
1                     First install (no previous version existed)
2 (or higher)         Upgrade (a previous version was already installed)
0                     Used in %preun/%postun — package is being fully removed
```
I guess we should pass 1 here and 0 on the uninstall path.

scripts/kata-install/osc-kata-install.sh
```diff
@@ -48,6 +48,15 @@ set_status_uninstalled() {
 	label_node "uninstalled"
 }
 
+extract_scriptlet() {
+    # Extract the script from the given phase in the RPM
+    # We substitute '$1' with '1' to pretend we are initial installation
```
gkurz: Is it appropriate for the `preuninstall` case ?

gkurz:
> Just make sure to add you S-o-b tags to each commit and make sure all commit titles follow a `foo: bla bla bla` format to silent `commit-msg-check` in CI.

`foo` should rather be an arbitrary subsystem name, not `fix` or `feat` as mentioned at :

https://github.com/openshift/sandboxed-containers-operator/blob/devel/docs/CONTRIBUTING.md#format

controllers/daemonset_reconcile.go
```diff
@@ -230,6 +328,12 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigInstallRequestDaemonSet
 		}
 	}
 
+	// Ensure RBAC resources exist for kata-install DaemonSet
+	if err := r.ensureKataInstallRBAC(); err != nil {
+		r.Log.Error(err, "Failed to ensure kata-install RBAC resources")
+		return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
```
gkurz: I used to be able to try out daemonset deployment mode in a regular non-HCP setup... this now prevents it because `ensureKataInstallRBAC()` fails...

Could we fall back on using the default SA when this happens ? Or could we have an environment variable based knob to fully skip this RBAC setup ?

controllers/fg_handler.go
```diff
@@ -14,18 +14,21 @@ const (
 	ConfidentialFeatureGate = "confidential"
 	LayeredImageDeployment  = "layeredImageDeployment"
 	DeploymentModeConfig    = "deploymentMode"
+	DaemonSetImageOverride  = "daemonSetImage"
```
gkurz: This setting isn't a feature gate. It is specific to the daemonset deployment mode. It should go to a separate `osc-daemonset-deployment-mode-cm` CM.

teodoroneata:
https://redhat.atlassian.net/browse/KATA-5433

scripts/kata-install/osc-kata-install.sh
```diff
@@ -160,6 +160,7 @@ install_kata() {
 		# RPMs only land after reboot (OSTree three-way merge). Copy the kata
 		# CRI-O drop-ins from the RPM's factory-defaults location now so CRI-O
 		# can see the runtime handlers without a reboot.
+		mkdir -p /host/etc/crio/crio.conf.d
```
snir911: can it happen? will it have the write permissions to allow the crio service to read from there?

scripts/kata-install/osc-kata-install.sh
```diff
@@ -169,9 +169,9 @@ install_kata() {
 		# This is notably necessary for CRI-O drop-ins so that the CRI-O
 		# restart below is effective
 		for rpm_path in "${install_rpms[@]}"; do
-			mapfile -t etc_files < <(rpm -qpl "$rpm_path" | grep "^/etc/")
```
snir911: i'd squash all the scriplets fixing commits

config/rbac/role.yaml
```diff
@@ -57,17 +57,25 @@ rules:
   - serviceaccounts
   verbs:
   - create
+  - delete
```
snir911: worth squashing the rbac commits also maybe

controllers/daemonset_reconcile.go
```diff
@@ -453,14 +454,21 @@ func (r *KataConfigOpenShiftReconciler) daemonSetForKataInstall(action KataDaemo
 		return nil, err
 	}
 
-	// Check DaemonSet config for custom image override
+	// Resolve DaemonSet image with 3-tier precedence:
+	// 1. ConfigMap override (osc-daemonset-config)
+	// 2. RELATED_IMAGE_KATA_DAEMONSET environment variable
```
snir911, line 459: can't we keep only the RELATED_IMAGE_KATA_DAEMONSET option? and if user needs he can override it directly, like we do with the other related images

config/peerpods/podvm/ami-helper.sh
```diff
@@ -207,11 +207,34 @@ delete_vmimport_role() {
...
+		echo "Bucket ${BUCKET_NAME} already exists and is owned by this account, adopting it"
+		return 0
+	fi
+
+	# Bucket doesn't exist, create it
 	if [[ ${REGION} == us-east-1 ]]; then
-		aws s3api create-bucket  --bucket ${BUCKET_NAME} --region ${REGION}
+		if ! aws s3api create-bucket --bucket ${BUCKET_NAME} --region ${REGION} 2>&1 | tee /tmp/bucket_create.log; then
+			# Check if error is "BucketAlreadyOwnedByYou"
+			if grep -q "BucketAlreadyOwnedByYou" /tmp/bucket_create.log; then
```
snir911: For my understanding this is not needed as if it's exist and owned by this account the first condition will return.
(i created similar pr as i was not aware to this: https://github.com/openshift/sandboxed-containers-operator/pull/2413/changes/3f0a201d67c1aa906a76c3c7c8520d21f7b0102e )

BTW as HCP should be supported only by kata first (no-pp) this commit is not essential for the HCP support , having said that, we better indeed have it ASAP :)

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
```diff
@@ -409,7 +409,7 @@ spec:
                 - name: PEERPODS_NAMESPACE
                   value: openshift-sandboxed-containers-operator
                 - name: RELATED_IMAGE_KATA_DAEMONSET
-                  value: quay.io/openshift_sandboxed_containers/osc-daemonset:1.14-apply-live
+                  value: quay.io/openshift_sandboxed_containers/osc-daemonset:1.13
```
snir911: I'd squash this commit also

scripts/kata-install/osc-kata-install.sh
c3d: There should be a lot of squashing / simplification once we collectively figure out a solution that really works. For now, I admit I'm leaving commits mostly to show "I tried this, then that". Squashing at the end is relatively easy.

config/rbac/role.yaml
c3d: Yes. Squashing will happen, but later unless you feel it's really necessary.

I already had a few instances where I dropped an AI-suggested commit when it turned out the AI had gotten it really wrong. So I prefer to keep separate commits for now as I keep experimenting.

controllers/daemonset_reconcile.go
c3d: > can't we keep only the RELATED_IMAGE_KATA_DAEMONSET option? and if user needs he can override it directly, like we do with the other related images

I think that's a good idea. The configmap override is how this started. The environment variable was added at @gkurz's suggestion to be more in line with the other images. With it, there is probably little point in keeping the ConfigMap override. That being said, Hypershift uses the `hypershift-override-images` CM, so in the HCP context this was sort of a "normal" pattern to have a CM.

config/peerpods/podvm/ami-helper.sh
c3d: > For my understanding this is not needed as if it's exist and owned by this account the first condition will return. (i created similar pr as i was not aware to this: [3f0a201](https://github.com/openshift/sandboxed-containers-operator/commit/3f0a201d67c1aa906a76c3c7c8520d21f7b0102e) )
> 
> BTW as HCP should be supported only by kata first (no-pp) this commit is not essential for the HCP support , having said that, we better indeed have it ASAP :)

This one was suggested by @makentenza. It's interesting that you both came up with a similar thing. Does that mean there are cases where it's necessary?

bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml
c3d: That one should just go. @makentenza thought it was a bad idea to give a 1.13 tag name that did not match. Ultimately, what we really want is a SHA-identified precise tag that we get from an official build, but that won't happen before this PR is closed, because we don't have a need for a daemonset image before that.

config/peerpods/podvm/ami-helper.sh
snir911: Apparently, IIRC we hit this in the CI once this is how i got to it

esposem:
The CSV in the PR (bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml:196-208) has incomplete RBAC verbs — it has create, get, update for serviceaccounts and clusterroles/clusterrolebindings, but is missing list and watch. The kubebuilder markers at controllers/openshift_controller.go:128-129 and config/rbac/role.yaml are correct (they include all verbs). The CSV was hand-edited instead of being regenerated with make bundle.

Can you please ensure you run `make bundle`?

esposem:
Also this PR does not seem to be rebased on devel branch, at least when I pull it locally it doesn't match (weird). Can you ensure it's rebased correctly? Otherwise we test with missing features

c3d:
> The CSV in the PR (bundle/manifests/sandboxed-containers-operator.clusterserviceversion.yaml:196-208) has incomplete RBAC verbs — it has create, get, update for serviceaccounts and clusterroles/clusterrolebindings, but is missing list and watch. The kubebuilder markers at controllers/openshift_controller.go:128-129 and config/rbac/role.yaml are correct (they include all verbs). The CSV was hand-edited instead of being regenerated with make bundle.
> 
> Can you please ensure you run `make bundle`?

Yes, this is the testing I am doing today. Before that I was testing by manually setting up RBAC and giving operator images. For what it's worth, testing in a OLM-compatible setting was what I was working on today.

    - quay.io/c3d/openshift-sandboxed-containers-operator:hcp-test-260713.1
    - quay.io/c3d/osc-daemonset:hcp-test-260713.1
    - quay.io/c3d/openshift-sandboxed-containers-operator-bundle:hcp-test-260713.1
    - quay.io/c3d/openshift-sandboxed-containers-operator-catalog:hcp-test-260713.1

c3d:
> Also this PR does not seem to be rebased on devel branch, at least when I pull it locally it doesn't match (weird). Can you ensure it's rebased correctly? Otherwise we test with missing features

It was rebased as of Monday (when I was on PTO, then July 14 was holiday). I am trying to not rebase once more because I am investigating a problem with the installation not being idempotent and another with the installation on OLM. Which I think is also what @snir911 is running into.

Can you be more precise on what you find weird on the rebase?

c3d:
This was superceded by #2519 and #2539

PR #2434, opened 2026-07-07
fix: controller: return empty Result when bubbling errors

### LINKED ISSUE #2433: controller-runtime: Error handlers return both RequeueAfter and error, violating semantics
## Problem

Throughout `controllers/daemonset_reconcile.go`, error handlers return both a `ctrl.Result{Requeue: true, RequeueAfter: duration}` and a non-nil error. According to controller-runtime semantics, **when `err != nil`, the Result is ignored** and the controller uses exponential backoff instead.

This is misleading because it appears to set a specific requeue delay, but that delay is never actually used.

## Affected Locations

There are 20 instances of this pattern in `controllers/daemonset_reconcile.go`:

```go
// Lines: 181, 189, 200, 204, 215, 230, 242, 256, 263, 272, 278, 312, 345, 354, 358, 369, 385, 392, 400, 420
return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
```

## Expected Behavior

When returning an error, use an empty Result to make the behavior explicit:

```go
return ctrl.Result{}, err
```

The controller will still requeue using its default exponential backoff mechanism.

## Reference

From controller-runtime documentation:
> If a non-nil error is returned, the Result is ignored and the request will be requeued using exponential backoff.

See also:
- https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/reconcile/reconcile.go
- https://www.golinuxcloud.com/requeue-requeueafter-error-handling-controller-runtime/

## Related

Identified during PR #2410 review: https://github.com/openshift/sandboxed-containers-operator/pull/2410#discussion_r3507799778

## Solution

Fixed in commit edd7d58c on branch csb/feat/daemonset-no-reboot-ddd-additions-coderabbit-fixes

c3d:
controller-runtime ignores the Result when `err != nil`, using exponential backoff instead. Returning both a `RequeueAfter` delay and an error is misleading - the delay is never applied.

Changed all error returns from:
```go
  return ctrl.Result{Requeue: true, RequeueAfter: duration}, err
```
to:
```go
  return ctrl.Result{}, err
```
This makes the error handling behavior explicit and matches controller-runtime semantics. The controller will still requeue, but using its default exponential backoff mechanism.

Found while addressing [code review comments](https://github.com/openshift/sandboxed-containers-operator/pull/2410#discussion_r3507799778) on PR #2410

Fixes: #2433

Assisted-By: Claude Sonnet 4.5

### DIFF
--- a/controllers/daemonset_reconcile.go
+++ b/controllers/daemonset_reconcile.go
@@ -70,7 +70,7 @@ func (r *KataConfigOpenShiftReconciler) processKataConfigDeleteRequestDaemonSet(
 	kataInstallDaemonSet, err := r.daemonSetForKataInstall(InstallKata)
 	if err != nil {
 		r.Log.Error(err, "Failed getting DaemonSet for Kata installation")
-		return ctrl.Result{Requeue: true, RequeueAfter: 15 * time.Second}, err
+		return ctrl.Result{}, err
 	}
 	err = r.Client.Delete(context.TODO(), kataInstallDaemonSet)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


c3d:
@coderabbitai review

PR #2454, opened 2026-07-09
konflux: add a tag to our images that mentions the PR it built from

littlejawa:
Identifying images on quay is hard. This commit makes sure we tag our image with the PR number it was built from. This should make it easier to identify what image we can use for testing a build before we merge the PR.

### DIFF
--- a/.tekton/build-pipeline.yaml
+++ b/.tekton/build-pipeline.yaml
@@ -94,6 +94,10 @@ spec:
     type: string
     default: .
     description: Target directories to scan with SAST tools. Multiple values should be separated with commas.
+  - name: additional-tags
+    type: array
+    default: []
+    description: Additional tags to apply to the built image (e.g. pr-<number> for PR builds)
   results:
   - description: ''
     name: IMAGE_URL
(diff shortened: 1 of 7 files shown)


littlejawa:
See: quay.io/redhat-user-workloads/ose-osc-tenant/osc-operator:pr-2454

What do you guys think? As it is, each new build will overwrite this tag... Should I add a timestamp? The commit revision?
Any suggestion is welcome.

c3d:
> See: quay.io/redhat-user-workloads/ose-osc-tenant/osc-operator:pr-2454
> 
> What do you guys think? As it is, each new build will overwrite this tag... Should I add a timestamp? The commit revision? Any suggestion is welcome.

Thanks a lot @littlejawa this will help a lot!

It's marked as "additional tags". So I suspect this means we can have more than one?

If we can, I would suggesting having both `pr-2454` and `pullreq-2454-2ee7ff0e8f8febc89beafda5004a421500777a92`. The reason for the latter is caching. If you overwrite `pr-2454` then a same cluster is likely to say "Ah, I have that already" and not pick up a new build. There are workaround, but for testing purpose, giving the extended name would be nice.

The benefit of a `pr-2454` is that you know what is the latest build for that PR easily.

The benefit if using `pr-` vs `pullreq-` is that you can search for one or the other depending on whether you want the "stable" name (`pr-2454`) or the "mutable" one.

WDYT?

littlejawa:
@c3d - You suggestions sounds good to me. I implemented it. Tell me what you think.

c3d (approved):
/lgtm let's see how it does in practice 😄

PR #2457, opened 2026-07-09
release: update CPE label version to 1.130

dbkreling:
ProdSec product-definitions registers the OSC 1.13 stream with CPE version 1.130. Update all Dockerfiles to match.


littlejawa (approved):
/lgtm
Thanks @dbkreling

PR #2478, opened 2026-07-10
fix(fbc/render.sh): prevent catalog-template.yaml from being left dity on render failure

### LINKED ISSUE #2307: fix(fbc/render.sh): prevent catalog-template.yaml from being left dirty on render failure
## Problem

When `fbc/render.sh` is run and `opm` fails mid-execution (e.g., because the user is not logged in to `registry.redhat.io`), the script exits immediately due to `set -xe`. This leaves `catalog-template.yaml` in a modified state — with the build-registry substitution applied and icon base64 data injected — because the swap-back and icon-removal steps never run.

This was observed in PR #2306 where the bundle image registry for v4.18 was left pointing to `quay.io/redhat-user-workloads/ose-osc-tenant/...` (the internal build registry) instead of `registry.redhat.io/openshift-sandboxed-containers/...` (the release registry), causing a catalog inconsistency.

## Suggested Fix

Instead of modifying `catalog-template.yaml` in-place, render from a temporary copy so the original template is never touched. Only `catalog.json` (the expected output) should be modified.

See the detailed diff proposed in: https://github.com/openshift/sandboxed-containers-operator/pull/2306#discussion_r3411684424

## References

- PR: https://github.com/openshift/sandboxed-containers-operator/pull/2306
- Comment: https://github.com/openshift/sandboxed-containers-operator/pull/2306#discussion_r3411684424

/cc @littlejawa

mmahut:
Fixes https://github.com/openshift/sandboxed-containers-operator/issues/2307

`render.sh` edits `catalog-template.yaml` in-place and only reverts after `opm` succeeds. A mid-run `opm` failure leaves the template dirty.

**- What I did**

Apply the substitutions to a temporary copy of the template and render from it and an EXIT trap removes the copy on failure. Only `catalog.json` is modiffied.

**- How to verify it**

Run `fbc/render.sh` while not logged in to `registry.redhat.io`: opm fails but `catalog-template.yaml` stays unchanged and no temp file is left. A normal run still produces the same `catalog.json`.

fbc/render.sh no longer leaves catalog-template.yaml modified when rendering fails.

### DIFF
--- a/fbc/render.sh
+++ b/fbc/render.sh
@@ -34,11 +34,17 @@ do
     BUILD_IMAGE=$(echo "$RELEASE_IMAGE" | sed "s|$RELEASE_REGISTRY|$BUILD_REGISTRY|")
     BUILD_IMAGE=$(echo "$BUILD_IMAGE" | sed "s|@|${BRANCH_BUILD_SUFFIX}@|")
 
+    # Render from a temporary copy of the template, so the original is never
+    # left dirty if a later step (e.g. `opm`) fails and the script exits.
+    RENDER_TEMPLATE="$PWD/$TEMPLATE_NAME.render.yaml"
+    trap 'rm -f "$RENDER_TEMPLATE"' EXIT
+    cp "$TEMPLATE_NAME" "$RENDER_TEMPLATE"
+
     # Switch to the build registry, so `opm` can pull freely.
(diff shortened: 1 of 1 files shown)


littlejawa (approved):
Thanks @mmahut, much safer.

/lgtm

PR #2561, opened 2026-07-24
osc-must-gather: use a newer builder image

littlejawa:


littlejawa:
As expected, the resulting image shows 0 vulnerabilities. Good to go :-)

PR #2627, opened 2026-08-10
Token based credentials improvements 

snir911:

**- What I did**

**- How to verify it**


wainersm:
## Konflux Build Triage — PR #2627

**Failed pipelines: 3/3** | **Assessment: Mixed — 2 retryable, 1 require fixes**
**Auto-retest:** Waiting for cooldown (retests after 14:16 UTC) | 0/2 retests used

### openshift-sandboxed-containers-enterprise-contract / osc-operator
- **Failed task:** `unknown`
- **Category:** Transient/Infrastructure (auto-learned) (retryable)
- **Log source:** GitHub check run summary only
- **Error:** `(no error text)`

**Suggested action:** Retry the build; collect full logs on next run to confirm timeout or infrastructure cause.
**PipelineRun:** [View in Konflux](https://red.ht/trusted)

---

### make-test / osc-operator
- **Failed task:** `unknown`
- **Category:** Transient/Infrastructure (auto-learned) (retryable)
- **Log source:** GitHub check run summary only
- **Error:** `(no error text)`

**Suggested action:** Retry the build; check cluster resources and timeout settings if retries continue to fail.
**PipelineRun:** [View in Konflux](https://red.ht/trusted)

---

### osc-operator-on-pull-request
- **Failed task:** `build-images`
- **Category:** Unknown (auto-learned) (not retryable)
- **Log source:** GitHub check run summary only
- **Error:** `(no error text)`

**Suggested action:** Retrieve full build logs from the Konflux/Tekton build run to identify the actual failure point.
**PipelineRun:** [View in Konflux](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/ose-osc-tenant/pipelinerun/osc-operator-on-pull-request-hpp74)

---

_Generated by konflux-build-triage_

config/peerpods/credentials-requests/ccoctl/credentials_request_aws_sts.yaml
```diff
@@ -0,0 +1,56 @@
...
+    namespace: openshift-sandboxed-containers-operator
+  serviceAccountNames:
+    - default
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: AWSProviderSpec
+    statementEntries:
+      - effect: Allow
+        action:
+          - ec2:*
```
wainersm, line 18: Hi @snir911 , shouldn't ec2 endpoint also be restricted to a few allowed operations?

controllers/credentials_controller.go
```diff
@@ -167,6 +168,29 @@ func (r *SecretReconciler) ccoDataMapping(ccoSecretData map[string][]byte) map[s
...
+	// Handle ccoctl AWS STS format: the "credentials" key contains an INI-style AWS config file
+	// produced by `ccoctl aws create-iam-roles`. Extract role_arn and web_identity_token_file
+	// and map them to the env-var names expected by the image creation job and CAA.
+	if credFile, ok := ccoSecretData["credentials"]; ok {
+		r.Log.Info("ccoDataMapping: ccoctl AWS STS INI-style config file identified, converting to the expected credentials variables format")
+		stsFieldMap := map[string]string{
+			"role_arn":                "AWS_ROLE_ARN",
+			"web_identity_token_file": "AWS_WEB_IDENTITY_TOKEN_FILE",
+		}
+		for _, line := range strings.Split(string(credFile), "\n") {
```
wainersm, line 181: Replacing this simple parser with a lib is something I was about to suggest but I was not sure (didn't want to introduce a new dep to the project). I chatted with Claude then:

```
Since gopkg.in/ini.v1 is already in the module graph (currently an indirect dependency — a go mod tidy promotes it to direct, no new third-party dep), consider using it here instead of hand-rolling the parse. The credentials blob is a real INI file with a [default] section, and this loop parses key=value lines regardless of section, so it doesn't actually scope to [default] (last-wins across sections if more than one ever appears) and doesn't handle comments/quoting. Reading [default] explicitly also lets us validate the required keys were present instead of silently emitting a partial peer-pods-secret:

import "gopkg.in/ini.v1"

if credFile, ok := ccoSecretData["credentials"]; ok {
    cfg, err := ini.Load(credFile)
    if err != nil {
        r.Log.Error(err, "ccoDataMapping: failed to parse ccoctl AWS credentials INI")
        return peerPodsSecretData
    }
    sec := cfg.Section("default")
    roleARN := sec.Key("role_arn").String()
    tokenFile := sec.Key("web_identity_token_file").String()
    if roleARN == "" || tokenFile == "" {
        r.Log.Info("ccoDataMapping: AWS STS credentials missing role_arn/web_identity_token_file")
    } else {
        peerPodsSecretData["AWS_ROLE_ARN"] = []byte(roleARN)
        peerPodsSecretData["AWS_WEB_IDENTITY_TOKEN_FILE"] = []byte(tokenFile)
    }
}

Not a blocker — the current output is a stable, known ccoctl format — but this would be more robust than line-splitting and folds in the missing-key validation.
```

controllers/credentials_controller.go
```diff
@@ -345,17 +344,33 @@ func (kh *KataConfigHandler) setupPeerPodsCredentials(ctx context.Context) (bool
...
+			return false, err
+		}
+		if stsConfigured {
+			kh.reconciler.Log.Info("peer-pods-secret configured successfully based on provided enviroment variables")
+			return true, nil
+		}
+
+		// Don't fall back to CCO workflow because CCO can't provision credentials in Manual mode.
+		// Return true to indicate "credentials are being handled" and avoid creating CredentialsRequest.
+		kh.reconciler.Log.Info("Cluster is in token-based auth mode but no STS env vars found, waiting for cco-secret from ccoctl workflow")
```
wainersm, line 367: Nit: something that confuses me a lot is the use of "sts" as same thing for Azure and AWS. isn't sts azure-specific then here it should be like "no STS/WIF env vars found"?

controllers/utils.go
```diff
@@ -376,3 +377,45 @@ func (r *KataConfigOpenShiftReconciler) getConfigMapVersion(name, namespace stri
...
+// Returns:
+//   - true if the cluster is in token-based auth mode
+//   - false if not in token-based auth mode or if resources don't exist/can't be fetched
+func isClusterInTokenBasedAuthMode(c client.Client) bool {
+	ctx := context.Background()
+
+	// 1. Check CloudCredential for Manual credentials mode
+	cloudCred := &operatorv1.CloudCredential{}
+	if err := c.Get(ctx, client.ObjectKey{Name: "cluster"}, cloudCred); err != nil {
+		// If CloudCredential doesn't exist or can't be fetched, assume not in token-based mode
```
wainersm, line 400: It bothers me this assumption of "if case of error assume not token-based mode". I think this might have bad consequences but then I am not expert. So I asked claude again:

```
Both CloudCredential/cluster and Authentication/cluster are cluster singletons that always exist, so a real NotFound shouldn't occur in practice — which means any error reaching this code is realistically transient, RBAC, or scheme-related. Collapsing them all into return false misclassifies the cluster as not token-based, and on a cluster that actually is STS/WIF that produces two silent failures: setupPeerPodsCredentials falls back to the CCO workflow and creates a CredentialsRequest that CCO can't fulfill in Manual mode, and configureCAAProvider skips addBoundSATokenVolume, so CAA comes up without the projected token.

Suggest switching to (bool, error), threading ctx, and having callers requeue on error. Full change:

1. controllers/utils.go — the function

func isClusterInTokenBasedAuthMode(ctx context.Context, c client.Client) (bool, error) {
      // 1. CloudCredential must be in Manual mode
      cloudCred := &operatorv1.CloudCredential{}
      if err := c.Get(ctx, client.ObjectKey{Name: "cluster"}, cloudCred); err != nil {
              // CloudCredential/cluster is a singleton that always exists; surface every error
              // (transient / RBAC / scheme) so the caller requeues instead of assuming "not token-based".
              return false, fmt.Errorf("getting CloudCredential/cluster: %w", err)
      }
      if cloudCred.Spec.CredentialsMode != operatorv1.CloudCredentialsModeManual {
              return false, nil
      }

      // 2. Authentication must have a non-empty OIDC issuer
      auth := &configv1.Authentication{}
      if err := c.Get(ctx, client.ObjectKey{Name: "cluster"}, auth); err != nil {
              return false, fmt.Errorf("getting Authentication/cluster: %w", err)
      }
      return auth.Spec.ServiceAccountIssuer != "", nil
}

(If you'd rather keep NotFound as a legitimate "not token-based" signal, replace the two error returns with: if k8serrors.IsNotFound(err) { return false, nil }; return false, fmt.Errorf(...) — but since these are singletons, treating all errors as requeue-worthy is safer.)

2. controllers/credentials_controller.go — setupPeerPodsCredentials (~line 348)

This func already returns (bool, error) and has a real ctx:

      isTokenBased, err := isClusterInTokenBasedAuthMode(ctx, kh.reconciler.Client)
      if err != nil {
              kh.reconciler.Log.Error(err, "failed to determine cluster auth mode; requeueing")
              return false, err
      }
      if isTokenBased {
              ...

3. controllers/peerpods.go — configureCAAProvider (~lines 252 and 261)

configureCAAProvider returns (*appsv1.DaemonSet, error) and is called inside the CreateOrUpdate mutate func in enablePeerPodsMiscConfigs, so returning the error propagates up and fails/requeues the reconcile naturally. There's no ctx param here (the enclosing code uses context.TODO()), so pass context.TODO() unless you want to thread ctx through as a small extra cleanup. Since the Azure and AWS cases are identical, simplest is to compute it once before the switch:

      tokenBased, err := isClusterInTokenBasedAuthMode(context.TODO(), r.Client)
      if err != nil {
              return ds, err
      }

      switch provider {
      ...
      case AzureProvider:
              if tokenBased {
                      r.addBoundSATokenVolume(ds)
              }
              return ds, nil
      case AWSProvider:
              if tokenBased {
                      r.addBoundSATokenVolume(ds)
              }
              return ds, nil
      }

That way a transient error requeues instead of silently building a DaemonSet without the bound-SA-token volume.
```

Dockerfile
```diff
@@ -36,6 +36,7 @@ COPY --from=builder /workspace/config/peerpods /config/peerpods
 
 RUN useradd  -r -u 499 nonroot
 RUN getent group nonroot || groupadd -o -g 499 nonroot
+RUN microdnf install -y tar && microdnf clean all
```
wainersm, line 39: Every new dependency we add in the project images can potentially drag us into the CVE handling hell. I am concerned about `tar` being a dependency that will have us to handle many CVEs... so I ask...guess who?

Here is claude's reply:
```
Could we avoid adding tar to the runtime image altogether? It's being pulled in solely so the guide can oc cp two small static YAML files out of the operator pod, and oc cp needs tar inside the container (it's implemented as oc exec -- tar cf - … | tar xf -). Since the base is ubi-minimal — which ships without tar — this also required the extra hermetic-build plumbing in this PR (rpms.in.yaml, rpms.lock.yaml, ubi.repo, and the .tekton/* changes) just to make the package installable. That's a fair amount of added surface for a copy of two files.

Two alternatives that need no tar in the image:

Option A — oc exec … -- cat (smallest change; only needs cat, which ubi-minimal already has):

oc exec -n openshift-sandboxed-containers-operator "${OPERATOR_POD}" -- \
  cat /config/peerpods/credentials-requests/ccoctl/credentials_request_azure_wif.yaml \
  > ./credrequests/credentials_request_azure_wif.yaml

(and the AWS equivalent for credentials_request_aws_sts.yaml). Keeps the "fetch from the running operator" UX exactly as documented.

Option B — oc image extract (no running pod, no exec, no tar): pulls the files straight from the operator image layer, e.g.

oc image extract "${OPERATOR_IMAGE}" \
  --path /config/peerpods/credentials-requests/ccoctl/:./credrequests/

Either way, the microdnf install -y tar line and the associated RPM/hermetic-build additions could be dropped, keeping the runtime image smaller and this PR's footprint smaller. Was oc cp a deliberate choice, or would one of these work?
```

wainersm:
Hi @snir911 , I finished my 1st review round \o/

It was half-human, half-machine: for some topics I was not sure I asked Claude. So I didn't just point claude to the PR and asked to review for me. :)

config/peerpods/credentials-requests/ccoctl/credentials_request_aws_sts.yaml
snir911: it is, however, it's hard to determine what exactly would cover all the CAA use cases, I'll try to revisit this.

controllers/credentials_controller.go
snir911: you find this option better?

controllers/credentials_controller.go
snir911: Indeed confusing :P STS stands for Short Token Service, it's normally used with AWS, for Azure/GCP - WIF (workload identity federation) is the common term. earlier i used "STS flow" as a general term for token based authentication, this is probably wrong, I'll try to make things more consistent..

Dockerfile
snir911: I can indeed use one of the two suggested options by our AI friend (i tried the first and it worked fine), however, 'oc cp' seems like the common openshift practice, this is the only reason i added this pkg, WDYT?

PR #2630, opened 2026-08-13
peer-pods: add operator webhook to inject permissive initdata for non-CVM pods

esposem:
Add a mutating admission webhook in the operator that automatically injects a permissive initdata annotation on peer-pod pods using non-confidential VM instance types. This allows exec and logs to work on standard (TrustedLaunch) VMs while keeping the restrictive default policy for confidential VMs (SEV-SNP).

The webhook reads CLOUD_PROVIDER from peer-pods-cm to determine the cloud platform (currently Azure, extensible to AWS and others). When the pod has no machine_type annotation, it falls back to AZURE_INSTANCE_SIZE to determine confidentiality. Pods that already carry a cc_init_data annotation are left untouched.

This effectively replaces the feature gate in Azure.

**- What I did**

**- How to verify it**

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -180,6 +180,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = controllers.SetupInitdataWebhook(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "PeerPodInitdata")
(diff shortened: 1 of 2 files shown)


esposem:
Reminder to self: also update the kata-remote  enable_annotation list in configuration.toml

PR #2647, opened 2026-08-18
Update network policy rbac

littlejawa:
We're getting failures in Conforma, because of a new rule for operators in OCP 5.0
We need to declare the RBAC for NetworkPolicy.

Running "make bundle" also modified unrelated files. This comes from the changes we made for 1.13.1 I think.
I'm keeping them in a separate commit, we can make a separate PR if needed.


gkurz (approved):
/lgtm

Even if the actual network policy support isn't here yet, I guess we can merge this if it helps with the conforma checks.

PR #2655, opened 2026-08-19
fix: regen bundle manifest to pull latest config

thejasn:
Re-run `make bundle` to update csv in bundle manifest to latest config.


littlejawa (approved):
/lgtm

Thanks @thejasn !

PR #2814, opened 2026-09-04
Bump to 1.14

littlejawa:
Doing the bump to make sure the builds we generate can be identified as "next release".

Also cleaning up the label management for the images: passing the version as a parameter, so that we have just one place to edit to bump the version there.


littlejawa:
Fixes [KATA-6040](https://redhat.atlassian.net/browse/KATA-6040)

gkurz (approved):
/lgtm

Thanks @littlejawa !

PR #2867, opened 2026-09-09
ci: add Claude Code commit validation skill

vvoronko:
## Summary

- Adds `osc-commit` Claude Code skill that enforces commit conventions from `docs/CONTRIBUTING.md`
- The skill validates build/lint, commit message format, and sign-off before creating commits
- Detects which Go module (operator root vs `test/e2e`) was affected and runs the appropriate checks

## What the skill enforces

| Step | Check | Details |
|---|---|---|
| 1. Pre-flight | Build & lint | Operator root: `make fmt vet && go build ./...`; E2E: `cd test/e2e && go build/vet` |
| 2. Staging | File review | No `git add -A`, re-stage after edits, no secrets |
| 3. Message format | Subsystem prefix | One of: `feat:` `fix:` `docs:` `test:` `ci:` `build:` `refactor:` `perf:` `chore:` `api:` `controller:` `webhook:` `peerpods:` `monitor:` |
| 3. Message format | Subject line | Max 80 characters including prefix |
| 3. Message format | Body | Required, lines max 150 chars (non-alpha-start lines exempt) |
| 3. Message format | JIRA reference | Optional `Fixes: rhjira#KATA-XXXX` |
| 4. Commit | Sign-off | Always `git commit -s`, never fabricate identity |
| 5. Post-commit | Validation | `./hack/commit-msg-check.sh origin/devel HEAD` |
| 6. Report | Summary | `git log --oneline -1`, `git diff --stat`, checker result |

## Verification

This commit was created using the skill's own steps as a dry run:

```
$ git status
Changes to be committed:
  new file:   .claude/skills/osc-commit/SKILL.md

$ ./hack/commit-msg-check.sh origin/devel HEAD
Checking commits from origin/devel to HEAD
All commit messages are valid.

$ git log --oneline -1
8e6a740e ci: add Claude Code commit validation skill

$ git diff --stat HEAD~1
 .claude/skills/osc-commit/SKILL.md | 101 +++++++++++++++++++++++++++
 1 file changed, 101 insertions(+)
```

No Go code changed — build/lint pre-flight was not applicable for this commit.

## Test plan

- [ ] Verify skill triggers when asking Claude Code to commit in this repo
- [ ] Verify correct module detection (operator root vs test/e2e)
- [ ] Verify commit-msg-check.sh passes on skill-produced commits

### DIFF
--- a/.claude/skills/osc-commit/SKILL.md
+++ b/.claude/skills/osc-commit/SKILL.md
@@ -1,0 +1,101 @@
+---
+name: osc-commit
+description: >-
+  Validates and creates commits in the sandboxed-containers-operator repo.
+  Enforces commit message format, sign-off, and pre-commit checks from
+  docs/CONTRIBUTING.md. Trigger when the user asks to commit changes.
+---
+
+# OSC Commit Skill
+
+When this skill is triggered, follow every step below before creating the commit.
+Do not skip steps. If any step fails, stop and report the failure.
+
+## 1. Pre-flight checks
+
+Determine which Go modules were affected by the staged changes, then run the
+appropriate build and lint commands.
+
+**Operator root module** (changes outside `test/e2e/`):
+
+```bash
+make fmt vet
+go build ./...
+```
+
+**E2E test module** (changes under `test/e2e/`):
+
+```bash
+cd test/e2e
+go build ./...
+go vet ./...
+```
+
+If both modules are affected, validate both. If gofmt or `make fmt` modifies
+files, re-stage them before proceeding.
+
+## 2. Stage files
+
+- Stage only the files relevant to the change. Never `git add -A`.
+- After editing any already-staged file, re-stage it.
+- Verify with `git status` — no unexpected files staged.
+- Never stage files that may contain secrets (`.env`, credentials, tokens).
+
+## 3. Compose the commit message
+
+Follow the format from `docs/CONTRIBUTING.md`:
+
+```
+<subsystem>: <short summary>
+
+<Body: what changed and why. Wrap at 150 chars.>
+
+Signed-off-by: (added by git commit -s)
+```
+
+### Rules
+
+- **Subsystem prefix required** — one of:
+  `feat:` `fix:` `docs:` `test:` `ci:` `build:` `refactor:` `perf:` `chore:`
+  `api:` `controller:` `webhook:` `peerpods:` `monitor:`
+- **Subject line max 80 characters** (including prefix)
+- **Body required** — explain what and why, not how
+- **Body lines max 150 characters** (lines starting with non-alpha chars are exempt)
+- If fixing a JIRA issue, add `Fixes: rhjira#KATA-XXXX` in the body
+- **Never add Co-Authored-By Claude tags**
+- **Never modify git config** (user.name, user.email)
+
+## 4. Commit
+
+Always use `git commit -s` to add the Signed-off-by automatically.
+Let `git commit -s` use git config for sign-off — never fabricate identity.
+
+Use a HEREDOC to pass the message:
+
+```bash
+git commit -s -m "$(cat <<'EOF'
+<subsystem>: <summary>
+
+<body>
+EOF
+)"
+```
+
+## 5. Post-commit validation
+
+Run the commit message checker:
+
+```bash
+./hack/commit-msg-check.sh origin/devel HEAD
+```
+
+If it fails, the commit message does not meet project standards.
+Do NOT amend — create a new commit with the corrected message after
+resetting the failed one (`git reset HEAD~1` then re-stage and re-commit).
+
+## 6. Report
+
+Show the user:
+- `git log --oneline -1` — the commit
+- `git diff --stat HEAD~1` — what changed
+- Result of `commit-msg-check.sh`


wainersm:
Hi @vvoronko , I don't think it should be a skill on its own, but rather some instructions in the project's `AGENTS.md` so that any LLM will auto-magically adhere to the convention of the project.

.claude/skills/osc-commit/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+## 1. Pre-flight checks
+
+Determine which Go modules were affected by the staged changes, then run the
+appropriate build and lint commands.
+
+**Operator root module** (changes outside `test/e2e/`):
+
+```bash
+make fmt vet
+go build ./...
```
littlejawa, line 23: I'm afraid it is more complicated than that.

First: we are building multiple things here. If the changes are in the `must-gather` or 
`bundle` subfolders, we should try building the must-gather or bundle image, not the main operator. Same goes for changes under `config/peerpods/podvm` which are related to the podvm-builder image.
This `go build` command will just miss anything changed there.

Second: we have a lot more than go code. We also have scripts that we could check (shfmt? shellcheck?). We could also consider running some yaml linter, because yaml files are modified much more often than go code.

Third: even if we just modify the operator's go code, `go build ./...` will not necessarily work. I've just tried to run `go build ./...` on a clean repo (after doing "make clean"), and it failed...  Operators are not like any "go" app :-( 

We run `make` or `make bundle` when we want to do local builds of the operator and bundle. This should be the preferred way to check if the modified code builds.
But running these `make` commands will make modifications that we don't want to commit... I'm not sure how to teach an agent how to make the difference between the things we want, and the things we don't :-/ My best workflow so far was to stage my changes first, then run a build, and see what was modified, discarding the unnecessary stuff...
I'm not sure you want to try teaching an agent to do that. It really should be the developer's responsibility.

The bottom line is: don't try :-)
Adding linters is fine, I think it's a good idea. But don't let the agent try to build it (yet). We have to figure out safer ways first.

.claude/skills/osc-commit/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+**E2E test module** (changes under `test/e2e/`):
+
+```bash
+cd test/e2e
+go build ./...
+go vet ./...
+```
+
+If both modules are affected, validate both. If gofmt or `make fmt` modifies
+files, re-stage them before proceeding.
```
littlejawa, line 35: But you didn't stage anything yet at this point? The stage step is after this?

.claude/skills/osc-commit/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+
+- **Subsystem prefix required** — one of:
+  `feat:` `fix:` `docs:` `test:` `ci:` `build:` `refactor:` `perf:` `chore:`
+  `api:` `controller:` `webhook:` `peerpods:` `monitor:`
+- **Subject line max 80 characters** (including prefix)
+- **Body required** — explain what and why, not how
+- **Body lines max 150 characters** (lines starting with non-alpha chars are exempt)
+- If fixing a JIRA issue, add `Fixes: rhjira#KATA-XXXX` in the body
+- **Never add Co-Authored-By Claude tags**
+- **Never modify git config** (user.name, user.email)
```
littlejawa, line 66: Do not repeat the rules here. Duplicating information means double-work when we need to change it.
The reference to the "doc/CONTRIBUTING.md" file should be enough for the agent to understand the rules. If not, we may want to clarify the doc. But don't duplicate it.

.claude/skills/osc-commit/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+Always use `git commit -s` to add the Signed-off-by automatically.
+Let `git commit -s` use git config for sign-off — never fabricate identity.
+
+Use a HEREDOC to pass the message:
+
+```bash
+git commit -s -m "$(cat <<'EOF'
+<subsystem>: <summary>
+
+<body>
```
littlejawa, line 79: AND add a "Co-authored-by [agent]" line.
I've seen "Assisted by" also, not sure what's the rule,  but any code written with an AI should have it explicitly marked as such in the sign-off.

.claude/skills/osc-commit/SKILL.md
wainersm: I'm worried about duplication of information as well. Today we have:

* doc/CONTRIBUTING.md
* README.md
* AGENTS.md

Each serves to a purpose (first two are for humans). We should cross-reference vs duplicate information.

.claude/skills/osc-commit/SKILL.md
wainersm: Open source projects use "Assisted-by" and I use that on all projects I work. Semantically speaking, I prefer "assisted" over "co-authored" because in the end the work belongs to the human being (that's the last understand at least :)

.claude/skills/osc-commit/SKILL.md
vvoronko: @littlejawa I see, will update this skill for test scope only

.claude/skills/osc-commit/SKILL.md
vvoronko: agree

PR #2880, opened 2026-09-10
test/e2e: run upstream tests from osc-release branch

wainersm:
The run_upstream_tests.sh which is a wrapper around kata containers tests currently tracks the downstrem tree (openshift/kata-containers) and main branch. It is wrong to use the main branch, so the primary goal here is to switch to osc-release (the correct one, the devel branch). As the branch was switched, the execution of the tests broken completely so I had to work on a fix. Apart from the fix, new tests were present on the new branch so I had to re-evaluate the list of tests (include new ones, remove flaky tests).

The full profile was executed 5x on the same cluster that was in turn built by one of our current prow jobs

### DIFF
--- a/test/e2e/run_upstream_tests.sh
+++ b/test/e2e/run_upstream_tests.sh
@@ -12,6 +12,12 @@ export AUTO_GENERATE_POLICY="no"
 export CONTAINER_RUNTIME="crio"
 export K8S_TEST_DEBUG="false"
 
+# RuntimeClass the workloads should run on. Upstream kata-containers uses a
+# kata-${KATA_HYPERVISOR} naming schema (e.g. kata-qemu), whereas OSC uses a
+# different one (kata, kata-remote, kata-cc, ...) and always defines "kata" as
+# the default. See the normalization step below.
+OSC_RUNTIMECLASS="${OSC_RUNTIMECLASS:-kata}"
+
 usage() {
     cat <<EOF
 Usage: $(basename "$0") [options]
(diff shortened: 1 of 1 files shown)


test/e2e/run_upstream_tests.sh
```diff
@@ -205,6 +218,18 @@ if [[ "$SKIP_SETUP" != "true" ]]; then
...
+    # The osc-release setup.sh rewrites the fixtures' `runtimeClassName: kata` to
+    # kata-${KATA_HYPERVISOR} (e.g. kata-qemu), but OSC exposes RuntimeClasses
+    # named kata, kata-remote, kata-cc, ... with "${OSC_RUNTIMECLASS}" (handler
+    # "kata") as the default. Without this, every pod is rejected with
+    # `RuntimeClass "kata-qemu" not found`.
+    echo "Pointing test workloads at the '${OSC_RUNTIMECLASS}' RuntimeClass..."
+    find runtimeclass_workloads_work -type f \
+        \( -name '*.yaml' -o -name '*.yaml.in' \) -print0 |
+        xargs -0 --no-run-if-empty sed -i -E \
+            "s/^([[:space:]]*runtimeClassName:[[:space:]]+)kata(-[^[:space:]]*)?[[:space:]]*\$/\1${OSC_RUNTIMECLASS}/"
```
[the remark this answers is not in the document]
wainersm: That is script is not exposed to malicious users, so not going to address that.

test/e2e/run_upstream_tests.sh
```diff
@@ -205,6 +211,18 @@ if [[ "$SKIP_SETUP" != "true" ]]; then
     echo "Running upstream setup.sh..."
     bash setup.sh || exit 1
 
+    # --- Point workloads at the OSC RuntimeClass ---
+    # The osc-release setup.sh rewrites the fixtures' `runtimeClassName: kata` to
```
ldoktor: This should probably state `The kata-containers's setup.sh` or even `The upstream kata-containers's setup.sh` to emphasize we don't have control over this file and need to keep this workaround.

test/e2e/run_upstream_tests.sh
```diff
@@ -205,6 +211,18 @@ if [[ "$SKIP_SETUP" != "true" ]]; then
...
+    # kata-${KATA_HYPERVISOR} (e.g. kata-qemu), but OSC exposes RuntimeClasses
+    # named kata, kata-remote, kata-cc, ... with "${OSC_RUNTIMECLASS}" (handler
```
ldoktor: This can be probably also better worded, like `OSC uses different naming but always defines the default one as "kata"` or something like that.

ldoktor:
Overall looks good to me, but I'd suggest improving the comments to emphasize why the workaround is needed (it's a different naming schema compare to upstream). As for the functional test failure, I'd suggest rebasing the PR as there were quite a few changes to prow recently.

test/e2e/run_upstream_tests.sh
wainersm: done

test/e2e/run_upstream_tests.sh
wainersm: done

PR #2944, opened 2026-09-14
tests: add Azure CoCo jobs

wainersm:
Triggering CoCo jobs for Azure.
OCP 4.21 and 4.22 jobs cannot be executed in parallel because they will be use too many Azure quota, so rebalanced the execution of the jobs.

Fixes: rhjira#[KATA-4846](https://redhat.atlassian.net/browse/KATA-4846)


dbkreling (approved):
LGTM

PR #2972, opened 2026-09-16
[WIP] Feature: Enable mixed cluster(BM+VM nodes)

mayamrinal:
OSC today assumes a homogeneous worker node fleet - all bare-metal or all VM. Someone running OpenShift on hybrid infrastructure (some workers on bare metal, others on VMs) cannot safely run sandboxed workloads because RuntimeClasses like `kata` and `kata-remote` share the same `scheduling.nodeSelector` (`node-role.kubernetes.io/kata-oc`). 

**- What I did**
Introduced `EnableMixedCluster` in `KataConfigSpec` to activate per-node-class sandbox scheduling. When enabled:

1. The operator deploys a NodeFeatureRule (`osc-node-type`) via NFD that labels worker nodes as kataconfiguration.openshift.io/node-type=bare-metal` or `virtual` based on the x86 CPUID hypervisor-present bit.
2. All bare-metal RuntimeClasses (`kata`, `kata-cc`, `kata-nvidia-gpu`, `kata-cc-nvidia-gpu`) receive `node-type: bare-metal` in their `scheduling.nodeSelector`. `kata-remote` receives `node-type: virtual`.
3. Existing RuntimeClasses on a running cluster are patched when `EnableMixedCluster` is toggled on (update path in `createRuntimeClass`).
4. The NodeFeatureRule is removed and RuntimeClass nodeSelectors are reverted when `EnableMixedCluster` is toggled off.
5. A pre-flight check verifies the NFD Operator is installed before proceeding.

**- How to verify it**
1. Install NFD Operator and create a `NodeFeatureDiscovery` CR
2. Deploy the operator with this change
3. Create a KataConfig with `enableMixedCluster: true`
4. Verify nodes are labeled:
   `oc get nodes -l kataconfiguration.openshift.io/node-type`
5. Verify RuntimeClass nodeSelectors include the `node-type` label:
   `oc get runtimeclass kata -o jsonpath='{.scheduling.nodeSelector}'`
   `oc get runtimeclass kata-remote -o jsonpath='{.scheduling.nodeSelector}'`
6. On a VM-only cluster, all nodes should be labeled `virtual` and `kata` pods should stay Pending (no BM node available)
7. Toggle `enableMixedCluster: false` and verify labels are removed and RuntimeClass nodeSelectors revert
8. Run unit tests: `go test ./controllers/ -run "^Test" -v`

Add `enableMixedCluster` support to restrict RuntimeClass scheduling per node class on clusters with both bare-metal and VM worker nodes.

**- TODO**
- [ ] Testing on mixed cluster
- [ ] **Nested virtualization support:** VM nodes where the hypervisor exposes
  nested virt can run Kata sandboxes via KVM-on-KVM. This iteration
  does not distinguish nested-virt-capable VMs from non-capable ones, all VM
  nodes are labeled `virtual` and routed to `kata-remote` (peer pods).


c3d:
@mayamrinal Thanks, this is quite interesting.

* What happens if `EnableMixedCluster` is false and you are running on a mixed cluster? I am wondering if this is not something we should treat as a bug rather than a conditional feature.

* I am concerned about the `kata-remote` having a `node-type: virtual`. AFAICT, `kata-remote` relies on cloud-provided infrastructure, but it does not *rely* on the nodes being virtual. At least in theory, we could have a cloud API adapter that provisions bare metal nodes instead of VMs, no? That does not work for CoCo, where the CoCo technology is designed to run in a VM. What about calling that case "peer" or "remote" or "cloud"?

* Related to the above: the nested case would be "virtual" as well. If you want to support nested, you will need a way to distinguish nested nodes from peer-pod / CAA nodes.

c3d:
Also wondering how this interacts with GPU and CoCo markers, e.g. discovering a SEV node with GPU?

mayamrinal:
@c3d, thanks a lot for raising the points

> What happens if EnableMixedCluster is false and you are running on a mixed cluster? I am wondering if this is not something we should treat as a bug rather than a conditional feature.

The flag exists as a backward-compatibility gate as this feature is dependent on NFD. Unconditionally adding node-type to RuntimeClass nodeSelectors would break every existing cluster that doesn't run NFD; nodes wouldn't have the label, sandbox scheduling stops.

`EnablePeerPods` is the exact same pattern, it's a feature that could theoretically be auto-detected (the operator could check if CAA is configured), but it's explicitly declared because the admin needs to control when and whether the peer-pod path is activated. `EnableMixedCluster` follows the same design principle: the admin declares intent, the operator acts on it.

>I am concerned about the kata-remote having a node-type: virtual. AFAICT, kata-remote relies on cloud-provided infrastructure, but it does not rely on the nodes being virtual. At least in theory, we could have a cloud API adapter that provisions bare metal nodes instead of VMs, no? 

Yes you are correct, theoretically we could have a cloud API adapter that provisions bare metal nodes instead of VMs. But these labels and node selectors are only added when we set `EnableMixedCluster` to true, i.e. we have both types of nodes(BM and VM) in the cluster. kata-remote pods will be handled by the VM node.

>That does not work for CoCo, where the CoCo technology is designed to run in a VM. What about calling that case "peer" or "remote" or "cloud"?

The label answers "what is this node?" that's NFD's job. The operator answers "what sandbox path runs here?"  that's KataConfig's job. Mixing policy into the label couples two concerns that that I think should remain independent.

>Related to the above: the nested case would be "virtual" as well. If you want to support nested, you will need a way to distinguish nested nodes from peer-pod / CAA nodes.

I agree, to introduce nested in the current implementation I will need a way to distinguish VM nodes that support nested from peer-pod / CAA VM nodes.

>Also wondering how this interacts with GPU and CoCo markers, e.g. discovering a SEV node with GPU?

The mixed-cluster label is additive to existing TEE and GPU labels, all conditions are ANDed in the RuntimeClass nodeSelector. 

For example, kata-cc-nvidia-gpu on a SEV-SNP node with GPU becomes:
```
scheduling:
  nodeSelector:
    node-role.kubernetes.io/kata-oc: ""
    amd.feature.node.kubernetes.io/snp: "true"
    nvidia.com/gpu.present: "true"
    nvidia.com/cc.mode.state: "on"
    kataconfiguration.openshift.io/node-type: "bare-metal"
```
The node must match all labels. No conflict with existing TEE or GPU detection.

Also, CoCo + GPU on VM nodes goes through kata-remote with pod annotations , not through kata-cc-nvidia-gpu. So restricting kata-cc-nvidia-gpu to bare-metal doesn't block that path.

---
Please fell free to push back/correct me if I am worng.

